PR #218, opened 2014-07-08
Remove remaining obsolete compatibility options

Gymnasiast:
Gnome-panel 2.32 contained code to upgrade from very old versions. These old versions have no corresponding MATE version, so the code doesn't do anything and is never called. Most of it is already removed, but this pull request removes the remainder.

Code taken from commits b9fc819b956319b6ae99fa13bf54c4f54d9e3a0e and e57b3afd9255671789ea6c1b6136bbad82573cdc of gnome-panel.

### DIFF
--- a/mate-panel/applet.c
+++ b/mate-panel/applet.c
@@ -65,8 +65,6 @@ mate_panel_applet_set_dnd_enabled (AppletInfo *info,
 		break;
 	case PANEL_OBJECT_APPLET:
 		break;
-	case PANEL_OBJECT_LOGOUT:
-	case PANEL_OBJECT_LOCK:
 	case PANEL_OBJECT_ACTION:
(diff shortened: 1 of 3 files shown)


stefano-k:
@Gymnasiast could you update this PR?

monsta:
@stefano-k: is there anything for that in ```org.mate.panel.object``` schema?

monsta:
So what should we do about it? Looks like we have to add some schema changes as well?

infirit:
merged as a single squashed commit. Don't know what needs to be done in the schema but feel free to open a new pull for that,

Gymnasiast:
Thanks, infirit!

monsta:
Hmm... when I edit ```object-type``` key for any existing applet via dconf-editor, I see just a string... nothing like enum with a set of allowed values...

Issue #149, opened 2013-12-12
panel buggy on multiple xscreens


ethanj:
I just did a clean install of mint 16, and configured a second xscreen by modifying xorg.conf .
When mate comes up, it correctly has a panel at the bottom of each screen, and I can add widgets to each.
However, any interaction with the widgets on the second screen shows up on the first screen rather than the second.
For instance, if I add a menu to the second screen's panel, then click on it, the popup which would normally show up above it shows up in the exact same location on the first screen, even if the first screen's panel does not contain a menu widget.
##

monsta:
@ethanj: this should be already fixed in 1.10 (see https://github.com/mate-desktop/mate-panel/commit/3f67cbdfeaa53820ed690dd25f52dc162e7511e2).

axelolmos:
I still had the problem with 1.10 (mate-panel-1.10.1-1.fc22.x86_64) and had to apply the patch I wrote.

monsta:
Well that's because your issue is different.

axelolmos:
This issue has gotten worse since Fedora 20.  In Fedora 20, multiple DISPLAYs would work with the mate-panel patch I proposed.  After that, (Fedora 21,22,23...) my patch fixes the menu, but on DISPLAYS greater than :0:0 the dropdown menu doesn't actually work, and the screen has a black background.

It really seems like no testing of DISPLAY > :0.0 is being done, and as new code is introduced, multiple DISPLAYs are breaking more and more.  I know testing multiple monitors, and drivers other than nouveau (e.g. NVIDIA) is a pain, but other modern OS's (Windows, Mac) support multiple monitors fine.

raveit65:
> the dropdown menu doesn't actually work
You mean hovering with the mouse over a menu entry does not work in f23?

axelolmos:
Yes, the menu pops down, but doesn't actually work.  In addition, the redraw of the menus does not work, so they just make a mess after they drop down as the desktop background is black and does not get updated.

I tested most of the desktops for Fedora 23:
NVIDIA Driver, Separate X Screens in xorg.conf:

KDE Plasma Workspaces           Black screen on 2nd screen
LXDE Desktop                            WORKS!  Needs menu bar on all windows.  
LXQt Desktop                             WORKS!  Needs menu bar on all windows.
Cinnamon Desktop                     Fallback mode, Grey screen on 2nd screen
MATE Desktop                           Black screen on 2nd screen, menu appears but doesn't work
[...]

PR #228, opened 2014-08-20
Patched panel-profile.c to have multiple Xscreens work.

axelolmos:
For newer versions of the panel, if you have multiple X screens, that is to say:
DISPLAY=:0.0 DISPLAY=:0.1 the panel will create two menus on screen 0, and none on screen 1.

The desired behavior is one menu per X screen.  This gets very important if you are running multiple video cards and multiple monitors.

When running only one X screen this patch shouldn't matter.

The stacking of menus and icons on the first screen has been happening since the gnome-panel >2.1 days.  It happens on anything that forked off of newer versions of Gnome 2.

This patch, although crudely written, restores the desired behavior.
Please modify it to your content and apply it to all versions of the panel.

Thank You.

### DIFF
--- a/mate-panel/panel-profile.c
+++ b/mate-panel/panel-profile.c
@@ -1392,6 +1392,7 @@ panel_profile_delete_object (AppletInfo *applet_info)
 
 	panel_profile_remove_from_list (type, id);
 }
+int screennum=0;
 
 static void
 panel_profile_load_object (char *id)
@@ -1399,6 +1400,7 @@ panel_profile_load_object (char *id)
 	PanelObjectType  object_type;
 	char            *object_path;
 	char            *toplevel_id;
+	char             newtoplevel_id[14];
 	int              position;
 	gboolean         right_stick;
 	gboolean         locked;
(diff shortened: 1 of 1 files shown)


thenewguy:
Any chance of this being fixed?

NiceandGently:
@monsta 
Can you please review this long outstanding PR ?

monsta:
In https://github.com/mate-desktop/mate-panel/issues/149 the menubar is mentioned only as example:

> However, any interaction with the widgets on the second screen shows up on the first screen rather than the second.

So what if the widget is a clock, for example? Or a volume control applet?
I look at the code here and don't see anything for that.

axelolmos:
I created this web page to illustrate the issue:
http://www.olmosconsulting.com/mate-panel

There's documentation there as to the nature of the problem, and how I solved it.

The easiest way to get a handle on this is have two monitors on your PC, and set up your xorg.conf so that you have two screens.  I have a sample xorg.conf where I do 16 screens at: http://www.olmosconsulting.com/mate-panel/xorg.conf

The code I submitted is a hack.  I simply made a guess as to what was going on, and tried to fix it.  When it worked the way I expected, I stopped writing.

That being said, the ability to use multiple video cards and multiple monitors with separate screens has been a feature of Unix / Linux operating systems running X / Xorg for decades.  It shouldn't die because of bug in an application higher up on the stack.

monsta:
> The code I submitted is a hack.

That's why we can't accept it as is. We need some more generic solution.

```
/* We make the possibly wrong assumption that each screen only has one menu bar
   This is how we know to increment the screen number.  */
```

What if there's no classic menubar at all?

linuxmint-mate

PR #261, opened 2014-11-26
2 fixes

monnerat:
- Issue 214
- Wrong system timezone detection.


applets/clock/system-timezone.c
```diff
@@ -634,6 +634,22 @@ system_timezone_read_etc_localtime_softlink (void)
                 return NULL;
 
         file = g_file_read_link (ETC_LOCALTIME, NULL);
+
+        if (*file != '/') {
```
dnk: use g_path_is_absolute

monnerat:
@dnk: thanks for suggestion. Done.

NiceandGently:
@stefano-k 
Can you please review this for 1.10.
I use the patch for several month in fedora and downstream report is closed since 2014-12-01.

PR #268, opened 2014-12-16
Set Cancel as default action when deleting a panel/drawer

djails:
When deleting a panel, the confirmation dialog box shows two buttons: Cancel and Delete.
The default button is Delete, this patch makes it Cancel to prevent accidental deletion.


monsta:
@stefano-k: wouldn't this close https://github.com/mate-desktop/mate-panel/issues/270 as well?

PR #289, opened 2015-03-05
Remove change_background_for_widget from GTK3 build (it is useless.)

rilian-la-te:
Working with style properties is useless, because GTK have only stubs here. 
GTK3 Background logic already in panel_background_prepare.


flexiondotorg:
@stefano-k Please review this for inclusion in MATE 1.10

Issue #269, opened 2014-12-22
Window List - stretched buttons on vertical panel


usv299792458:
In Gnome 2 I used to have this layout.
![gnome](https://cloud.githubusercontent.com/assets/1787194/5520527/93f9f95a-89ae-11e4-830b-1f6afe95a8ee.png)

Unfortunately, in MATE panel, in Window List, window buttons are stretched vertically.
![mate](https://cloud.githubusercontent.com/assets/1787194/5520528/9a8f63cc-89ae-11e4-8ca0-e5b896ac7a23.png)

My setup:
MATE version 1.8.1 (Debian Jessie).
Steps to reproduce:
Select TraditionalOK theme.
Remove all panels except one, make it right-sided, set width to 30 pixels, run an application.
##

mihai-bica:
I had the same problem and I run "mate-panel --reset" (which resets the panel layout) and the vertical windows list mate applet got fixed. I think that this is a configuration problem.

usv299792458:
Hello mihai-bica,
Thanks for sharing your results.
Unfortunately,
> mate-panel --reset
does not fix the issue for me: it just makes top and bottom panels as by default. When I reconfigure right panel, buttons are stretched again.
I also noticed that windows are reorganized from one-column to two-column layout (in fast intervals) when is panel is on the right, and this causes "trembling" of window buttons, which makes them completely unusable.
Igor.

mihai-bica:
Hi, you are right. mate-panel --reset does not fix the problem.

However, I downloaded the source code and did this modification:
mihai@mihai-vb ~/git/mate-panel $ git diff
diff --git a/applets/wncklet/window-list.c b/applets/wncklet/window-list.c
index 721aef2..53db860 100644
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -126,6 +126,7 @@ static void applet_change_orient(MatePanelApplet* applet, MatePanelAppletOrient
        wnck_tasklist_set_orientation (tasklist->tasklist, new_orient);
 #endif
 #endif
+       wnck_tasklist_set_orientation (tasklist->tasklist, new_orient); //MIHAI
        tasklist_update(tasklist);
 }
 
@@ -451,6 +452,8 @@ gboolean window_list_applet_fill(MatePanelApplet* applet)
 #endif
 #endif
 
+       wnck_tasklist_set_orientation (tasklist->tasklist, tasklist->orientation); //MIHAI
+
[...]

PR #300, opened 2015-05-01
Fixing  Window List - stretched buttons on vertical orientation #269

mihai-bica:
Hi,
This is fixing the bug #269.  I am wondering if it is OK to remove the `ifdef WNCK_CHECK_VERSION`. I did a grep for `WNCK_CHECK_VERSION` and didn't find it defined on my system.

The fix was tested on linux mint 17 and on debian 8. 

Thanks,
Mihai

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -122,7 +122,11 @@ static void applet_change_orient(MatePanelApplet* applet, MatePanelAppletOrient 
 
 	tasklist->orientation = new_orient;
 
-	wnck_tasklist_set_orientation (WNCK_TASKLIST (tasklist->tasklist), new_orient);
+#ifdef WNCK_CHECK_VERSION
+#if WNCK_CHECK_VERSION (3, 4, 6)
+	wnck_tasklist_set_orientation (tasklist->tasklist, new_orient);
+#endif
+#endif
 	tasklist_update(tasklist);
 }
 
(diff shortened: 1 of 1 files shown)


monsta:
Not sure how it was tested, but ```wnck_tasklist_set_orientation``` is available only in [libwnck >= 3.4.6](https://git.gnome.org/browse/libwnck/tree/libwnck/tasklist.c#n851) while GTK+2 build of mate-panel uses [libwnck 2.30](https://github.com/mate-desktop/mate-panel/blob/master/configure.ac#L81).

mihai-bica:
I tested this on branch 1.8 because it compiles easily on Mint 17. I cannot build and test the code on branch master because I have some dependency problems on my machine.

The function `wnck_tasklist_set_orientation` is available from libwnck 1: 
    mihai@mihai-vb ~ $ `strings /usr/lib/libwnck-1.so.22.3.31 | grep wnck_tasklist_set_orientation`
    `wnck_tasklist_set_orientation`
    mihai@mihai-vb ~ $
    It is also available in the headers of package `sudo apt-get install libwnck-dev`: 
    mihai@mihai-vb ~ $ `grep wnck_tasklist_set_orientation /usr/include/libwnck-1.0/libwnck/tasklist.h`
    `void wnck_tasklist_set_orientation(WnckTasklist *tasklist, GtkOrientation orient);`
    mihai@mihai-vb ~ $


However, the function `wnck_tasklist_set_middle_click_close` is available only from libwnck 3, so I reverted the change and kept the ifdef around `wnck_tasklist_set_middle_click_close`.

monsta:
I've checked it out - it's a [Debian-specific patch](https://sources.debian.net/src/libwnck/2.30.7-2/debian/patches/01_tasklist_orientation.patch/). It's also inherited in Ubuntu so it's the same in Mint 17.

We can't accept it upstream (unless every distro out there accepts the same patch).

mihai-bica:
Hi, I agree here:
1) Theoretically, the bug is already fixed for libwnck >= 3.4.6 . The problem appears if the OS has libwnck < 3.4.6
2) Should we close the bug? (I highly suspect that it will not reproduce on libwnck >= 3.4.6. Didn't test on libwnck >= 3.4.6)
3) This is not the place to commit this fix.

If I still want to have this fixed in Debian/Ubuntu/Mint I can either wait until they use libwnck >= 3.4.6 or commit this as a patch to Debian...

Thanks for your investigation.

monsta:
Don't look at wnck version 3.x.x, it's for GTK+3 build of MATE which is still experimental.
GTK+2 build of MATE (the default) uses wnck 2.30.x. It cannot use wnck 3.x.x - that is by design.

PR #304, opened 2015-05-21
GTK3: fix typo in gtk_style_context mate-custom-panel-background

NiceandGently:


rilian-la-te:
Why it is a typo? As far as I remember, I use this class as an internal and mark it by leading dash. Now it is confusung a bit.

NiceandGently:
using
.-mate-custom-panel-background { }
in a theme css is very unusual. I never saw that in any GTK3 application.
The right selector is  .mate-custom-panel-background { }
without a dash.
If you use that in forked themes, please correct that.

rilian-la-te:
It SHOULD NOT BE USED in themes.
Leading dash is for indicating that.
It is service style like recent -gtk-dpi CSS property (this purpose only
for monitor DPI).
Using this selector in themes will break transparency in panels.

Leading dash for that.

NiceandGently:
> It SHOULD NOT BE USED in themes. Leading dash is for indicating that. It is service style like recent -gtk-dpi CSS property (this purpose only for monitor DPI). Using this selector in themes will break transparency in panels.

I don't use use that in my  themes , maybe you ? 
But it can be used.

rilian-la-te:
Yes, but using it will break panel functionality.

PR #309, opened 2015-06-22
add missing help file for menubar

NiceandGently:


NiceandGently:
note:
There are some non working links inside the help, but i don't want unmark them as links to avoid editing po files. They didn't work with 1.8 already, so we can ignore this.
about connect-server applet:
We can add the missing file here or add the working 1.8 help for caja too and correct the link than in code.
I would prefer this solution because of old help for caja is translated.

Issue #280, opened 2015-02-20
GdkWindow type-hint property not set


Defavlt:
*If this issue has already been mentioned (didn't find anything searching the issues currently open), feel free to close this.*

[Taken from source](https://gitorious.org/projects/topmenu/issues/5):

    [..] unfortunately mate-panel does not set this hint via Gdk, rather accessing the X11 property directly.
    [I] Suspect the most promising one is patching mate-panel to properly set the GdkWindow type-hint property.

javispedro:
Just to clarify, I am talking about whether this entire function https://github.com/mate-desktop/mate-panel/blob/master/mate-panel/panel-xutils.c#L43 (panel_xutils_set_window_type) could be replaced with this plain Gtk+ call: https://developer.gnome.org/gtk3/stable/GtkWindow.html#gtk-window-set-type-hint . 

Not really a bug, but it does look like redundant code.

Will try to test when I find the time...

monsta:
Can anyone confirm if the proposed change fixes the Topmenu issue (and does not introduce any regressions here in mate-panel)?
If so, please make a pull request with the fix if you can.

javispedro:
Just made that pull request. I didn't find any regressions so far, but seeing this changes the order the hint is set (Gtk+ delays setting it until right before the map notify), it may (or may not) introduce race conditions in other window managers.

Other than that, both functions do exactly the same, and the Gtk+ one has been on Gtk+2 since 2.0, so it's time to get rid of the panel one (which is full of #ifdef GTK3s to begin with).

It also works with TopMenu.

PR #317, opened 2015-06-28
replace panel_xutils_set_window_type with gtk_window_set_type_hint

javispedro:
Both functions have exactly the same semantics, so prefer the Gtk+ builtin. That's one function less to worry about Gtk+2 vs 3 compatibility (or Wayland).

In addition, this also fixes issue #280.


flexiondotorg:
@monsta Can you test this please?

monsta:
Looks alright.

PR #318, opened 2015-06-29
Fix Force-Quit panel applet to avoid hang with GTK3

City-busz:
This patch implements handling of XInput2 extension events in the applet.
Without this patch, the applet fails to recognize such events (either mouse
click for killing application, or escape key for exiting), hence freezing the
desktop.

Also fix the event mask in call of gdk_device_grab() for keyboard.

Author: Sébastien Villemot <sebastien@debian.org>
http://bugs.debian.org/698740

Original commit in gnome-panel:
https://git.gnome.org/browse/gnome-panel/commit/?id=5a8583a77aab91fe1a369becc9453b27bf30ee70

### DIFF
--- a/mate-panel/panel-force-quit.c
+++ b/mate-panel/panel-force-quit.c
@@ -32,6 +32,8 @@
 #include <X11/Xlib.h>
 #include <X11/keysym.h>
 
+#include <X11/extensions/XInput2.h>
+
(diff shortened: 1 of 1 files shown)


NiceandGently:
confirmed, this PR fixes the issue :)

thank you

PR #333, opened 2015-08-02
UI GTK3: use a size request for scrolled window in panel-run dialog

NiceandGently:
- this fixes program list in GTK3


NiceandGently:
Works well with gtk2 and has the same size.
Note, using expands or fil doesn't work here, because of scrollbar changes with gtk+-3.16

PR #337, opened 2015-08-06
Na tray gtktable + gtkgrid patch 1

braikar:
It makes very little difference, I added the ifs for gtkgrid


braikar:
Sorry for the mistake, still getting used to the system..

PR #339, opened 2015-08-10
Fix missing gettext setup for panel applet common parts

obache:
This is required for following situations:
1. MATELOCALEDIR is not a default location
2. applet's textdomain is not "mate-panel"

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -2005,6 +2005,7 @@ mate_panel_applet_init (MatePanelApplet *applet)
 	applet->priv->size   = 24;
 
 	applet->priv->panel_action_group = gtk_action_group_new ("PanelActions");
+	_MATE_PANEL_APPLET_SETUP_GETTEXT(FALSE);
 	gtk_action_group_set_translation_domain (applet->priv->panel_action_group, GETTEXT_PACKAGE);
 	gtk_action_group_add_actions (applet->priv->panel_action_group,
 				      menu_entries,


monsta:
Hmm... I see this is already called (with either ```TRUE``` or ```FALSE``` args) in https://github.com/mate-desktop/mate-panel/blob/master/libmate-panel-applet/mate-panel-applet.h - isn't that enough to handle the described situations?

obache:
No. It only affect to applet own parts, not for common parts in libmate-panel-applet.
So It is true only for applets bundled with mate-panel, because GETTEXT_PACKAGE="mate-desktop" for such applets.

But for example, applets from "mate-applets", GETTEXT_PACKAGE != "mate-desktop", and "mate_panel_applet_init()" in libmate-panel-applet/mate-panel-applet.c is built as a part of "mate-desktop", so GETTEXT_PACKAGE="mate-desktop" for the context, then "bindtextdomain" for "mate-desktop" will not be done for such applets and translations for common ui parts (ex, "Remove From Panel") will not be appeared unless MATELOCALEDIR points to system default localedir.

infirit:
Let me understand this, Your distro's install gettext files in one place (lets say /usr/share/locale) which includes libmate-panel-applet. Then you install an applet with the gettext files somewhere else?

If not please explain what you are doing and where things are installed..

obache:
No, counter situation, gettext files of mate-panel are not installed in default place of system gettext (say /usr/share/locale) and applet's textdomain is not "mate-panel".
It is the relevant condition to reproduce the issue.

lukefromdc:
Is this something we still need or should it be closed? This has sat open for over five years

obache:
@lukefromdc Steps to reproduce the behaviour
1. build MATE packages with something like `--prefix=/opt/mate` (locale data files will not be installed into system standard locale data directory)
2. start MATE desktop with panel appletes ( using libmate-panel-applet but not a part of mate-panel package)
3. `_Remove From Panel` menu item is translated? ( with `ja` locale, it should be `パネルから削除(_R)` )

If it is fine now, this issue is already fixed with different way (I'm not using MATE so long time).

lukefromdc:
That's not something I can test myself, as while I DO run MATE (it's my main DE), I do all my testing on the same machine I use for everything else and don't have any virtual machines set up. On top of all else I have unused locales purged to limit size of backup filesystem images that are taken at intervals and used to restore the system if things go really bad from developing on a production machine-or just one of the bad updates that can happen on Debian Sid.

Does anyone on the team know if we still need this? I have no way to determine it for myself

raveit65:
> If it is fine now, this issue is already fixed with different way (I'm not using MATE so long time).

We can re-open it when the issue from 5 years ago is reproducible again.

PR #347, opened 2015-09-05
This branch enables the use of custom themes for the panel and applet menus/widgets in gtk3 builds

lukefromdc:
This fork extends the .mate-panel-menu-bar selector in gtk3 themes when built with gtk3. In mate-panel built from master at this point, panel main menu themes using gtk3 can be customized with the .mate-panel-menu-bar .menu selector in the gtk3 theme. This fork extends this to the panel context menu and applet menus controlled by applet.c or mate-panel-applete.c . Also included is support for transparency in these themes. In all cases but the calendar that require no additional theming elements, as .mate-panel-menu-bar is applied to the toplevel widget along with code from this element in the panelplug section of mate-panel-applet.c to get/set the screen for RGB or RGBA. In the case of the calendar this made the background transparent with other themes, so it is set to follow the #MatePanelPopupWindow widget name and if not present in the theme default to the theme background color.

One commit that others might want to revert or to rewrite so as to allow user choice in the matter is disabling week numbers in calendars. With transparent themes they become confusing but they have been in GNOME 2's calendar as long as I can remember. An option to switch them on or off would be better.

If this is merged similar forks of mate-panel-applets and other mate packages affecting tray or applet menus are also waiting, but for convenience of other devs I am only creating this one pull request for discussion. 

This fork is kept current with github.com/mate-desktop/mate-panel as much as possible


monsta:
@raveit65: what do you think about this?

raveit65:
I like the idea to to control the applets and panel context menus too with a theme, different from general theme settings.
Same for transparency support for menus.
I will test it over the weekend.
But what i don't like are the commits...........This is a Hack!
A Hack implement that it is not the best soution and it could be do better ;)

raveit65:
I don't think we should disable week numbers in general, here we ned a setting for it if we wand to merge this.

lukefromdc:
I've been using this every day since I wrote it, and before that a previous, much clunkier version
since March. I had an invitation over the forums to offer this and figured it would make
sense to put in this pull request. I will try to find a way to make the calendar week numbers
selectable as that has been asked for. 

The code here is the best I was able to do at the time. Does someone have a better idea
about how to implement this?  For my own use the themablity is mandatory, as I am 
using MATE as a (much) higher performance replacement for Cinnamon.

It was on these programs that I had to learn to hack GTK/ C coding in the first place, so if
it looks rough that is way. I was very motivated to get this themability and have had it working
well on my own machines for many months.

lukefromdc:
I just reverted the committ that hardcoded calendar week numbers not to show, turns out org.mate.panel.applet.clock lists this as a key. Will rebuild and test this momentarily. The key does not show in dconf-editor but can be user-set with gsettings from the command line.

lukefromdc:
I just found clock.c is not using the value set in org.mate.panel.applet.clock for show-week, I've got more work to do to get that setting used.

lukefromdc:
The schema itself can control whether or not the week numbers show, but running org.mate.panel.applet.clock:/usr/share/glib-2.0/schemas/  show-week-numbers false
or
org.mate.panel.applet.clock:/usr/share/glib-2.0/schemas/  show-week-numbers true
has no effect. Only editing the schema, then recompiling schemas has an effect at all,
and the value does not show up in dconf-editor even though it has always been in this
relocatable schema. There should be a trivial fix for this, I just don't know what it is

lukefromdc:
The just-added commits enable selecting calendar week number display from the clock preferences GUI. Just tested this, it works on my machine here. Feel free to use this with or without any of my other work.

raveit65:
Luke,
can you please give some style selector examples how do you select the menus, panel-context-menu and applet-menus?
ie.
.mate-panel-menu-bar PanelMenuButton .menu .menuitem,
.mate-panel-menu-bar .menubar .menu .menuitem {
    background-color: shade (@theme_bg_color, 1.2);
}
which styles both menus.
With Gtk3-inspector i can't select the context-menus for example.
This will make it easier to review you PR.

about clock changes:
Using the the new checkbutton in clock pref menus doesn't have any effect.
I suggest to open a different PR for this. You can create another branch from master in you r git repo for this.
Please use 'git rebase -i' (interactive + use vim editor) to clean this PR. With that you can delete commits without having a revert commit, and change 2 commit in one which is needed for f2e08f2 and ca21ece for example.
Or reorder the commits, see http://stackoverflow.com/questions/4981061/how-to-re-order-commits-in-git-non-interactively
Start with the the latest commit before your PR in your repo.
'git rebase -i 9d76273d550a6ec24e716beb63e3c31c6f71552b'

For the future i suggest to create a different branch for PRs, to avoid to have the merge commits from upstrem master in the PR.
If you have questions please ask.

lukefromdc:
Here's how they are done in my theme, the code below sets the menu theme on 
panel main menus  on current master, on my version it also controls the panel
context menu and all applet menus controlled by libmate-panel-applet. When the 
rest of my forks plus my network-manager-gnome fork are used too it controls every
menu that comes from the panel except the window list button right click menus,
I think the WM makes those but am not sure. I have attached my entire panel.css
so you all can look at it too.

/*Keep box-shadow off anything from the panel*/

.mate-panel-menu-bar .window-frame.csd.popup{
		box-shadow: none; 
}

.mate-panel-menu-bar .menu,
.mate-panel-menu-bar .menu,
.menu#cpufreq-menu /*For hacked Ubuntu indicator-cupfreq until it catches up */
 {
	background-color: alpha(black,0.35);
	border-image: none;
	border-style: solid;
	border-color: #888a85;
	border-width: 3px 3px 3px 3px;
	border-radius:10px;
	font: normal;	
	padding: 0px 0px 0px 0px;  /*Kill double borders on some menus */
}

lukefromdc:
The clock changes have worked in every machine I tested them in. I wonder what you did differently? I am using gtk3.18 which is now in Debian Unstable, the code is just a cut and paste from the show seconds box and should work the same way. If you missed the final commit you have a cut and paste error in the prefs menu and it won't open. The last commit fixes that.  Maybe the whole set of these didn't work right when applied to the master branch by itself?  I will set up another fork of master with just these commits and see what happens.

On my machines, if you right click on the clock, select "preferences" and in "general" you should get an entry for "show week numbers in calendar" which is checked by default. Week numbers should show up in the calendar. Uncheck it, open the calendar again and they are gone. You can even check and uncheck with the calendar open, but that will cause the window to be redrawn high over the panel until it is closed and then re-opened, at which point it appears in its normal position.

raveit65:
Opps, i did a mistake with defining the clock patch in spec file for the rpm.
Toggling the weeks in clock prefs works :)
Same with using dconf-editor.
So this part of the PR is fine.
I need to test more the rest, it looks that all is different depends on which theme is used :)
The selector .mate-panel-menu-bar .window-frame.csd.popup works for all menus i have test, so i think your changes are OK.

I can clean up your PR this time, but for the future this is not acceptable.
Using git at command line is not difficult and there a lot of good how-tos at the web.
The basic beginning to have the repo local is always
git clone <ssh link-which you find on githup page>

If it is secure for you visit us at irc freenode channel mate-dev,
here we can give you help.
You're welcome

raveit65:
@monsta 
PR looks good from my side, please can you review it from code side?
... before is start to rebase the PR
PS: for some reasons the gsettings key is not always shown in dconf-editor,
but same for all other gsettings keys of clock applet....sometime you see the keys...sometimes not.... weird.
So it looks more like an general issue.

lukefromdc:
OK, I will have to take some time to figure out how to rebase the others.
Probably what I would have to do is first find out how to get a "clean" rebase,
then rebase the other packages, them submit the pull requests for them

Question: I suppose those should be separate branches used only for the PR, then
not updated?  I normally merge all "master" updates back into my own forks, then download
the changed files and build. Should the running updates stay on those forks, with 
the PR's being made from copies of those forks that are not then updated with the 
changes in master?   If these changes are merged, obviously there is no further need for 
these forks and I would not have to deal with them unless I came up with something totally
new, which would then base on master.

I've got debian packages built from snapshots of all this at 

https://archive.org/details/DebianPackagesForMate-desktopWityGtk3AndCustomPanelTheme

The GTK theme I use in there and offers a quick way to test the theming and pull code
templates that can be used to write other themes. Just untar it and drop the theme
directory into /usr/share/themes if not using a Debian based distro.

I don't see any security issues concerning freenode, but I have never used it and know
nothing about chat or IRC. I did not even have network access at home until a couple
years ago. I cut my teeth on doing media work offline and then uploading it via first
public access computers and then public wi-fi access points.

raveit65:
I've clean up the the PR local.
@monsta 
need a feedback before merging

raveit65:
merged

Issue #305, opened 2015-05-25
GTK+ 3: panel main menu don't close properly


raveit65:
The panel freezes after adding a launcher form menu or desktop, in result most panel elements can not be executed until the panel is restarted by a 'killall mate-panel'.
Backtrace:
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/gtk3_mate-panel_while_adding_launcher.txt
Also the panels spams the xsession.errors.log during every session start.

```
(mate-panel:1620): GdkPixbuf-CRITICAL **: gdk_pixbuf_get_height: assertion 'GDK_IS_PIXBUF (pixbuf)' failed

(mate-panel:1620): Gtk-CRITICAL **: gtk_widget_get_preferred_width_for_height: assertion 'height >= 0' failed

(mate-panel:1620): GdkPixbuf-CRITICAL **: gdk_pixbuf_get_height: assertion 'GDK_IS_PIXBUF (pixbuf)' failed

(mate-panel:1620): Gtk-CRITICAL **: gtk_widget_get_preferred_width_for_height: assertion 'height >= 0' failed

(mate-panel:1620): GdkPixbuf-CRITICAL **: gdk_pixbuf_get_height: assertion 'GDK_IS_PIXBUF (pixbuf)' failed

(mate-panel:1620): Gtk-CRITICAL **: gtk_widget_get_preferred_width_for_height: assertion 'height >= 0' failed

(mate-panel:1620): GdkPixbuf-CRITICAL **: gdk_pixbuf_get_height: assertion 'GDK_IS_PIXBUF (pixbuf)' failed

(mate-panel:1620): Gtk-CRITICAL **: gtk_widget_get_preferred_width_for_height: assertion 'height >= 0' failed

(mate-panel:1620): GdkPixbuf-CRITICAL **: gdk_pixbuf_get_height: assertion 'GDK_IS_PIXBUF (pixbuf)' failed

(mate-panel:1620): Gtk-CRITICAL **: gtk_widget_get_preferred_width_for_height: assertion 'height >= 0' failed

(mate-panel:1620): GdkPixbuf-CRITICAL **: gdk_pixbuf_get_height: assertion 'GDK_IS_PIXBUF (pixbuf)' failed

(mate-panel:1620): Gtk-CRITICAL **: gtk_widget_get_preferred_width_for_height: assertion 'height >= 0' failed
```

Not shure if both issues has the same cause.
##

raveit65:
[...]

PR #351, opened 2015-09-25
eliminate criticals

dnk:
gdk_window_set_geometry_hints accept 0 as flags when geometry is null
return 1px width/height while image is not loadded


dnk:
reduce warnings for https://github.com/mate-desktop/mate-panel/issues/305

raveit65:
Confirmed, the commit fixes the warnings, thank you.

raveit65:
i see no regressions after a day.

Issue #84, opened 2013-03-13
Panel objects get resorted if there is no space left


LeoUnglaub:
If you connect your laptop to a beamer with a quit low resolution you have the problem that some icons in the panel get rearanged/reordered. Even if they are locked on to the panel.

![Auswahl_012](https://f.cloud.github.com/assets/734684/253674/f480505c-8bdc-11e2-8604-adf12f1f96a5.jpg)

mirkoperillo:
confirm this issue on ubuntu mate 14.04 (mate 1.8.1) with my 24'' external monitor

V3n3RiX:
no fix for this yet? even in dev branch, we can backport it to older versions

opnoack:
This is a really annoying bug. Is this going to be fixed anytime soon?

janmisek:
+1

harlock974:
Issue still occurs in Mint LMDE2 (mate-panel 1.8.1+dfsg1-3). I still use the same workaround described on 30 dec 2013.

Skydive:
Still occurring. +1
Will this be fixed soon? The workaround does not work for me.

clefebvre:
Similar bug in GNOME2 https://bugs.launchpad.net/ubuntu/+source/gnome-panel/+bug/44082

clefebvre:
Patch reported to fix the issue in GNOME2 https://bug314235.bugzilla-attachments.gnome.org/attachment.cgi?id=51170

clefebvre:
Upstream bug report https://bugzilla.gnome.org/show_bug.cgi?id=341441

clefebvre:
code commit in GNOME2: https://github.com/GNOME/gnome-panel/commit/8871b8e00929df310936f0d6ba8b7b88044e2e4f

clefebvre:
Sadly it looks like this is already in mate-panel...

opnoack:
Is there any update so far? I'd really like to make MATE my main desktop but with this bug it's annoying to work with external displays.

milankni:
It seems to be related to issue: align applets with relative position #62

monsta:
Well, since there were no more comments since @clefebvre's commit, I guess it has been indeed fixed.

opnoack:
In which version of MATE will it be included?

monsta:
Already released in 1.12.

cpalaiop:
About:The MATE Panel 1.12.1
[...]

PR #355, opened 2015-10-15
Don't change position/right-stick of locked applets

clefebvre:
Until position calculations are refactored to fix the issue of the panel applets
getting reordered on resolution changes don't save position/right-stick on locked applets.

This partially fixes https://github.com/mate-desktop/mate-panel/issues/84 (it fixes it as long as applets remain locked. Unlocked right-stick applets are still vulnerable ... we need a much bigger solution for this, possibly redesign the way we handle coordinates).


raveit65:
+1 from me
I think the most users have locked clock and systray from origin configuration, so this PR is an improvement.

PR #362, opened 2015-11-04
Make MAX_ITEMS_OR_SUBMENU configurable.

j-rueger:
Hi everyone!

It annoyed me for some time that I could only have 8 bookmarks in the places menu before a submenu was created, even though my screen resolution would allow for quite a few more...
Upon investigating, I could only find this: http://ubuntuforums.org/showthread.php?t=1066964
I was thinking "This should really be a setting in dconf or something". Some tinkering later, I got it working and here it is! ;)

Please review. I'm sure I made some stupid rookie mistake or two.
For example the new gschema.xml; I had to manually copy it to /usr/share/glib-2.0/schemas and run glib-compile-schemas after make install. What's the right way to fix that?
Do I need to check the GSettings pointer for NULL before using g_settings_get_int() ?
If the schema does not exist, the pointer is NULL and I think get_int() returns 0 or something negative, so you get a submenu even for a single bookmark. It does not crash or anything, so do I need to handle this case?

Thanks guys!

### DIFF
--- a/data/org.mate.panel.menubar.gschema.xml.in
+++ b/data/org.mate.panel.menubar.gschema.xml.in
@@ -45,5 +45,10 @@
       <summary>Menu items icon size</summary>
       <description>Set the size of icons used in the menu. The panel must be restarted for this to take effect.</description>
     </key>
+    <key name="max-items-or-submenu" type="i">
+      <default>8</default>
+      <summary>Threshold of menu items before submenu is created</summary>
+      <description>Maximum number of menu items (i.e. bookmarks) that are displayed without being put in a submenu.</description>
+    </key>
   </schema>
 </schemalist>
--- a/mate-panel/panel-menu-items.c
+++ b/mate-panel/panel-menu-items.c
@@ -75,6 +75,7 @@ struct _PanelPlaceMenuItemPrivate {
 
 	GSettings   *caja_desktop_settings;
 	GSettings   *caja_prefs_settings;
+	GSettings   *menubar_settings;
 
 	GtkRecentManager *recent_manager;
 
@@ -351,7 +352,7 @@ panel_menu_items_create_action_item (PanelActionButtonType action_type)
 }
 
 static void
-panel_place_menu_item_append_gtk_bookmarks (GtkWidget *menu)
+panel_place_menu_item_append_gtk_bookmarks (GtkWidget *menu, int max_items_or_submenu)
 {
 	typedef struct {
 		char *full_uri;
@@ -378,7 +379,7 @@ panel_place_menu_item_append_gtk_bookmarks (GtkWidget *menu)
 
 	/* We use a hard limit to avoid having users shooting their
 	 * own feet, and to avoid crashing the system if a misbehaving
-	 * application creates a big bookmars file.
+	 * application creates a big bookmarks file.
 	 */
 	for (i = 0; i < MAX_BOOKMARK_ITEMS; i++) {
 		char      *contents;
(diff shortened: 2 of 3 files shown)


flexiondotorg:
@monsta @raveit65 Can you take a look at this for inclusion in 1.14?

j-rueger:
Hi tarakbumba and thanks for your review!
I added the line to configure.ac as a new commit, hope that is acceptable. Should I merge the current upstream master into my repo?

Regarding menubar.gschema.xml, I thought about that back then and decided against it. It does not really fit there thematically (imho) and all the other files are divided into menubar and menuitems respectively, so I went with that.

Also sorry for the late answer, I had a busy week ;)

raveit65:
Looks good for me, tested with gtk3 build.
Why not naming the schema 'locationbar', because it controls only menuitems there?
So it's more clear what it does.
The menubar schema have keys for menuitems too, so it is bit confusing seeing a schema menuitem in dconf-editor without the keys for the menubar.
Beside from this i can merge the PR.
@monsta 
@j-rueger 
Any thoughts.

> Should I merge the current upstream master into my repo

no need of it, your PR works good on top of master. I will rebase it against master before the 'git merge --ff-only' command.

j-rueger:
Looks like a friendliness-to-user vs. friendliness-to-programmer issue. I went with the latter because dconf is kind of under-the-hood.

Of course I can change that if you prefer. How do you like the attached patch? (renamed it "...places-menu.gschema.xml" and so on)

How should I handle that in case you approve? Another patch on top of my repo? Can you apply the patch file as-is? Should I make another (clean) PR?

Thanks!

raveit65:
Well, you can use 'git rebase -i *commit-id- before-your-commits*' (interactive). Than you can edit/reword/squash  every commit from your PR, or sort those.
Squash means put this commit into previous commit.
After all is fine local use 'git push -f' (force)
Than this PR will be updated automatically.
Don't worry this is easy :)
Hope you know basic vim commands.
Doing this would help us a lot, otherwise we have to do this work for you :)

Thank you

monsta:
I'm a bit late and probably missed something in the discussion... but why not simply make a new key in ```org.mate.panel.menubar``` schema and so make the things simpler?

For example, look at https://github.com/mate-desktop/mate-panel/pull/406 where @XRevan86 added ```icon-size``` and ```item-icon-size``` keys to that schema. (You'll need to have mate-panel 1.13.1 or just current git master branch to have these changes.)

j-rueger:
@monsta
I felt it didn't fit in with the other options in menubar. With Revans commits now however and 'cause everyone and their pony suggested I should put it there... Well here you go ;)

I also merged the current master and force-pushed to my repo.

OK, guys I hope this is it!

raveit65:
Thank you for your patient with us :)

j-rueger:
@monsta @raveit65 Of course! Thank **you** for all your work keeping my very favourite DE alive and kicking!

monsta:
Thanks. Some notes:

- there's no need to check for menubar schema as it's already installed along with panel itself;
- even if it wouldn't exist, there are no checks before ```g_settings_get_int``` calls;
- there's a possible memory leak due to ```menubar_settings``` object not being unreferenced;
- key type should really be unsigned;
- no need to keep the old ```MAX_ITEMS_OR_SUBMENU``` define around now.

I've addressed these issues in https://github.com/mate-desktop/mate-panel/commit/db46e948f57ad16170afeaf72d418c508ef63d1e.

PR #374, opened 2015-11-20
Don't set the panel transparent by default

City-busz:
This fixes the background for GTK+ 3 themes which don't support mate-panel explicitly. Themes could still overwrite the background if they want in the usual way.

To get back exactly the previous state, you can add the following lines to your themes:
Before other declarations: `.mate-panel-menu-bar {background: transparent;} `
After other declarations (these background settings were ignored previously): `PanelToplevel, PanelMenuBar, PanelSeparator {background: transparent;} `

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1979,33 +1979,39 @@ mate_panel_applet_setup (MatePanelApplet *applet)
 void _mate_panel_applet_apply_css(GtkWidget* widget, MatePanelAppletBackgroundType type)
 {
 	GtkStyleContext* context;
-	GtkCssProvider  *provider;
 
 	context = gtk_widget_get_style_context (widget);
-	gtk_widget_reset_style(widget);
+	gtk_widget_reset_style (widget);
 
 	switch (type) {
 	case PANEL_NO_BACKGROUND:
-		gtk_style_context_remove_class(context,"mate-custom-panel-background");
+		gtk_style_context_remove_class (context, "mate-custom-panel-background");
 		break;
 	case PANEL_COLOR_BACKGROUND:
 	case PANEL_PIXMAP_BACKGROUND:
-		provider = gtk_css_provider_new ();
-		gtk_css_provider_load_from_data (provider,
-						".mate-custom-panel-background{\n"
-						" background-color: rgba (0, 0, 0, 0);\n"
-						" background-image: none;\n"
-						"}",
-						-1, NULL);
 		gtk_style_context_add_class (context, "mate-custom-panel-background");
-		gtk_style_context_add_provider (context,
-						GTK_STYLE_PROVIDER (provider),
-						GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
 		break;
 	default:
 		g_assert_not_reached ();
 		break;
 	}
+}
+
+void _mate_panel_applet_prepare_css (GtkStyleContext *context)
+{
+	GtkCssProvider  *provider;
+
+	provider = gtk_css_provider_new ();
+	gtk_css_provider_load_from_data (provider,
+					 ".mate-custom-panel-background{\n"
+					 " background-color: rgba (0, 0, 0, 0);\n"
+					 " background-image: none;\n"
+					 "}",
+					 -1, NULL);
+	gtk_style_context_add_provider (context,
+					GTK_STYLE_PROVIDER (provider),
+					GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
+	g_object_unref (provider);
 }
 #endif
 
@@ -2045,10 +2051,10 @@ mate_panel_applet_init (MatePanelApplet *applet)
 	gtk_widget_set_visual(GTK_WIDGET(applet->priv->plug), visual);
 	GtkStyleContext *context;
 	context = gtk_widget_get_style_context (GTK_WIDGET(applet->priv->plug));
-	gtk_style_context_remove_class (context,GTK_STYLE_CLASS_BACKGROUND);
 	gtk_style_context_add_class(context,"gnome-panel-menu-bar");
 	gtk_style_context_add_class(context,"mate-panel-menu-bar");
 	gtk_widget_set_name(GTK_WIDGET(applet->priv->plug), "PanelPlug");
+	_mate_panel_applet_prepare_css(context);
 #endif
 	g_signal_connect_swapped (G_OBJECT (applet->priv->plug), "embedded",
 				  G_CALLBACK (mate_panel_applet_setup),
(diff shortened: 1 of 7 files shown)


lukefromdc:
I just tested this, it needs more work. Panels background color setting is buggy with this in that the theme background color cannot have an alpha value or the gtk theme background color is shown under it. On the other hand, custom panel colors set in the properties dialog MUST have an alpha value (even barely under 100%) unless no panel background is set in the gtk them-or the panel comes up entirely transparent as soon as it is restarted. Widgets such as buttons, the tray, and other panelplugs are exceptions and get their intended backgrounds. Changing anything in the properties dialog will bring up the intended color. Also, if no panel background is set in the gtk theme, custom panel colors can be 100% opaque.  

Image backgrounds act the same as 100% opaque colors: they come up transparent on restart under part of the panel if any background is set in the gtk theme.

No matter what is in the theme, if the panel is started with a custom background and an attempt is made to select the system theme background from the panel properties dialog, the default GTK theme background color is applied to the panel background. Any different background set in the theme is ignored. Restarting the panel with the system theme selected will then bring up the correct background as set in the theme. Something is being read at the wrong time and/or there is a conflict of the non-removed system css provider with the custom one, a conflict which is resolved by a re-read of the custom value.

lukefromdc:
The bugs I encountered here are mostly but not entirely this one:
https://github.com/mate-desktop/mate-panel/issues/375
There are some differences with this patch that remain unique additional issues:
1: Setting a transparent or partially transparent panel background in the GTK theme no longer works.
The theme's default gtk background color, not the desktop background, is under any background set in the theme, even using the selectors you advised. My guess is this is the gtk window background and that needs to be transparent even if the panel is not.
2: Switching back to the theme bg from a custom bg requires a panel restart to fully apply it

The bugs related to issue 375 might be gtk3.18 or gtk3.16/18 only bugs, I did not have panel background image issues in Spring 2015 but was using gtk3.14 in those days.

lukefromdc:
https://github.com/lukefromdc/mate-panel/tree/set-default-background-hardcoded2

Is a simpler way to do this, gives about the same appearance in Adwaita and Raleigh without making issue #375 any worse. Note that no hardocded background will match the foreground text of all themes-and Adwaita has both black and white text and has issues with this. Dark themes without MATE support do too. A full fix would be to set an entire default theme for the panel with priority fallback, but that's a lot more data to hardcode or else a default theme file could be installed in /etc and the provider set to read it. This would affect the panel and all of it's menus, resulting in a default "shell theme" rather like they way GNOME keeps the shell theme separate from the gtk theme. it would be overriden by any panel theme set in the GTK theme, as is the case with both your PR and my proposed alternative.

Right now a custom panel background with an alpha value so it can start properly is the only workaround for the non-MATE themes. I will see if I can dig into #375 myself, it appears to be a timing issue.

lukefromdc:
Update: all of the issues with custom backgrounds are a gtk3.18 only problem, probably the same or a similar bug as that which causes the cpufreq applet to turn transparent when the governor is changed by a user, again only with gtk3.18.

raveit65:
By the way, offtopic on
see https://blogs.gnome.org/mclasen/2015/11/20/a-gtk-update/
If this comes all themes needs a complete rewrite and code needs adjusted too,
or the whole universe use adwaita theme.
offtopic off
Sorry guys but the bg stuff is far away from my interest now.

lukefromdc:
This will certainly force me to rewrite my theme-and means I will have to download
and build gtk3 from source (done this before) to do so. No way in hell I'm giving up
the theme I've used since 2007.

On a plus note, once this is done it will fix a number of ugly bugs such as not being able
to do alternating colors in treeviews. If this works I should be able to more closely match
these to how they looked in gtk2.

lukefromdc:
I just built and installed gtk3.19.2 with these changes, looks so far like I got exactly NO 
breakage in my theme, just a few more options I am about to play with, particularly
to see if I can "stripe" that treeview as the documentation now states. 

Caja is working fine, so is pluma, the panel, etc. This is without changing anything yet.
More changes to come in this code but it looks to me like it will be nowhere near the
negative impact we thought, given that the transition to using "css nodes" was said to 
be "mostly done" but 3.19.2, which I just installed. Alas, no bugfix for #375 in it either...

lukefromdc:
Update: I've got quite a bit of work to do, but nothing I can't handle. Gtk did not
reload right away, so once I have my results I will upload them to Archive. I do not
understand the build system enough to do a theme package on Github right now.

raveit65:
A quick test with BlueMenta and Blue-Submarine themes shows me any breakage, tested in f23 (gtk+-3.18.x).
So it seems that no adjustmenent are needed with mate-themes, but i need to test the other themes.
But why selector 'background'?
I never saw that in themes, only background-color or background-image

Bad News: it seems that gtk3-3.18.5-3.fc23 breaks any 'hover button/menuitem state' (button:hover or menuitem:hover).
I'm realy tired of following/using gtk+ toolkit
Only one thing is shure.....the next breakage

raveit65:
Normaly, the term background is used like this:
```
.background {
                          background-color: ....;
                          color: ..........;
                          etc..........;
}
```

raveit65:
correction: the breakage of 'hover button/menuitem state' is a regression since gt+-3.18.3
Reproducible with adwaita theme too.
.......LOL

raveit65:
@City-busz 
again, why {background: transparent;} 
and not
{background-color: transparent;}
?

lukefromdc:
I tried both with the same results: did not work. This is actually the first
time I've seen it specified like that.

All these specs are going to change with GTK 3.20, I've been playing with
gtk3.19 and had to rebuild almost everything in my theme. Details in the 
forums shortly.

lukefromdc:
Here is my forum article on the gtk3.19 situation, including a link
to my newly ported theme.

http://forums.mate-desktop.org/viewtopic.php?f=16&t=3610

raveit65:
> I tried both with the same results: did not work. This is actually the first time I've seen it specified like that.

I think @City-busz posted somthing wrong here.
gtk3-inspector shows me those style classes now
.background.mate-panel-menu-bar.gnome-panel-menu-bar
So using 
```
.mate-panel-menu-bar {
    background-image: none;
}
```
before other settings for the panel should override .background.
I used in mate-themes always something similar at the beginning, see 
https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/BlueMenta/gtk-3.0/mate-applications.css#L344
For this reason the PR doesn't have any affect for mate-themes.

Btw, i noticed that the issue ' applet crashes if moving the transparent slider to fast' is also fixed with this PR.

raveit65:
no further comunication anymore.....i will merge the PR

monsta:
> Btw, i noticed that the issue ' applet crashes if moving the transparent slider to fast' is also fixed with this PR

No it's not :smile:

raveit65:
> No it's not :smile:

set bg color --> logout/login --> all fine here

PR #383, opened 2015-12-16
GTK3 workspace-switcher: add fallback color

City-busz:
Without setting a color, the actual workspace is not highlighted, if the theme does not set WnckPager:selected explicitely (e.g. with Adwaita).

mate-themes are unaffected by this change.


monsta:
Thanks, fix confirmed with Adwaita. Merged in https://github.com/mate-desktop/mate-panel/commit/956d36c07bf236619dc6a13a0efd41f864943520.

PR #384, opened 2015-12-25
Fix rendering of user backgrounds in gtk3.18+

lukefromdc:
User set panel backgrounds other than colors with an alpha value fail on restarting the panel using gtk3.18 or 3.19. This makes picture and fully opaque color backgrounds set by the user work again and for some reason does not break transparent color backgrounds either. Tested with gtk3.19.5, should work with gtk3.18 as this bug was exactly the same but needs testing with it. 

Note that ONLY the final commit here represents a changed file from mate-desktop/master, other was just fast forwarding my master tracking branch to today

### DIFF
--- a/mate-panel/panel-background.c
+++ b/mate-panel/panel-background.c
@@ -554,7 +554,11 @@ panel_background_composite (PanelBackground *background)
 		break;
 	}
 
+#if GTK_CHECK_VERSION (3, 0, 0)     /*panel user background fix for gtk3.18+*/
+	background->composited = FALSE; /*this is actually WRONG but fixes rendering of user selected color BG*/
+#else
 	background->composited = TRUE;
+#endif
 
 	panel_background_prepare (background);
 


monsta:
I wonder what will happen with 3.14... I don't have such installation though, so maybe @raveit65 could test it.

lukefromdc:
I'm about to find out. I will download and build a gtk3.14 tarball. If I don't get
build errors I should know shortly.

lukefromdc:
Just tested this with GTK 3.14, no obvious problems.

raveit65:
I can't reproduce the issue with gtk+-3.16.
But yes, with 3.18 the panel is always transparent after a restart until i move the slider a bit.

lukefromdc:
This issue began with gtk3.18, the gtk3.14 test was to ensure my fix didn't create
new issues with older GTk versions.

raveit65:
Thanks,
i rebased and merged the PR.
merged https://github.com/mate-desktop/mate-panel/commit/410de37dad6f46c6d28a3ebaf832efa1f31e3141

Issue #386, opened 2015-12-27
gtk3.19 now breaks calendar window position


lukefromdc:
On 12-22-2015, GTK3.19 received this commit which causes both the calendar window and the panel properties dialog to always come up at the top left of the screen, ignoring the functions that normally position them:

https://git.gnome.org/browse/gtk+/commit/?id=2438a06d54636e5074c29bd696e3e81e90288b8dRevert "window: Remove suspicious branch"
This reverts commit 67ab00e01ec06d7b05a82c3d80b235ac6c4abfd2.

See https://bugzilla.gnome.org/show_bug.cgi?id=759705
for a case where this makes a difference.

If I read that right, the features they removed were meant to support  gtk_widget_realize , which only is used in na-tray-applet.c but that is not where the problems are. I send an email to the author of this commit with no response, it appears this will NOT be reverted, so barring some later change before GTK 3.20 fixing this we will have to find a way to work around this. Menus position fine, that code might offer clues. I have not been able to fix this in initial experiments, so for now I am locally reverting the offending commit in my gk3 buiids. Needless to say, MATE cannot depend on a patched version of GTK3 a year from now.

lukefromdc:
This function was added to gtkwindow.c

+  if (need_resize)
+    {
+      GtkWindowGeometryInfo *info = gtk_window_get_geometry_info (window, TRUE);
+      GtkAllocation allocation = { 0, 0 };
+      GdkRectangle configure_request;
+      GdkGeometry new_geometry;
+      guint new_flags;
+      gboolean was_realized;
+
+      /* We are going to go ahead and perform this configure request
+       * and then emulate a configure notify by going ahead and
+       * doing a size allocate. Sort of a synchronous
+       * mini-copy of gtk_window_move_resize() here.
+       */
+      gtk_window_compute_configure_request (window,
+                                            &configure_request,
+                                            &new_geometry,
+                                            &new_flags);
+      
+      /* We update this because we are going to go ahead
+       * and gdk_window_resize() below, rather than
+       * queuing it.
+       */
+      info->last.configure_request = configure_request;
+      
+      /* and allocate the window - this is normally done
+       * in move_resize in response to configure notify
+       */
+      allocation.width  = configure_request.width;
+      allocation.height = configure_request.height;
+      gtk_widget_size_allocate (widget, &allocation);
 
+      /* Then we guarantee we have a realize */
+      was_realized = FALSE;
+      if (!_gtk_widget_get_realized (widget))
[...]

PR #387, opened 2015-12-28
GTK 3: Fix calendar, props dialog position in gtk 3.19

lukefromdc:
These fixes https://github.com/mate-desktop/mate-panel/issues/386 and I discovered the panel properties dialog was also affected, fixed that too. 

This gtk3.19 change looks like it's to stay, these windows now need to call gtk_widget_realize () or they go to the top left side of the screen. Tested in current gtk3.19 and also tested with gtk3.14 to ensure nothing broke with older versions, worked fine in both on my system.


lukefromdc:
Closing this as the bug in question was fixed in gtk3.19.5 as of 1-3-2016 with this commit:
https://git.gnome.org/browse/gtk+/commit/?id=e93d64e4c3cdc71972bc92106763f29bfdc0abbf

let's hope it stays this way...

PR #388, opened 2016-01-03
applets/clock: add missing pkgconfig flags for NetworkManager

gktrk:
mate-panel-1.8 fails to compile with NetworkManager >=1.0.8 due to a
missing include directive ('-I'). Substitute network manager CFLAGS
and LIBS supplied by the pkgconfig macro in configure.ac and use it
for applets/clock.

Signed-off-by: Gokturk Yuksek <gokturk@binghamton.edu>


monsta:
MATE 1.8 isn't supported these days. Please ask your distro maintainer to make a patch.

PR #392, opened 2016-01-04
Gtk3: Don't use deprecated gtk_misc_set_alignment

pudh4418:


monsta:
> gtk_label_set_xalign() aligns the text inside the widget, while gtk_widget_set_halign() aligns the GtkLabel widget as a whole

Hmm but which behavior does ```gtk_misc_set_alignment``` have?

monsta:
Ok, I see that [API docs](https://developer.gnome.org/gtk3/stable/GtkMisc.html) tell me that x/y align is relevant only for GtkLabel. Is that correct?

raveit65:
merged https://github.com/mate-desktop/mate-panel/commit/067dbe7b68fa625ea87ef6e259539fff2baaeaf7
Thank you

PR #393, opened 2016-01-05
background: make code more readable

monsta:
@raveit65 @lukefromdc here I've tried to fix crashes on changing compositing on/off or on moving transparency slider too fast.
Well, I didn't succeed at that... panel applets still crash :disappointed:
But in the meanwhile I've removed some hacks from the code... So I'd like to check whether I've broken something or not - hence this PR. Please test it for possible regressions.

### DIFF
--- a/mate-panel/panel-background-monitor.c
+++ b/mate-panel/panel-background-monitor.c
@@ -84,13 +84,6 @@ static PanelBackgroundMonitor **global_background_monitors = NULL;
 
 static guint signals [LAST_SIGNAL] = { 0 };
 
-#if GTK_CHECK_VERSION(3, 0, 0)
-gboolean gdk_window_check_composited_wm(GdkWindow* window)
-{
-	return gdk_screen_is_composited(gdk_window_get_screen(window));
-}
-#endif
-
 static void
 panel_background_monitor_finalize (GObject *object)
 {
--- a/mate-panel/panel-background-monitor.h
+++ b/mate-panel/panel-background-monitor.h
@@ -49,9 +49,6 @@
 typedef struct _PanelBackgroundMonitorClass PanelBackgroundMonitorClass;
 typedef struct _PanelBackgroundMonitor      PanelBackgroundMonitor;
 
-#if GTK_CHECK_VERSION(3, 0, 0)
-gboolean                gdk_window_check_composited_wm          (GdkWindow* window);
-#endif
 GType                   panel_background_monitor_get_type       (void);
 PanelBackgroundMonitor *panel_background_monitor_get_for_screen (GdkScreen *screen);
 GdkPixbuf              *panel_background_monitor_get_region     (PanelBackgroundMonitor *monitor,
--- a/mate-panel/panel-background.c
+++ b/mate-panel/panel-background.c
@@ -450,10 +450,8 @@ composite_color_onto_desktop (PanelBackground *background)
 
 	cr = cairo_create (surface);
 
-	if(!gdk_window_check_composited_wm(background->window)){
-		gdk_cairo_set_source_pixbuf (cr, background->desktop, 0, 0);
-		cairo_paint (cr);
-	}
+	gdk_cairo_set_source_pixbuf (cr, background->desktop, 0, 0);
+	cairo_paint (cr);
 
 	gdk_cairo_set_source_rgba (cr, &background->color);
 	cairo_paint (cr);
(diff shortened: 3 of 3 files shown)


lukefromdc:
I did not find any problems using gtk3.19.5, panel behaves exactly the same as before here

raveit65:
All with gtk3:
Looks good so far, no applet crashes on baremetal with switching compositor on/off with patched cairo.
But i noticed applet craches with using 'killall mate-panel' several times successively,
Stacktrace looks a bit different as the others from compositor switching.
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/applet-crashes/backtrace-wnck-applet

In build log (gtk2) there are some warnings about mateconf in gtk-docs, we should kill theme.
Some gir warnings i will fix later to avoid to break your merge.
And some warnings are left.

I think we can adopt this commit from g-p to eliminated
```
Jan 06 15:08:23 mother org.mate.panel.applet.ClockAppletFactory[1723]: (clock-applet:7955): GLib-GObject-WARNING **: gsignal.c:2451: signal 'size_request' is invalid for instance '0x17319e0' of type 'GtkLabel'
```
https://git.gnome.org/browse/gnome-panel/commit/applets/clock/clock.c?id=a6cadc6fd74c04dfd379e65e7f06db4fcd216402

monsta:
> The program 'wnck-applet' received an X Window System error.\nThis probably reflects a bug in the program.\nThe error was 'BadGC (invalid GC parameter)'.

Oh great, a new type of X error for us. If it goes like this I'll know them all :laughing:
Jokes aside, this stacktrace is completely made of GTK+ functions... starting from the main loop which is running in the applet. So there's nothing I can possibly do here.

Did the same crash happen before my changes?

raveit65:
No, this comes up after you ask me to test 1.13 from the fix-it-maybe branch.
But only in a VM, and i never noticed a prob with killall before in VMs.

raveit65:
You killed 'real transparancy' :-1: 
I do not see a window anymore when moved under the bottom panel.

lukefromdc:
Some of this background work also may have caused the recent na-tray
corruption which I have seen only on my Intel Atom netbook, only when
running Marco 1.12 or 1.13.

https://github.com/mate-desktop/mate-panel/issues/394

This might be a panel bug or might be hitting a bug in Marco, so I also filed
this issue against Marco, hopefully tomorrow I will know the exact version of\
marco where this first appears.

https://github.com/mate-desktop/marco/issues/247

No trace of this on my big desktops in Compiz or any version of Marco, no
trace on the laptop if compiz is used either.

monsta:
Oops :confused:

Ok, so this needs more work. Good that I didn't push this to master. I'll check what can be done.

@lukefromdc: I didn't know you encountered that bug with the code from this PR... do you also get it with latest git master?

monsta:
BTW, do you guys get different results when marco's compositing is on and off? Or is it the same?

raveit65:
@monsta 
It doesn't matter if is use compositor with marco or not, or using another window manager (compiz gwd gtk3), the bottom panel use only a copy from the wallpaper and is not really transparent anymore. It's like with gtk2 version.

raveit65:
mate-panel from latest master does not have this issue ;)

monsta:
That's weird... the hacks I've tried to remove were supposed to introduce different behavior when compositing is on and off.

Now that you mentioned it, I can only make panel "truly transparent" (so that I could see the content of the window moved beside it) when compositing is on in marco. If compositing is off, I see the bottom of desktop background instead. That's with git master.

But anyway, I'm surely not merging this PR as is :smile:

monsta:
Ok, here goes https://github.com/mate-desktop/mate-panel/pull/396 - just code reordering, no attempts to remove hacks or change logic. Please test it just in case.

lukefromdc:
I tested both the branch and also tested the latest GIT master, both had
the same problem in the tray. There was a lot of background and profile
work that went into master for 1.13.

lukefromdc:
Still same tray issue on the netbook with master, the old branch, or the
new branch. Will try today's version of marco next.

lukefromdc:
See

https://github.com/mate-desktop/marco/issues/247

for Marco connections to the panel tray issue. At first I thought it was only
version 1.13, but the issue went all the way back to 1.11.1, appearing when
a branch to fix crashes on moving windows was merged. Moving the tray
or moving the neighboring window switcher applet invokes the corruption
most of the time unless i roll Marco back to 1.11.0.

PR #396, opened 2016-01-10
some code reordering for better readability

monsta:


raveit65:
@monsta 
looks good for me.
No applet crashes in VM when using 'killall mate-panel', and real transparency works again.

PR #398, opened 2016-01-19
GTK3: multiple background bugfixes

lukefromdc:
Fix setting image or alpha backgrounds from the GTK theme 

Fix rendering of the system panel bg immediatly after switching to it from a user bg the panel was started with a user BG

Make it  possible to theme the panel at all in GTK 3.19.7

Keep the .mate-panel-menu-bar and .gnome-panel-menu-bar style classes on the panel toplevel unconditionally. GTK Inspector showed that the toplevel was actually being themed by PanelWidget and not by PanelToplevel, causing multiple issues. This was why it was not possible to set an image or an alpha background in the GTK theme. The .mate-custom-panel-background style class was being applied to the toplevel, but not mate-panel-menu-bar and .gnome-panel-menu-bar . This was also why resetting the panel from a custom to a system background required restarting the panel.
 
I found all of this because of a bug or changed feature (not sure which) in GTK 3.19.7 which prevented the panel from being themed at all. Fixing this right also fixes all the bugs mentioned above. First restart after setting a custom theme will sometimes but not reliably crash an applet, not sure if that is a pre-existing bug though. At any rate, setting ANY custom background was just broken by gtk3.19.7. don't know if they will fix that or we will have to. That one can wait.


raveit65:
merged https://github.com/mate-desktop/mate-panel/commit/e4d45d7212088f9ff5845135c83d21ef3042dda7
thanks again

PR #405, opened 2016-02-10
fix some deprecations

raveit65:
I hope i did not break anything :)
Most stuff is gtk3 only but a few commits affect gtk2 too.
@monsta @lukefromdc 
Can you please test it?

### DIFF
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -156,7 +156,9 @@ create_hig_frame (CalendarWindow *calwin,
                   GCallback   callback)
 {
         GtkWidget *vbox;
+#if !GTK_CHECK_VERSION (3, 0, 0)
         GtkWidget *alignment;
+#endif
         GtkWidget *label;
         GtkWidget *hbox;
         GtkWidget *button;
@@ -195,11 +197,17 @@ create_hig_frame (CalendarWindow *calwin,
                 gtk_label_set_use_markup (GTK_LABEL (label), TRUE);
                 gtk_container_add (GTK_CONTAINER (button), label);
 
-		alignment = gtk_alignment_new (1, 0, 0, 0);
+#if GTK_CHECK_VERSION (3, 0, 0)
+                gtk_widget_show_all (button);
+
+                gtk_box_pack_end (GTK_BOX (hbox), button, FALSE, FALSE, 0);
+#else
+                alignment = gtk_alignment_new (1, 0, 0, 0);
                 gtk_container_add (GTK_CONTAINER (alignment), button);
-        	gtk_widget_show_all (alignment);
+                gtk_widget_show_all (alignment);
 
                 gtk_container_add (GTK_CONTAINER (hbox), alignment);
+#endif
 
                 g_signal_connect_swapped (button, "clicked", callback, calwin);
         }
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -264,10 +264,14 @@ static void
 clock_location_tile_fill (ClockLocationTile *this)
 {
         ClockLocationTilePrivate *priv = PRIVATE (this);
-	GtkWidget *align;
+#if !GTK_CHECK_VERSION (3, 0, 0)
+        GtkWidget *align;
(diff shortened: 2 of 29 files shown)


lukefromdc:
I just built this  in gtk3.19, everything works. I noticed the "edit" button in the opened up calendar with the world map showing is now on the far right side of the calendar rather than next to the "Locations" expander, which could confuse someone.

Also, in running the panel from terminal I still get GtkWidget:focus-line-width and  GtkWidget:focus-padding showing up as deprecated style properties loading from data.  Changing themes does not change this, further verifying this to be still loading from data somewhere. Terminal gets spammed with these, making other work from terminal (as in background experiments) difficult

raveit65:
> Also, in running the panel from terminal I still get GtkWidget:focus-line-width and GtkWidget:focus-padding showing up as deprecated style properties loading from data. 

This comes from an applet somewhere, maybe from mate-applets. For the applets from mate-panel i've removed those deprecated styles properties, ie. with 
https://github.com/mate-desktop/mate-panel/commit/0828507731be133f03703ac6403959f74e897f34
We need to remove this also from mate-applets.

raveit65:
Btw. caja have some deprecated properties too.

lukefromdc:
I just tested applets by removing them and adding them back, the window-list is the source of the remaining focus-line-width  and focus-padding style properties . No other applet in my setup did this, though the clock always gets 

`(mate-panel:4970): GLib-GObject-WARNING **: /build/glib2.0-2.46.2/./gobject/gsignal.c:2516: signal 'size_request' is invalid for instance '0x1bbe300' of type 'GtkLabel'`

MOVING an applet sometimes gets a 

`(mate-panel:3978): Gtk-WARNING **: Allocating size to GtkAlignment 0x280d2a0 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?`

This also sometimes pops up on first adding an applet that takes a lot of space like hardware sensors or workspace switcher, probably because of something having to move.

lukefromdc:
Just tested the window list after building the panel with the lines focus-line-width  and focus-padding physically removed from every last instance in the code-and they STILL came up when running with the window list in the panel from terminal. This means it must be coming in with something loaded by the window list applet, but it only appears in "colorsel" in mate-desktop.   Tried compiz, marco, and even gtk2 based IceWM in case it was coming in from the window manager itself, no difference.

raveit65:
Let us open a new report to work on this warnings because they are not related to this PR.
Best way to find the 'focus-line-width' or other deprecated properties is to 'grep' on a tarball.
I guess for the other warning we need to run mate-panel or other mate apps with 'export G_DEBUG=fatal-criticals' in 'gdb' to find out at which place in the code a  gtk_widget_get_preferred_width/height() is expected.

raveit65:
@monsta 
i updated
do not use deprecated gdk_cursor_new
add missing 'static' to avoid a build warning

Anything left?

raveit65:
@monsta 
Maybe you find some time to fix this deprecated warning?
```
panel-profile.c:1700:2: Warnung: »gdk_display_get_n_screens« ist veraltet [-Wdeprecated-declarations]
  n_screens = gdk_display_get_n_screens (display);
  ^
In file included from /usr/include/gtk-3.0/gdk/gdkscreen.h:32:0,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:31,
                 from panel-profile.h:29,
                 from panel-profile.c:28:
/usr/include/gtk-3.0/gdk/gdkdisplay.h:52:13: Anmerkung: hier deklariert
 gint        gdk_display_get_n_screens      (GdkDisplay  *display);
             ^
```
I tried it for myself but no luck.
You can take a look at 'GTK+3: do not use deprecated n_screens', to see how it goes.

monsta:
> I noticed the "edit" button in the opened up calendar with the world map showing is now on the far right side of the calendar rather than next to the "Locations" expander, which could confuse someone.

Well it's so in GTK+2:

calendar

I think it looks alright...

lukefromdc:
Good enough

raveit65:
merged

PR #407, opened 2016-02-14
GTK+-3.0 workspace-switcher: Remove switching workspace on scroll events

raveit65:
This is now done by WnckPager itself.
And it fixes a build warning.

taken from:
https://git.gnome.org/browse/gnome-panel/commit/?id=8a8c60f

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -250,6 +250,7 @@ static void applet_change_background(MatePanelApplet* applet, MatePanelAppletBac
 #endif
 }
 
+#if !GTK_CHECK_VERSION (3, 0, 0)
 static gboolean applet_scroll(MatePanelApplet* applet, GdkEventScroll* event, PagerData* pager)
 {
 	GdkScrollDirection absolute_direction;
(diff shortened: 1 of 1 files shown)


raveit65:
Note, switching workspaces with mouse scrollwheel work s w/o this code.
@monsta 
Please test.

lukefromdc:
I just tested this with gtk 3.19.8, found no issues. Only observable difference was that the workspace selector was slightly faster to render the selected workspace color.  I didn't know of any way of switching workspaces with the scroll wheel though, at least not in compiz. Only in "expo" mode does that do anything on my setup, though doing so does work far better and more responsively with than without this patch. Found that by experimentation just now.

raveit65:
Workspace switching via scrollwheel does not work with compiz for some reasons, same with compiz-reloaded (0.8.12).

raveit65:
Only commit name is updated.

monsta:
Scrolling works fine for me (with marco).

raveit65:
merged https://github.com/mate-desktop/mate-panel/commit/0be1bd917ac00a4553bc6d3ace836aed44cd99a4

PR #411, opened 2016-02-27
Gtk3.20 user background fix 

lukefromdc:
based on code from gnome-panel, magic line turned out to be
` gdk_window_ensure_native (window);`
Also had to be applied in mate-panel-applet.c in order to draw user backgrounds on panel applets. There had to be applied from a function containing the type switch to keep it from killing the system background on applets. Kept everything behind gtk3.19 build time switches.

Had to reapply the theme default BG for themes like Adwaita that do not support MATE explicitly, also limited that to GTK 3.19 or later only. 

Two issues remain: Panel must be restarted to reapply custom background after any theme change, and after returning to the system theme (like in GTK3.18 until an earlier fix of mine). These should affect gtk3.19 and later only, previously NO user background worked with it.


raveit65:
Nice finder :)
I tested it with f24 and your PR does what it promised.
I noticed 2 things:
1. Changing back from custom color or custom background to theme background works only for the panel itself, not for the applets. Indeed a panel restart is needed.
2. If custom color is selected changing transparency applies only for the panel itself but not for applets (ie.wnck applets). Here a simple switch to theme bg and back to custom color applies ths right transparency value to the applets.

Maybe the latest 2 commits from g-p are needed to fix this, i will try them.

Ok, as we are the only ones whose run gtk+-3.19 and any other won't test your PR, i will merged them. All is for 3.19 only so it is Ok for me.
Thank you

raveit65:
rebased and merged
https://github.com/mate-desktop/mate-panel/commit/b21ae316e9c500db40129870abb1bd9e0013e950
https://github.com/mate-desktop/mate-panel/commit/0fef6a917e9f60c52d0e4fb5b0d5ef9f8371a5f7
https://github.com/mate-desktop/mate-panel/commit/979e5aacd36f44b095426cd07e50eb77e85b50f2

I will leave this report open as reminder to fix the open issues.

PS: you should really learn how to use the 'git rebase -i' command, to squash 2 commits in 1 ;)
It's not difficult.

lukefromdc:
Here's the source of the applet issue: calling
`gdk_window_set_background_pattern(window,NULL);`
is necessary to remove the custom background from applets when returning to the system background. Removing it will leave the custom background in place and is no fix. The problem is that once
 `gdk_window_ensure_native (window);`
has been called on the applet windows, this turns them transparent. Needed is a way to force the GTK theme to be drawn on them at this time, or to revert the forced gtk native windows when the custom theme is deselected. GNOME handles this another way but I have yet to find it

PR #415, opened 2016-03-10
Gtk 3.20-custom applet bg fix2

lukefromdc:
GTK 3.20-Apply the gnome-panel commits to restore applet backgrounds on gnome-panel (with considerable modificaton to mate-panel-applet.c due to differences) to gtk3.19 and later builds of mate-panel. Remove gdk_window_ensure_native from mate-panel-applet.c and use the new panel plug files from GNOME intact. They and their associated code in mate-panel-applet permit custom backgrounds to show on applets, and backgrounds on applets are changed immediately from custom to system and back again. So are changes to alpha values in a color background.

Keep this whole mess off of GTK 3.18 and earlier builds, otherwise panel applets in an alpha background are almost twice as dark due to double application of the transparent color. 

based on 
github.com/GNOME/gnome-panel/commit/3115f77b536a7c79c7d43ded0591e2b8f45219c4

I recommend building the clock, tray, wncklet, and fish in-process as this stops most of the applet crashes on panel startup with a custom background(especially image backgrounds) selected.

With a png image custom background I got these warnings but the panel rendered normally and so did all the applets:

** (mate-panel:32657): WARNING **: Failed to get pattern 20974650,40,0

** (mate-panel:32657): WARNING **: Failed to get pattern 20974650,393,0

** (mate-panel:32657): WARNING **: Failed to get pattern 20974650,1143,0

** (mate-panel:32657): WARNING **: Failed to get pattern 20974650,1588,0

** (mate-panel:32657): WARNING **: Failed to get pattern 20974650,1677,0

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -103,7 +103,7 @@ case "$with_gtk" in
 esac
 
 AC_SUBST(GTK_API_VERSION)
-
+AM_CONDITIONAL([HAVE_GTK3], [test "x$GTK_API_VERSION" = "x3.0"])
 dnl pkg-config dependency checks
 
 PKG_CHECK_MODULES(EGG_SMCLIENT, ice sm gtk+-$GTK_API_VERSION)
--- a/libmate-panel-applet/Makefile.am
+++ b/libmate-panel-applet/Makefile.am
@@ -22,7 +22,9 @@ libmate_panel_applet_4_la_SOURCES =		\
 	mate-panel-applet-factory.h		\
 	mate-panel-applet-factory.c		\
 	mate-panel-applet-gsettings.c		\
-	mate-panel-applet-gsettings.h
+	mate-panel-applet-gsettings.h       \
+	panel-plug.c			\
+	panel-plug-private.h
 
 libmate_panel_applet_4_la_LIBADD  = \
 	$(LIBMATE_PANEL_APPLET_LIBS) \
--- a/libmate-panel-applet/Makefile.am
+++ b/libmate-panel-applet/Makefile.am
@@ -14,7 +14,7 @@ libmate_panel_appletinclude_HEADERS =	\
 	mate-panel-applet.h			\
 	mate-panel-applet-gsettings.h		\
 	mate-panel-applet-enums.h
-
+if HAVE_GTK3
 libmate_panel_applet_4_la_SOURCES =		\
 	$(BUILT_SOURCES)		\
 	mate-panel-applet.h			\
@@ -25,6 +25,16 @@ libmate_panel_applet_4_la_SOURCES =		\
 	mate-panel-applet-gsettings.h       \
 	panel-plug.c			\
 	panel-plug-private.h
+else
+libmate_panel_applet_4_la_SOURCES =		\
+	$(BUILT_SOURCES)		\
+	mate-panel-applet.h			\
+	mate-panel-applet.c			\
+	mate-panel-applet-factory.h		\
+	mate-panel-applet-factory.c		\
+	mate-panel-applet-gsettings.c		\
+	mate-panel-applet-gsettings.h
+endif
 
 libmate_panel_applet_4_la_LIBADD  = \
 	$(LIBMATE_PANEL_APPLET_LIBS) \
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -47,7 +47,9 @@
 #include "mate-panel-applet-factory.h"
 #include "mate-panel-applet-marshal.h"
 #include "mate-panel-applet-enums.h"
-
+#if GTK_CHECK_VERSION (3, 19, 0)
+#include "panel-plug-private.h"
+#endif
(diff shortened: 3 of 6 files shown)


raveit65:
Cool, nice work :)
your PR does what it promised.........
.......... but it breaks gtk2 build.
Can you please apply this patch on top of your branch?
https://gist.github.com/raveit65/92ec88892f9828f59e84
This fixes the gtk2 build.

Btw., i don't see the warnings in f24. maybe they will arrive wll next gtk+ release or they come from an applet which i don't use.
@monsta 
Can yo please do a code review?
Would be nice to use this fix for 1.14

lukefromdc:
OK, I'll get that done tomorrow, I am on the road tonight on a netbook.

lukefromdc:
Patched, committed, built, and running right now on GTK 3.19. Building anythinng and
working with the files on a netbook was not fun but it worked. Also now have verification
on one more system that it works-and that it doesn't turn the panel into a CPU or RAM
hog. Only  4 more MB loaded to use a png background at 1024x600 than the system
solid color BG.

monsta:
I can at least confirm it didn't break anything with 3.18 :smile:
Custom background loads fine, applets don't crash, no warnings.

monsta:
GTK+2 build:

```
  CC       mate-panel-applet.lo
mate-panel-applet.c:2646:1: warning: return type defaults to 'int' [-Wreturn-type]
 mate_panel_applet_set_background_widget (MatePanelApplet *applet,
 ^
mate-panel-applet.c:2646:1: warning: conflicting types for 'mate_panel_applet_set_background_widget' [enabled by default]
In file included from mate-panel-applet.c:46:0:
mate-panel-applet.h:108:6: note: previous declaration of 'mate_panel_applet_set_background_widget' was here
 void mate_panel_applet_set_background_widget(MatePanelApplet* applet, GtkWidget* widget);
      ^
```

Indeed, ```void``` is missing in definition of ```mate_panel_applet_set_background_widget```.

monsta:
I also suspect there will be need to fix make dist as two new source files should be included in the tarball as well...

raveit65:
> I also suspect there will be need to fix make dist as two new source files should be included in the tarball as well...

Well, this can we fix later if distcheck will really fail.

raveit65:
@lukefromdc 
It's Ok, i will add the missing void during merge

raveit65:
Thank you, all are merged
https://github.com/mate-desktop/mate-panel/commit/a4d5480a2f4f7524fe2b5599dd99842465d4eeb3
https://github.com/mate-desktop/mate-panel/commit/3ce73cc9a7e6f98454f52d3a746a731094a44ae9
https://github.com/mate-desktop/mate-panel/commit/ea38ff3e660049756adb98eb34ad1ab41aa08ee6
https://github.com/mate-desktop/mate-panel/commit/d7fda0e930816abe3696d86d88fd539e43125e9d
https://github.com/mate-desktop/mate-panel/commit/b6684bd302bd6794771eb2569b2b7b1539f8be09

monsta:
@lukefromdc: I've missed one thing... we can't use [G_DECLARE_FINAL_TYPE](https://developer.gnome.org/gobject/unstable/gobject-Type-Information.html#G-DECLARE-FINAL-TYPE:CAPS) as it's only available after GLib 2.44. In theory, we target 2.36 now, in practice, I need to support LMDE (Jessie-based) which has 2.42... :confused:

raveit65:
We can fix that like this which is maybe easier.
https://gist.github.com/raveit65/96677f4a57f7851eafcd
Are panel-plug.c and panel-plug-private.h needed for < gtk+-3.19 ?

lukefromdc:
Those are used only in GTK 3.19

raveit65:
@monsta 
If you do not have any objection or want to fix it in another way, i will apply the linked patch.
I result building is fixed for rhel/centos/debian and we can create tarballs again.

monsta:
The only possible issue is that it won't work on systems where GTK+3 is >= 3.19 but GLib is still < 2.44.
I'm not sure if such systems exist though.

lukefromdc:
Fine by me if OK by @monsta  , I'll also apply it against 
https://github.com/lukefromdc/mate-panel/tree/gtk3.18-3.20-move-bg-handling
(which uses the same libmate-panel-applet files) and should work fine if it works
on current master.

lukefromdc:
There are now so many incompatable GTK versions that it may quickly
become impossible to cover all corner cases. Most likely this one would
come up if someone built GTK 3.20 or later on a system originally set
up wityh an older Glib and never updated. Probably they would expect
issues, and update Glib in response. The real question is will any DISTRO
use this combination?

monsta:
I agree, we can't cover such corner cases (e.g. upgrading GTK+3 from some PPA on Ubuntu).
Let's merge it then.

raveit65:
> The only possible issue is that it won't work on systems where GTK+3 is >= 3.19 but GLib is still < 2.44.
I'm not sure if such systems exist though.

I don't think, glib and gtk+ are mostly bundled.

lukefromdc:
I just tested this on top of  my PR 
https://github.com/mate-desktop/mate-panel/pull/419
for fixing the remaining panel BG bugs, no problems from
it at all, just as I expected. Built fine, runs fine on GTK 3.20.

PR #417, opened 2016-03-20
GTK+3.20: add new CSS name for WnckPager

City-busz:
The old one will no longer work. The new name was added here:
https://git.gnome.org/browse/libwnck/commit/?id=36c9e4a823cf78f2f5ffe7863acb68f015ea430a

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -567,6 +567,7 @@ gboolean workspace_switcher_applet_fill(MatePanelApplet* applet)
 
 	provider = gtk_css_provider_new ();
 	gtk_css_provider_load_from_data (provider,
+                                         "wnck-pager:selected,\n"
                                          "WnckPager:selected {\n"
                                          "background-color: #4A90D9; }",
                                          -1, NULL);


raveit65:
We have already a style class wnck-pager
https://github.com/mate-desktop/mate-panel/commit/25e25157410ff41965dd38f426538fed6c934913
Which works for all gtk+ versions.

raveit65:
@albertsmuktupavels
Why not using the old name WnckPager in libwnck as css name?
In result old theme settings will work.
With a new name older themes needs to be adjusted.

albertsmuktupavels:
Older themes will need to be adjusted anyway, no? Lowercase name just looks better with all other css names. Also using type as css name could make illusion that it is still possible to match css by type names which is not true.

City-busz:
The wnck-pager style class in commit https://github.com/mate-desktop/mate-panel/commit/25e25157410ff41965dd38f426538fed6c934913 is currently applied to the whole applet, not to the WnckPager element, so I can't use it. I opened another pull request to use class names consistently, which allows to use the class selector for fallback color CSS: https://github.com/mate-desktop/mate-panel/pull/418. If you accept it, this pull request can be ignored.

raveit65:
> Also using type as css name could make illusion that it is still possible to match css by type names which is not true.

If this is true why i can use this in themes.
```
/* the grid left from wnckpager and wncktasklist */
MatePanelAppletFrameDBus {
    background-image: -gtk-scaled(url("assets/panel-grid.svg"));
    background-repeat: no-repeat;
    background-position: left;
}
```
after adding the css name MatePanelAppletFrameDBus in our code ? ;)

raveit65:
@City-busz 
I think i prefer the other PR, i will test this.

albertsmuktupavels:
And how someone looking at your theme can know that you have used type name as css name somewhere in code? Looking in theme I see type name... So any contributor might want to style GtkButton... And that will not work, will it?

menubar, button, headerbar, PanelApplet, menuitem, MatePanelAppletFrameDBus, window, label, notebootk, wnck-pager, combobox, spinbutton...

Would not that list look better when using panel-applet (or panelapplet), mate-panel-applet-frame? Using type names now will make css look inconsistent.

raveit65:
I'm working since 24 of january to port all themes (8) to 3.20, mostly 6-8 hours a day, only to follow ideas from damage.inc whose are mostly paid by Linux companies.
I do that that all in my free time and i'm really tired about.
Sorry, i don't change anything only for cosmetic reasons.

raveit65:
The other PR is merged https://github.com/mate-desktop/mate-panel/commit/f1ad9e27d125359efe2984e21b166bab3d392e81

PR #418, opened 2016-03-21
GTK+3.20: adjust CSS classes for WnckPager

City-busz:
This allows to use class names consistently. The old WnckPager CSS selector will not work with GTK+ 3.20.

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -563,14 +563,16 @@ gboolean workspace_switcher_applet_fill(MatePanelApplet* applet)
 #if GTK_CHECK_VERSION (3, 0, 0)
 	GtkStyleContext *context;
 	context = gtk_widget_get_style_context (GTK_WIDGET (applet));
+	gtk_style_context_add_class (context, "wnck-applet");
+	context = gtk_widget_get_style_context (pager->pager);
 	gtk_style_context_add_class (context, "wnck-pager");
 
 	provider = gtk_css_provider_new ();
 	gtk_css_provider_load_from_data (provider,
-                                         "WnckPager:selected {\n"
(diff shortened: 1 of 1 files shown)


monsta:
I see the change you've mentioned in https://github.com/mate-desktop/mate-panel/pull/417 is quite new. Do we need to require some minimal wnck version for this? How will mate-panel work with older wnck if this PR (or https://github.com/mate-desktop/mate-panel/pull/417) is applied?

City-busz:
This PR works with any versions of gtk3 and libwnck as we add classes, which is supported consistently. The PR #417 will work only if build libwnck from git master, or use gtk3 older than 3.19.

So if you prefer to support all combinations, then this PR is better.

monsta:
Ok, thanks.

raveit65:
This breaks greatly settings in mate-themes (all themes). And i have zero motivation to rewrite it again, after 3 month of porting themes to 3.20
Can you please describe what you want to style?
The background of the of wnck-pager (normal/checked/hover state) is pretty nice to select with current code state.

raveit65:
Ok, new settings work like this with the PR
```
PanelApplet.wnck-applet .wnck-pager { }
```
........if it helps poor and unattractive adwaita theme :P

raveit65:
After installing unreleased libwnck from master the whole selector looks like this.
PanelApplet.wnck-applet wnck-pager.wnck-pager { }
Before:
PanelApplet.wnck-applet widget.wnck-pager { }
So this PR makes sense.
I'm not sure if we really need style class 'wnck-applet', btw. what it can style, but it is nice to have.
I will merge it like this.

raveit65:
merged https://github.com/mate-desktop/mate-panel/commit/f1ad9e27d125359efe2984e21b166bab3d392e81
Thank you

PR #419, opened 2016-03-21
Gtk 3.18 &  3.20-move bg handling

lukefromdc:
This branch fixes the temporary loss of custom panel backgrounds when changing GTK themes in GTK 3.20. Also stopped the applet crashes on theme change or loading the panel with a custom background applied, so applied it to GTK 3.18 as well. Do NOT apply it to GTK3.16 or earlier builds, as moving the panel background handling to the toplevel does not work with them.
Based on gnome-panel's
https://github.com/GNOME/gnome-panel/commit/47741777c3115b9700fd762ab7543a81d2ed92dc
 move background handling from PanelWidget to PanelToplevel 
and
https://github.com/GNOME/gnome-panel/commit/08e2e24c21694fc835c98166ff1b0cf5bba42c1c
set rgba visual on PanelToplevel not PanelWidget 

Tested with gtk 3.14, 3.16, 3.18, and today's newly-released GTK 3.20. Needs testing in GTK 2 as I don't have a GTK 2 build environment set up

Note that some early commits were named as gtk3.18/3.10 by mistake, those are GTK 3.18/3.20 commits. I cannot change the names from the website except by starting over with a new branch.

### DIFF
--- a/mate-panel/panel-applet-frame.c
+++ b/mate-panel/panel-applet-frame.c
@@ -123,8 +123,12 @@ mate_panel_applet_frame_draw (GtkWidget *widget,
 	gtk_style_context_get (context, state,
 			       "background-image", &bg_pattern,
 			       NULL);
-	background = &frame->priv->panel->background;
 
+#if GTK_CHECK_VERSION (3, 18, 0)
+	background = &frame->priv->panel->toplevel->background;
+#else
+	background = &frame->priv->panel->background;
+#endif
 	if (bg_pattern && (background->type == PANEL_BACK_IMAGE ||
 	    (background->type == PANEL_BACK_COLOR && background->has_alpha))) {
 		cairo_matrix_t ptm;
@@ -223,9 +227,11 @@ mate_panel_applet_frame_update_background_size (MatePanelAppletFrame *frame,
 	    old_allocation->width  == new_allocation->width &&
 	    old_allocation->height == new_allocation->height)
 		return;
-
+#if GTK_CHECK_VERSION (3, 18, 0)
+	background = &frame->priv->panel->toplevel->background;
+#else
 	background = &frame->priv->panel->background;
-
+#endif
 	if (background->type == PANEL_BACK_NONE ||
 	   (background->type == PANEL_BACK_COLOR && !background->has_alpha))
 		return;
@@ -647,8 +653,11 @@ mate_panel_applet_frame_change_background (MatePanelAppletFrame    *frame,
 
 	if (frame->priv->has_handle) {
 		PanelBackground *background;
-
+#if GTK_CHECK_VERSION (3, 18, 0)
+		background = &PANEL_WIDGET (parent)->toplevel->background;
+#else
 		background = &PANEL_WIDGET (parent)->background;
+#endif
 #if GTK_CHECK_VERSION (3, 0, 0)
 		panel_background_apply_css (background, GTK_WIDGET (frame));
 #else
@@ -744,7 +753,11 @@ _mate_panel_applet_frame_update_flags (MatePanelAppletFrame *frame,
 		 * it */
 		PanelBackground *background;
 
+#if GTK_CHECK_VERSION (3, 18, 0)
+		background = &frame->priv->panel->toplevel->background;
+#else
 		background = &frame->priv->panel->background;
+#endif
 		mate_panel_applet_frame_change_background (frame, background->type);
 	}
 }
@@ -800,8 +813,11 @@ _mate_panel_applet_frame_get_background_string (MatePanelAppletFrame    *frame,
 			break;
 		}
 	}
-
+#if GTK_CHECK_VERSION (3, 18, 0)
+	return panel_background_make_string (&panel->toplevel->background, x, y);
+#else
 	return panel_background_make_string (&panel->background, x, y);
+#endif
 }
 
 static void
--- a/mate-panel/panel-background.c
+++ b/mate-panel/panel-background.c
@@ -554,13 +554,8 @@ panel_background_composite (PanelBackground *background)
 		break;
 	}
 
-#if GTK_CHECK_VERSION (3, 18, 0)
-    /* FIXME, Hack, panel user background fix for gtk+-3.18+ */
-    /* this is actually WRONG but fixes rendering of user selected color BG */
-	background->composited = FALSE;
-#else
 	background->composited = TRUE;
-#endif
+
 
 	panel_background_prepare (background);
 
@@ -1500,4 +1495,3 @@ panel_background_change_background_on_widget (PanelBackground *background,
 	}
 }
 #endif
-
--- a/mate-panel/panel-menu-bar.c
+++ b/mate-panel/panel-menu-bar.c
@@ -274,8 +274,11 @@ static void panel_menu_bar_size_allocate(GtkWidget* widget, GtkAllocation* alloc
 		return;
 	}
 
+#if GTK_CHECK_VERSION (3, 18, 0)
+	background = &PANEL_MENU_BAR(widget)->priv->panel->toplevel->background;
+#else
 	background = &PANEL_MENU_BAR(widget)->priv->panel->background;
-
+#endif
 	if (background->type == PANEL_BACK_NONE || (background->type == PANEL_BACK_COLOR && !background->has_alpha))
 	{
 		return;
@@ -483,7 +486,9 @@ void panel_menu_bar_popup_menu(PanelMenuBar* menubar, guint32 activate_time)
 
 void panel_menu_bar_change_background(PanelMenuBar* menubar)
 {
-#if GTK_CHECK_VERSION (3, 0, 0)
+#if GTK_CHECK_VERSION (3, 18, 0)
+	panel_background_apply_css(&menubar->priv->panel->toplevel->background, GTK_WIDGET(menubar));
+#elif GTK_CHECK_VERSION (3, 0, 0)
 	panel_background_apply_css(&menubar->priv->panel->background, GTK_WIDGET(menubar));
 #else
 	panel_background_change_background_on_widget(&menubar->priv->panel->background, GTK_WIDGET(menubar));
--- a/mate-panel/panel-profile.c
+++ b/mate-panel/panel-profile.c
@@ -620,8 +620,12 @@ panel_profile_load_background (PanelToplevel *toplevel)
 	gboolean             rotate;
 
 	panel_widget = panel_toplevel_get_panel_widget (toplevel);
-	background = &panel_widget->background;
 
+#if GTK_CHECK_VERSION (3, 18, 0)
+	background = &panel_widget->toplevel->background;
+#else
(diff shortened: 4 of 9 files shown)


lukefromdc:
With this branch I was unable to force applets to crash on ANY gtk version tested (3.14-3.20) for some reason, not sure why. This finishes the background handling fixes for GTK 3.20 including applying custom BG's to the panel hide buttons and not losing custom backgrounds when the GTK theme is changed.

raveit65:
Are the gtk checks in the resulting diff ok?
https://patch-diff.githubusercontent.com/raw/mate-desktop/mate-panel/pull/419.diff
Btw. i tested it, looks good in a quick test in f24 gtk+-3.19.12.

monsta:
> Note that some early commits were named as gtk3.18/3.10 by mistake, those are GTK 3.18/3.20 commits. I cannot change the names from the website except by starting over with a new branch.

No problem, it's easily fixable via git's interactive rebase :smile:

lukefromdc:
I have never once been able to log in to github from the terminal or
do anything that can't be done by a random, non-logged in user except
via the website.

What I can do is this: make a new branch from scratch applying the same
changes to today's master, and provide a link to it, about an hour's work.
The only reason I didn't start over when I made this is it was already 3AM,

raveit65:
No, it's OK.
renaming the commit message or squashing 2 commits in one is a simple task which we can do during merge.
You did a lot of wok with the PR itself ;)

lukefromdc:
Thanks, I appreciate that. I've probably got 40 hours or more into all that background
work that the GTK 3.20 transition broke. I could not have done it without Albert's 
gnome-panel work. This is one reason I favor trying to keep the code between these
two panels from diverging too far on recent GTK versions and in general favor using
gnome's approaches to fixing things broken by GTK changes when possible.

raveit65:
Looks good with gtk2 too.
@monsta 
If you don't see issues i will merge and rebase it.
Than i can make a new release and we have about 2 weeks for more testing the whole tarball before 1.14 release.

monsta:
Sure, I'll test it later.

BTW, we need to pay attention to drawers here. Drawers are special toplevels that are attached to the normal toplevels (panels). People still use them (and report issues), so I don't want to remove them like gnome guys did.

monsta:
Works ok with GTK+ 3.18.9 so far, but... it kills the custom bg color on a drawer. I've tested only color with various transparency values, not the custom bg image. With git master, color and alpha are applied correctly. With this PR, I always get the default bg from GTK+ theme.

monsta:
Another issue with 3.18... for some reason, some parts of panel don't change the color when told to do so:

mate-gtk3 18

On the top panel, the menubar and the empty place on panel (between the menubar and the window-pickler) are still with GTK+ theme.
On the bottom panel, it's seen on the drawer's button and the empty place on panel.

Again, enabling/disabling compositor doesn't affect anything.

monsta:
Sorry, my bad. I totally forgot there was an old Mint-X theme in use - which is meant for 3.14.

But then I switched to TraditionalGreen (for 3.18) and the issue remained. In the end, I found out that Menta for 3.18 doesn't have this issue. So it might be theme-specific.

I've also found that when using Menta, the transparency looks different with and without compositor.

Without compositor:
menta-compositor-off

With compositor:
menta-compositor-on

Notice the different transparency level on menubar, drawer and the empty parts of panel when compositor is on.

monsta:
Yes, and same happens with drawer's background: the color is alright with Menta but isn't with Traditional themes.

raveit65:
The different with transparency is normal.
with compositor --> real transparency
without compositor --> you see an transparent image of the desktop
Same what you see in gtk2 build.
You can test this with moving a window under the bottom panel.
I'm in doubt that this can be look equal, and this is not related to the PR.
You see this effect with older versions too.
Well, i will look into TOK themes for gtk+-3.18.

raveit65:
@monsta 
The reason for the the different with themes is here.
https://github.com/mate-desktop/mate-themes/blob/gtk3.18/desktop-themes/BlueMenta/gtk-3.0/mate-applications.css#L350
With most themes i make first every element transparent, to avoid unwanted settings from general themeing in gtk-widgets.css, before i start to style the panel.
For some reasons i didn't do that for TOK themes.

lukefromdc:
I just found that drawers are treated as another panel so far as custom 
theming is concerned. You can right click for the panel prefs dialog and theme
them quite independently of the panel. Each panel can similarly be themed
independently, the "Mutiny" setup Ubuntu is playing with uses that.

I just tested a drawer in GTK 3.20, and it worked just fine with custom backgrounds
so long as you remembered to theme it. Otherwise the default is applied, just like
a new panel.

This means the panel drawer help in mate-user-guide needs updating.

raveit65:
> Works ok with GTK+ 3.18.9 so far, but... it kills the custom bg color on a drawer. I've tested only color with various transparency values, not the custom bg image. With git master, color and alpha are applied correctly. With this PR, I always get the default bg color from GTK+ theme.

Hmm, are you sure that the Drawer's drawer ever use the custom-bg-color from the panel?
In gtk2 build (latest 1.12) i see exactly the same behavior with fedora 23.

Anyways, a quick test with gtk3-inspector shows me that it is possible fix TOK themes for 3.18, but i don't start with that before the PR is merged.

lukefromdc:
It looks like what is happening is this: The custom background code only removes the
default theme on those elements that get the .mate-custom-panel-background style
class. If a background is set on an element that isn't the panel toplevel and isn't
a panelplug and will accept a background (not all will), the result would be a background
drawn over the custom them. This should not be hard to fix.

In GTK 3.18 I only have BlackMATE, Adwaita, and my own theme installed so I did not
see this.  I did not test without a compositor, didn't think of that. Was focussed on testing
all the GTK versions from 3.14 though what was then 3.19.

raveit65:
> I just found that drawers are treated as another panel so far as custom theming is concerned. You can right click for the panel prefs dialog and theme them quite independently of the panel. Each panel can similarly be themed independently, the "Mutiny" setup Ubuntu is playing with uses that.

:)   Nice founder , i didn't know that as i never use the drawer.
But this shows that we see here the expected behavior, or not ;)

lukefromdc:
Theme issues with custom backgrounds are not new. Back in 2008, a custom background in GNOME 2's panel with the UbuntuStudio
theme was partially covered by elements that wrote over the panel and thus could not match a partially transparent panel I played with.
 In short, it is easy to break custom panel backgrounds with a theme that is not written with them in mind. One of the first things I did 
when I wrote the panel part of my theme was to fix all the elements that wrote backgrounds over a custom panel background.

This sort of thing could be stopped in GTK3 by identifying each panel element that can get a background, and making sure that
mate-custom-panel-background is applied to every one of them. That's another bug and another PR though, and one for all GTK 3 
versions. Not sure how to deal with this in GTK 2 where I first saw it back in 2008 though.

raveit65:
Ok, new theme settings for TOK themes
https://gist.github.com/raveit65/1983d718f57d68d9f56a
which soles the theme issue.
@monsta 
I guess you won't like this ;)
So, please give me a hint to merge this PR over the weekend.

raveit65:
https://github.com/mate-desktop/mate-themes/commit/dbf2018e2940571b2345cca4dd72f31fcf9317a4

lukefromdc:
Different background opacities for panel and applets were something I was
unable to duplicate in GTK 3.18 with marco in a test today with compositing
on or off, with a transparent blue background. 

There was a point in developing the earlier applet background PR where I 
got such problems in all the older GTK versions. Limiting the panelplug code
to GTK 3.19 only took care of that in every test on my systems, though only
compositing WMs were ever used. Today's GTK 3.18 test without compositing
didn't show anything different in my own theme, BlackMATE, or Adwaita.

@monsta, I wonder what exactly were you doing differently. Was that a transparent 
image used as a background? I tested opaque colors, transparent colors, and an 
opaque image. I will now test a transparent image and see what happens, as the
codepath is different for image and color backgrounds. If I can dupilicate this issue,
I will then seek a solution.  If this is a GTK 3.18 only issue and doesn't bother GTK
3.20, changing the build conditions to go GTK 3.20 only would prevent it at the 
price of possibly not stopping applet crashes. I'll poke around and see what I can
find.

lukefromdc:
I just managed to duplicate this different applet transparency issue. Unless I find the
cause I will prepare a second branch that limits the changes to GTK 3.20, which does
not have this problem. Will investigate GTK 3.18 specific stuff first tonight, if that 
doesn't work the alternate branch will be available before I go to bed.

It's worth taking the time to do this right.

lukefromdc:
I think I now know what the problem is: In order to use the panel BG code with 
gtk 3.18. I probably also need to use the panelplug code with GTK 3.18, will 
give that a try.

lukefromdc:
New PR https://github.com/mate-desktop/mate-panel/pull/421 applies the panel-plug work from https://github.com/mate-desktop/mate-panel/pull/415 to GTK 3.18 as well and includes all the rest of this PR with a bit of cleanup.  These need to always be used together, as they render transparent backgrounds differently.

Note that this requires Glib 2.44 or later, but that is used all the way back to Ubuntu Vivid (15.04) which used GTK 3.14.  Alternate approach would be to simply limit both the panel plug and panel background work to GTK 3.20 or later. That would work but sacrifice the reduction in applet crashes in GTK 3.18. Only turning compositing on and off seems to crash applets when this is in use.

This change also permits GTK 3.18 builds to be used with GTK 3.20 with some themes or with a custom background. GTK 3.20 theme support elements are of course absent in such cases.

raveit65:
other PR is mergerd
https://github.com/mate-desktop/mate-panel/commit/18cd1f0c30c8a6bd5c25a4a5942d199afaf0cf6f
https://github.com/mate-desktop/mate-panel/commit/17a0277ecbc5156e5900efaffab4dcd88062ba98
https://github.com/mate-desktop/mate-panel/commit/448750ef63b8806323a4f4a9ed79e126cab206d6
https://github.com/mate-desktop/mate-panel/commit/40bb634d564dbf752fcc3420d3d92cef94eea592
https://github.com/mate-desktop/mate-panel/commit/da9c05196c082312c18cb74fd21e1a4534d56f53
https://github.com/mate-desktop/mate-panel/commit/78790736f054ac2539e090ceea0e2f324dac075e
https://github.com/mate-desktop/mate-panel/commit/3fd59e58fe9119023bb177da6cf945d501f8804e
https://github.com/mate-desktop/mate-panel/commit/566242b000341de4b424fe215da6cff7b5ce6ad1
https://github.com/mate-desktop/mate-panel/commit/9bf088cdca91e1feed8276fc55d2d47a460b983b
https://github.com/mate-desktop/mate-panel/commit/7cc67fd58e249109937ab876e852099d4be6b6e8
https://github.com/mate-desktop/mate-panel/commit/0ac60dc13d55944c363eab68d5a34dcba326704a
https://github.com/mate-desktop/mate-panel/commit/c14023e3713eb8d53a2426124e3d7907479d3ecb

PR #421, opened 2016-03-25
Gtk 3.18 & 3.20-move bg handling with correct applet colors in GTK 3.18

lukefromdc:
Rebase https://github.com/mate-desktop/mate-panel/pull/419 to current master,  then apply the panel-plug changes from https://github.com/mate-desktop/mate-panel/pull/415 to GTK 3.18 as well. These need to always be used together, as they render transparent backgrounds differently. 

Otherwise GTK 3.18 users get applets that are a different opacity than the panel if a transparent custom background (image or color) is used. This is because some versions (not the newest) of GTK 3.18 have issues of double application of an alpha value. Had trouble with that in my theme, and in testing I noticed that with compositing (my default) transparent images render darker than in GTK 3.16 or 3.20, at least this way the difference is all the way across panel and applets.

Note that this requires Glib 2.44 or later, but that is used all the way back to Ubuntu Vivid (15.04), meaning distros released in Spring 2015 or later. To break this would require GTK 3.18 (as in Ubuntu 16.04) or later plus mate-panel 1.14 in an older distro than that. Alternate approach would be to simply limit both the panel plug and panel background work to GTK 3.20 or later. That would work but sacrifice the reduction in applet crashes in GTK 3.18.  Only turning compositing on and off seems to crash applets when this is in use.

This change also permits GTK 3.18 builds to be used with GTK 3.20 with some themes or with a custom background. GTK 3.20 theme support elements are of course absent in such cases.

This branch fixes the temporary loss of custom panel backgrounds when changing GTK themes in GTK 3.20. Also stopped the applet crashes on theme change or loading the panel with a custom background applied, so applied it to GTK 3.18 as well. Do NOT apply it to GTK3.16 or earlier builds, as moving the panel background handling to the toplevel does not work with them.
Based on gnome-panel's
GNOME/gnome-panel@4774177
move background handling from PanelWidget to PanelToplevel
and
GNOME/gnome-panel@08e2e24
set rgba visual on PanelToplevel not PanelWidget 

Tested with a partially transparent image background, partially transparent color background, and theme background  in GTK 3.16, 3.18, and  3.20

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -47,7 +47,7 @@
 #include "mate-panel-applet-factory.h"
 #include "mate-panel-applet-marshal.h"
 #include "mate-panel-applet-enums.h"
-#if GTK_CHECK_VERSION (3, 19, 0)
+#if GTK_CHECK_VERSION (3, 18, 0)
 #include "panel-plug-private.h"
 #endif
 #define MATE_PANEL_APPLET_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), PANEL_TYPE_APPLET, MatePanelAppletPrivate))
@@ -71,7 +71,7 @@ struct _MatePanelAppletPrivate {
 	MatePanelAppletOrient  orient;
 	guint              size;
 	char              *background;
-#if !GTK_CHECK_VERSION (3, 19, 0)
+#if !GTK_CHECK_VERSION (3, 18, 0)
 	GtkWidget         *background_widget;
 #endif
 
@@ -1023,7 +1023,7 @@ mate_panel_applet_popup_menu (GtkWidget *widget)
 }
 
 #if GTK_CHECK_VERSION (3, 0, 0)
-#if !GTK_CHECK_VERSION (3, 19, 0)
+#if !GTK_CHECK_VERSION (3, 18, 0)
 static void
 mate_panel_applet_get_preferred_width (GtkWidget *widget,
 				       int       *minimum_width,
@@ -1119,7 +1119,7 @@ mate_panel_applet_size_allocate (GtkWidget     *widget,
 	GtkBin        *bin;
 	GtkWidget     *child;
 	int            border_width;
-#if !GTK_CHECK_VERSION (3, 19, 0)
+#if !GTK_CHECK_VERSION (3, 18, 0)
 	int            focus_width = 0;
 #endif
 	MatePanelApplet   *applet;
@@ -1127,7 +1127,7 @@ mate_panel_applet_size_allocate (GtkWidget     *widget,
 	if (!mate_panel_applet_can_focus (widget)) {
 		GTK_WIDGET_CLASS (mate_panel_applet_parent_class)->size_allocate (widget, allocation);
 	} else {
-#if !GTK_CHECK_VERSION (3, 19, 0)
+#if !GTK_CHECK_VERSION (3, 18, 0)
 		/*
 		 * We are deliberately ignoring focus-padding here to
 		 * save valuable panel real estate.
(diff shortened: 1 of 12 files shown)


monsta:
Good, the transparency with compositor on is now more consistent across the panel elements.

> Note that this requires Glib 2.44 or later

I see this version check only in panel plug code - but that was already added by @raveit65 before.
Or did I miss any other place where GLib 2.44 is required?

raveit65:
> Good, the transparency with compositor on is now more consistent across the panel elements.

I think the real cause of this is that that some applets paint their own bg, ie. brightness applet.
Means with transparency you see 2 backgrounds, one from the panel and one from the applet.
There are some commits in g-p where this is removed from applets.
But this is another problem and maybe my  guess is wrong.

monsta:
> Do NOT apply it to GTK3.16 or earlier builds, as moving the panel background handling to the toplevel does not work with them.

So it's not technically possible there?
I was hoping to reduce number of ```#if```s sometime later (during 1.15 dev phase)...

lukefromdc:
In my testing, an attempt to do this in GTK 3.16 or GTK 3.14 breaks custom
backgrounds entirely for those GTK versions. Take the #ifs out and see
for yourself. You will get transparent portions or the original theme applied
over a custom backgroud, this was why I never submitted a PR for my first
move-background-handling branch, which moved it unconditionally.

Cleaning up all that #if crap would take reducing the number of supported GTK
versions, which would be reasonable at some point in the future. In fact, it
might be worth looking at what scenarios will ever build MATE 1.16 with
GTK 3.16 or earlier. If no distro will do this and Ubuntu PPA's etc won't 
be either (3.18 in the LTS release), then all the 3.18 #ifs could go and the
minimum GTK 3 version set to 3.18. 

That still leaves the 3.19/2.20 theme elements, which cause undefined symbol
issues if used with GTK 3.18 or earlier which do not include them. Some of the
#ifs are the unavoidable price of supporting mutually incompatable GTK versions,
GNOME gets out of this by supporting only one GTK version at a time but distros
don't have a need to mix and match GNOME with its own toolkit like they do all
the other DE's. 

We will have to deal with 3.18/3.20 #if probably all the way until Ubuntu 18.04 is
used for a new version of Mint, assuming Mint stays with Ubuntu 16.04 as a base
for two years like they did before.

lukefromdc:
I simply kept @raveit65's work concerning Glib 2.44, since something
in the panelplug code required it.

raveit65:
> If no distro will do this and Ubuntu PPA's etc won't be either (3.18 in the LTS release), then all the 3.18 #ifs could go and the minimum GTK 3 version set to 3.18.

I build official packages for Rhel7/Centos7 which use gtk+-3.14 and has a support until 2024 or so ;)
So breaking compatibility for < 3.18 don't  give me an option to change mate to gtk3 after this is done for fedora.
But i guess a switch to gtk3 only in future will reduce the if define hell.

raveit65:
merged
https://github.com/mate-desktop/mate-panel/commit/18cd1f0c30c8a6bd5c25a4a5942d199afaf0cf6f
https://github.com/mate-desktop/mate-panel/commit/17a0277ecbc5156e5900efaffab4dcd88062ba98
https://github.com/mate-desktop/mate-panel/commit/448750ef63b8806323a4f4a9ed79e126cab206d6
https://github.com/mate-desktop/mate-panel/commit/40bb634d564dbf752fcc3420d3d92cef94eea592
https://github.com/mate-desktop/mate-panel/commit/da9c05196c082312c18cb74fd21e1a4534d56f53
https://github.com/mate-desktop/mate-panel/commit/78790736f054ac2539e090ceea0e2f324dac075e
https://github.com/mate-desktop/mate-panel/commit/3fd59e58fe9119023bb177da6cf945d501f8804e
https://github.com/mate-desktop/mate-panel/commit/566242b000341de4b424fe215da6cff7b5ce6ad1
https://github.com/mate-desktop/mate-panel/commit/9bf088cdca91e1feed8276fc55d2d47a460b983b
https://github.com/mate-desktop/mate-panel/commit/7cc67fd58e249109937ab876e852099d4be6b6e8
https://github.com/mate-desktop/mate-panel/commit/0ac60dc13d55944c363eab68d5a34dcba326704a
https://github.com/mate-desktop/mate-panel/commit/c14023e3713eb8d53a2426124e3d7907479d3ecb
Thank you :)

monsta:
BTW, I looked into ```configure.ac``` of GTK+ 3.18 and saw it already requires GLib 2.45.8 to build. So maybe we can even remove these GLib checks and leave only GTK+ ones.

PR #423, opened 2016-03-28
clock: update tooltip only when new text is different

### LINKED ISSUE #397: Feature request: Don't show tooltip in clock applet
One more option to not show the tool tip with a date.

Currently this is annoying: scrolling something down by scroll bar, mouse is over the clock applet and then scrolling up and then tooltip is displayed somewhere on the screen.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/29783125-feature-request-don-t-show-tooltip-in-clock-applet?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

zaps166:
This can prevent problems with OpenGL on some drivers (see #397)

This also accidentally closes #397, because the problem being is updating the tooltip on every second.

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -682,6 +682,8 @@ update_clock (ClockData * cd)
 static void
 update_tooltip (ClockData * cd)
 {
+        char *tip;
+        char *old_tip;
         if (!cd->showdate) {
 		struct tm *tm;
 		char date[256];
@@ -689,7 +691,6 @@ update_tooltip (ClockData * cd)
                 char *zone;
                 time_t now_t;
                 struct tm now;
-                char *tip;
 
 		tm = localtime (&cd->current_time);
 
@@ -722,17 +723,23 @@ update_tooltip (ClockData * cd)
 
                 tip = g_strdup_printf (utf8, zone);
 
-                gtk_widget_set_tooltip_text (cd->panel_button, tip);
                 g_free (utf8);
-                g_free (tip);
         } else {
-		if (cd->calendar_popup)
-			gtk_widget_set_tooltip_text (cd->panel_button,
-						     _("Click to hide month calendar"));
-		else
-			gtk_widget_set_tooltip_text (cd->panel_button,
-						     _("Click to view month calendar"));
+                if (cd->calendar_popup)
+                        tip = _("Click to hide month calendar");
+                else
+                        tip = _("Click to view month calendar");
         }
+
+        /* Update only when the new tip is different.
+         * This can prevent problems with OpenGL on some drivers */
+        old_tip = gtk_widget_get_tooltip_text (cd->panel_button);
+        if (g_strcmp0 (old_tip, tip))
+            gtk_widget_set_tooltip_text (cd->panel_button, tip);
+
+        g_free (old_tip);
+        if (!cd->showdate)
+                g_free (tip);
 }
 
 static void


zaps166:
Any news about this?

Reproducible when clock is updated (e.g. on every second if seconds are visible). Causes problems with smoothness on OpenGL + NVIDIA drivers + 'Sync to VBlank' + 'Allow Flipping' at full screen applications (standard gaming settings) or when OpenGL compositor is used.

raveit65:
Sorry for delay, but we're are a very small team and a day does not have 128 hours.
Sometimes a PR is not on my screen........
I added it to my TODO list.

zaps166:
@raveit65 Thanks :) Also sorry...

raveit65:
@zaps166 
Does the issue exists with m-p gtk2 or gtk3 version?

zaps166:
Gtk2 version. If bug exists on gtk3, it should fix it; otherwise this patch will do nothing for gtk3.

raveit65:
Ok good, because i couldn't reproduce the issue here with gtk3 and nvidia driver (340xxx legacy) if i enables seconds.
Here the tooltips looks ok.
Or do i need special settings for nvidia driver?
Ok, i need to reboot to gtk2 version.

zaps166:
@raveit65 Also the same is here: https://github.com/mate-desktop/mate-sensors-applet/pull/37

Network speed applet behaves OK :)

---

This patch fixes another (annoying) bug if seconds are enabled: press mouse button and move mouse cursor over the clock applet - after a second the tooltip automatically shows even if mouse button is still pressed (but it shouldn't).

raveit65:
Hmm, it's hard to see the effect that the tooltip is not at the right place, here in Mate-1.16 gtk2
Can ypou give me some detail instruction to reproduce the issue, please?
.....and please for dummies :)

raveit65:
How do you see a panel with glxgears -fullscreen ?

zaps166:
`glxgears -fullscreen` - I can see a panel, but if you set it stay on top in WM then panel is not visible. Probably you can run any game or smooth animation on full screen (Allow Flipping and Sync to VBlank must be enabled in NVIDIA panel, VSync must not be disabled by software).

- OpenGL compositing enabled (e.g. XFWM4 GIT version) - in both cases the issue happens (fullscreen/windowed).
- No compositing - the issue happens only when panel is invisible (real fullscreen).

Just tested on NVIDIA 370.28, GTX 960, GTK2. I've got also triple buffer enabled in xorg.conf, but probably this is not important.

zaps166:
Exactly the same is when window title flickers - e.g. notification if you have a message on communicator - every change on task bar causes frame skipping (2x per second) :D

zaps166:
[tooltip-position-bug.mkv.zip](https://github.com/mate-desktop/mate-panel/files/541963/tooltip-position-bug.mkv.zip)

Unpack and watch the video :) You can see when the button is pressed - the rectangle on desktop is visible. Tooltip must not be visible if mouse button is pressed. This is annoying when you scrolling a web page down in web browser using scroll bar :D

zaps166:
And another movie with nvidia problem - once per second, the frame is skipped and next is duplicated - this causes that image is not smooth (video is 60FPS). This happens also in games.

[nvidia-1second-framedrop-and-dup.mkv.zip](https://github.com/mate-desktop/mate-panel/files/542023/nvidia-1second-framedrop-and-dup.mkv.zip)

monsta:
Alright, bug in first video reproduced with GTK+2 :smile:
I had to enable seconds in clock prefs - otherwise it's not reproducible at all (tooltip never shows while mouse button is being pressed).

zaps166:
> I had to enable seconds in clock prefs - otherwise it's not reproducible at all (tooltip never shows while mouse button is being pressed).

Probably you must wait for text changed, up to 60 seconds :)

monsta:
Ok, and glxgears has some slight skips when seconds are enabled in clock... even though the panel isn't visible in fullscreen mode. :open_mouth: 

I'll test the PR later.

zaps166:
@monsta @raveit65 Thanks for testing this :)

Short about patch: call `gtk_widget_set_tooltip_text()` only when tooltip text is changed (tooltip text contains only day, year and month, so it shouldn't be updated on every second :smile: )

monsta:
Ok, tested it...
- now tooltips aren't shown when mouse button is held
- tooltips also don't get misplaced somewhere on screen
- glxgears is smoother than before, but I still see some skips, though much rare than without this fix

I suspect that glxgears is affected by similar mate-sensors-applet issue - I have that applet on my panel. I'll test https://github.com/mate-desktop/mate-sensors-applet/pull/37 as well to check that.

@raveit65: so, if you won't find any regressions, you can merge it.

raveit65:
@monsta 
Ok, it is a bugfix for gtk2 and i will cherry-pick it to 1.16 branch then, as 1.16 is the last gtk2 release, if you don't mind.

monsta:
Yes, it's ok for 1.16 too.

raveit65:
Thank you.
merged 299fd12061be8eec8cdb97b7b0afb0307ef86944

PR #429, opened 2016-04-25
gtk3: build with gweather-3.0

dnk:


monsta:
Version 3.18 is unacceptable, we still need to support distros where libgweather is at 3.14...

Example: https://packages.debian.org/source/jessie/libgweather

raveit65:
clock applet crashes during weather data update.
stacktrace:
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/backtrace-clock-applet-gweather
Also the weather icon is gone.
Tested with libgweather-3.20.0.

raveit65:
No crashs any more with updating weather info.
But it seems that libgweather don't calculate sunrise/set very well because i don't have a icon yet.
And we really need a line wrap for locations, libgweather displays City, State and County in one line which can be annoying on notebooks with small displays. See
bildschirmfoto zu 2016-06-17 13-11-28

raveit65:
I don't get any weather informations (temp) any more with libgweather.

monsta:
Ok, it needs more work and conflict resolving, let's leave it for 1.17 dev phase.

lukefromdc:
I just tested this with today's rebase, still does not show weather information but other than that builds and runs fine.

monsta:
I know, it's still broken. I used this PR as a base for my testing branch, and then tried to fix it using https://git.gnome.org/browse/gnome-panel/commit/?id=cd1a0db671b6eb57efcf2770287882e625ae1753 as a reference. Hopefully I'll make the weather show again. Can't say how much time it will take though :smile:

monsta:
Looking more at upstream changes ([clock](https://git.gnome.org/browse/gnome-panel/commit/?id=cd1a0db671b6eb57efcf2770287882e625ae1753) and [weather applet](https://git.gnome.org/browse/gnome-applets/commit/?id=8a413573e7ec34f19e7724737183caa25dc709eb)), I get the feeling it will take too much effort to try porting our clock and weather applet to libgweather for 1.18 release. There's more work to do, just renaming some functions and types isn't enough to make it work. Also there are some differences in GSettings keys handling, and lots of code changes that aren't directly related to weather. So I'm inclined to postpone this work until some next release, convert libmateweather to GTK+3 instead, possibly pick some fixes from upstream, etc.

@raveit65 @lukefromdc what do you think guys?

raveit65:
I am fine with using libmateweather as libgweather support only one weather station for my city berlin, we have 3 ;-)

lukefromdc:
Libmateweather actually works all the way with GTK3 right now, it's just a matter of a GTK3-only migration to stop mixed builds that kill the clock and weather applets. If I turn on showing weather and a location, everything just plain works. Using libgweather would commit us to dealing with whatever GNOME decides to do in the future, no matter how much work it made. On the other hand, it would save nearly 100MB of installed space when someone has both GNOME and MATE installed.

When only one is installed, libgweather is about 22MB, libmateweather unpacks to about 95MB, presumably due to supporting far more locations. Possibly GNOME has sacrificed completeness for smaller size or easier maintainance?  Where installed space is an issue libmateweather can be manually "stripped" to only keep the locations you use. This is what I do, keeping only the Turtle Island(N America) related files due to my need to keep my installed OS image size down for easy backup and compression to tarballs.

With modern hard drives and SSD's more functionality is probably a lot more important than 80MB of disk space,  we've already done a better job than GNOME of dealing with changes in the weather servers, and we never know what GNOME will do next. I vote for keeping libmateweather.

If we go with libgweather, we will get bug reports about no longer supported locations, as well as about whatever new issues the port creates.

monsta:
Well, looking at upstream ```data/Locations.xml.in```, I see that it's about as large as ours. I guess there's no significant difference between location lists. What's different is how location translations are stored...

We still use one ```Locations.XX.xml``` file per language, possibly compressed if ```--enable-locations-compression``` build option is used. Upstream changed that to only have the base ```Locations.xml``` file, plus a set of translation files in ```/usr/share/locale``` (see commit [here](https://git.gnome.org/browse/libgweather/commit/?id=a046fc7be69226d768712ecb91f5662bc9288c1d)).

For comparison, here are Russian translation files for both code and locations in my Jessie-based system:
```
$ LC_ALL=C ls -l /usr/share/locale/ru/LC_MESSAGES/libmateweather.mo
-rw-r--r-- 1 root root 14316 Dec 13 12:36 /usr/share/locale/ru/LC_MESSAGES/libmateweather.mo
$ LC_ALL=C ls -l /usr/share/libmateweather/Locations.ru.xml.gz 
-rw-r--r-- 1 root root 188687 Dec 13 12:37 /usr/share/libmateweather/Locations.ru.xml.gz

$ LC_ALL=C ls -l /usr/share/locale/ru/LC_MESSAGES/libgweather-*.mo
-rw-r--r-- 1 root root  13671 Oct 13  2014 /usr/share/locale/ru/LC_MESSAGES/libgweather-3.0.mo
-rw-r--r-- 1 root root 233708 Oct 13  2014 /usr/share/locale/ru/LC_MESSAGES/libgweather-locations.mo
```

monsta:
Ok, migration PR is at https://github.com/mate-desktop/libmateweather/pull/36.

I'm closing this one for now.

lukefromdc:
Thanks for the tip about --enable-locations-compression, by combining that with --enable-all-translations-in-one-xml I was able to limit installed size to just 7.8MB, just 400KB larger than the stripped Turtle Islans (N America)-only builds. Build of dev-gtk3-migration with both options works fine. ./configure --help warns that the one big locations file is slow to parse but that does not seem to be an issue on modern machines. Even with the CPU throttled all the way down, when adding a location the list seemed to open instantly. On panel start weather took about 1 sec to be displayed but that is no big deal and probably is always the case given the list opened so fast to add a locationl

Issue #444, opened 2016-06-05
Focus and keyboard


Texou-:
Hi,

Just to have a trace of what we experiencec yesterday. When pressing alt-ctrl-tab, the focus should go to panels, and it does but randomly. Once on a panel, tab should move the focus on various icons (applets, etc). Actually, this behavior works randomly. It should be stabilized in next releases.

Regards,

monsta:
I confirm this and confirm that #446 fixes that. I can actually move between applets with Tab/Shift-Tab and use Enter to invoke left-click action on an applet.

raveit65:
merged
https://github.com/mate-desktop/mate-panel/commit/e14d3305b6ea36bb68db6c410494824e3c64ab5d

Issue #441, opened 2016-05-31
mate panel has accessibility issues with some of it's applets, as well as freezing periodically


coffeeking:
steps to reproduce: start orca in mate. If needed, install it first. Make sure the following applets are on your panel. clock, power manager, universal access. Switch to the panel using control+alt+tab. Tab around the panel. You'll notice that the clock applet simply says "click to select month calendar". The universal access applet simply says "1 of 1" and so does the power manager applet if there is no battery present. The panel is also very verbose when tabbing along it, usually saying "bottom expanded edge panel. Bottom expanded edge panel panel" with every tab. The last issue is a habbit of freezing. The panel itself doesn't seem to have frozen, since the applets visually update themselves, but it's the best termI can come up with for it. Once this happens, orca can no longer see any of the applets on it until you restart it with "mate-panel --replace." This is an old issue. It existed in gnome 2.x before mate was forked, and has persisted in mate. It exists in the latest release, mate 1.14. I'll be happy to help you guys solve it any way I can.

coffeeking:
hi
I'd like to add the following applets to this bug. The sound volume applet is accessible to orca, but only if you press control+f1, which is a very well hidden trick that isn't documented anywhere. Would it be possible to change the applet to show the current volume and output device when the applet is hovered over? Also allow the ability to change output, and input devices from the applet? I'd be happy to file a feature request if you like. The system load indicator applet is also inaccessible. This one simply says "1 of 1" when orca hovers over it, and orca can't read any of the items in it's menus, which presumably details things like cpu load, memory usage, etc.
Thanks
Kendell Clark

Joanmarie:
Some (extremely quick) observations made without a MATE environment (sorry!):

> You'll notice that the clock applet simply says "click to select month calendar".

Orca gets what it says by first checking the AtkObject name (because interactive widgets should have accessible names). Then it falls back on the AtkObject description (because we cannot count on developers actually providing accessible names, but they often provide tooltips, the content of which Gtk+ exposes via the AtkObject description).

Whatever is displayed visually on the clock applet should be exposed via the AtkObject name. If you do that, Orca should speak what users see.

The clock applet should also be keyboard accessible (I'm not saying it's not; merely mentioning it).

> The universal access applet simply says "1 of 1" and so does the power manager applet if there is no battery present.

Missing AtkObject name again.

[...]

Issue #440, opened 2016-05-29
updateing to the latest version of mate, the mate panel becomes unuseable with orca screen reader


mhussaincov:
I am  able to get to the panel but when I tab around to get to the cpu applet or the network applet ore any other applet, i'm unable to.
there is no feedback from orca, i'm not sure if this is a mate issue or not but it happened after upgradeing to the latest version of mate
my distribution is
sonar  which is intern based on manjaro which is based on arch.
thanks,
Majid Hussain

flexiondotorg:
What version of MATE are Manjaro shipping and do you know if it is built against GTK2 or GTK3?

mhussaincov:
the version of mate being run  is
mate] v1.14, and we are using gtk2
the version of mate panel is
mate-panel-1.14.0-1
I hope this is of help?
Majid Hussain

coffeeking:
hi all
I can confirm this bug. After updating to mate 1.14.1 and the latest accessibility stack, which includes at-spi2-atk 2.20, at-spi2-core 2.20, atk 2.20 and orca 3.20, the mate panel is completely inaccessible. I can no longer tab my way along it and read the applets contained in it. To make sure this wasn't my setup, I've deleted the panel configuration, both through the mate interface by using "delete this panel" and by resetting it's gsettings using dconf reset -f /org/mate/panel. Nothing seems to work. I can't mark this bug as critical, but I can't stress this enough. This is a critical bug for us disabled users. In particular, I am a co developer of the sonar gnu linux distribution for blind and disabled users, and this bug will keep us from releasing until it's fixed. Please don't worry, we're still in alpha stage so there's time. I don't wish to demand that this bug get fixed, that's not how you behave in an open source community, but it is a very important bug and I'll do all I can to help you guys fix it. Would you guys mind contacting joanmeri diggs, the orca maintainer? She can tell you what mate panel is doing internally with the accessibility stack so you guys can fix it. You can contact her at, <jdiggs@igalia.com> If you guys don't have the time, I'd be happy to do so and post her comments hear.

raveit65:
I would simply try first to downgrade the components, to find out the culprit.
It should be possible to run Mate with with a 1.12.x panel version.
Or downgrade the accessibility stack to a version where it did work.

raveit65:
Btw. you're are very brave if you're are using a rolling release distro as base for sonar.
You will always the first one who run in bugs.

coffeeking:
hi
[...]

PR #445, opened 2016-06-07
fix missing focus for panel widgets

raveit65:
fixes
partially https://github.com/mate-desktop/mate-panel/issues/441
https://github.com/mate-desktop/mate-panel/issues/440
https://github.com/mate-desktop/mate-panel/issues/444

@monsta @lukefromdc 
Please test, this is a urgent PR for mentioned reports.
I tested it it with both toolkit versions, and applets don't crash in gtk3 version if you change gtk theme if a custom background is selected, which was the intention for 9bf088c

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -460,6 +460,7 @@ panel_widget_class_init (PanelWidgetClass *class)
 	widget_class->style_set = panel_widget_style_set;
 #endif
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
I just testesd the latest code from this branch, no applet crashes and no panel background problems on theme changes. Can't test it with Orca, I don't have it installed or know how to use it.

PR #446, opened 2016-06-07
fix missing focus for panel widgets

raveit65:
credits and thanks to Joanmarie from orca project

fixes
partially https://github.com/mate-desktop/mate-panel/issues/441
https://github.com/mate-desktop/mate-panel/issues/440
https://github.com/mate-desktop/mate-panel/issues/444

@monsta @lukefromdc @coffeeking @Texou-
Please test, this is a urgent PR for mentioned reports.
I tested it it with both toolkit versions, and applets don't crash in gtk3 version if you change gtk theme if a custom background is selected, which was the intention for 9bf088c

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -460,6 +460,7 @@ panel_widget_class_init (PanelWidgetClass *class)
 	widget_class->style_set = panel_widget_style_set;
 #endif
 
+	widget_class->focus = panel_widget_real_focus;
 	container_class->add = panel_widget_cadd;
 	container_class->remove = panel_widget_cremove;
 }


monsta:
All good in my GTK+2 build.

lukefromdc:
OK,. I am about to hit the road for the dah but can test this tonight

lukefromdc:
I was able to build and run this on my netbook, noticed no change
in behavior but have not had applet crashes to begin with using
GTK3.21 or 3.20. No change in crashing the panel  on removing
the inhibit applet (ONLY that applet) from the panel either.

lukefromdc:
Got this error on the netbook and on the desktop when removing the inhibit applet:
`traps: mate-panel[4355] general protection ip:42e22f sp:7ffe398a99c0 error:0 in mate-panel[400000+8b000]`
removing ANY applet crashes the panel with or without the latest from this branch,

lukefromdc:
I don't have Orca installed so no test of that was possible

lukefromdc:
On testing at home from the dev-accessable branch I had no trouble with applet crashes on theme change, and no trouble with panel backgrounds custom or system. Took me a while to figure out what I was being asked to test, no trouble at all with this in GTK 3.21 so no regressions found. If this fixes Orca it's probably good to go

raveit65:
> Took me a while to figure out what I was being asked to test

The question was more if there was a reason why you had removed that line here
https://github.com/mate-desktop/mate-panel/commit/9bf088cdca91e1feed8276fc55d2d47a460b983b#diff-61790045ddf6fdb86bd773bd32a4bb67L443

raveit65:
merged 
f18ef408e34afca9e41b239f4f91857a5016f543

lukefromdc:
I was working at the limits of my ability. Goal was to get panel custom backgrounds
working right with GTK 3.20 and above as they did in gnome-panel, so I was trying 
to port over what GNOME had done with as few changes as possible. Found it 
helped with applet crashes a  lot, so I set the selectors for GT 3.18, the oldest
it would work with. Gtk 3.16 and below would not render custom panel backgrounds
with those changes.

It was a brutal, ugly cut and paste, then debug effort to avoid being forced to remove
features that had been broken by GTK 3.20 changes. I think in the end it will be worth 
it. The GTK change that caused this was aimes at speed improvements, and in the 
3.21 cycle a lot of underhood work has trimmed nearly 20MB from the finished GTK3
binary. That much less code to load for mostly the same features has to be less to 
run. Downside was that deprecated Gtkstyle  got a change that makes some uses of
'it segfault, and GNOME may not have the time to debug deprecated anything.

Thus more work for us, but possibly a better, faster, lighter end user system. BTW, we
NEED these bug reports to know what else is breaking! Not being an orca user there
is now way in hell I could have found this during my efforts, and I do NOT like cutting
off access to people who need these features. What we really don't need is a bunch
of broken features getting all the way to Ubuntu MATE 16.10, a future Fedora relese
or worse yet into a future RHEL, Debian Stable, or Ubuntu MATE LTS release.

Joanmarie:
> Not being an orca user there is now way in hell I could have found this during my efforts

This wasn't an Orca issue. What broke was keyboard navigation (i.e. being able to use the Tab key to move amongst the widgets). All Orca does in this case is speak the new focus in response to the focus changing. Why Orca entered into the picture at all is simply because most users click on everything; users who are blind rely upon keyboard navigation.

There is admittedly a lot more to accessibility than keyboard navigation, and a lot more to a screen reader than speaking focus changes.That said, if you test proposed changes to be sure you can still interact with the application using the keyboard, you will be preventing potentially serious regressions for Orca users. Like this one.

lukefromdc:
That too is something that I never knew was possible, as few keyboard
shortcuts have been part of my normal workflow. This time around tne
two were  connected, next time around they may not be. Most of my tests
of that round  of changes focussed on theme changes and applet crashes
in four diffferent versions of GTK3, with no ability to build on GTK2.

For testing by all of us, it would be good to have a comprehensive set of
different use cases and features to test, as many of us are unaware of some
of the features. For instance, the panel drawer can set a custom theme just
like different panels can rather than following the panel it is attached to.  I 
found that out by accident while chasing a bug report that the drawer wasn't
following custom panel themes!

Thus, we need to come up with a list of all the features and how to use 
all of them. I do not even know how to tab between applets, and obviously
I need to find out, then test. Probably i should also install Orca and figure
out how to use it, as breakage there is serious,

raveit65:
No problem, everything is fine. I think it was simply a typo because that line was never removed from gnome-panel :smile: 
Btw: after 4 years building packages for fedora i think it's normal business to break sometime packages and fix it after ;)

raveit65:
Btw. how to use orca
1. install it
2. enable accessible support in mate-at-properties (checkbox)
3 click on prefered apps in that window
4. enable orca auto start
5 restart session
After new login you should hear orca
Than use crtl-alt-tab(tab+shift) to select the panels
and use tab or arrow keys to navigate.

But honestly a orca user had to teach me that.

lukefromdc:
Got python errors in console at setup after installing-and no sound on 
anything. Orca shows up in Conky as running but I'n not getting sound
from it from ANY app.

Also shift-tab/ctrl-alt-tab are doing nothing in Compiz, had to run 
marco --replace to make them do anything at all. I have essentially
zero experience with this sort of keyboard navigation. Arrow keys to 
switch workspace and alt-F2 for the run dialog, that's about it.

Given those problems I got no change with mate-panel 1.15.0 but 
will upload the .deb anyway to my usual repo with a note to Orca
users to try it.

monsta:
BTW, I wonder how Ctrl-Alt-Tab works for you guys. Because I still stumble upon https://github.com/mate-desktop/marco/issues/118...

coffeeking:
hi
It works most of the time but I do run into that bug a lot. Could fixing
that help? If so, I can comment and help fix it. I can't code but I can
do most anything else to help
Thanks
Kendell Clark


monsta wrote:

lukefromdc:
I was unable to do anything with it. Not part of my normal workflow and
in my Compiz build does nothing at all. In Marco I was unable to get it
to do anything with the panel except select it and put a border around it.
Not sure how to use it though

monsta:
Well, if it switches to panel right away, then you can press Tab/Shift-Tab to select applets on the panel, and then press Enter to invoke left-click action on an applet. That's what was broken before, was reported in https://github.com/mate-desktop/mate-panel/issues/444, and was fixed by this PR.

However, if you have more than one panel (e.g. top and bottom ones), you might find out you can't switch to another panel with Ctrl-Alt-Tab. When you first press these keys, focus switches to the first panel. Next you still hold Ctrl and Alt and press Tab once more, but the focus won't switch to another panel. Instead, the switch dialog just disappears. That's what https://github.com/mate-desktop/marco/issues/118 is about. In other words, Ctrl-Alt-Tab dialog doesn't behave the same way as Alt-Tab dialog.

Ok, that's getting offtopic here, let's continue discussion at https://github.com/mate-desktop/marco/issues/118  :smile:

raveit65:
> Next you still hold Ctrl and Alt and press Tab once more, but the focus won't switch to another panel.

At this point you need to use crtl+alt+( shift+tab) to select the another panel.
Maybe several times, but it is possible to select the second panel ;)

PR #462, opened 2016-07-28
port to GtkGrid

raveit65:
Note, adding a second ui file was needed as the old ui file use GtkTable.
Than i decided to fix all deprecations there.
We only need to find a way to make sure that both ui files are added to the tarball.
Is EXTRA_DIST the right place for this?
@monsta @lukefromdc @XRevan86 @flexiondotorg 
Please test and review.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -99,6 +99,7 @@ case "$with_gtk" in
 esac
 
 AC_SUBST(GTK_API_VERSION)
+
 dnl pkg-config dependency checks
 
 PKG_CHECK_MODULES(EGG_SMCLIENT, ice sm gtk+-$GTK_API_VERSION)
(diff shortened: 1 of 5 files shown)


lukefromdc:
Just tested this, seems to work fine, panel overall seemed very well behaved when I worked it through various sizes from the reworked properties dialog.  Had to deal with this however:
https://github.com/mate-desktop/mate-common/issues/19
a problem in  mate-common caused by the round of Debian Unstable updates I got today (two weeks worth, no landline at home). A workaround got my build done but I cannot reliably evaluate changes to any package's version of configure.ac or autogen.sh until this is fixed

raveit65:
We need to install both UI files otherwise distcheck will fail when creating mate-panel.pot.

raveit65:
merged

PR #463, opened 2016-07-31
clock: remove useless duplication

kajzersoze:
Unless I'm missing something I don't see a reason for setting "use_markup = FALSE;" at 643 and then again at 648


monsta:
Thanks, looks like some leftover from gnome.

PR #464, opened 2016-08-03
drop ussage of MateColorButton, and use GtkColorChooser and GdkRGBA for gtk3

raveit65:


raveit65:
Please review.

monsta:
Something's not right in GTK+2...

Dialog after clicking MateColorButton:
panel-mate-color-dialog

Dialog after clicking GtkColorButton:
panel-gtk-color-dialog

raveit65:
This is correct for GtkColoButton.
The patterns are extra stuff from GtkColorchooser which was added to MateColorButton by Stefano.
http://git.mate-desktop.org/mate-desktop/commit/?id=c3fc9b3e52b0c31c6c880b753ed370d9b6187c7b
The origin GtkColorButton for gtk2 don't have this patterns.

raveit65:
I guess you like the patterns, so will keep MateColorButton for gtk2.
Does make any different for me as we can't drop m-d dependency for mate-panel.

raveit65:
@monsta 
Can we finished that PR?

lukefromdc:
I just tested this, works fine with GTK 3.21. Unfortunately we've got a GTK 3.21 regression in custom panel bg handling I discovered only because I was testing this: custom BG not applied to a transparent panel in the clock button, tray, and show desktoo button + restart needed for solid color or image to fully apply it.  

https://github.com/mate-desktop/mate-panel/issues/468

I am almost certain this is more fallout from 
https://git.gnome.org/browse/gtk+/commit/?id=2c7b21718f18ab743cd79784ae28cad582f58503

monsta:
I'm actually fine with either dialog, with or without palette. Just didn't know that palette isn't on by default in GtkColorSelection. [That property](https://developer.gnome.org/gtk2/stable/GtkColorSelection.html#GtkColorSelection--has-palette) can be turned on, but here we have no access to GtkColorSelection, it seems to be managed internally in GtkColorButton code...

So it's ok, you can merge it.

raveit65:
https://github.com/mate-desktop/mate-panel/commit/d45dd88c9028fa69e543ded6909cd1da1a7ae842
https://github.com/mate-desktop/mate-panel/commit/70e2fc7377121a7b905bb444634f9226a79e649b
https://github.com/mate-desktop/mate-panel/commit/3a0f5c2fb9a270c7f987626f2973d12794dfe89c

PR #466, opened 2016-08-07
Fix FTBFS with Gtk 3.18

mspaulding06:
Not sure if this is the right fix, but was able to get it building with Gtk 3.18 this way.


raveit65:
I have a local commit which fixes this too and some deprecated gtk_device_ungrab for 3.20.
But i'm waiting for that another PR from me is reviewed.

monsta:
Looks alright to me. Not sure how we missed this mistake... :smile:

raveit65:
Hmm, complete fix is here https://github.com/mate-desktop/mate-panel/pull/478

PR #478, opened 2016-09-04
avoid deprecated device_ungrab

raveit65:
Gtk_Seat is already used but this way it fixes also deprecated device_ungrab.
Taken from https://git.gnome.org/browse/gnome-panel/log/?qt=grep&q=device
Also, some Gtk_Version_Checks are corrected from previous commits.
See, https://github.com/mate-desktop/mate-panel/pull/466

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -444,11 +444,12 @@ static void panel_toplevel_begin_grab_op(PanelToplevel* toplevel, PanelGrabOpTyp
 	GdkCursor     *cursor;
 #if GTK_CHECK_VERSION (3, 0, 0)
 	GdkDisplay *display;
-	GdkDevice *pointer;
-	GdkDevice *keyboard;
 #if GTK_CHECK_VERSION (3, 20, 0)
 	GdkSeat *seat;
+	GdkSeatCapabilities capabilities;
 #else
+	GdkDevice *pointer;
+	GdkDevice *keyboard;
 	GdkDeviceManager *device_manager;
 #endif
 #endif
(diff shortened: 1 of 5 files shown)


lukefromdc:
Just tested this, it worked fine on Debian Unstable with GTK 3.21

mate-panel/panel-toplevel.c
```diff
@@ -448,7 +448,10 @@ static void panel_toplevel_begin_grab_op(PanelToplevel* toplevel, PanelGrabOpTyp
 	GdkDevice *keyboard;
 #if GTK_CHECK_VERSION (3, 20, 0)
 	GdkSeat *seat;
+	GdkSeatCapabilities capabilities;
 #else
+	GdkDevice *pointer;
```
monsta: pointer and keyboard are already defined at lines 447 and 448 - should they be removed from there?

mate-panel/panel-toplevel.c
```diff
@@ -553,6 +564,8 @@ static void panel_toplevel_end_grab_op (PanelToplevel* toplevel, guint32 time_)
 #if GTK_CHECK_VERSION (3, 20, 0)
 	GdkSeat *seat;
 #else
+	GdkDevice *pointer;
```
monsta: same as in my previous comment... duplication of two variables

monsta:
In GTK+2 build all is fine, the code is basically unchanged.

raveit65:
@monsta 
double vars are removed.

monsta:
Ok, all seems good now.

raveit65:
Alright , i will merge it.

PR #483, opened 2016-09-28
clock-applet: mark current day

Safari77:
shows up **bold** in Green-Submarine theme.

clocka

### DIFF
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -86,12 +86,41 @@ static GtkWidget * create_hig_frame 		  (CalendarWindow *calwin,
 		  				   const char *key,
                   				   GCallback   callback);
 
+static void calendar_mark_today(GtkCalendar *calendar)
+{
+	time_t now;
+	struct tm tm1;
+	guint year, month, day;
+
+	gtk_calendar_get_date(calendar, &year, &month, &day);
+	time(&now);
+	localtime_r (&now, &tm1);
+	if ((tm1.tm_mon == month) && (tm1.tm_year + 1900 == year)) {
+		gtk_calendar_mark_day (GTK_CALENDAR (calendar), tm1.tm_mday);
+	} else {
+		gtk_calendar_unmark_day (GTK_CALENDAR (calendar), tm1.tm_mday);
+	}
+}
+
+static gboolean calendar_update(gpointer user_data)
+{
+	GtkCalendar *calendar = user_data;
+	calendar_mark_today(calendar);
+	return G_SOURCE_REMOVE;
+}
+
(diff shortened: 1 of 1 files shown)


lukefromdc:
I just tested this with GTK 3.22. It works, though the bold text is a bit subtle.  Which GTK version did you use? Green-submarine and blue-submarine have an issue in GTK 3.20 where the background window is transparent blue or green (pretty but I don't think as intended) and the text is so light on the calendar as to be difficult to read and thus the bold barely shows up. 

In many themes with GTK 3.22 (probably 3.20 as well) there is no support for clicking on a day to indicate it, though without Evolution support there may be no reason for that. Anyway, that can be done entirely in the theme and when the calendar opens defaults to the current day, as it does now in BlackMATE and in my "UbuntuStudio_Legacy" theme. See the panel and calendar coding in BlackMATE for details.

raveit65:
> Green-submarine and blue-submarine have an issue in GTK 3.20 where the background window is transparent blue or green (pretty but I don't think as intended)

This is intended since 3.20 ;) see https://github.com/mate-desktop/mate-themes/blob/gtk3.20/desktop-themes/Blue-Submarine/gtk-3.0/mate-applications.css#L541
Well, font color could be a bit darker, can you open a report as reminder please?
Looks like reporter use an older gtk+ version.

lukefromdc:
I do like that transparent green-and am glad to see my hard work on making that sort of thing possible used outside my own theme! Submitted this issue about the too-light text that spoils the effect:
https://github.com/mate-desktop/mate-themes/issues/143

I've seen no problems from this PR, it works in every theme, including themes like Adwaita that do not explicitly support MATE.  This can be done in themes by themselves (probably...) but would require changing every theme and would not work in non-MATE themes that way. Alternately we can merge this PR, it does add a bit of code but is theme-independent and really helps those themes that do not highlight the portion of the calendar containing the current day.

Safari77:
I have gtk 3.22. Actually, I didn't know in some other themes current day was already highlighted 😁

I have been thinking of adding more features, like ```gtk_calendar_set_detail_func``` but let's see..

lukefromdc:
Given that themes can highlight the current day, should we merge this or leave it for themes? 
Pros: works with all themes, subtle enough not to disrupt any.
Cons: only one, an increase in total volume of code

lukefromdc:
I just did a further test of using themes to highlight the current date. The font cannot be easily bolded in the theme, only a marker set that also is moved when another day is clicked on (selected day, as used when Evolution was integrated). This PR keeps the current day bold no matter what, so the current day can be distinguished even if the selected day has been moved.

Will merge this if nobody objects by tonight

raveit65:
Sorry, for late response, i agree with this PR.

Issue #491, opened 2016-10-19
GTK3 clock applet width changing


MDykstra:
Using GTK3 version of Mate 1.16.0 on Arch Linux.
Open calendar in clock applet, then fold down **'Locations'**. Notice small width:
![clock-applet-1](https://cloud.githubusercontent.com/assets/1680211/19501458/2e57b1a4-95a6-11e6-8ea8-69cd0c2512b0.png)

Toggle by clicking **'date, time'** twice on panel. Notice larger width; the correct width as it was under GTK2:
![clock-applet-2](https://cloud.githubusercontent.com/assets/1680211/19501536/bb122fa2-95a6-11e6-9511-35e5e274a92e.png)

Now fold back **'Locations'** and notice confusion about width:
![clock-applet-3](https://cloud.githubusercontent.com/assets/1680211/19501579/fe8de348-95a6-11e6-96e1-ee3c6d36b88c.png)

lukefromdc:
Perhaps this is an interaction with one or more themes? I've used mate-panel
with every version of GTK3 back to 1.14, and every version of MATE back to 
1.9 (development) and never had this happen. I have the day of the week, the
date, and the time in 12 hour format without seconds in the clock button.

raveit65:
Confirmed with f24 gtk+-3.20, blue-submarine, Menta, BlackMATE, etc.
Here the calendar get a smaller width if you un-expand the location GtkExpander.
Looks like we need a minimum width for the calendar window which have the same width than the world map.
I never noticed that as locations are always expand here.

lukefromdc:
Ok, that explains why I never see this. In my theme (UbuntuStudio_Legacy) the calendar is expanded by padding around the days to fit a grid background image. The "Locations" section gets pulled out to that size, not the other way around. Thus, this can be controlled in themes, or with a hardcoded minimum width in the code. Might have to be different for GTK 3.14-18 and for GTK 3.20-22, the theming syntax is quite different as we all know.

raveit65:
Any chance to set this in code as i see that reporter don't use our themes?

MDykstra:
I use slightly altered `Greybird` theme from `Shimmer Project` inteded for Xfce. It was recently updated for GTK 3.20-22 in a separate branch. I noticed toggling `Locations` a few times, that the calendar extends outside the screen the second time the world map appears (with this theme).

lukefromdc:
Try https://github.com/mate-desktop/mate-panel/pull/492

[...]

PR #492, opened 2016-10-19
Calendar Window: force minimum size

lukefromdc:
Force minimum width for calendar window to stop jumping (on locations open/close) in themes that don't enlarge calendar from the rather small default. Set the minimum height smaller than any theme would set it as it otherwise just puts blank space under the last row rather than spacing out the numbers.

Fix for https://github.com/mate-desktop/mate-panel/issues/491

### DIFF
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -123,6 +123,9 @@ calendar_window_create_calendar (CalendarWindow *calwin)
 	struct tm                  tm1;
 
 	calendar = gtk_calendar_new ();
+	#if GTK_CHECK_VERSION (3, 0, 0)
+	gtk_widget_set_size_request(GTK_WIDGET(calendar), 260, 100);
+	#endif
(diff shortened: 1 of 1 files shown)


raveit65:
Thanks, that works great with gtk3 build but with gtk2 build the vertical size is wrong.
bildschirmfoto zu 2016-10-20_19-17-14
So, i suggest to make it gtk3 only as we need it for 1.16 branch for distros which swichted Mate to gtk3.

lukefromdc:
OK, just limited this to GTK3-only. The selectors for that will have to come back out for the dev-gtk3-migration branch down the road

lukefromdc:
Just rebased dev-gtk3-migration on master after this commit, then removed the no longer needed GTK3 build selectors from that branch

monsta:
Thanks - I've noticed that issue some time ago but later forgot about it. The fix works fine here with 3.22.

lukefromdc:
OK, I'll take care of that

PR #497, opened 2016-11-23
fix deprecated warnings

sc0w:


lukefromdc:
Builds and runs fine here with GTK 3.22.4 , warnings on restarting mate-panel (with my applet set) are now down to just these:
```
Nov 23 23:32:41 ubuntu mate-inhibit-ap[28573]: gsignal.c:2523: signal 'expose-event' is invalid for instance '0x5579ac7401c0' of type 'GpmInhibitApplet'
Nov 23 23:32:41 ubuntu mate-inhibit-ap[28573]: gsignal.c:2523: signal 'expose-event' is invalid for instance '0x5579ac7401c0' of type 'GpmInhibitApplet'
Nov 23 23:32:41 ubuntu trashapplet[28583]: Allocating size to GtkImage 0x565109a3e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
Nov 23 23:32:41 ubuntu trashapplet[28583]: Allocating size to GtkImage 0x565109a3e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
Nov 23 23:32:41 ubuntu trashapplet[28583]: Allocating size to GtkImage 0x565109a3e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
Nov 23 23:32:41 ubuntu trashapplet[28583]: Allocating size to GtkImage 0x565109a3e140 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
Nov 23 23:32:41 ubuntu clock-applet[28581]: gsignal.c:2523: signal 'size_request' is invalid for instance '0x560cd4218de0' of type 'GtkLabel'
Nov 23 23:32:41 ubuntu clock-applet[28581]: gsignal.c:2523: signal 'size_request' is invalid for instance '0x560cd4218de0' of type 
```

Mostly  gtk_widget_get_preferred_width/height() spam at this point

raveit65:
@sc0w 
Without testing, do we have icons in menu entries and buttons now?

sc0w:
@raveit65 yes, apparently, I don't see differences

sc0w:
@raveit65 yes, I can see:

before:

before

later:

later

I will investigate this

monsta:
Unfortunately, ```gtk_menu_item_new_with_mnemonic``` creates [GtkMenuItem](https://developer.gnome.org/gtk3/stable/GtkMenuItem.html), which doesn't have an icon, unlike [GtkImageMenuItem](https://developer.gnome.org/gtk3/stable/GtkImageMenuItem.html).

monsta:
As for stock items, they should be replaced only with corresponding icon names. List of replacements can be found [here](https://developer.gnome.org/gtk3/stable/gtk3-Stock-Items.html).
However, some stock items don't have a replacement icon (thank you very much gnome devs), and that's where we didn't touch the code - or else we simply lose an icon on a button.

lukefromdc:
I just saw this myself and can confirm this issue, which I consider a blocker. Here's what GTK devs are advising people to do instead, it's a bit complex, the icon named is just an example:


https://developer.gnome.org/gtk3/stable/GtkImageMenuItem.html

Description
A GtkImageMenuItem is a menu item which has an icon next to the text label.
This is functionally equivalent to:
```
GtkWidget *box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 6);
GtkWidget *icon = gtk_image_new_from_icon_name ("folder-music-symbolic", GTK_ICON_SIZE_MENU);
GtkWidget *label = gtk_label_new ("Music");
GtkWidget *menu_item = gtk_menu_item_new ();

gtk_container_add (GTK_CONTAINER (box), icon);
gtk_container_add (GTK_CONTAINER (box), label);

gtk_container_add (GTK_CONTAINER (menu_item), box);
```

gtk_widget_show_all (menu_item);

Note that the user may disable display of menu icons using the “gtk-menu-images” setting, so make sure to still fill in the text label. If you want to ensure that your menu items show an icon you are strongly encouraged to use a GtkMenuItem with a GtkImage instead.

GtkImageMenuItem has been deprecated since GTK+ 3.10. If you want to display an icon in a menu item, you should use GtkMenuItem and pack a GtkBox with a GtkImage and a GtkLabel instead. You should also consider using GtkBuilder and the XML GMenu description for creating menus, by following the GMenu guide. You should consider using icons in menu items only sparingly, and for "objects" (or "nouns") elements only, like bookmarks, files, and links; "actions" (or "verbs") should not have icons.

lukefromdc:
The above recommended (by GTK devs) code change would still disable the existing user option to turn icons in context menus on or off

lukefromdc:
Given that GTK 3.22 is the last version of GTK3 and deprecated things like stock items are only actually removed from GTK4 (3.89 right now), we don't actually have to remove them at all unless we ever plan to use GTK4. As I said above, it's possible to replace GtkImageMenuItem at the price of not so easily being able to turn icons on and off, as Monsta said, replacing stock items is harder because not all have corrosponding replacements.

To keep all our functionality we'd first have to parse the same gsettings value as before to switch between an icon or a blank image if the user has turned icons off. The code would also need to default to a blank image and not the "missing icon" image for those menu items that don't get an icon and ensure that all the labels line up either way.

Next, we'd have to line up the icon names for replacing stock items and where no replacement exists pick one.

Finally we'd use the suggested functionally equivalent code block and it should all work. The lines of code would be considerably increased in C for the same menu items, though after it is all compiled I don't know if it would be  as big in increase in the size of the actual binary code due to similar results from different source. Using multiple GTK items to do the work of one would certainly write bigger and possibly slower binaries, however. 

Thus, the option of just marking these with a "FIXME" so they could be quickly found for any later GTK4 port also exists.

sc0w:
@monsta @lukefromdc thanks for the information

I found https://github.com/GNOME/gnome-panel/commit/e16c2484c3cba8ae5cdd22d693d4150afc80fdc0 applied on https://github.com/GNOME/gnome-panel/commit/7807bf814c1170d4cd84df8698d0e8a6719d7251 but I don't know how to adapt to `applet.c`

(the gnome developers and the deleted features...)

lukefromdc:
It may even be that GNOME deprecated GtkImageMenuItem simply because they themselves no longer wanted to use icons in menus and thus did not want to maintain it

sc0w:
well, I like the icons, so, closing...

monsta:
Yes, for now these deprecated features work, so no need to break them. I'd suggest to pay attention to drawing deprecations instead... these are things that might be broken (and often they are, as we noticed multiple times already).

monsta:
> You should consider using icons in menu items only sparingly, and for "objects" (or "nouns") elements only, like bookmarks, files, and links; "actions" (or "verbs") should not have icons.

This is the thing that irritates me the most in that doc. Is it even explained anywhere? I never found any rationale.

PR #501, opened 2016-11-29
Fix: the panel does not display the icon on the mouse with dnd

### LINKED ISSUE #457: the panel does not display the icon on the mouse when I try to move them
debian testing mate 1.14.1
mate-themes 3.20.8
mate-panel 1.14.1
gtk 3.20.6
theme: traditional ok

arrastrariconospanel
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36002368-the-panel-does-not-display-the-icon-on-the-mouse-when-i-try-to-move-them?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

sc0w:
Fixes #457

### DIFF
--- a/mate-panel/applet.c
+++ b/mate-panel/applet.c
@@ -747,6 +747,15 @@ applet_button_press (GtkWidget      *widget,
 		     GdkEventButton *event,
 		     AppletInfo     *info)
 {
+	gboolean     applet_locked;
+	PanelWidget *panel_widget;
+
+	panel_widget = mate_panel_applet_get_panel_widget (info);
+	applet_locked = panel_widget_get_applet_locked (panel_widget,
+							info->widget);
+
+	if (!applet_locked) mate_panel_applet_set_dnd_enabled (info, TRUE);
+
 	if (event->button == 3)
 		return applet_do_popup_menu (widget, event, info);
 


lukefromdc:
This is a partial fix only. Some applets, notably panel launchers now can be dragged and dropped with a left click/hold,  with the icon staying on the cursor. These also render correctly when dropped into another panel. Selecting "Move" in the context menu and right click and hold they slide within the panel as before. 

The window list and the tray can now be moved in the panel by left click and hold, but the icon slides in the panel rather than staying on the mouse. These do not render correctly after moving to another panel until the panel is restarted. Might be more of these but they are all I found tonight acting this way.

Still others either ignore the left click entirely or for functional reasons capture the left click, both sets can only be moved by selecting "move" in the context menu, and they then slide in the panel.

Here is the list from tonight's test:

Can move on left-click and hold, icon stays on mouse cursor(works perfectly:

Shut down
Force Quit
all launchers
search for files
Run Application
Drawer

Can move left click and hold but applet slides in panel. Incomplete display until restart if moved to another panel:

Window-list
Tray


Cannot move on left-click and hold, requires click "move" in context menu:

trash (no response)
command (no response)
battery (no response)
calendar (no response)
fish (no response)
geyes (no response)
show-desktop (no response)
sticky-notes (no response)
timer (no response)
separator (no response)
netspeed (no response)
hardware monitor (no response)
Inhibit(captures click)
CpuFreq(captures click)
menubutton(captures click)
menubar(captures click)
window-switcher(captures click)
window-selector(captures click)
system-monitor(captures click)
weather(captures click)
brightness(captures click)
dictionary(captures click)
disk mounter(captures click)
invest(captures click)

sc0w:
@lukefromdc thanks for your tests, I will investigate it

sc0w:
@lukefromdc 

If you can't move the icons without my PR with drag and drop, you can't move it with my PR with drag and drop

It only works with items that you can move with drag and drop, this is the expected for me on my PR

Some icons we can't move with drag and drop, so, I consider it as a new issue

- There is something that you can move with drag and drop and you can't see the icon over the mouse with my PR?

- There is something that works without my pr, and doesn't work with my pr?

lukefromdc:
I didn't notice any adverse effects at all, this could be merged as-is to fix the icons that worke perfectly. The window list and tray can be moved with dnd, my guess is they are in another part of the code. That could be added to this PR, be the next PR. or be left alone. How did they behave in GTK2?  For end users updating to 1.16 (in Debian based distros) or 1.18(in those that built 1.16 with GTK2) what they will notice in changes in behavior rather than any particular behavior.

lukefromdc:
In looking through the different cases covered by  mate_panel_applet_set_dnd_enabled ,it looks like the behavior with your PR may be the originally intended behavior. When the case is "applet" the function exits with "break." I will boot a live disk with GNOME 2 on something to see how the window-list and tray (the two oddball cases) are supposed to behave.

lukefromdc:
I just tested an old Mint 14/MATE setup using GTK2, the dnd behavior was essentially the same as with this PR except for the bug in rendering the tray or window list after moving to another panel not being present, and with GTK2  a generic icon was shown on the cursor with launchers rather than the correct icon as with this PR.

The issue with moving the  tray or window list to another panel is a separate issue and should not hold this PR. The behavior in dragging and dropping them within the same panel is the same in GTK2 and with this PR. The behavior with the launchers is actually better, in that the correct icon for each launcher is used.

This is ready to go then, the remaining issues I reported came from GNOME 2 itself having inconsistant behavior in terms of panel drag and drop.

PR #513, opened 2017-01-10
Fix most GTK warnings from mate-panel

lukefromdc:
This stops most GTK warnings from the panel itself, except that themes putting a non-zero width border on the clock button will still throw a 1px underallocation warning.  With no border there no warnings from clock but themers should not lose this. The rest of the warnings from the panel itself are gone.

### DIFF
--- a/applets/wncklet/showdesktop.c
+++ b/applets/wncklet/showdesktop.c
@@ -387,9 +387,7 @@ gboolean show_desktop_applet_fill(MatePanelApplet* applet)
 	GtkActionGroup* action_group;
 	gchar* ui_path;
 	AtkObject* atk_obj;
-#if !GTK_CHECK_VERSION (3, 20, 0)
 	GtkCssProvider *provider;
-#endif
 
 	mate_panel_applet_set_flags(applet, MATE_PANEL_APPLET_EXPAND_MINOR);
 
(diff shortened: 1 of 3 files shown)


lukefromdc:
Related PR's
https://github.com/mate-desktop/mate-applets/pull/221 stops warnings from the trash and CPUfreq applets
https://github.com/mate-desktop/mate-power-manager/pull/208 stops warnings from the inhibit applet

applets/wncklet/window-list.c
```diff
@@ -384,11 +384,26 @@ gboolean window_list_applet_fill(MatePanelApplet* applet)
 	TasklistData* tasklist;
 	GtkActionGroup* action_group;
 	gchar* ui_path;
+    GtkCssProvider  *provider;
```
raveit65: Indents are different to which are used in file here and lines below.

applets/wncklet/showdesktop.c
```diff
@@ -419,8 +417,20 @@ gboolean show_desktop_applet_fill(MatePanelApplet* applet)
 	sdd->button = gtk_toggle_button_new ();
 
 	gtk_widget_set_name (sdd->button, "showdesktop-button");
-#if !GTK_CHECK_VERSION (3, 20, 0)
-	provider = gtk_css_provider_new ();
+    provider = gtk_css_provider_new ();
```
raveit65: Indents are wrong here.

lukefromdc:
There are a lot of places where tabs were used in the indents in the surrounding code, no surprise I'd miss one. This is easily fixed

lukefromdc:
https://github.com/mate-desktop/mate-panel/commit/1dd29732f03f96ab579823f1326e9405331ebceb
Fixes indents in this: the file was full of tabs and by mistake I has used the normal spaces

monsta:
I presume the changes in 5bdda73212808452a1355906b4b47615a2cb7024 should actually be for GTK+ >= 3.20?

lukefromdc:
They should work with any GTK3 version, I usually test with multiple GTK versions including GTK 3.18 before pushing something like this though I have forgotten at this point how I tested this one in particular

monsta:
Are you sure? In similar commits 07b3132af934bd61629be064206d7d12e59baa88 and d06d561dfca50f269c6113bf72fd10e063713942 there's a different CSS code used for 3.20/22 and earlier versions. The CSS code in 5bdda73212808452a1355906b4b47615a2cb7024 looks like it's for 3.20/22 only. I've decided to add a version check when backporting it to 1.16 branch (259556826d99300aa67b33a2af2cdd6155b9f704).

raveit65:
#tasklist-button {} selector can be use for <3.20 too.
from https://github.com/mate-desktop/mate-panel/commit/5bdda73212808452a1355906b4b47615a2cb7024

lukefromdc:
All of the named widget selectors (those starting with #) can be used with any GTK3 version. It's css nodes that are GTK 3.20 or later only

monsta:
Selector is the same, I see, but ```padding``` and ```margin``` are used only for 3.20/22 in 07b3132af934bd61629be064206d7d12e59baa88 and d06d561dfca50f269c6113bf72fd10e063713942.

raveit65:
```
+                                         " padding: 0px;\n"
+                                         " margin: 0px;\n }",
```
I don't think that a reset to 0px is bad for < 3.20 in general.
But i don't know why the author use no version check here but for other commits.

lukefromdc:
If you think the version check is needed I have no objection to adding it.  Did this actually cause any themes to render differently in GTK 3.18 or earlier?

monsta:
Thanks. No, I didn't see any warnings, it's just for consistency.

PR #517, opened 2017-01-20
NotificationArea: Add Status Notifier support

cwendling:
The StatusNotifier implementation part is based off GNOME-Panel's *status-notifier* applet: https://git.gnome.org/browse/gnome-panel/tree/modules/external/status-notifier


cwendling:
@XRevan86 as for displaying the menu when the Item doesn't implement *Activate*, it seems a bit complicated.  Apparently it's not possible to know whether the method is implemented before actually calling it, and at the point whether the call might return with an error we're in *SnItemV0* which is not responsible for dealing with UI stuff (parent *SnItem* is).
So, I'd say it *could* be done, but would be fairly ugly: it would abuse the fairly nice split that currently exists, and require popping up a menu out of any normal event handling.

lukefromdc:
I don't know much about this but a build and run did not show any behavior changes or problems with my install over GTK 3.22 on Debian Unstable. How would I test the new feature?

cwendling:
> How would I test the new feature?

Well, you need something that registers a StatusNotifier item.  On recent Qt5s, it's the default for status icons, however I don't know much apps that do that.  On Qt4, you need the `sni-qt` package, and then apps like Mumble or Owncloud will start using it.
There are other apps that use it when available, like RedShift-GTK, or that can be configured to use it, like the Bluetooth applet (by activating the *AppIndicator* module).

lukefromdc:
I don't have ANY of the mentioned apps or any of the associated accounts, but building for basic review and resize test the latest commits now

lukefromdc:
Merged -ff-only with
https://github.com/mate-desktop/mate-panel/commit/7d39b2e82f46777efa67224f078c1cec9e827654
through
https://github.com/mate-desktop/mate-panel/commit/1c7cea2fed9c95bdf3b1831f8287ecdc551152b6

probably should have squashed these

muktupavels:
Just small note - code in gnome-flashback is under GPLv3+

cwendling:
@muktupavels hi!  and first, thanks for the work on that :)

Oh, indeed I missed that the GF watcher code was GPLv3+.  @lukefromdc @XRevan86 is this a problem?  it effectively would make mate-panel v3+ I guess.

lukefromdc:
I rather like the GPL v3 myself, and anything licensed "v2 or later" can be relicensed under v2. Thus I favor relicensing under v3, as long as that does not cause BSD distros to drop MATE. So long as no BSD (permissive) licensed programs depend on mate-panel and link to its libraries I don't see why they would have a problem installing it on top of their BSD licensed stack. Not a lawyer, that's just my take.

The main difference is v3 was designed to stop "Tivoization" where a commerical user locks a device to a particular built binary, using firmware to block other versions from running. Tivo did that to keep out modified operating systems their Cable TV buddies didn't like.

If there is a license issue, I also have no problem with a revert, but with the disgust I have for the current wave of locked devices I do like the GPLv3. Not a lawyer. but it would seem to me that no commercial user would "need" to lock a device to a particular version of MATE anyway, hell most of that kind of stuff would use a minimalist proprietary GUI.   This would be an issue if someone was selling a locked tablet with MATE on it that locked the entire operating system image to the firmware, that's about it. The goal of GPLv3 is an attempt to force vendors not to lock things.

cwendling:
@lukefromdc FTR, only the files under `libstatus-notifier-watcher/` are GPLv3, and those are not critical to be part of the applet.  They provide a DBus service that the applet needs, but that would ideally be provided by the session or specific daemon.  GNOME Flashback has it part of `gnome-flashback` which is roughly the equivalent to the session daemon.

> Thus I favor relicensing under v3, as long as that does not cause BSD distros to drop MATE.

Does the BSD guys hate the GPLv3 more than they do the GPLv2?

Apart from that, I myself have nothing against v3 either, but would hate to force it on you guys.

Issue #521, opened 2017-01-29
blueman: tray icon too small, and whithout icons with right button


sc0w:
see the differences, the blueman icon, and the order of the network and sound icons

mate panel 1.16

![beforematepanel](https://cloud.githubusercontent.com/assets/7734191/22400743/bfb76150-e5bd-11e6-9016-aa36581a07f0.png)

mate panel with the latest changes on the master branch ( I have tested with https://github.com/mate-desktop/mate-panel/commit/3edb41ee3e632cec5fcda70d34f4a613e24adf52 as latest commit)

![latermatepanel](https://cloud.githubusercontent.com/assets/7734191/22400756/26f509e4-e5be-11e6-91da-9de8e2944f59.png)

------------------------------------------------------------------------

and now, see the differences, right click on blueman: (with icons)

mate panel 1.16

![bluemanbefore](https://cloud.githubusercontent.com/assets/7734191/22400790/f191bc2e-e5be-11e6-9e7d-9d4c48dae8d9.png)

mate panel with the latest changes on the master branch: (where is the icons?)

![bluemanlater](https://cloud.githubusercontent.com/assets/7734191/22400793/f776c684-e5be-11e6-97e1-d0230c919796.png)

**EDIT:** This commit makes the issue: 7d39b2e (PR https://github.com/mate-desktop/mate-panel/pull/517 )

-------------------------------------------------------------------------------------

cwendling:
> see the differences, the blueman icon

My guess here would be that before the icon was stretched, and since 195c2a8d155cac5bd10d59a481259c161cea6c57 and 3edb41ee3e632cec5fcda70d34f4a613e24adf52 we decided not to scale raster icons.
Would be interesting to know which sizes are available for that icon on your system.

> and the order of the network and sound icons

Oops, that was an overlok on my part.  Actually all the fixed-place tray icons are now reversed, I'll provide a patch.

> mate panel with the latest changes on the master branch: (where is the icons?)

Does you apps menu have icons?  There is an (in)famous GTK setting for showing icons in menus or not, and it defaults not to on recent versions.  This new menu should follow that setting, I'm not sure about the previous one.

BTW, the previous menu has an odd blank space on the left.

Lastly… how was the Bluetooth icon displayed before?  'cause it's weirdly placed in 1.16, as in theory if the previous code worked for it it should have been placed between volume and network icons.  But maybe it just didn't work, not sure.

sc0w:
[...]

PR #522, opened 2017-01-30
notification area: Fix system tray icons forced ordering

cwendling:
The order was reversed when porting from the previous sorting method.

Part of #521.


lukefromdc:
Merged --ff-only via
https://github.com/mate-desktop/mate-panel/commit/0149809d9ea5d8f4d72b8f9a2bed139bfda98730

No problems in my builds, no unwanted behavioral changes observed.

PR #525, opened 2017-02-07
Don't spam accessible name changes when it is translated

cwendling:
Fix the logic guarding against updating the name and description so
that it properly checks against strings in the same locale.

This prevents spamming accessibility 'accessible-name' change events,
avoiding some noise when interacting with the panel.

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -1646,7 +1646,7 @@ static const char* panel_toplevel_construct_description(PanelToplevel *toplevel)
 	};
 
 	if (toplevel->priv->attached)
-		return N_("Drawer");
+		return _("Drawer");
 
 	switch (toplevel->priv->orientation) {
 	case PANEL_ORIENTATION_TOP:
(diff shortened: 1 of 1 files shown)


mate-panel/panel-toplevel.c
```diff
@@ -1701,8 +1701,8 @@ static void panel_toplevel_update_description(PanelToplevel* toplevel)
 	panel_a11y_set_atk_name_desc (
 		GTK_WIDGET (toplevel->priv->panel_widget),
 		toplevel->priv->name ? toplevel->priv->name :
-				       _(toplevel->priv->description),
-		_(toplevel->priv->description));
+				       toplevel->priv->description,
+		toplevel->priv->description);
```
cwendling: Those 2 weren't useful, as `toplevel->priv->description` was already translated from [l1695](https://github.com/mate-desktop/mate-panel/pull/525/files#diff-3bfcfd676793593c71a6a0e86c54c4c9R1695).

cwendling:
@monsta these descriptions are also used when focusing a panel with <kbd>Shift+Ctrl+Tab</kbd>, but indeed too frequent updates won't be an issue visually.

monsta:
You meant Ctrl-Alt-Tab? Yes, I don't see any changes in look & feel there.

cwendling:
@monsta ah yeah sorry, <kbd>Alt</kbd>, not <kbd>Shift</kbd>.
But all good then :)

Issue #538, opened 2017-03-20
Launcher does not respect icon selection


hoz816:
Mate Panel 1.18 on Arch Linux.  After upgrading, custom icons selected with 1.16 have reverted to default.  The selection of a different icon using the launcher properties dialog fails to result in a change using 1.18.

lukefromdc:
I've never had this problem with any version of Caja and have 1.18 running now.  My Caja install is from a series of nearly every commit update back to a few months before 1.10 came out. Thus, I am wondering if this is a migration issue caused by a file somewhere being corrupted. Try creating a new user for test purposes, logging into that session, creating a folder or file on the desktop, and changing the icon.

Question: was your 1.16 install built with GTK2 or with GTK3?

raveit65:
Hmm, i am using always the default icon for launcher or some self made launchers, so i never saw that with any MATE version.
But i can confirm that it is impossible to change  the icon for firefox launcher , for examle.
Or any other launcher.
In the launcher preferences gui i see another icon, but not in the panel.
@hoz816 
Was 1.16 compiled with gtk2 or gtk3 in archlinux?

raveit65:
Ok, it is working with 1.16 compiled against gtk+-3 in fedora 25.
So it isn't unrelated of using gtk+2 or 3.

raveit65:
mate-panel-1.17.0 does not have the issue. So the regression is introduced somewhere between 1.17.0 and 1.17.1.
git bisect should help us here.

raveit65:
Seems the culprit is commit is
```
[rave@f26 mate-panel-compile]$ git bisect good
7ee450758e4f2bee13aa8ff967a5795a6dcdc91f is the first bad commit
commit 7ee450758e4f2bee13aa8ff967a5795a6dcdc91f
Author: xxxx
Date:   Mon Jan 2 23:29:52 2017 +0300

    support desktop actions
```

raveit65:
@dnk 
Can you please take a look at it ?

dnk:
After configuring custom icon desktop file looks:
...
Icon=google-chrome   <-- default icon
...
Icon[en_US]=/usr/share/icons/Faenza/apps/48/gnome-info.png <- custom icon

I'll fix it.

lukefromdc:
Got this in my email and mistook it for a caja issue. This is confirmed here with mate-panel 1.18

raveit65:
Fix is in master and will publish with next point release.

PR #540, opened 2017-03-20
setup custom icon

dnk:
restore ability to configure launcher's custom icon which degradated after switching on g_app_info_get_icon
also update launcher state after saving launcher's properties.

https://github.com/mate-desktop/mate-panel/issues/538


raveit65:
Guys, please test, this fixes the regression with 1.18 tarball.

lukefromdc:
Works here, merged

PR #541, opened 2017-03-22
Simplify panel description shown when switching panels with the WM

### LINKED ISSUE #530: Change text when switching between panels with the keyboard
When you switch between panels with the keyboard, the strings are long so it reduces the productivity for the blinds persons that need to listen to the whole string to determine if the panel is the top or the bottom one.
In English, it's not a big issue because "top" is the first announced element, but in French the top ("supérieur") ends up at the end of the sentence, and thus is read last by the Orca screen reader.

At this time there are 4 detailed elements for the top bar (the same for the bottom, left and rieght bars):
* "Top Expanded Edge Panel"
* "Top Centered Panel"
* "Top Floating Panel"
* "Top Edge Panel"

Would it be possible to reduce it to only one string "Top bar" due the low usefulness of the "extended", "edge" or "floating" information for a sighted person?

Best regards.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/42488144-change-text-when-switching-between-panels-with-the-keyboard?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cwendling:
Fixes #530.

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -1611,73 +1611,24 @@ void panel_toplevel_update_edges(PanelToplevel* toplevel)
 
 static const char* panel_toplevel_construct_description(PanelToplevel *toplevel)
 {
-	int orientation, type;
-
-	static const char *description[4][4] = {
-		{
-	/* translators: these string will be shown in MetaCity's switch window
-	 * popup when you pass the focus to a panel */
-			N_("Top Expanded Edge Panel"),
-			N_("Top Centered Panel"),
-	     		N_("Top Floating Panel"),
-	     		N_("Top Edge Panel"),
-		},
-
-		{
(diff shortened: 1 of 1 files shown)


raveit65:
First i was a bit confused that orca anounced first the selected panel in alt-tab gui and than the panel itself.
But Ok, this seems to  be needed for blind users.

raveit65:
merged https://github.com/mate-desktop/mate-panel/commit/2798c330871e830ea13a5c1b035b891143a31166
Thank you

Issue #420, opened 2016-03-23
Workspace GUI option "Allow workspace wrap around in switcher" does nothing.


BrotherAl:
After some research into the the issue I've discovered that the GUI back end controlled by [workspace-switcher.c](https://github.com/mate-desktop/mate-panel/blob/master/applets/wncklet/workspace-switcher.c) calls `wrap_workspaces_changed`. This attempts to set or clear a boolean value called `wrap-workspaces` which does not exist in the schema, and as a result does nothing. The preferences GUI suggests that the paging system only supports a horizontal layout of virtual desktops but the code in Marco clearly shows support for multiple rows and columns. The preferences GUI will have to be changed to reflect this and the `wrap_workspaces_changed` as well as `wrap_workspaces_toggled` functions will have to be rewritten to support the proper gsetting in **org.mate.Marco.WrapStyle**. This key is not a simple boolean but an enumerated list with the selections of "no wrap", "classic", and "toroidal" 
##

BrotherAl:
I'm wondering if the code from Marco should be changed to a boolean. Is there a reason to have a difference between "classic and "toroidal"?

muesli4:
I initially wrote this functionality. Upon updating to version 1.18 I noticed that it has been removed (See: https://github.com/mate-desktop/mate-panel/commit/d0ddaedffa7d417a699cafce73c1e418515c4a28#diff-6a52aa73583e754f64a0683c23ceba46). I don't know why you throw away code that worked perfectly fine and was very usable, just to leave it in a broken state.

raveit65:
well, shit happens.
This has to be done by a mistake.
@monsta 
Huston, we have a problem :-)

raveit65:
Ok, enable the function via gsettings is working.
So only the GUI is broken, Here changing the option doesn't change the value of the gsettings key.

raveit65:
> I'm wondering if the code from Marco should be changed to a boolean. Is there a reason to have a difference between "classic and "toroidal"?

There is a different see https://github.com/mate-desktop/marco/pull/303 for explanation

muesli4:
I have trouble following the code. Did the scroll event somehow move to https://github.com/mate-desktop/marco?

muesli4:
> There is a different see mate-desktop/marco#303 for explanation

[...]

PR #542, opened 2017-03-23
Reenable workspace switcher applet wrapping behavior

muesli4:
See https://github.com/mate-desktop/mate-panel/issues/420. (Note: I accidentally rebased, but that shouldn't be a big deal.)

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -210,6 +210,113 @@ static void applet_change_background(MatePanelApplet* applet, MatePanelAppletBac
                 type == PANEL_NO_BACKGROUND ? GTK_SHADOW_NONE : GTK_SHADOW_IN);
 }
 
+/* Replacement for the default scroll handler that also cares about the wrapping property.
+ * Alternative: Add behaviour to libwnck (to the WnckPager widget).
+ */
+static gboolean applet_scroll(MatePanelApplet* applet, GdkEventScroll* event, PagerData* pager)
+{
+	GdkScrollDirection absolute_direction;
+	int index;
+	int n_workspaces;
+	int n_columns;
+	int in_last_row;
+
+	if (event->type != GDK_SCROLL)
+		return FALSE;
+
+	index = wnck_workspace_get_number(wnck_screen_get_active_workspace(pager->screen));
+	n_workspaces = wnck_screen_get_workspace_count(pager->screen);
+	n_columns = n_workspaces / pager->n_rows;
+
+	if (n_workspaces % pager->n_rows != 0)
+		n_columns++;
+
+	in_last_row    = n_workspaces % n_columns;
+
+	absolute_direction = event->direction;
+
+	if (gtk_widget_get_direction(GTK_WIDGET(applet)) == GTK_TEXT_DIR_RTL)
+	{
+		switch (event->direction)
+		{
+			case GDK_SCROLL_DOWN:
+			case GDK_SCROLL_UP:
+				break;
+			case GDK_SCROLL_RIGHT:
+				absolute_direction = GDK_SCROLL_LEFT;
+				break;
+			case GDK_SCROLL_LEFT:
+				absolute_direction = GDK_SCROLL_RIGHT;
+				break;
+		}
+	}
+
+	switch (absolute_direction)
+	{
+		case GDK_SCROLL_DOWN:
+			if (index + n_columns < n_workspaces)
+			{
+				index += n_columns;
+			}
+			else if (pager->wrap_workspaces && index == n_workspaces - 1)
+			{
+				index = 0;
+			}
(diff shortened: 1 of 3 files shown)


raveit65:
hmm, the PR does not fix it here.
I can't enable wrap via applet gui but i got a new build warning
```
workspace-switcher.c: In function 'applet_scroll':
workspace-switcher.c:240:3: warning: enumeration value 'GDK_SCROLL_SMOOTH' not handled in switch [-Wswitch]
   switch (event->direction)
   ^~~~~~
```

muesli4:
It works perfectly fine on my machine, I built a package and used it since then.

For the warning: That case is missing in the handler, however not in the WnckPager applet. I already created a patch for *libwnck* (see https://bugzilla.gnome.org/show_bug.cgi?id=780490) to share that functionality. Then I would create another patch for *mate-panel* to make use of it (which would then fix all compiler warning). Sadly, getting your code in is a slow process.

monsta:
Second ```switch``` that starts at line 254 doesn't handle ```GDK_SCROLL_SMOOTH``` as well, but here it's not so harmless. There's an assertion in the ```default```.

raveit65:
OK, my fault , i miss read your PR and bugreport. The function is only working with mousescrolling again but does not have any influence on keyboard settings for it.
```
org.mate.Marco.general wrap-style
```
crtl+alt+left/right

monsta:
Funny code in libwnck... so they added smooth scrolling support in:
https://git.gnome.org/browse/libwnck/commit/?id=82f4f0985d7ddb41bd3f4ddd1398ce37a88308f9

But they already had this change at the moment:
https://git.gnome.org/browse/libwnck/commit/?id=de74b83bfdb78af536e71c58fc297a1eb3b7efe0

This stays in their code up to this day, so I guess we could add the same dummy handling here. Not sure if ```GDK_SCROLL_SMOOTH``` could ever reach this function, but I prefer to have some protection against unexpected values.

lukefromdc:
Merged with
https://github.com/mate-desktop/mate-panel/commit/9b8f9e8147e27e19eb08a262a140b937051545b9

raveit65:
@monsta 
I recall that we use such a 'dummy handling' somewhere in code.
Can you please add this?
No need to introduce a warning.
@muesli4 
I maintain libwnck in fedora.
@muktupavels 
Can you please take a look at https://bugzilla.gnome.org/show_bug.cgi?id=780490  ?

muktupavels:
@raveit65 you want contact Marco Trevisan, he is libwnck maintainer. I added small review about style / indentation. Otherwise patch looks good, I have nothing against it.

lukefromdc:
OK, as of the moment the functionality works at the price of introducing the warning. The proposed patch for libwnck will stop the warning. If it is not merged, do we want to revert this here or keep the functionality and leave the warning? I would vote for the latter, and for trying to find a way to trap the unhandled call within mate-panel.

muktupavels:
What warning? from where?

lukefromdc:
Raveit got this warning:
```

workspace-switcher.c: In function 'applet_scroll':
workspace-switcher.c:240:3: warning: enumeration value 'GDK_SCROLL_SMOOTH' not handled in switch [-Wswitch]
   switch (event->direction)
```

muktupavels:
Add empty case and no more warning, easy no?
If that should not be reached than add also assert.
Or just add code that does something?

libwnck returns FALSE in that case so empty case should be enough to fix warning. I guess real question here is why libwnck returns FALSE for GDK_SCROLL_SMOOTH and then later in switch tries to handle that which will be never reached.

muesli4:
@raveit65 The workspace switcher applet from *mate-panel* and *marco* both use the `WnckPager` widget. However the issue at hand is only related to *mate-panel*.

@muktupavels The thing is that *libwnck* already provides a handler which handles the `GDK_SCROLL_SMOOTH` case. See variable `absolute_direction` where it either gets `GDK_SCROLL_LEFT` or `GDK_SCROLL_RIGHT`. Once I get my code into *libwnck* this missing case should no longer matter. Thanks for reviewing by the way.

monsta:
Hmm... are we in a hurry? I would prefer the merge happen only after addressing all the issues. Oh well, I'll fix it in master then.

monsta:
Ok, https://github.com/mate-desktop/mate-panel/commit/9705869fdcf0f31bc05b8454cd7ec60b7b489234 should do it.

raveit65:
I can confirm that https://github.com/mate-desktop/mate-panel/commit/9705869fdcf0f31bc05b8454cd7ec60b7b489234 fixes the build warning.

lukefromdc:
I originally merged this just before the build warning was posted here, having missed that myself. PR functioned and had been approved by another, so I went ahead with it. Thanks for cleaning up the warning

PR #543, opened 2017-03-24
status-notifier-watcher: Allow re-registration of already registered items

### LINKED ISSUE #539: Duplicate icons in notification area
Mate Panel 1.18 on Arch Linux.  An undersized icon is displayed in the notification area when the clipboard manager (clipit) is first started.  When a clipboard function is first accessed (on copy for instance), a second, larger version of the icon is displayed next to it.  This behavior was not present prior to upgrading. 
screenshot at 2017-03-19 20-46-21
screenshot at 2017-03-19 20-45-39

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43214804-duplicate-icons-in-notification-area?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

cwendling:
The [SNI specification](https://www.freedesktop.org/wiki/Specifications/StatusNotifierItem/StatusNotifierWatcher/) doesn't explicitly state what should happen when trying to register again an already registered item, so it would seem only logical to forbid it.  Unfortunately, libappindicator tries re-registering pretty often, and even falls back to System Tray if it fails.  So in practice we need to be forgiving and pretend it's OK, or we get [a](https://github.com/mate-desktop/mate-panel/issues/539) [duplicate](https://ubuntu-mate.community/t/17-04-nm-applet-works-fine-but-shows-2-icons-in-the-system-tray/12020) System Tray item for this one.

Fixes #539.

### DIFF
--- a/applets/notification_area/libstatus-notifier-watcher/gf-sn-watcher-v0.c
+++ b/applets/notification_area/libstatus-notifier-watcher/gf-sn-watcher-v0.c
@@ -284,10 +284,24 @@ gf_sn_watcher_v0_handle_register_item (GfSnWatcherV0Gen      *object,
 
   if (watch != NULL)
     {
+      /* the specification doesn't explicitly state what should happen when
+       * trying to register the same item again, so it would make sense to
+       * forbid it.  Unfortunately libappindicator tries re-registering pretty
+       * often, and even falls back to System Tray if it fails.
+       * So in practice we need to be forgiving and pretend it's OK. */
+#if 0
       g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR,
                                              G_DBUS_ERROR_INVALID_ARGS,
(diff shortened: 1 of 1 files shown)


flexiondotorg:
I've patched `mate-panel` 1.18.0 using this pull request, built it in a PPA and tested it on hardware that exhibited the issue. The duplicate icons is fixed. Thanks @cwendling :-)

cwendling:
@muktupavels You might be interested by this, or a version of it.

muktupavels:
@cwendling Thanks!

Issue #536, opened 2017-03-19
Nonresponsive line is left above workplace switcher


DavidGriffith:
At least as far back as 1.8.1 this was true:  When a workplace-switcher is placed in a panel along the top of the screen, every pixel from the bottom of the switcher to the edge of the screen is counted as part of the switcher.  Now (at least 1.16.1) there is a tiny strip of pixels above the workplace-switcher where clicks are not recognized.

At least for me, this causes trouble because I typically switch workplaces by jamming the mouse pointer up against the top of the screen where the switcher is located.  It lands in this dead zone and my clicks go nowhere.  If I move left and right while pushing up, the highlighting of the workplace rapidly flashes on and off.

DavidGriffith:
Can I get some commentary / help on this one?

lukefromdc:
I can't duplicate this in any theme with GTK3. Are you using MATE 1.16 built with GTK3 or with GTK2? Current 1.18 is GTK3-only, and 1.16 is offered with GTK3 by Debian and Ubuntu but GTK2 by MATE, with other distros no doubt varying as well.

DavidGriffith:
This is a freshly-installed instance of Debian 9.  As far as I can tell, it's compiled GTK3 only:

```
thoth:/usr/bin$ ldd mate-panel | grep gtk
        libcanberra-gtk3.so.0 => /usr/lib/x86_64-linux-gnu/libcanberra-gtk3.so.0 (0x00007ff8dc851000)
        libgtk-3.so.0 => /usr/lib/x86_64-linux-gnu/libgtk-3.so.0 (0x00007ff8dbd29000)
```

Is there a guide somewhere on how to produce your own MATE debs?  I tried last night, but the effort got very messy.

lukefromdc:
No guide that I know of. I use Checkinstall, one .deb per source, and have to manually repack them to get rid of icon cache, compiled gschema, and files in /usr/share/mime that should not be copied back to the packages

raveit65:
Does the issue still exists?

DavidGriffith:
> Does the issue still exists?

It has been mostly fixed.  What remains is that if an icon can an irregular perimeter, the pointer will not register if it's in any of the indents.  The Firefox icon is a good example of this.

balazs-endresz:
[...]

PR #546, opened 2017-03-29
status-notifier: Fix handling of icon-padding style property

cwendling:
It should only affect the space between items, not all around, and
that padding should be part of the item itself, not be outside of it.

### DIFF
--- a/applets/notification_area/status-notifier/sn-host-v0.c
+++ b/applets/notification_area/status-notifier/sn-host-v0.c
@@ -125,7 +125,7 @@ add_registered_item (SnHostV0    *v0,
   item = sn_item_v0_new (bus_name, object_path);
   g_object_ref_sink (item);
 
-  g_object_bind_property (v0, "icon-padding", item, "margin",
+  g_object_bind_property (v0, "icon-padding", item, "icon-padding",
                           G_BINDING_DEFAULT | G_BINDING_SYNC_CREATE);
   g_object_bind_property (v0, "icon-size", item, "icon-size",
                           G_BINDING_DEFAULT | G_BINDING_SYNC_CREATE);
--- a/applets/notification_area/status-notifier/sn-item-v0.c
+++ b/applets/notification_area/status-notifier/sn-item-v0.c
@@ -78,6 +78,7 @@ enum
   PROP_0,
 
   PROP_ICON_SIZE,
+  PROP_ICON_PADDING,
 
   LAST_PROP
 };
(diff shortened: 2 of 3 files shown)


cwendling:
@monsta @raveit65 this should fix the padding handling for SNI items

lukefromdc:
I can't properly test this one simply because I don't have anything using the status notifier

DavidGriffith:
It seems to me that something similar is going on with #536?

raveit65:
Btw. using a padding larger than 1 pixel isn't a good idea because it breaks 'outline' support, see https://github.com/mate-desktop/mate-themes/issues/156#issuecomment-270734976
Of course this is fixed with this PR for one direction.

monsta:
@cwendling: maybe you know which apps can be tested in Debian Testing or Unstable? This would help @lukefromdc as he's using Unstable.

cwendling:
@monsta you mean which apps will register as SNI?  Well, yeah: any Qt5 app, any Qt4 app after installing sni-qt (*Mumble*, *ownCloud*), any app using appindicator.  Also not sure what it uses, but *redshift-gtk* too.  And I'm sure @XRevan86 has a fair list to add here, even though he's on OpenSUSE IIUC, but the apps are likely the same.

cwendling:
BTW I just tested and outlines are fine for me with SNI items even for large values of *icon-padding*.

monsta:
I meant the apps that are built with appindicator support in Debian... for example, nm-applet and transmission-gtk aren't built with it, unlike in Ubuntu.

cwendling:
With libappindicator I'm not sure, but the apps mentioned (*mumble*, *owncloud*, *redshift-gtk*) all register with SNI.  For Qt4 ones, install *sni-qt* first.

raveit65:
> BTW I just tested and outlines are fine for me with SNI items even for large values of icon-padding.

Only with using 1px :-)
Try 10px for example.....
But this isn't related to this PR.

raveit65:
Fusion-icon use libappindicator

cwendling:
@raveit65 I tried a value of 6px and 12px, and while for non-SNI items the left and right sides are indeed cut off (I use the applet in the top panel, so horizontal), the SNI ones are not cut off at all, and display as I expected them, that is around the clickable area (so, including the padding).

Here is a screenshot with `na-tray-applet { -NaTrayApplet-icon-padding: 12px; }`:
nasnipadding

raveit65:
Ok, good, i forgot always that there is a different between SNI and normal icons :-).
Anyway, this is the reason why we use 1px in our themes.

raveit65:
Thanks again
merged https://github.com/mate-desktop/mate-panel/commit/a4f21ec70b26aae76a071bdd323e0ea0faf7bf98

PR #553, opened 2017-04-06
notification-area: Fix item orientation

cwendling:
Properly forward orientation to the items.  This especially fixes SNI items height in vertical panels, but affects several details on all items.

### DIFF
--- a/applets/notification_area/na-box.c
+++ b/applets/notification_area/na-box.c
@@ -243,14 +243,19 @@ na_box_realize (GtkWidget *widget)
   NaBox *self = NA_BOX (widget);
   GdkScreen *screen;
   GtkOrientation orientation;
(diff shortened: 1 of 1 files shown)


cwendling:
Second commit is kind of unrelated, but fixes a small bug when reorienting a panel live, which actually is technically similar.

monsta:
@flexiondotorg: this fixes invisible tray icons on vertical panel, see [here](https://ubuntu-mate.community/t/17-04-nm-applet-works-fine-but-shows-2-icons-in-the-system-tray/12020/32) or [here](https://ubuntu-mate.community/t/how-to-reopen-windows-from-tray/12373/11). You can test with nm-applet, transmission-gtk, redshift-gtk, caffeine. I think it would be nice to add the patch to 17.04.

Issue #552, opened 2017-04-04
GTK3 Panels do not resize to fit display when rotated using xrandr


wyatt8740:
After rotating my display via `xrandr --output LVDS1 --rotate left`, my MATE panel bars stick off the edge of the visible screen. The only way to make them fit inside the screen is to restart mate-panel manually.

I think this worked when building mate-panel based on GTK2.

I am using Debian's Mate Desktop 1.16.1 in Debian Sid.

Normal landscape view:
![Normal landscape view](http://i.imgur.com/CpsgCFJ.png)

After rotation:
![After rotation](http://i.imgur.com/wRVLJvw.png)

After rotating and running `killall mate-panel`:
![After rotating and running 'killall mate-panel'](http://i.imgur.com/CMGAD5n.png)

One other thing that might just be due to my theme, but I can't seem to make the 'hide' buttons narrower than in the screenshot. They were about half as wide in GTK2. I am guessing there's some CSS I could use that I am not currently aware of?

raveit65:
This fixes the hide buttons.
https://github.com/mate-desktop/mate-themes/commit/11f70c602cac40eb58e372e11f67194243302dbb

monsta:
I think I see the same issue when I move panel from horizontal to vertical (e.g. from bottom to left) via panel prefs dialog...

wyatt8740:
@raveit65 That doesn't really fix the hide buttons. I was referring to how wide they are compared to what I'm used to. Your patch just makes them look darker.

Of course, until recently I was using a modified version of the Ubuntu Lucid theme, but I can't seem to make the background color for the toolbars change either. This impacted the size of the hide buttons there too.

raveit65:
Nope, i don't know what you see, but this is how they should look for mate-themes.

wyatt8740:
Hm, OK.

Still, rotation is broken.

lukefromdc:
Any time something like this is applied only on restarting the panel, you are looking at something not being redrawn when it should have been. This seems to have gotten worse with the last three GTK3 cycles.

wyatt8740:
@lukefromdc so is this GTK3's fault? Or something we can fix on the MATE end?

lukefromdc:
All of the other such issues had to be fixed or worked around on our end

monsta:
[...]

PR #557, opened 2017-04-10
properly fit expanded panel to smaller screen size

monsta:
when changing panel orientation or rotating the screen itself,
panel might keep its old size, so some applets might appear
behind screen edge (until you restart the panel). same thing
might happen when enabling hide buttons.

fixes https://github.com/mate-desktop/mate-panel/issues/552

upstream commit with detailed problem description:
https://git.gnome.org/browse/gnome-panel/commit/?id=fc24d8fa623c4cf841eb5168cb1c250b4a84d4da

upstream commit with the actual fix:
https://git.gnome.org/browse/gnome-panel/commit/?id=4511f4466df0b163906794b4a32196226ccf90f3

yes, it's somewhat messed up there :)

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -1285,15 +1285,7 @@ panel_widget_get_preferred_size(GtkWidget	     *widget,
 		g_free (panel->applets_using_hint);
 	panel->applets_using_hint = NULL;
 
-	if (!panel->packed) {
-		if (panel->orient == GTK_ORIENTATION_HORIZONTAL) {
-			minimum_size->width = panel->size;
-			natural_size->width = panel->size;
-		} else {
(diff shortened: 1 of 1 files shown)


raveit65:
Not sure what symptoms should be fixed, my results:
- panel applets stay at the same positions if i change resolutions
- panel applets stay at the same positions if i un-expand the panel and enable hide buttons.
I think this was broken.
- rotating with xrandr seems not working in a VM and on bare metal i have too much applets at top panel, in result i don't see every applet if rotate to left.

raveit65:
Ok, it seems that launchers overlaps menubar applet, if i rotate the screen to left or right.
Any ways, after rotating back to normal all applets are there and at the same position.
So the PR seems to be fine.

PR #558, opened 2017-04-12
status-notifier: Do not hide passive items

### LINKED ISSUE #537: 1.18: Notification Area does not display some icons
Today I upgraded MATE to 1.18 on Arch Linux, and mate-panel 1.18 shows not all app icons.

For example, deadbeef audio player (gtk2 version) or KDE Connect has no tray icons for 1.18. Also Telegram icon is not visible sometimes. But other apps (Thunderbird, Psi+, Remmina etc.) still has icons in 1.18.

If I downgrade mate-panel to 1.16, then icons are returned.

If I run xfce4-panel instead of mate-panel, then icons are also visible.

cwendling:
Although the specification states that most hosts will hide those,
several items report them as Passive all the time and thus never show
up.  So for now, just show everything.

Fixes #537.

### DIFF
--- a/applets/notification_area/status-notifier/sn-item-v0.c
+++ b/applets/notification_area/status-notifier/sn-item-v0.c
@@ -238,7 +238,6 @@ update (SnItemV0 *v0)
   AtkObject *accessible;
   GtkImage *image;
   SnTooltip *tip;
-  gboolean visible;
   gint icon_size;
 
   g_return_if_fail (SN_IS_ITEM_V0 (v0));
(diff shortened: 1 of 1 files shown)


lukefromdc:
I also cannot test this as I have none of those apps. Should we wait for someone who has them to test this or go ahead and merge it?

raveit65:
i trust you :-)
...merged
https://github.com/mate-desktop/mate-panel/commit/36c54308f7b0e05f24b9b90fb2f79546608667b2
Thanks

Issue #559, opened 2017-04-12
Run dialog: Saving history in wrong order


muesli4:
Steps to reproduce: Run `echo 1`, `echo 2`, `echo 3` and check `org.mate.panel.general.history-mate-run`. I will provide a simple fix in an instant.

raveit65:
https://github.com/mate-desktop/mate-panel/commit/0d7df5455d544b527ecc8d09bc24f84a555382f8

PR #560, opened 2017-04-12
Run dialog: Fix wrong history order

muesli4:
Fixes https://github.com/mate-desktop/mate-panel/issues/559.

Initially the `org.mate.panel.general.history-mate-run` key was saved with
the most recent command last. In combination with wrong code this led to
scrambling the history.

This commit changes the history storage in a way that most recent
elements are stored first. The code adapts to this change by changing
only one function to prepend instead of append in the list store. (Any
old histories might therefore be wrong for a short time, but that's
probably worth it to fix this bug.)

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -132,7 +132,7 @@ _panel_run_get_recent_programs_list (PanelRunDialog *dialog)
 	     items[i] && i < PANEL_RUN_MAX_HISTORY;
 	     i++) {
 		GtkTreeIter iter;
-		gtk_list_store_prepend (list, &iter);
+		gtk_list_store_append (list, &iter);
 		gtk_list_store_set (list, &iter, 0, items[i], -1);
 	}
 


raveit65:
Does what it promised.
merged https://github.com/mate-desktop/mate-panel/commit/0d7df5455d544b527ecc8d09bc24f84a555382f8
Thank you

monsta:
Nice, I already see the improvement after testing a few commands there. Thanks :slightly_smiling_face:

raveit65:
about discussion in https://github.com/mate-desktop/mate-panel/commit/0d7df5455d544b527ecc8d09bc24f84a555382f8#commitcomment-21801512
and
https://github.com/mate-desktop/mate-panel/pull/561
From my point of view this change was correct, If i sit in front of my monitor pressing the down key most recent entry will select.
Pressing the down key again i get older entries.
And with the up key i can select the most recent entry again.
So what is wrong with this new behaviour?
And here the search in mate-terminal don't save any previous search.

raveit65:
Btw, what we see in in terminal is the bash history and not the most recent used command.
I don't think that can't be compared.

muesli4:
@raveit65 Personally I'm fine with the way it is now and I fully agree with you. But if someone wants consistent behavior with *bash history*, I'm willing to provide a solution (maybe as opt-in).

raveit65:
Ok, if you have a simple solution than please provide it.
But honestly, why should i want the less recent entry first?
This is a **drop-down-menu** which suggest to use the down button to get the most recent entry.
In opposite to **bash history** where i go back in history (up) and forward (down) to latest used command.
For me both behaviour are are right from logic view point and i would never mix or compare them.

lukefromdc:
Problem is that someone using the arrow key will not be looking at the drop-down menu in most cases. In fact, I was unaware of that menu until recently, and for years treated this the same as in terminal though not always getting things back as expected

lukefromdc:
I don't object to cherry-picking this to 1.18 though, because the previous behavior was never consistant anyway

raveit65:
Thanks for the option.
I hope everybody agree now to back port 
- Run dialog: Fix wrong history order
- Run dialog: Add option to reverse history view

to 1.18, as first one is really a bugfix.
@monsta thoughts?

monsta:
You'll probably have to pick https://github.com/mate-desktop/mate-panel/commit/5661e088e300173c64331c5350710b69774b4a99 as well to avoid conflicts. It will also help with translations, as master and 1.18 branches will have the same number of new/changed translatable strings, so you'll be able to pull the latest translation updates into both branches.

That said, I didn't push these new/changed strings to Transifex yet, so we'll have to do it first and then wait a few days to let the translators do their job...

Also let's not forget fixes from https://github.com/mate-desktop/mate-panel/commit/3171e41dcd4e0eea310acdb9c1d18c3a3adbd5fa and https://github.com/mate-desktop/mate-panel/commit/05e192158ea0ec58bb4876667c1ec8ddeb324078.

raveit65:
I agree with picking all them to 1.18, feel free to do that.
I would like to see a new release before f26 beta freeze, which is at 2017-05-16.
So we can wait a bit for translations.

raveit65:
Ok i cherry-picked them all to 1.18

monsta:
Ok good, I've pushed the updated strings to Transifex.

PR #561, opened 2017-04-12
Run dialog: Allow changing history size

muesli4:
A key has been added to the gsettings schema that controls history size and the code of the run dialog has been changed to use that key. The default of 10 is preserved and the setting is only exposed via gsettings.

Note that this *pull request* includes https://github.com/mate-desktop/mate-panel/pull/560 because it is based on it.

### DIFF
--- a/data/org.mate.panel.gschema.xml.in
+++ b/data/org.mate.panel.gschema.xml.in
@@ -23,7 +23,12 @@
     <key name="history-mate-run" type="as">
       <default>[]</default>
       <summary>History for "Run Application" dialog</summary>
-      <description>This is the list of commands used in "Run Application" dialog.</description>
+      <description>This is the list of commands used in "Run Application" dialog. The commands are sorted descendingly by recency (e.g., most recent command comes first).</description>
+    </key>
+    <key name="history-max-size-mate-run" type="u">
+      <default>10</default>
+      <summary>Maximum history size for "Run Application" dialog</summary>
+      <description>Controls the maximum size of the history of the "Run Application" dialog. A value of 0 will disable the history.</description>
     </key>
     <key name="toplevel-id-list" type="as">
       <default>[]</default>
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -155,31 +155,30 @@ _panel_run_save_recent_programs_list (PanelRunDialog   *dialog,
 
 	history_max_size = g_settings_get_uint (dialog->settings, PANEL_RUN_HISTORY_MAX_SIZE_KEY);
 
+	items = g_array_new (TRUE, TRUE, sizeof (gchar *));
 	if (history_max_size > 0) {
-		items = g_array_new (TRUE, TRUE, sizeof (gchar *));
 		g_array_append_val (items, lastcommand);
 		i++;
 
 		model = gtk_combo_box_get_model (GTK_COMBO_BOX (entry));
 
-		if (gtk_tree_model_get_iter_first (model, &iter)) {
-			char *command;
-			do {
-				gtk_tree_model_get (model, &iter, 0, &command, -1);
-				if (g_strcmp0 (command, lastcommand) == 0)
-					continue;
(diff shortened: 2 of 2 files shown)


lukefromdc:
I just tested this, found two minor issues: First is that the saved history length is actually the user set value plus one. Other than that it works. 

There is a UI issue that remains and may explain why the saving was originally in inverse order: Using the up-arrow key with the run dialog brings up the oldest item in the list, the down arrow key brings up the last item used (first in list). This is opposite the behavior in terminal, where up arrow first brings up the last item used, then sequentially older items and down arrow reverses back towards newer items provided that you are already in the terminal history. I propose we reverse these keybindings unless everyone is already used to them.

My guess is that the original inverse order was to match terminal, at the price of inverting the order shown in menu-if the menu was available at all when this was written.

muesli4:
Thanks for looking into it.

@lukefromdc I can't reproduce issue number one. The thoughts about the second issue crossed my mind (more related to https://github.com/mate-desktop/mate-panel/pull/560). I'm not sure what the right decision is: Either direction is a sacrifice for consistency between terminal or GUI. Changing the keybindings could be a solution (if implementable with GTK+).

Due to the bug we can be almost sure that the functionality hasn't been used. Now is probably a good time to change it.

lukefromdc:
The "plus one history length" issue should not be a blocker for this, though I've still got it on another session after rebooting. The reverse would be a bigger issue because it would transform zero to -1, not sure what that would do.

I think the history order was deliberately reversed, so the reversed keybindings would reverse it again.

lukefromdc:
If the keybindings are reversed, than the run dialog keybindings will match the terminal keybindings, and the menu and saved commands will show the most recent command at the top, next to the entry as they do with this PR as it stands. This seems to be consistant.

lukefromdc:
OK, on further testing I've now got the saved history length equal to the specified length and not that length plus one. Cleared that history several times, not sure what was going on. Opened dconf-editor, set history length to 2, cleared the string by setting to the default, and this time it saved two commands, not the three I got earlier in the day. Still don't know what did that but its seems to be gone now.

lukefromdc:
Oh I see what I was dealing with now: the plus one size was only with history length 1, which is what I tested earlier. About to test the new commit.

lukefromdc:
The last commit fixed the wrong history length(2) issue in the case of history length set to 1. A length of zero works as expected, saving nothing, and a negative number gets switched to the maximum supported length. This works quite well, leaving only the question of whether to deal with the arrow key/UI vs terminal UI issue here, on another PR, or not at all.

raveit65:
Seems to works fine.
What is the maximum supported length (-1 value) ?
Because personal i don't see any reason to limit the history size.
I don't think that i ever will use more than 20 applications for normal daily usage.

lukefromdc:
Maximum possible length comes up as 4294967295 when a larger value or any negative number of reasonable size is selected.  I tried some deliberately unreasonable values too, so see what would happen: really large positive values just get truncated to 4294967295, and really long negative numbers just reduce the length of the string and if a large enough negative number you get zero. This sort of testing is one way to smoke out potential buffer overflows.

raveit65:
Ok, you mean it is good to have another value than -1/+1 here?
For personal usage i will increase the default to 20 or 25.

lukefromdc:
I just wanted to reduce the risk of a magic number the lets some joker at a multiuser machine run mate-panel --replace from terminal, then use that magic value to get a buffer overflow and a root shell

muesli4:
@raveit65 @lukefromdc The maximum value is only limited by `guint`. `-1` is technically an integer underflow to `4294967295`. That value should be a reasonable maximum in all uses cases. Concerning a *buffer overflow* attack: I can only think of one with a *signed integer* (which doesn't mean there isn't one).

For the default value: I preserved the fixed value of `10` as default. Personally I think a bigger value is more reasonable. But I would recommend against a very high value, as performance can degrade over time (saving and loading from gsettings, filtering, and displaying each entry as a widget in the `GtkComboBox`).

PR #564, opened 2017-04-18
Reverse history

muesli4:
As discussed in https://github.com/mate-desktop/mate-panel/pull/560, this patch adds a `GSettings` key `history-reverse-mate-run` to view the run dialog history in reverse order (*i.e. history stays the same*). Thus with this key enabled (by default disabled) the history is displayed with the least recent entry first and allows to select the most recent entry with the up key. **Important:** This patch includes https://github.com/mate-desktop/mate-panel/pull/561 as it is on top of it.

Downsides: Changing the setting will reverse the history when the run dialog is open. This can be solved by registering a signal handler to the `GSettings` object. (I didn't bother to do it, because it seemed too much work for too little gain.)

### DIFF
--- a/data/org.mate.panel.gschema.xml.in
+++ b/data/org.mate.panel.gschema.xml.in
@@ -30,6 +30,11 @@
       <summary>Maximum history size for "Run Application" dialog</summary>
       <description>Controls the maximum size of the history of the "Run Application" dialog. A value of 0 will disable the history.</description>
     </key>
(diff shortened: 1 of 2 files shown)


lukefromdc:
This works great, and the setting is respected. Seems to be ready to go

lukefromdc:
Works with GTK 3.18 too in another test

raveit65:
Thanks that works great.
At least it's a question of personal taste or habit.

Issue #225, opened 2014-08-05
The Menu Bar panel applet inconsistently populates Places


flexiondotorg:
Hi,

The Places pop-out in the Menu Bar panel applet doesn't display the folders that are created by `xdg-user-dirs-update`. This is inconsistent with the Places shown in Caja, which does include the xdg folders.

Here is the original issue from the Ubuntu MATE bug tracker on Launchpad.

  * https://bugs.launchpad.net/ubuntu-mate/+bug/1341148

Regards, Martin.

monsta:
Hmm, Caja shows these dirs for me, but I could only find a place in the code where it loads the old location - https://github.com/mate-desktop/caja/blob/master/src/caja-bookmark-list.c#L120...

flexiondotorg:
@raveit65 @lukefromdc Given then recent update in https://github.com/mate-desktop/caja/pull/762, can you take a look at this for consistency?

raveit65:
Indeed, menubar applet use old location

raveit65:
@flexiondotorg 
All menus applets (Menubar, Main menu, mate-menu [mint-menu-fork)) uses the old location.
Maybe this is handled by mate-menu package?

raveit65:
Found it it is defined here
```
mate-panel/panel-menu-items.c:60:#define BOOKMARKS_FILENAME      ".gtk-bookmarks"
```

flexiondotorg:
Was just about to post that :-)

monsta:
So it's fixed for the panel now, but I wonder where Caja gets the list of these places. Looks like they're hardcoded somewhere. You don't need to have them in the bookmarks file, Caja will show them anyway. :slightly_smiling_face:

muesli4:
I just found out that Caja looks in `~/.gtk-bookmarks`. I will look at that.

PR #568, opened 2017-04-25
panel: Use new GTK bookmarks location if using GTK+ >= 3.5.1

raveit65:
This is a runtime check, obviously.

https://bugzilla.gnome.org/show_bug.cgi?id=675393

taken from:
https://git.gnome.org/browse/gnome-panel/commit/?id=19a23e8
https://git.gnome.org/browse/gnome-panel/commit/?id=af73783

Fixes https://github.com/mate-desktop/mate-panel/issues/225#issuecomment-296941888

We like to follow caja changes.

### DIFF
--- a/mate-panel/panel-menu-items.c
+++ b/mate-panel/panel-menu-items.c
@@ -57,7 +57,6 @@
 #include "panel-util.h"
 #include "panel-schemas.h"
 
-#define BOOKMARKS_FILENAME      ".gtk-bookmarks"
 #define MAX_BOOKMARK_ITEMS      100
 
 G_DEFINE_TYPE(PanelPlaceMenuItem, panel_place_menu_item, GTK_TYPE_IMAGE_MENU_ITEM)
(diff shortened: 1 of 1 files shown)


raveit65:
This works with Main Menu and Menubar, it looks like that other menus like mate-menu (mintmenu-fork) needs an update too.

raveit65:
@XRevan86 
Can you please take a look at mate-menu, please?

monsta:
We already require GTK+ >= 3.14 :smile:

raveit65:
well, i can change that, but i thaught it is a good idea to remind at the date where the new location was added by gtk+.

raveit65:
> I think it's enough to have a link to upstream commit in our commit message. I don't see any benefits from having some code that will never be executed...

What do you mean?
What should i do?
Please change it :-)

monsta:
Sure, I'll add a commit for squashing later :slightly_smiling_face:

raveit65:
Yes, this works too :-)

Issue #520, opened 2017-01-27
Analogue clock in time applet stretched when hovering over set button


mrferda:
I have set up multiple locations in the clock applet.
When you hover over one of the locations to change to it, the analogue clock is stretched when the button appears (see attached screenshot).
![clock](https://cloud.githubusercontent.com/assets/24938498/22389154/8c93be42-e4e4-11e6-85dd-2d86a641ca98.png)

bcv:
[me too!](https://www.bountysource.com/issues/41407980-analogue-clock-in-time-applet-stretched-when-hovering-over-set-button?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)

raveit65:
i saw that too, but i was too lazy to file out a report ;-)

phillip-white-sociomantic:
Same for me

lukefromdc:
I just found this is theme-sensitive, the issue does not appear in my UbuntuStudio_Legacy theme but did appear in a test in green-submarine

ekotoyants:
Its still [stretching](https://kotoyants.com/public/img_20170620105521.gif) but in different direction.

raveit65:
Not here ;-)

lukefromdc:
That's  a different problem: the entire calendar window being stretched by the "set" button appearing on hover. What theme are you using?

PR #577, opened 2017-05-20
Fixing damaged clock image

raveit65:
Fixes https://github.com/mate-desktop/mate-panel/issues/520
Tested with master for gtk+-3.22 and 1.18 branch for gtk+-3.22 and 3.14
Please test.

### DIFF
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -327,7 +327,6 @@ clock_location_tile_fill (ClockLocationTile *this)
         gtk_box_pack_start (GTK_BOX (box), priv->current_marker, FALSE, FALSE, 0);
         gtk_box_pack_start (GTK_BOX (box), priv->current_spacer, FALSE, FALSE, 0);
         priv->button_group = gtk_size_group_new (GTK_SIZE_GROUP_VERTICAL);
-        gtk_size_group_set_ignore_hidden (priv->button_group, FALSE);
         gtk_size_group_add_widget (priv->button_group, strut);
 
         /*
@@ -340,7 +339,6 @@ clock_location_tile_fill (ClockLocationTile *this)
          * but they are never visible together).
          */
         priv->current_group = gtk_size_group_new (GTK_SIZE_GROUP_BOTH);
-        gtk_size_group_set_ignore_hidden (priv->current_group, FALSE);
         gtk_size_group_add_widget (priv->current_group, priv->current_marker);
         gtk_size_group_add_widget (priv->current_group, priv->current_spacer);
 
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -309,6 +309,7 @@ clock_location_tile_fill (ClockLocationTile *this)
         priv->current_label = gtk_label_new ("");
         gtk_widget_show (priv->current_label);
         gtk_widget_set_no_show_all (priv->current_button, TRUE);
+        gtk_widget_set_valign (priv->current_button, GTK_ALIGN_CENTER);
         gtk_container_add (GTK_CONTAINER (priv->current_button), priv->current_label);
         gtk_widget_set_tooltip_text (priv->current_button,
                                      _("Set location as current location and use its timezone for this computer"));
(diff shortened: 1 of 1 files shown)


lukefromdc:
Merged with 
https://github.com/mate-desktop/mate-panel/commit/69f396b25e993e04fe7efcdb8ff99c687d14d568
and
https://github.com/mate-desktop/mate-panel/commit/4de86661f4a6f67285333c67342c287f99deffeb

monsta:
Guys, does the second commit affect anything? I just wonder whether it might be needed in 1.18 branch or not.

lukefromdc:
I never tested them separately,but  the second commit removes unneeded code at the least.  The function removed is said not to work anymore so premumably it should not be used. Bringing it back when applying this to 1.18 would be a potential source of a bug it seems. At the very least it would require new testing to split it up.

raveit65:
clock: drop deprecated gtk_size_group_set_ignore_hidden
isn't needed to fix the issue, For some reasons i didn't seperate it in PR.
But take a look at https://developer.gnome.org/gtk3/stable/GtkSizeGroup.html#gtk-size-group-set-ignore-hidden.
This function is deprecated with 3.20 and it seems that it doesn't work with older gtk+.
I use this commit as patch in fedora with 1.18.x already.
So i am +1 for adding this to 1.18.

raveit65:
Btw. it is dropped from gnome-panel too.

monsta:
Ok, done in https://github.com/mate-desktop/mate-panel/commit/15f2b5650b00e947e49121bf9e9f31ffe9de4343.

ekotoyants:
There is still [problem](https://kotoyants.com/public/img_20170620105521.gif) with this applet.

lukefromdc:
That's not a problem with the analog clock image, it's the "set" button taking too much space and stretching the entire calendar window popup. I just tested for this with GTK 3.22.15 and every theme I have (including all the MATE themes) and could not duplicate it. What version of mate-panel, and what version of GTK 3?

ekotoyants:
Did you configure a location that has different week day like in example (there is Monday in Kahului when there is Tuesday in Odessa - 13 Hours Time Difference)?

Arch Linux
gtk3 3.22.16-1
mate-panel 1.18.3-1
BlueMenta Theme

monsta:
Took me a while to find the existing report about this new issue. It was reported in the wrong repo: https://github.com/mate-desktop/mate-applets/issues/101.

raveit65:
I added Odesa and Kahului location (my home is berlin), and use Odesa at home location.
Hovering with the mouse over the edit button of Kahului does not produce any failure here.
I can't reproduce it with BlueMenta.
@ekotoyants 
Are you sure this is BlueMenta?
The button style doesn't look like BlueMenta, maybe Adwaita or Arc-theme?
Which version of mate-themes?
I suggest to open an new report at mate-panel for this.

ekotoyants:
Sorry my fault, I am using customized BlueMenta with "Controls" Clearlooks and "Icons" Faenza.

raveit65:
A customized BlueMenta.......
Than you have a error with handling the border of the button for different button states.
.... a so called jumping button.
To be sure can you test origin BlueMenta from last release for gtk+-3.22, please?
http://pub.mate-desktop.org/releases/themes/3.22/mate-themes-3.22.12.tar.xz
You can simply copy the Bluementa theme folder to /usr/share/themes/ and reload the theme.
And move your custom theme first to another dir.

monsta:
Kahului is already 06:01 at the moment, meaning it's the same day (Wednesday), so the line isn't long enough to trigger the issue.

raveit65:
Ok, adding Brisbane (Thuesday in Australia) trigger the issue here.
@ekotoyants 
Can you please open a new report for it?
Sorry, we can't move the existing report from mate-applets to mate-panel at github.
...than i can confirm the issue.

ekotoyants:
[Screen](https://kotoyants.com/public/img_20170622100616.gif) with BlueMenta and DualScreen (tested on both single and dual screen).

ekotoyants:
Should I open a new report in mate-panel? What about [mate-applets#101](https://github.com/mate-desktop/mate-applets/issues/101)? It is still open won't it be a duplicate report?

raveit65:
Yes, please at mate-panel, than we can close the other one which is at the wrong place.

ekotoyants:
[mate-panel#594](https://github.com/mate-desktop/mate-panel/issues/594)

Issue #494, opened 2016-11-02
segfault when some files at /usr/share/applications changes


Safari77:
When I updated google-chrome-stable, mate-panel 1.16.0 segfaulted in gtk3-3.22.2 function gtk_image_definition_unref because def is NULL:

```
def->empty.ref_count--;
```

I can crash it simply by doing
```
echo "" >> /usr/share/applications/google-chrome.desktop
```

```
audit: CONFIG_CHANGE auid=500 ses=4 op="updated_rules" path="/usr/opt/google/chrome/chrome" key=(null) list=4 res=1
audit: ANOM_ABEND auid=500 uid=500 gid=500 ses=4 pid=9653 comm="mate-panel" exe="/usr/bin/mate-panel" sig=11
kernel: mate-panel[9653]: segfault at 4 ip 00007fa24ae54c10 sp 00007ffdf23d4dc8 error 4 in libgtk-3.so.0.2200.2[7fa24ac4e000+6f0000]
journal: Negative content width -5 (allocation 1, extents 3x3) while allocating gadget (node button, owner GtkToggleButton)
journal: Negative content width -2 (allocation 1, extents 0x3) while allocating gadget (node box, owner GtkBox)
journal: gsignal.c:2523: signal 'size_request' is invalid for instance '0x3457bd0' of type 'GtkLabel'
dbus-daemon[4448]: [system] Activating service name='org.mate.SettingsDaemon.DateTimeMechanism' requested by ':1.464' (uid=500 pid=11463 comm="/usr/libexec/mate-panel/clock-applet ") (using servicehelper)
journal: Negative content width -9 (allocation 1, extents 5x5) while allocating gadget (node button, owner GtkToggleButton)
dbus-daemon[4448]: [system] Successfully activated service 'org.mate.SettingsDaemon.DateTimeMechanism'
```

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  gtk_image_definition_unref (def=0x0) at gtkimagedefinition.c:249
(gdb) bt
#0  0x00007fa24ae54c10 in gtk_image_definition_unref (def=0x0) at gtkimagedefinition.c:249
#1  0x00007fa24ade85db in gtk_drag_source_site_destroy (data=0x557e4debecf0) at gtkdragsource.c:117
#2  0x00007fa248c2a9f3 in g_datalist_clear (datalist=<optimized out>) at gdataset.c:273
[...]

Issue #555, opened 2017-04-07
mate-panel: gtk_image_definition_unref(): mate-panel killed by SIGSEGV


smoge:
https://bugzilla.redhat.com/show_bug.cgi?id=1409226

lukefromdc:
I have not seen this on Debian Unstable with GTK 3.22 and any version of mate-panel compatable with it.  Wonder what is different. Could one of the many build time warnings point to something that Federa's core library build options or kernel configuration are stricter about? This is not the first time I've seen a bug report of segfaults that could not be duplicated on Debian against a package-and I think but am not sure last time around was also on Fedora.

Also, can this be duplicated with mate-panel 1.18 or (git master) 1.19? Also, might something have gotten updated since mate-panel was last built in Fedora that caused this?

lukefromdc:
There was mention in the Fedora report of someone first seeing the crash after a Chrome update, I don't have Chrome installed. One user was using Xnomad, another was not, yet another said Fedora 26 did not have this and called Fedora 25 "buggy."  Also need to know exactly what applets are being used, and if any were built in-process. A crashing in-process applet takes down the panel, an out of process applet usually does not.

raveit65:
I can't reproduce that in fedora.
Updating the system with 'dnf update' does not have any influence on the panel.
But i only use package from fedora and not packages from external 'google repos` like chrome.

monsta:
Might be the same as https://github.com/mate-desktop/mate-panel/issues/494 because the backtraces are similar (not exactly the same though).

smoge:
It happens with me all the time on Fedora 25. I use XMonad as the WM, but other people reported having this problem without XMonad as well. A few other people reported this problem, but it seems to happen with just a few of us for some reason. The maintainer thinks it might be related to Chrome because it's from a third-party repo. I can't confirm this. Fedora's maintainer couldn't help much. Let me know if I can provide more information.

raveit65:
>  Fedora's maintainer couldn't help much.

Fedora mantainer asked for a new stacktrace and you didn't provide this info.
How/Why should a volunteers help you if you ignore questions?

raveit65:
Btw. in this stacktrace is ktelnet mentioned at the end of Thread 1
https://bugzilla.redhat.com/attachment.cgi?id=1270253
```
[...]

PR #581, opened 2017-05-23
menu: fix crash when some .desktop file has empty Icon field

monsta:
fixes https://github.com/mate-desktop/mate-panel/issues/494

### DIFF
--- a/mate-panel/menu.c
+++ b/mate-panel/menu.c
@@ -1177,14 +1177,12 @@ create_menuitem (GtkWidget          *menu,
 
 		if (matemenu_tree_entry_get_icon (entry) != NULL) {
 			const char *icon;
-			char       *icon_no_ext;
+			GIcon      *gicon;
 
 			icon = matemenu_tree_entry_get_icon (entry);
-			if (!g_path_is_absolute (icon)) {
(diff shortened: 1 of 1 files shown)


raveit65:
...we have another open report, hopefully the fix helps here too.
https://github.com/mate-desktop/mate-panel/issues/555

raveit65:
Ok, it seems that both issues are fixed with this commit.
echo "" >> /usr/share/applications/google-chrome.desktop
or 
yum-deprecated update / dnf update
are working fine now if ggz-gtk-client is installed.
I just tested this with mate-panel-1.19.0 / 1.18.2 and 1.16.2.
Looks like it is saved to back port it to 1.18/16.
Thanks

monsta:
Oops, `gtk_drag_source_set_icon_gicon` doesn't exist in GTK+2. I'll have to correct the port to 1.16...

raveit65:
@monsta 
We have a little remaining issue.
```
echo "" >> /usr/share/applications/google-chrome.desktop
```
generates this warning in .xsession-errors.
```
(mate-panel:7038): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```
Same happens if you add the menubar to the panel.

PR #582, opened 2017-05-24
menu: remove incorrect code from the popup menu of a menuitem

monsta:
fixes https://github.com/mate-desktop/mate-panel/issues/305

should be tested with GTK+ 3.14/16/18

### DIFF
--- a/mate-panel/menu.c
+++ b/mate-panel/menu.c
@@ -417,71 +417,6 @@ add_menu_to_panel (GtkWidget      *menuitem,
 	matemenu_tree_item_unref (directory);
 }
 
-/*most of this function stolen from the real gtk_menu_popup*/
-static void
-restore_grabs(GtkWidget *w, gpointer data)
-{
-	GtkWidget *menu_item = data;
-	GtkMenu *menu = GTK_MENU (gtk_widget_get_parent (menu_item));
-	GtkWidget *xgrab_shell;
-	GtkWidget *parent;
-
-	/* Find the last viewable ancestor, and make an X grab on it
-	 */
-	parent = GTK_WIDGET (menu);
-	xgrab_shell = NULL;
-	while (parent) {
-		gboolean viewable = TRUE;
-		GtkWidget *tmp = parent;
-
-		while (tmp) {
-			if (!gtk_widget_get_mapped (tmp)) {
-				viewable = FALSE;
-				break;
-			}
-			tmp = gtk_widget_get_parent (tmp);
-		}
-
-		if (viewable)
-			xgrab_shell = parent;
-
-		parent = gtk_menu_shell_get_parent_shell (GTK_MENU_SHELL (parent));
-	}
-
-	/*only grab if this HAD a grab before*/
-	/* FIXME fix for GTK3 */
-#if 0
-	if (xgrab_shell && (GTK_MENU_SHELL (xgrab_shell)->have_xgrab))
-          {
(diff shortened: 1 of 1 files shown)


lukefromdc:
This PR had no harmful effects on a GTK 3.22 build, no obvious behavioral change. Needs a test with GTK 3.18 or earlier from someone able to duplicate the nonresponsive panel.

monsta:
@raveit65: can you give it a try please? This is the last remaining fix, and it can be used for 1.18.

raveit65:
Already tested :-)
It fixes the freeze of the panel with m-p-1.18.x on rhel7 with MATE gtk3 from my corps repo.
> This is the last remaining fix,

??? I am in doubt, because menu doesn't close normal if you click on the desktop or in another application window.
A click on the panel is needed to close the menu.
Sorry, i see no reason to revert http://pkgs.fedoraproject.org/cgit/rpms/mate-panel.git/tree/mate-panel_0023-GTK-3-panel-Remove-popup-menu-for-items-in-applicati.patch for fedora builds.
So this is only a partial fix for https://github.com/mate-desktop/mate-panel/issues/305
It should be re-opened and renamed after the PR is merged.

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/a94e7d192e39fda1d5bba1921c676fec699ac533
after a GTK 3.18 test. 

Seemed to work fine here, though with current glib et all I was unable to duplicate the issue of the popup not closing on clicking the desktop, even from a newly added launcher. I don't recall having trouble with this when I was using GTK 3.18 and earlier. Still, incorrect code is incorrect code, and it obviously caused issues on some systems so it's gone.

monsta:
I should've said "the last remaining open PR"...

In #305 we actually had two issues instead of one. The panel freezing was the one people complained about, as I saw. That's what the report was initially about. I was going to close it and open a fresh one about the menu not closing. I didn't expect any fix for it anytime soon.

Last two days of experimenting showed me that the second issue might be actually solved soon. I figured out how to correct the hacks to make them work in GTK+3 :slightly_smiling_face:
These are still hacks, they're needed because GtkMenu/GtkMenuShell apparently isn't suitable for creating a popup menu on already opened popup menu (that's what is done when you right-click on a menuitem). But at least the changed hackish code seems to work with GTK+ 3.22 here...
I'll try it with 3.18 as well and then make another PR. If you wish to test it already, there's "menu-hacks" branch with the current work on it.

lukefromdc:
I just tested menu-hacks with GTK 3.22, seemed to work fine except for one instance where I was clicking around the menu and the mouse became unresponsive and I could not raise the run dialog. Killed mate-panel from another VT, and the mouse slowly recovered.  I had not yet killed and restarted the panel however! There had also been some menus not wanting to close properly after upgrade and before killing the panel. Once the panel had been restarted (and fully loaded the changes)  I could not duplicate any of those issues. Leaving this installed for further testing

monsta:
I always restart the panel after installing the modified version, so I never had such issues...

I've tried to reproduce it by rolling back to git master and restarting the panel, then building and installing menu-hacks but not restarting the panel this time. Clicked here and there, nothing bad happened...

lukefromdc:
We are probably fine then, and I probably ran into a hardware issue with the mouse freeze. I think we can put that one aside for now and get on with this fix. I've been running it for days just fine now.

PR #583, opened 2017-05-29
status notifier: follow same menu theme as rest of panel

lukefromdc:
Support any menu theme set up for the panel's menus differing from the rest of the Gtk theme using the .mate-panel-menu-bar style class Without this, anything using the status notifier will have mismatched menus if the GTK theme sets a different menu style for the panel (as mine does)

Tested with the "drop all elevated privileges" menu from mate-polkit built with appindicator support, which has an unrelated issue with the status icon (not the menu) not disappearing after dropping priviliges
https://github.com/mate-desktop/mate-polkit/issues/35

 Needs testing with another program using it, preferably one producing multiple menu items

### DIFF
--- a/applets/notification_area/status-notifier/sn-dbus-menu.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu.c
@@ -352,9 +352,24 @@ static void
 sn_dbus_menu_constructed (GObject *object)
 {
   SnDBusMenu *menu;
+  GtkWidget *toplevel;
+  GdkScreen *screen;
+  GdkVisual *visual;
+  GtkStyleContext *context;
 
   G_OBJECT_CLASS (sn_dbus_menu_parent_class)->constructed (object);
   menu = SN_DBUS_MENU (object);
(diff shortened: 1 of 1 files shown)


monsta:
I can try it, but do we have such theme(s) in mate-themes package?

lukefromdc:
There is no such theme in the mate themes, but my own theme at 
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy
uses a totally different theme for panel menus, with grey menus on applicatiions and semi-transparent black panel menus copied from early gnome-shell and Cinnamon's theme based on that.  It will give a very visible test. 

All of our panel theme support began as local patches I used to support that exact theme. I would be happy to see it added to the mate-themes at some point, I suspect others might like it too.

lukefromdc:
Rebased and merged with 
https://github.com/mate-desktop/mate-panel/commit/c27ba68a23fed89c9b9065c7ddd9cbfa3dee5095

Issue #504, opened 2016-12-04
Window list, tray rendered incompletely until restart after moving to another panel


lukefromdc:
Using mate-panel 1.17 and GTK 3.22, if I drag the window list or tray from one panel to another, the drop works but only the grab handle is rendered until the panel is restarted. Dragging the window list to another panel can also crash the workspace switcher applet.
##

raveit65:
yes, i see that with other applets too, ie. cpu-freq applet.

raveit65:
And gtk+-3.20 is also affected.

monsta:
@lukefromdc: yes, we got some new info on this in https://github.com/mate-desktop/mate-panel/issues/516.

lukefromdc:
I still have no idea what causes this

monsta:
I found out it's reproducible with GTK+ 3.14 and 3.18 as well. Not with all applet types though... A few special applet types like logout button, lock screen button, drawer, etc. are rendered on the new panel properly.

monsta:
Surprise, I can reproduce it in 1.16 with GTK+2... :open_mouth:

monsta:
Here's the backtrace from GTK+2 build of 1.16.

```gdb
(gdb) bt full
#0  0x00007f9691f15428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
        resultvar = 0
        pid = 3944
        selftid = 3944
#1  0x00007f9691f1702a in __GI_abort () at abort.c:89
        save_stage = 2
        act = 
          {__sigaction_handler = {sa_handler = 0x10004d11b, sa_sigaction = 0x10004d11b}, sa_mask = {__val = {0, 140284674413088, 3, 0, 140284670683708, 18446744073709551615, 0, 140733790577632, 7315242843790799616, 140, 140284671125245, 140284674070384, 140, 140733790567552, 140284706310528, 140}}, sa_flags = -1846105745, sa_restorer = 0x7f9692398ee1}
        sigs = {__val = {32, 0 <repeats 15 times>}}
#2  0x00007f9692337111 in g_thread_abort (status=<optimized out>, function=function@entry=0x7f9692398ee1 "pthread_setspecific")
    at /build/glib2.0-prJhLS/glib2.0-2.48.2/./glib/gthread-posix.c:77
[...]

PR #584, opened 2017-05-29
fix moving applets between panels

monsta:
fixes https://github.com/mate-desktop/mate-panel/issues/504

have to repeat my change from 024f89bc7c8074f753120d9c8ec4bd839745984b
because of fallout from 225702b3c6a609065465c66f0f2739fc1d2184a9

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -2754,10 +2754,12 @@ panel_widget_reparent (PanelWidget *old_panel,
 	if (gtk_container_get_focus_child (GTK_CONTAINER (old_panel)) == applet)
 		focus_widget = gtk_window_get_focus (GTK_WINDOW (old_panel->toplevel));
 
-	g_object_ref (applet);
-	gtk_container_remove (GTK_CONTAINER (old_panel), applet);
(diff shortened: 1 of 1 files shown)


raveit65:
LOL
```
Do not touch until GTK+4
```

monsta:
Does it actually work in gnome-panel? I recall I've tested it some time ago, and the issue was there... unless my memory fails me :slightly_smiling_face:

raveit65:
> Does it actually work in gnome-panel?

I didn't test it, i saw only this 2 commits
https://git.gnome.org/browse/gnome-panel/commit/?id=f9d94265eea95e2b319a90c1896f94b1bf48ba4d
https://git.gnome.org/browse/gnome-panel/commit/?id=2ad0b229e3d08cfdc3c91ffccb515409d5502adc

monsta:
Yes, I saw these commits, that's indeed the same code as we had before.
I found gnome flashback session 3.20 in one of my VMs (GTK+ is 3.22) and tested the panel there. It's not the latest version, but moving applets between panels actually works... my memory fails me then :slightly_smiling_face:

PR #589, opened 2017-06-06
Fix closing of main menu

lukefromdc:
Fix https://github.com/mate-desktop/mate-panel/issues/305

### DIFF
--- a/mate-panel/menu.c
+++ b/mate-panel/menu.c
@@ -669,60 +669,7 @@ drag_end_menu_cb (GtkWidget *widget, GdkDragContext     *context)
 
   if (xgrab_shell)
     {
-      gboolean status;
-      GdkDisplay *display;
-      GdkDevice *pointer;
-      GdkDevice *keyboard;
-#if GTK_CHECK_VERSION(3, 20, 0)
-      GdkSeat *seat;
-#else
-      GdkDeviceManager *device_manager;
-#endif
-      GdkWindow *window = gtk_widget_get_window (xgrab_shell);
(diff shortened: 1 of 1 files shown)


monsta:
Please note that the branch also has a fix for restoring the menu focus after you drag-n-drop a menu item to the panel. However, this seems only relevant for GTK+ 3.20/22.

With GTK+2 and 3.14/18 the menu automatically closes when you drag-n-drop a menu item from it. Therefore, no focus restore is needed.
With GTK+ 3.20/22 the menu doesn't close, but it loses the focus. I couldn't determine where the focus moves then, but it's surely not the menu and not the currently active app. After that you'll have to use the mouse to get out of that... the keyboard doesn't seem to do anything.

So, at first I removed another hack from the code, and the focus started to move to the currently active app. You can see this behavior if you use "remove-weird-code" branch. (I left that branch for testing, I won't merge it as it's not a complete fix.)
Later I found out I can actually improve the hack to bring the focus back to the menu. That is what the second commit here does.

I don't know why the menu behaves different in various GTK+ versions, but I'd like you to test it with these versions to make sure I didn't break anything. :slightly_smiling_face:

lukefromdc:
GTK 3.18 and GTK 3.22 builds both seem to work fine, the menu closing in GTK 3.18 and staying open in GTK 3.22 after a drag and drop to the panel is confirmed and has the advantage that this makes it easier to drag multiple launchers to the panel.  Both menubar and menubutton tested-and I found a bug in my theme with the menubar and GTK 3.18 that I will now fix...

monsta:
@raveit65: I hope the d-n-d fix is ok for you too :slightly_smiling_face:

PR #593, opened 2017-06-20
mate-panel: Dynamically apply a wmclass to control panel shadow

ikeydoherty:
In order to further facilitate mate-desktop/marco#327, this change
will dynamically set the wmclass on each panel start to one respecting
alpha settings.

That is to say, in only the themed (PANEL_BACK_NONE) mode, do we want
to actually have a shadow on mate-panel. To do this, we'll maintain
the "mate-panel" wmclass, which will be dealt with by Marco on map.
For all other cases we'll change the wmclass to "mate-panel-alpha"
to indicate that we're an RGBA window with non theme styling, and
as such do not expect any shadowing.

Signed-off-by: Ikey Doherty <ikey@solus-project.com>

### DIFF
--- a/mate-panel/panel-profile.c
+++ b/mate-panel/panel-profile.c
@@ -544,6 +544,13 @@ panel_profile_load_background (PanelToplevel *toplevel)
 
 	image = get_background_image (toplevel, &fit, &stretch, &rotate);
 
+	/* Only allow transparency for non themed panels */
+	if (background_type == PANEL_BACK_NONE)
+		panel_toplevel_update_wmclass (toplevel, FALSE);
+	else
+		panel_toplevel_update_wmclass (toplevel, TRUE);
+
+
 	panel_background_set (background,
 			      background_type,
 			      &color,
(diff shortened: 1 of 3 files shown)


ikeydoherty:
This PR is dependent on https://github.com/mate-desktop/marco/pull/327

lukefromdc:
I just tested this, it seems to create no problems with my theme (which keeps shadows OFF all panel items explicitly in either Marco with the related PR or with compiz. In fact, I didn't see panel shadows with any theme, but presumably that's just mate-themes not trying to set a shadow that doen't work with current code. I wanted to make sure I didn't get a nasty square shadow border around my translucent menus with rounded corners, nor on the panel itself. No issues, so I'm guessing this just adds suppport for using a shadow in the theme.

The code in the  PR includes
` "/* Only allow transparency for non themed panels */"`
Though I found it is possible (as it should be) to explicitly set a transparent theme in the GTK theme as well as with the panel preferences dialog. Found some theming issues with doing so, looking to see if that's just my theme or something else. No difference in that behavior with or without this PR though.

lukefromdc:
Found a workaround for including an opaque image background on the panel in the GTK theme: draw it on  PanelToplevel>grid  and not on the toplevel itself, where the rendering bug it. Draw it also on the panel plugs as they cannot be transparent in the GTK theme. I haven't seen any GTK themes with a transparent/translucent panel, but images were used in the GNOME 2/GTK2 days. Ubuntu's Mutiny layout uses a transparent panel-but I'm pretty sure that's a custom panel background as otherwise the GTK theme would have had to change.

Issue #596, opened 2017-06-25
Theme default transparent color panel backgrounds don't work on applets


lukefromdc:
#### Expected behaviour
Applying a transparent or translucent background to the panel in the GTK theme (not the panel prefs dialog) should turn the entire panel that color. If the color is transparent or translucent the desktop should be visible underneath it.

#### Actual behaviour
Transparent or translucent background shown on panel, but renders over black background in panel applets. If a transparent background is selected by a user from the panel preferences dialog it works fine,  but if applied from the GTK theme the applets render over a black background. The alpha value is actually applied and can be seen, but it is over a black background window and not over the desktop background as the panel is

#### Steps to reproduce the behaviour

Install mate-panel built with GTK 3.20 or 3.22. Probably also will occur with GTK 3.18 but did not test

Edit any GTK theme to apply a transparent color background to the panel. Either apply it to all of .mate-panel-menu-bar/gnome-panel-menu-bar or to BOTH PanelToplevel (themes the panel but not the applets) and #PanelPlug(themes the applets)

#### MATE general version
1.19 from GIT master

#### Package version
locally compiled

#### Linux Distribution
Debian Unstable with GTK 3.22.15

lukefromdc:
Fixed by https://github.com/mate-desktop/mate-panel/pull/597

PR #597, opened 2017-06-25
GTK 3.18+ Make transparent panel BG set in GTK theme work again

lukefromdc:
Transparent panel backgrounds set in GTK themes have not worked since the GTK 3.20 PanelPlug work. This was because NULL in gtk_window_set_background_pattern breaks on alpha values. Force this transparent and the GTK background is properly drawn

Fixes https://github.com/mate-desktop/mate-panel/issues/596

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1538,7 +1538,8 @@ mate_panel_applet_change_background(MatePanelApplet *applet,
 	_mate_panel_applet_apply_css(GTK_WIDGET(applet->priv->plug),type);
 	switch (type) {
 	case PANEL_NO_BACKGROUND:
-		gdk_window_set_background_pattern(window,NULL);
+		pattern = cairo_pattern_create_rgba (0,0,0,0);     /* Using NULL here breaks transparent */
+		gdk_window_set_background_pattern(window,pattern); /* backgrounds set by GTK theme */
 		break;
 	case PANEL_COLOR_BACKGROUND:
 		gdk_window_set_background_rgba(window,color);


lukefromdc:
Needs testing with GTK 3.18 and 3.20. Code changed is not in the GTK 3.18 or earlier build path but should be checked. No existing MATE theme uses a transparent panel BG so this is not critical but GTK does support this. I checked every MATE theme and my own in GTK 3.22 and no undesired side effects were visible.

Still working on image support, which is more complex.

lukefromdc:
GTK 3.20 worked fine, GTK 3.18 had issues with drawing the plugs more transparent than the panel toplevel when this was used. That can be fixed in the theme, an alpha of 0.8 in plugs works well with 0.5 in the panel for example. We could document that somewhere, or alternately I can put this behind a GTK 3.20 and later-only selector. GTK 3.18 always had issues with screwy rendering of alpha values and this turned out to be another example.

lukefromdc:
I will go ahead and merge this. Works perfectly in GTK 3.20 and in 3.22, and in GTK 3.18 the behavior is much better than before and the mismatch can be themed out if someone wants to write a theme with a semitransparent panel. 

The existing behavior cannot be themed out and comes from code that is logically incorrect, as NULL in gtk_window_set_background_pattern inherits from the parent window, but this is the plug toplevel window with nothing to inherit from, thus cairo's default black underneath whatever the theme puts on the plug window. That's what breaks transparency support. Custom backgrounds work because the code path does not hit that.

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/0b60b4bc66c8aae24a255b3bf82e07a59a2a9d07

PR #600, opened 2017-06-27
Add a gsettings key to enable/disable SNI Support

clefebvre:
SNI Support triggers many changes and regressions.

This PR introduces a gsettings key to be able to disable it.

### DIFF
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -64,15 +64,22 @@ static GfStatusNotifierWatcher *sn_watcher_service = NULL;
 static GfStatusNotifierWatcher *
 sn_watcher_service_ref (void)
 {
-  if (sn_watcher_service != NULL)
-    g_object_ref (sn_watcher_service);
-  else
+  GSettings *settings;
+  settings = g_settings_new ("org.mate.panel");
+
+  if (g_settings_get_boolean (settings, "enable-sni-support") == TRUE)
     {
-      sn_watcher_service = gf_status_notifier_watcher_new ();
-      g_object_add_weak_pointer ((GObject *) sn_watcher_service,
-                                 (gpointer *) &sn_watcher_service);
+      if (sn_watcher_service != NULL)
+        g_object_ref (sn_watcher_service);
+      else
+        {
+          sn_watcher_service = gf_status_notifier_watcher_new ();
+          g_object_add_weak_pointer ((GObject *) sn_watcher_service,
+                                     (gpointer *) &sn_watcher_service);
+        }
     }
 
+  g_object_unref (settings);
   return sn_watcher_service;
 }
 #endif
--- a/data/org.mate.panel.gschema.xml.in
+++ b/data/org.mate.panel.gschema.xml.in
@@ -84,5 +84,10 @@
       <summary>Disable Force Quit</summary>
       <description>If true, the panel will not allow a user to force an application to quit by removing access to the force quit button.</description>
     </key>
+    <key name="enable-sni-support" type="b">
+      <default>true</default>
+      <summary>Enable SNI support</summary>
+      <description>If true, the panel provides support for SNI.</description>
+    </key>
   </schema>
 </schemalist>


alexarnaud:
Le 27/06/2017 à 16:11, Clement Lefebvre a écrit :
> SNI Support triggers many changes and regressions.

Do you have report all regression you've found ? It's essential for 
low-vision and blind users to have SNI support so it would be great if 
we could fix regressions related to the SNI support. SNI allows keyboard 
navigation between apps into the notification area (a blind person user 
uses keyboard-only).

So if you want you're distribution fully usable by keyboard users you 
shouldn't disable SNI support.

I'm really sorry for the off-topic.

Kind regards.

raveit65:
> SNI Support triggers many changes and regressions.

Can you give some more details about what is wrong with SNI Support?

rilian-la-te:
For example, to use alternative SNI applet:-))

clefebvre:
Hi, sure.

I've got two examples in mind:

- With SNI support nm-applet switches to an indicator and is no longer left-clickable. Without, it falls back to behaving like it did in the past.
- With SNI support indicator-keylock always show its icon in the notification area, no matter what states the lock keys are. Without it, it works correctly.

The issue isn't really with how the SNI support is implemented, it's with the fact that there's a whole lot of apps out there which indicator support is poor and which work better without than with.

I can't remember other examples from the top of my head, but we've implemented indicator support in Cinnamon a long time ago, and after a few releases and a huge amount of negative feedback, we've resorted to disable it by default.

This PR does no harm, it simply adds the ability to disable it. I also think it should be disabled by default, but that's a distribution decision so I left it ON here in this PR.

I would also strongly recommend to have an option in mate-control-center to enable/disable this support graphically. This requires new translations though so it's best to implement this on its own within the dev cycle in time for 1.20.

lukefromdc:
I tested this, it worked both with and without status notifier support. I don't have anything installed that gives a visible difference, however.

raveit65:
> I've got two examples in mind:

> - With SNI support nm-applet switches to an indicator and is no longer left-clickable. Without, it falls back to behaving like it did in the past.

I can't reproduce this in fedora, nm-applets works like before here, linuxmint issue ?

> - With SNI support indicator-keylock always show its icon in the notification area, no matter what states the lock keys are. Without it, it works correctly.

Do you mean our own implementations?
```
[rave@mother ~]$ gsettings get org.mate.peripherals-keyboard-xkb.general duplicate-leds
false
```
Or some third-party (ubuntu) indicator?
I ask because this needs to be debug if you speak about MATE stuff.

clefebvre:
@raveit65 could be an Ubuntu specificity, something to do with libindicator. In any case, disabling SNI support certainly helps here.

rilian-la-te:
@clefebvre No, this is not Ubuntu-specific)). 
Just use ```nm-applet --indicator```.

I am using my ```vala-panel-sntray-mate```, and it switches left and right click actions (if we compared it to KDE).

I am supporting @clefebvre 's pull request, because it will allow us to develop alternative SNI applets, and we will not have duplicated trays then.

clefebvre:
Thanks @rilian-la-te, I had a look at nm-applet.. it's patched by Ubuntu to make --indicator the default behaviour.

```
$ more indicator-by-default.patch 
From: Mathieu Trudel-Lapierre <mathieu@canonical.com>
Subject: Have the appindicator enabled by default.
Forwarded: not-needed

---
 src/main.c |    8 +++-----
 1 file changed, 3 insertions(+), 5 deletions(-)

Index: b/src/main.c
===================================================================
--- a/src/main.c
+++ b/src/main.c
@@ -31,7 +31,7 @@
 
 gboolean shell_debug = FALSE;
 gboolean with_agent = TRUE;
-gboolean with_appindicator = FALSE;
+gboolean with_appindicator = TRUE;
 
 static void
 usage (const char *progname)
@@ -61,11 +61,9 @@ int main (int argc, char *argv[])
 			shell_debug = TRUE;
 		else if (!strcmp (argv[i], "--no-agent"))
 			with_agent = FALSE;
-		else if (!strcmp (argv[i], "--indicator")) {
+		else if (!strcmp (argv[i], "--no-indicator")) {
 #ifdef WITH_APPINDICATOR
-			with_appindicator = TRUE;
-#else
-			g_error ("Error: --indicator requested but indicator support not available");
+			with_appindicator = FALSE;
 #endif
 		}
 	}

```

So that explains why this particular app behaves differently in other distributions.

flexiondotorg:
Ubuntu MATE experiences issues where indicator-applet and notification area are both active. Upon starting the session some indicators that have fallback to na-tray appear in the tray (when they shouldn't), other indicators register with indicator-applet and some (such as nm-applet) appear twice in na-tray and the indicator. 

Passive indicators (update-manager) are always shown in na-tray, when they should not be.

Using this patch to disable SNI "solves" these issues in Ubuntu MATE. However I do realise it will impact visually impaired users and is masking an underlying issue.  I want to use this patch in Ubuntu MATE but would also like to see the underlying issue resolved. 

So do merge this or let distro maintainers carry this commit as a distro patch while the root cause is fixed? 

_Sent from my OnePlus ONEPLUS A5000 using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

rilian-la-te:
@flexiondotorg , this issues will be faced with any alternative SNI implementation, not only indicator-applet. And this issues is not ubuntu-related.

raveit65:
Ok, i think i will use it for fedora with 1.18 too as it make the SNI applet of dnfdragora visible in tray, this is the new mageria/fedora Packet-Manager GUI for dnf.

alexarnaud:
Le 29/06/2017 à 14:15, raveit65 a écrit :
> Ok, i think i will use it for fedora with 1.18 too as it make the SNI 
> applet of dnfdragora visible in tray, this is the new mageria/fedora 
> Packet-Manager GUI for dnf.

OK, if everybody wants to disable SNI support until the fix of the 
existing serious bugs I propose to change the implementation a little bit :
- disable by default SNI support
- enable SNI support if accessibility is enabled or if SNI support is 
enabled

Is it OK for you ?

Best regards.
-- 
Alex ARNAUD

clefebvre:
> until the fix of the existing serious bugs

These will never be fixed. On our side, these issues aren't specific to MATE's SNI support, they're inherent to indicators. Their API is inferior to statusicons by design and doesn't support as many features, so enabling this support de-facto lowers the interactions users can have with software supporting them.

The only way you would fix that, would be by either patching the indicator framework or by removing support in the many applications which are using it.

This is outside of the scope for MATE. What we need here is simply a way to disable that support, that's what I implemented in this pull request.

alexarnaud:
Le 14/07/2017 à 19:32, Clement Lefebvre a écrit :
>     until the fix of the existing serious bugs
> 
> These will never be fixed. On our side, these issues aren't specific to 
> MATE's SNI support, they're inherent to indicators. Their API is 
> inferior to statusicons by design and doesn't support as many features, 
> so enabling this support de-facto lowers the interactions users can have 
> with software supporting them.
> 
> The only way you would fix that, would be by either patching the 
> indicator framework or by removing support in the many applications 
> which are using it.

OK, if you're right some issues will stay present in the future. I'm not 
aware of the technical stuff about SNI so I can't answer on this subject.

> This is outside of the scope for MATE. What we need here is simply a way 
> to disable that support, that's what I implemented in this pull request.

You should think of keyboard-only users as blind people they need to be 
able to use the notification area. If we want to build an accessible 
desktop for all we need to think of people with need keyboard 
interaction. I don't ask you impossible things, just a little change for 
an experimented developer like you.
Do you imagine another way to take into account keyboard-only users ?

Best regards.
-- 
Alex ARNAUD

flexiondotorg:
I've been using this in Ubuntu MATE 17.10 for several weeks now with the appropriate gschema override and it does remedy the issues I was experiencing.

Issue #595, opened 2017-06-23
Image backgrounds in GTK3 theme do not actually work


lukefromdc:
#### Expected behaviour
An image background for PanelToplevel or .mate-panel-menu-bar set in the GTK theme should be displayed on the panel, same as if it was a user-set custom background. In GTK2 builds of GNOME 2, some themes used an image background, notably UbuntuStudio. 

#### Actual behaviour
An image background will show up fine on the PanelPlugs if they are specified or .mate-panel-menu-bar is used to specify it, but on the panel toplevel it will appear "stretched thin" and almost transparent.  

#### Steps to reproduce the behaviour
Install mate-panel built with GTK 3.22 (maybe 3.18 or later-not tested), and modify the selected GTK theme to set an image backgrond on PanelToplevel and on  #PanelPlug, or with the image background set on on .mate-panel-menu-bar . Use an opaque image to most clearly show the problem.  Ensure the panel background is set to "non-use system theme" and restart mate-panel.

#### MATE general version
1.19.0 from git master as of 6-22-2017

#### Package version
locally compiled

#### Linux Distribution
Debian Unstable with locally built gtk3.22.15 and glib 2.53.1 and Debian's provided libcairo2 version 1.14.8-1

lukefromdc:
There is a codepath in panel-background.c to explicitly handle image backgrounds, but it does not actually workThere is a bug in cairo scaling that makes opaque images translucent if "stretched" too far, rather like rubberized almost opaque material might do in the real world. 

![mate-panel_with_image_bg_in_gtk_theme](https://user-images.githubusercontent.com/14036792/27496608-b19c56c8-5824-11e7-8177-cfe7a1903267.png)

Thus with the current code and a vertical panel, two different rendering problems appear with an image file of a bar with a gradient on it: only the center of the gradient shows on the panel toplevel, and it is nearly transparent due to being stretched by a factor of 1/1920 on my monitor, which apparently means it is pulled to 1920 times its original length-and not all pixels get the color. In GTK2 with circa 2008 cairo what appears to have been this same code worked, and the original UbuntuStudio theme got away with a panel background image just 1 px wide which the code rendered across the entire screen.

 Here is the offending part of "panel_background_prepare" in panel-background.c:

	```
case PANEL_BACK_NONE:
		if (background->default_pattern) {
			/* the theme background-image pattern must be scaled by
			* the width & height of the panel so that when the
			* backing region is cleared
[...]

PR #601, opened 2017-06-28
Make bg image set in GTK theme work

lukefromdc:
Make panel image backgrounds set in the GTK theme work, like they did in GTK2. Tested with all supported GTK3 major versions 4.14 through 3.22.  Both image and gradient backgrounds work, best rendering in recent GTK versions but all will work. Image file bg is stretched to fit, same as on a user-set image background.

Fixes https://github.com/mate-desktop/mate-panel/issues/595

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1776,18 +1776,23 @@ static void _mate_panel_applet_prepare_css (GtkStyleContext *context)
 	GtkCssProvider  *provider;
 
 	provider = gtk_css_provider_new ();
+
 	gtk_css_provider_load_from_data (provider,
-					 ".mate-custom-panel-background{\n"
+					 "#PanelPlug {\n"
+					 " background-repeat: no-repeat;\n" /*disable in gtk theme features */
+					 " background-size: cover; "        /*that don't work on panel-toplevel */
+					 " }\n"
+					 ".mate-custom-panel-background{\n" /*prepare CSS for user set theme */
 					 " background-color: rgba (0, 0, 0, 0);\n"
 					 " background-image: none;\n"
 					 "}",
 					 -1, NULL);
+
 	gtk_style_context_add_provider (context,
 					GTK_STYLE_PROVIDER (provider),
 					GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
 	g_object_unref (provider);
 }
-
 static void
 mate_panel_applet_init (MatePanelApplet *applet)
 {
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1776,7 +1776,7 @@ static void _mate_panel_applet_prepare_css (GtkStyleContext *context)
 	GtkCssProvider  *provider;
 
 	provider = gtk_css_provider_new ();
-
+#if GTK_CHECK_VERSION (3, 18, 0)
 	gtk_css_provider_load_from_data (provider,
 					 "#PanelPlug {\n"
 					 " background-repeat: no-repeat;\n" /*disable in gtk theme features */
@@ -1787,7 +1787,14 @@ static void _mate_panel_applet_prepare_css (GtkStyleContext *context)
 					 " background-image: none;\n"
 					 "}",
 					 -1, NULL);
-
+#else
+gtk_css_provider_load_from_data (provider,
+					 ".mate-custom-panel-background{\n" /*prepare CSS for user set theme */
+					 " background-color: rgba (0, 0, 0, 0);\n"
+					 " background-image: none;\n"
+					 "}",
+					 -1, NULL);
+#endif
 	gtk_style_context_add_provider (context,
 					GTK_STYLE_PROVIDER (provider),
 					GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
--- a/mate-panel/panel-background.c
+++ b/mate-panel/panel-background.c
@@ -112,7 +112,9 @@ panel_background_prepare (PanelBackground *background)
 			cairo_surface_t *surface;
 			double width, height;
 
+			surface = NULL;
 			cairo_pattern_get_surface(background->default_pattern, &surface);
+			cairo_surface_reference(surface);
 			width = cairo_image_surface_get_width (surface);
 			height = cairo_image_surface_get_height (surface);
 
@@ -124,6 +126,7 @@ panel_background_prepare (PanelBackground *background)
 
 			gdk_window_set_background_pattern (background->window,
 						       background->default_pattern);
+			cairo_surface_destroy(surface);
 		} else
 			gdk_window_set_background_rgba (
 				background->window, &background->default_color);
--- a/mate-panel/panel-background.c
+++ b/mate-panel/panel-background.c
@@ -113,19 +113,32 @@ panel_background_prepare (PanelBackground *background)
 			double width, height;
 
 			surface = NULL;
+			width = 1.0;
+			height = 1.0;
 			cairo_pattern_get_surface(background->default_pattern, &surface);
-			cairo_surface_reference(surface);
-			width = cairo_image_surface_get_width (surface);
-			height = cairo_image_surface_get_height (surface);
+			/* catch invalid images (e.g. -gtk-gradient) before scaling and rendering */
+			if (surface != NULL ){
+				cairo_surface_reference(surface);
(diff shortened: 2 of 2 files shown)


clefebvre:
@lukefromdc which theme do you use to test?

In the code, do we need the "\n" in the CSS? they look like they could be removed.

lukefromdc:
Test theme is
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy/tree/dev-panel-bg
which is a testing branch of my UbuntuStudio_Legacy theme that will later be merged to master.

The /n endings for each line in the CSS is consistant with other in-code CSS loaded elsewhere in MATE, including the CSS already present in the same file. I would imagine GTK does not care one way or the other, but I've never seen the code written without it.

lukefromdc:
I just found that a -gtk-gradient background causes a segfault, and is used in Traditional Green and Traditional OK. The gradients never worked there, can be made to work by a change to a linear gradient. Within GTK, the gtk-gradient is deprecated (it predated the css gradient support) and can only be read by a special function. I will propose fixes for these two themes, their backgrounds will then work, and I will have to come up with error-handling code to catch this, and ideally to print a warning that -gtk-gradient is not supported.

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/601/commits/0e44ef5ae406562657b0843860a6a52cbd052df0  Stops the segfaults by catching any unsupported images such as -gtk-gradient and warns the user/theme developer with text in terminal (though not in syslog) to use a css gradient or an image.

The lets the traditional-green and TOK themes work as they stand, though the panel gradients don't work and a solid color results. This code change makes it work, but the difference is so subtle as to be invisible to my eye. Testing with the colors changed in gtk.css confirmed the new code in mate-applications.css to be functional however:

The old code for the traditional themes panel background:
```

.mate-panel-menu-bar,
MatePanelApplet > menubar,
PanelToplevel.background,
PanelWidget,
PanelFrame,
PanelAppletFrame,
MatePanelAppletFrameDBus,
MatePanelAppletContainer,
MatePanelApplet {
    background-image: -gtk-gradient (linear,
                      left top,
                      left bottom,
                      from (@menubar_gradient_a),
                      color-stop (0.50, @menubar_gradient_b),
                      to (@menubar_gradient_c));
```

The new code for the traditional themes panel background:
```

.mate-panel-menu-bar,
MatePanelApplet > menubar,
PanelToplevel.background,
PanelWidget,
PanelFrame,
PanelAppletFrame,
MatePanelAppletFrameDBus,
MatePanelAppletContainer,
MatePanelApplet {
    background-image: linear-gradient( 
					  to bottom,
                      @menubar_gradient_a,
                      @menubar_gradient_b 50%,
                      @menubar_gradient_c);
    background-color: @theme_bg_color;

```

raveit65:
I can confirm that Traditional themes crashes the panel here, btw. i can't start the panel any more with this theme configuration.
Your proposal for the background-image works well here.
But do we need this?
Or should we simply use a bg color like before?
Personal i am fine with this theme change, let's see what Traditional themes maintainer thinks.
@sc0w ^^^^

raveit65:
I can confirm that Traditional themes crashes the panel here, btw. i can't start the panel any more with this theme configuration.
Your proposal for the background-image works well here.
But do we need this?
Or should we simply use a bg color like before?
Personal i am fine with this theme change, let's see what Traditional themes maintainer thinks.
@sc0w ^^^^

lukefromdc:
Please retest with the last commit from last night in place. That one catches the error and allow the panel to start and run with the Traditional themes as they stand now. I agree that such a tiny and nearly invisible gradient is probably a waste, but that's the theme maintainer's choice. My guess is it was a workaround for a long ago bug.

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/601/commits/771c8b23e1b31f058d4066b553789dead4e92f97
Cleans up image rendering in GtkPlugs in the gtk3.14 and 3.16 cases where the css for the later versions did not work as the plug handling changes with GTK 3.18 and later. Tested with all supported GTK versions and an image background file.

raveit65:
> Please retest with the last commit from last night in place. That one catches the error and allow the panel to start and run with the Traditional themes as they stand now.

This works fine now, the panel don't get a segfault if -gtk-gradient () is used.

raveit65:
I tested the full commit stack again on 3.22 and 3.14, seems to be OK.
But honestly, using a bg image in theme needs a rewrite of the panel theme settings.
Not sure if i ever will find the time to do that in a theme.
Maybe this work gives you the experience to fix the usage of an image bg in panel preferences.
Any ways, nice work :-)

monsta:
I found a lot of "background-image: -gtk-gradient" in 3.18 themes. Tested them all, no crashes.

```
$ grep -RI '\-gtk-gradient' /usr/share/themes | cut -d '/' -f 5 | sort | uniq
BlackMATE
BlueMenta
Blue-Submarine
GreenLaguna
Green-Submarine
Menta
Mint-X
Mint-X-Aqua
Mint-X-Blue
Mint-X-Brown
Mint-X-Grey
Mint-X-Orange
Mint-X-Pink
Mint-X-Purple
Mint-X-Red
Mint-X-Sand
Mint-X-Teal
TraditionalGreen
TraditionalOk
TraditionalOkTest
```

monsta:
I'd replace g_printerr with g_warning to have the process name (i.e. mate-panel) automatically prepended to the message. Currently the messages don't give any hints about it:
```
unsupported panel image background such as -gtk-gradient
use an image file or a standard css gradient
```

lukefromdc:
All those other instances of -gtk-gradient work because they are read directly by GTK, which uses  gtk_gradient_resolve() to create the cairo_pattern_t . An attempt to read a -gtk-gradient with this function:

```
gtk_style_context_get (context, state,
	                       "background-color", &bg_color,
	                       "background-image", &bg_image,
	                       NULL);

```
in panel-toplevel.c does not return a valid cairo_pattern_t, and calling cairo_image_surface_get_width () on the result of cairo_pattern_get_surface() applied to the invalid cairo_pattern_t caused the segfault. Thus the code to catch that case and warn the user. I'll test replacing g_printerr with g_warning now

When I tried letting GTK draw the background directly on the panel (by disabling code that overrides it), it got rendered on top of the menus and panel launchers and most items that were not in a plug. That would have given the simplest code and supported anything GTK supported but just didn't work. Could probably be done by going back through all those other widgets but would be a lot of work and could create a whole new set of corner case bugs and regressions.

lukefromdc:
The g_warning for an unsupported image works fine, yielding this text in terminal on a -gtk-gradient:

```
** (mate-panel:30032): WARNING **: unsupported panel image background such as -gtk-gradient


** (mate-panel:30032): WARNING **: use an image file or a standard css gradient

```

That is enough to let theme developers know what they've done wrong.

To read a -gtk-gradient, I'd have to find a way to detect it in the theme in panel-toplevel, then replace the invalid results of  gtk_style_context_get() with  a call to gtk_gradient_resolve() and do it only in that case and no other.  Other unsupported images such as a missing image file or maybe an odd file format are also possible, so it's not enough to test for a valid cairo_pattern_t and the GTK docs say this sort of handling should never be necessary, that all the gtk_gradient functions are deprecated and normally used only inside GTK.

https://developer.gnome.org/gtk3/stable/gtk3-GtkGradient.html

At any rate, that could be another PR late if it can even be made to work. Also, you are right: the user-background image handling is extremely complex, yet does not handle options GTK can like a tiled background. I could not use it to handle a theme background because it requres a GdkPixbuf, not a cairo_pattern_t and I could find no way to create one from the theme. Gtk uses the actual path to the file.

lukefromdc:
squashed, rebased, and merged with
https://github.com/mate-desktop/mate-panel/commit/a1445bc208e22260c27ce67ad4ebce6040ff5a3c

monsta:
I'd like to make the warning fit on one line and, if possible, make it tell about other unsupported properties. As I understand it, `-gtk-gradient` isn't the only unsupported property?

lukefromdc:
I have yet to find any other unsupported properties in a theme, so I do not know what they all are but suspect anything understood by GTK3 and not by Cairo would fail.  
That's why I wrote the warning the way I did. Feel free to change it to just
 "unsupported -gtk-gradient in panel image background "
and see if any reports of other problems come in if you want

monsta:
Ok, reworded it in https://github.com/mate-desktop/mate-panel/commit/4e2f9b1c421ef94c5912cc294771accc4191ab73, hope it's better.

lukefromdc:
I like that change, makes it a bit clearer where the issue is. Thanks

Issue #594, opened 2017-06-22
Hovering over timezone text, causes panel to scale beyond screen limit


ekotoyants:
When you have cities in your location list with different weekday, the text is very long and the window stretches beyond the screen limit.

Examples for [single](https://kotoyants.com/public/img_20170620105521.gif) and [dual](https://kotoyants.com/public/img_20170622100616.gif) monitor configuration. 

**Packages:**
mate-panel 1.18.3
gtk3 3.22.16

**OS:**
Arch Linux
Mate 1.18

PR #602, opened 2017-06-29
Dev clock

raveit65:
This fixes https://github.com/mate-desktop/mate-panel/issues/594
and need to test with different languages.
I tested it with german locale which use mostly long words.
So i guess it should work with other languages too, otherwise i can increase the values a bit, but this increase the size of the whole window too.

### DIFF
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -481,6 +481,7 @@ calendar_window_init (CalendarWindow *calwin)
 	gtk_window_set_type_hint (window, GDK_WINDOW_TYPE_HINT_DOCK);
 	gtk_window_set_decorated (window, FALSE);
 	gtk_window_set_resizable (window, FALSE);
+	gtk_window_set_default_size (window, 337, -1);
 	gtk_window_stick (window);
 	gtk_window_set_title (window, _("Calendar"));
 	gtk_window_set_icon_name (window, CLOCK_ICON);
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -123,7 +123,7 @@ calendar_window_create_calendar (CalendarWindow *calwin)
 	struct tm                  tm1;
 
 	calendar = gtk_calendar_new ();
-	gtk_widget_set_size_request(GTK_WIDGET(calendar), 260, 100);
+	gtk_widget_set_size_request(GTK_WIDGET(calendar), 330, 100);
 	options = gtk_calendar_get_display_options (GTK_CALENDAR (calendar));
 	if (calwin->priv->show_weeks)
 		options |= GTK_CALENDAR_SHOW_WEEK_NUMBERS;
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -318,6 +318,7 @@ clock_location_tile_fill (ClockLocationTile *this)
         priv->current_marker = gtk_image_new_from_icon_name ("go-home", GTK_ICON_SIZE_BUTTON);
         gtk_widget_set_halign (priv->current_marker, GTK_ALIGN_END);
         gtk_widget_set_valign (priv->current_marker, GTK_ALIGN_CENTER);
+        gtk_widget_set_margin_start (priv->current_marker, 75);
         gtk_widget_set_no_show_all (priv->current_marker, TRUE);
 
         priv->current_spacer = gtk_event_box_new ();
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -288,6 +288,7 @@ clock_location_tile_fill (ClockLocationTile *this)
         gtk_box_pack_start (GTK_BOX (head_section), priv->city_label, FALSE, FALSE, 0);
 
         priv->time_label = gtk_label_new (NULL);
+        gtk_label_set_width_chars (GTK_LABEL (priv->time_label), 20);
         gtk_widget_set_margin_end (priv->time_label, 3);
 #if GTK_CHECK_VERSION (3, 16, 0)
         gtk_label_set_xalign (GTK_LABEL (priv->time_label), 0.0);


raveit65:
You need to add another location which use another day than your location.
Try Brisbane (australia) or Wellington (New Zealand)

lukefromdc:
I was using two locations added so one could be hovered for selection, but I have never been able to see the issue

lukefromdc:
I found https://github.com/mate-desktop/mate-panel/issues/603 in testing this, BTW

lukefromdc:
I added those two locations, and got a theme-independent jumping button by just a couple px. With master, turns out my own theme stays stable and all the others jump beyond the screen borders.

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/602/commits/5c52d4db732b04f2efe5c8a11813cf6d2d52bdc6 stops the jump with the two locations listed and in English, more testing would be a good idea. Works perfectly here

raveit65:
Without your commit i have a jumping with using Kahului (Hawai), here is Thuesday.
The german translation for Thueday = Donnerstag is to large. I need to increase size of time_label.
And i will test your improvement.

raveit65:
I rebased PR to
```
gtk_label_set_width_chars (GTK_LABEL (priv->time_label), 20);
and
gtk_widget_set_margin_start (priv->current_marker, 63);
```
A smaller value than 63 for priv->current_marker, the home icon which is at the same place than the (set-) button, does not work in my language.
The german translation for set = Einstellen is much larger than in englisch.

This works with american english and german language for me.
I will test some other languages over the day.
But testing in other languages by other ppls is welcome.

raveit65:
Rebased again for ukraine language which seems to use the largest word for the (set-)button.
I tested this with a lot of languages, including main languages like english, chinese, spain or french.
Now:
```
gtk_label_set_width_chars (GTK_LABEL (priv->time_label), 20);
and
gtk_widget_set_margin_start (priv->current_marker, 75);
```
Well the width of the GUI looks a bit larger now.

lukefromdc:
That last change (to a 74 px margin) really makes the window jump. When the locations are opened with long labels and the clock near the right end of the panel, the window jumps partially off-screen. In the process, the background grid image in my theme gets dragged with it out of position. On closing and reopening the screen edge constrains the jump. Finally, a re-test in Menta verified that this is theme-independant, the grid in mine just helps illustrate what is going on.

The attached tarball is a video showing the jumping I got, in my theme and in menta. This is not ready

It seems that setting that margin to anything beyond 60px brings back the jumping.
[jumper_dev-clock.mp4.tar.gz](https://github.com/mate-desktop/mate-panel/files/1116145/jumper_dev-clock.mp4.tar.gz)

raveit65:
Ok,
i can reproduce the off-screen issue if the the clock is open without locations first.
But i don't see the button jumping.
Well the large values for priv->current_marker is needed for rusian and ukraine languages which use a large word in the button.
I will try set a width for the whole window which is used if the clock is open in calendar only mode.
...tomorrow.

raveit65:
Setting a default width fixes the window jumps if you first open only the calendar and use the expander to view the locations.
Changing the size_reguest of the calendar fixes the window-offset  if you first open only the calendar.
I tested this with your theme too, sadly the calendar_grid.svg looks weird now.
Maybe be you can fix that with using another padding values in your theme.
Sorry, i don't know another solution.
All that shows me that the clock window has in general a bad design.
If you do not agree i can use this as patches for fedora.

raveit65:
@monsta Can you please test this too?
....because your language use a large word for the (set)-button ;-)
.... which is one of the 2 causes of this report https://github.com/mate-desktop/mate-panel/issues/594

raveit65:
Hmm, now BlackMate has some issues here. :-/

raveit65:
Another rebase to fix BlackMate.

raveit65:
Btw. Do you have an idea why the locations icons are so large with BlackMate?
Otherwise
```
gtk_widget_set_size_request(GTK_WIDGET(calendar), 330, 100);
```
should be sufficient.

raveit65:
Ok this setting make the icons so large with BlackMate
https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/BlackMATE/gtk-3.0/gtk-widgets.css#L3393

An this setting would override it
```
#MatePanelPopupWindow > frame > box > box > box > box > widget image,
#MatePanelPopupWindow > frame > box > box > box > box > widget label {
    padding: 0px;
}

#MatePanelPopupWindow > frame > box > box > box > box > widget button > label {
    padding: 3px;
}
```

lukefromdc:
If the final fix for this moves the grid around on my theme, I can easily move it back, just so long as it is always moved to the same place and does not jump. If it moves only when locations are open that's less of an issue as I normally run the calendar with them closed.

raveit65:
I rebased it to final
```
gtk_widget_set_size_request(GTK_WIDGET(calendar), 330, 100);
```
With your theme i saw only that the grid nedds some padding changes, no jumping anymore.
Hope this works on your box too.
For BlackMate we need to add the probosed setting to mate-application.css.
I can do that if you agree.

lukefromdc:
I have no trouble with adding the necessary revisions to BlackMATE

lukefromdc:
I like the latest revision, as it seems to solve a longstanding problem in my theme. I will make a new calendar grid bg image to match it.

Until now, the size of the calendar window was controlled by the font, meaning a font change, at times even the difference between Debian and Ubuntu could screw up that calendar grid's scaling. This fixed size stops that, and removes a potential source of breakage in my theme. I would suggest adding a minimum height to the calendar window in other themes, though it is up to each theme maintainer to decide to prioritize space vs a window closer to square.   Use of a larger font to take advantage of the space makes a more readable calendar.

lukefromdc:
Here's how my calendar looks with the new grid and this PR, quite nice:

calendar_fixed

Vertical size cannot be directly set in themes it turns out, as "calendar" does not respond to "min-height" . It comes up in GtkInspector but is ignorerd. The fonts and the padding around them can control it, as they thus still do in my theme. That means I still have to hardcode fonts for the calendar in my theme so user/OS set themes can't break it though.  Good enough, let's merge this and apply the fix to BlackMATE.

lukefromdc:
Adding this code to BlackMATE is what fixes both the oversized clock face (image) and the jump(label):

```
/*fix oversize calendar window locations icons and jump */
window#MatePanelPopupWindow image, label {
	padding: 0px;
}

```

This is necessary because those get set to 3px elsewhere in the theme and I didn't want to risk causing a bug elsewhere. It was found by a process of elimination

lukefromdc:
I just tested this with all the supported GTK versions in the process of updating my theme. 

Here are my test results:

GTK 3.22, 3.20: perfect, no jump, all themes fine after revision to BlackMATE and to UbuntuStudio_Legacy(my theme)

GTK 3.14, 3.16, 3.18: No jump on opening locations, small  jump to a few px narrower on right side on hovering over the upper of two locations set(Wellington). Basically workable but not perfect. Themes work fine

raveit65:
Ok, i will merge it like it is and will do a test with 3.14 the next days, or if i find time.
Theme settings for the panel are complete different before the rewrite for 3.20.
At least there is no distro which use MATE gtk3 on < 3.20, as far i know.
So that is negligible at the moment.

raveit65:
BlackMate is updated, some days we need to clean up the obsolete panel settings.

raveit65:
Btw. i am very happy that you found a solution for your theme, clock looks nice there.

lukefromdc:
We've got one remaining problem, and that seems to be the icon for "heavy rain" rendering way oversize in all themes, forcing a larger window, a larger clock face-and a jump to the left to accomodate the window. Kicks my grid out of alignment. 

It's an old bug predating this commit. A rollback to last month's mate-panel gave a nasty vertically-distorted clockface for Wellington with that heavy rain forecast. Now at least it is round, so we've made progress on a second issue as well. 

I found that Wellington had a forecast for heavy rain today, and replacing that location with (dry) Washington DC "fixed" the oversize rendering. When adding Wellington pack, the problem returned, again ONLY on Wellington with the heavy rain forecast. Other clockfaces were normal size, and without Wellington the window worked as yesterday.

Note that a location with heavy rain in the forecast must be found to test this!

I will now go looking in libmateweather to see if we are shipping an icon that won't scale.

lukefromdc:
I think I found the problem: libmateweather only ships a partial icon set, and icons in the user's theme are otherwise (maybe always?) used. My icon theme is to blame, I don't have a small version of that icon so the bug is on my end and I can fix it here and be done with it. The problem did not appear in the full Menta theme with the icons, so that's that.

lukefromdc:
Confirmed: fixing my icon theme (missing smaller icons) fixes the Wellington/rain forecast issue. Odd though that libmateweather seems to draw different icons from different places

raveit65:
Cool, that you could fix it, thx

PR #606, opened 2017-07-06
Fix GTK theme image bg support on vertical panel

lukefromdc:
Themers can use the .vertical and .horizontal style classes to rotate a background on vertical panels the same way scrollbars are themed. Apply these style classes to PanelPlugs too so this can actually be used.

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1533,6 +1533,8 @@ mate_panel_applet_change_background(MatePanelApplet *applet,
 				    GdkRGBA* color,
 				    cairo_pattern_t *pattern)
 {
+	GtkStyleContext* context;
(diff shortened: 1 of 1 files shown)


lukefromdc:
Tested in Debian Unstable with both GTK 3.22 and GTK 3.16 to ensure both codepaths work. Test theme is https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy  includng latest commit https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy/commit/8c04b88f0ace512fe449484090b2048f3aee09e6 to reorient the background on vertical panels.
One vertical and one horizontal panel test case (same as for Mutiny) with applets in PanelPlugs in each

raveit65:
merged https://github.com/mate-desktop/mate-panel/commit/b7d2033dff3f68d6ac67290c8b90203906e6762f
Thank you

Issue #585, opened 2017-05-31
[1.16] Notification area icons disappears when apply a background image to the panel


caringi:
When I try to set a background image for a panel, if this panel has a notification area, all their icons disappears.
![mate_panel_problem_before_background](https://cloud.githubusercontent.com/assets/2509235/26624810/5c2c0984-45f2-11e7-8363-e082de2f7fff.png)
![mate_panel_problem_after_background](https://cloud.githubusercontent.com/assets/2509235/26624817/61853acc-45f2-11e7-858f-b167a3fc0812.png)

lukefromdc:
What version of mate-panel, and is it built with GTK2 or GTK3? If GTK3, what GTK3 version?  MATE 1.18 always uses GTK3, and some distros such as Ubuntu and Debian build MATE 1.16 with GTK3. MATE 1.14 and earlier is almost always supplied built with GTK2 by distros.

Also, what kind of image? I can set a png or jpg image and never get this, the notification area icons show up fine on my panel. I have compiled mate locally with GTK3 all the way back to 1.10 and never seen this, but if this is with GTK2 that is why.

caringi:
I'm using MATE shipped with Fedora 25 (1.16.2).
It's using GTK3 (3.22.15):

```
$ ldd /usr/bin/mate-panel|grep libgtk
	libgtk-3.so.0 => /lib64/libgtk-3.so.0 (0x00007fce1f570000)

$ pkg-config --modversion gtk+-3.0
3.22.15
```
The image is in png format, I'm sending it attached, but it happens with all the images that I've tried (doesn't matter the image format)...

I have been using different versions of MATE in different systems in the last years and never seen this behavior before.

![panel_background1](https://cloud.githubusercontent.com/assets/2509235/26670046/62881ff2-46b0-11e7-9cef-f7498097f8fa.png)

lukefromdc:
I just tested your image, tray icons worked fine for me, using GTK 3.22.15 on Debian Unstable. I've never seen this, can't duplicate it. Can one of the other Fedora users here test this?

monsta:
@raveit65 please check if it happens in your Fedora versions :slightly_smiling_face:

raveit65:
I can confirm that with
```
[rave@f25 ~]$ rpm -qa mate-panel
mate-panel-1.16.2-2.fc25.x86_64
```
But not with
```
[rave@f26 ~]$ rpm -qa mate-panel
[...]

PR #607, opened 2017-07-07
System tray: fix hidden icons on image/transparent color BG

lukefromdc:
Includes https://github.com/mate-desktop/mate-panel/pull/606 to support rendering GTK theme backgrounds on vertical panels by attaching the .vertical and .horizontal style classes to PanelPlugs as it is based on that branch.

Also force a relative background on na-tray child objects to force the icons to appear on a custom image background or in the case of a noncompositing WM a transparent color bg (which acts like an image BG). Fix https://github.com/mate-desktop/mate-panel/issues/585

### DIFF
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -76,7 +76,7 @@ na_tray_child_realize (GtkWidget *widget)
       gdk_window_set_composited (window, TRUE);
       cairo_pattern_destroy (transparent);
 
-      child->parent_relative_bg = TRUE;
+      child->parent_relative_bg = FALSE;
     }
   else if (visual == gdk_window_get_visual(gdk_window_get_parent(window)))
     {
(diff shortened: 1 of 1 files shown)


raveit65:
Sorry, a rebase from master was needed.

lukefromdc:
Not a problem, I'm quite used to rebasing as this stuff evolves in parallel

lukefromdc:
From what I have examined, it looks like this callback in na-tray.c is the the only one that  calls na_tray_child_force_redraw when the applet has not been moved or resized. It runs when the applet is just sitting there. When it   only works with a relative bg, this means the icons don't get redrawn with real transparency, which seems to be the case always with GTK 3.10 or later. The modern versions of GTK3 use transparency and composit internally regardless of the window manager, and changes elsewhere may have created a race condition leading the the covered icons.

```
static gboolean
idle_redraw_cb (NaTray *tray)
{
  NaTrayPrivate *priv = tray->priv;

  g_hash_table_foreach (priv->trays_screen->icon_table,
                        (GHFunc) na_tray_child_force_redraw, NULL);
  
  priv->idle_redraw_id = 0;

  return FALSE;
}
```

lukefromdc:
Squashed (yielding a 1 line commit) and merged with
https://github.com/mate-desktop/mate-panel/commit/ab786fa5282be8169b9c94d258cc57d07944ed13

Issue #604, opened 2017-07-05
[1.18] Cascading applet crashes on login when using transparent panel background


epgfm:
#### Expected behaviour

Upon login, panel and applets should load normally

#### Actual behaviour

Upon login, a random amount of applets are crashing after receiving signal 5. (SIGTRAP) 

#### Steps to reproduce the behaviour

Unclear, this does not seem to affect everyone so it might be hardware related or config-specific.
One factor though is that this happens only with a transparent panel background.

#### MATE general version

1.18 

#### Package version

mate-panel 1.18.1 ( 1.18.1-0ubuntu2 )

#### Linux Distribution

Ubuntu MATE 17.04

#### Additional information

Originally discussed here:

https://ubuntu-mate.community/t/zesty-multiple-applet-crashes-on-login-when-using-transparent-panels/13192

At least one other user is affected by this.

I've noticed the applets on the left side of the top panel (Brisk menu, firefox launcher) seem unaffected by this: I don't remember those ever crashing.

Hardware info:

```
CPU:       Quad core Intel Celeron N3150 (-MCP-) cache: 1024 KB 
           clock speeds: max: 2080 MHz 1: 801 MHz 2: 571 MHz 3: 745 MHz
           4: 558 MHz
Graphics:  Card: Intel Atom/Celeron/Pentium Processor x5-E8000/J3xxx/N3xxx Integrated Graphics Controller
           Display Server: X.Org 1.19.3 drivers: modesetting (unloaded: fbdev,vesa)
           Resolution: 1366x768@60.06hz
           GLX Renderer: Mesa DRI Intel HD Graphics 400 (Braswell)
           GLX Version: 3.0 Mesa 17.0.3
```

Configuration: Marco + Compton

Attaching .crash files and the panel layout files to this report.

[ouroumov-tweak.layout.tar.gz](https://github.com/mate-desktop/mate-panel/files/1124069/ouroumov-tweak.layout.tar.gz)
[ouroumov-tweak.panel.tar.gz](https://github.com/mate-desktop/mate-panel/files/1124068/ouroumov-tweak.panel.tar.gz)
[_usr_lib_mate-applets_mate-multiload-applet.1000.crash.tar.gz](https://github.com/mate-desktop/mate-panel/files/1124072/_usr_lib_mate-applets_mate-multiload-applet.1000.crash.tar.gz)
[...]

Issue #369, opened 2015-11-11
GTK3: applets crashing on transparent panel if switching compositor off


raveit65:
Steps to reproduce:
Choose a transparency panel bg color.
Switch marcos compositor off
Most applets are crashing and you was ask to reload them.
##

monsta:
Confirmed. Only with stickynotes though, and it took me a lot of tries to trigger it.

Dammit. It fails exactly at the same place as in that bug with moving transparency slider too fast...
And X function that fails is properly wrapped in push/pop, so it should never crash there. I've moved push/pop closer to that function call but it didn't help at all.

```
(gdb) bt full
#0  0x00007f4acd09f88b in g_logv (breakpoint=1) at /build/glib2.0-ocmJ1Y/glib2.0-2.46.2/./glib/gmessages.c:324
        domain = 0x0
        data = 0x0
        depth = 1
        log_func = 0x7f4acd09f050 <g_log_default_handler>
        domain_fatal_mask = <optimized out>
        masquerade_fatal = 0
        test_level = <optimized out>
        was_fatal = <optimized out>
        was_recursion = <optimized out>
        msg = 0x1ae9610 "The program 'stickynotes-applet' received an X Window System error.\nThis probably reflects a bug in the program.\nThe error was 'BadPixmap (invalid Pixmap parameter)'.\n  (Details: serial 5358 error_cod"...
        msg_alloc = 0x1ae9610 "The program 'stickynotes-applet' received an X Window System error.\nThis probably reflects a bug in the program.\nThe error was 'BadPixmap (invalid Pixmap parameter)'.\n  (Details: serial 5358 error_cod"...
        i = 2
#1  0x00007f4acd09f88b in g_logv (log_domain=0x7f4ace77482e "Gdk", log_level=G_LOG_LEVEL_ERROR, format=<optimized out>, args=args@entry=0x7fff195aeae0)
    at /build/glib2.0-ocmJ1Y/glib2.0-2.46.2/./glib/gmessages.c:1081
        domain = 0x0
        data = 0x0
        depth = 1
        log_func = 0x7f4acd09f050 <g_log_default_handler>
        domain_fatal_mask = <optimized out>
        masquerade_fatal = 0
        test_level = <optimized out>
        was_fatal = <optimized out>
        was_recursion = <optimized out>
[...]

PR #612, opened 2017-07-11
fix applet crashes on transparent user-set panel bg

lukefromdc:
Avoid cairo function that seemed to be causing crashes related to setting geometry, use code copied from panel-background.c instead. Fixes https://github.com/mate-desktop/mate-panel/issues/604

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1350,8 +1350,10 @@ mate_panel_applet_get_pattern_from_pixmap (MatePanelApplet *applet,
 
 	width = gdk_window_get_width(window);
 	height = gdk_window_get_height(window);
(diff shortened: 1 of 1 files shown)


raveit65:
This fixes also https://github.com/mate-desktop/mate-panel/issues/369 for me

epgfm:
Sorry it took me a while to manage to test this. (building noob here)
I can confirm this fixes #604 on bare-metal using Marco + Compton.
Thanks a lot @lukefromdc, I feel like I have a stable system again, it's been a while. ^^

monsta:
I guess you can close https://github.com/mate-desktop/mate-panel/issues/613 then.

lukefromdc:
In all the work on this, I found an old GNOME commit that while it did not fix this problem when used by itself stopped applet crashes GNOME was getting on background changes. It is a fix for the variable pattern in the same function getting used uninitialized if the background is changing and can't be read.
https://github.com/mate-desktop/mate-panel/pull/619

PR #619, opened 2017-07-16
libmate-panel-applet: Fix another potential source of applet crashes

lukefromdc:
Properly initialize pattern, GNOME found this stopped applet crashes on bg change(though not the same crashes we just fixed)

https://github.com/GNOME/gnome-panel/commit/88b9beeb8ab69072fe6912af33c620143d17b0a3
We could return a random value from
panel_applet_get_pattern_from_pixmap(), and that was very nasty.
based on

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1329,6 +1329,8 @@ mate_panel_applet_get_pattern_from_pixmap (MatePanelApplet *applet,
 	cairo_t         *cr;
 	cairo_pattern_t *pattern;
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
Merged with https://github.com/mate-desktop/mate-panel/commit/cf46fb7f03d6ab0e106fa1cbfbf180fbbb9a5b75

monsta:
Hmm but this is already done at line 1367...

lukefromdc:
OK, I see that now, further down in the function. We can remove one or the other.

lukefromdc:
Reverted this as it is a duplicate of code already present

monsta:
I was a bit late with my comment :slightly_smiling_face:

Issue #468, opened 2016-08-08
GTK 3.21 regression: custom background problems with transparent panel with in-process applets


lukefromdc:
Somewhere between July 19 and July 22, a change to GTK 3.21.4 caused application of custom color backgrounds to fail on the clock button, tray, and show desktop button. Also corruption of window list button display until panel restarted. Solid color and image backgrounds still work, but the panel must be restarted to fully apply this.

Looks like 
https://git.gnome.org/browse/gtk+/commit/?id=2c7b21718f18ab743cd79784ae28cad582f58503
again, as my June 9 build tested OK but not the June 10 build-and all my "cajafix" builds from before we fixed the desktop also gave perfect custom background application.
##

lukefromdc:
Found a workaround: For GTK 3.21, build all applets out of process. Only in-process applets are affected, so building for GTK 3.22 with all applets out of process fixes this. In-process applets get redraw problems when custom backgrounds are used with GTK 3.21 so distros will need to know never to build applets in-process with this GTK version. Didn't test the fish but the clock, tray, and wncklet components were all affected, and I had them all in-process. Rebuilt with them out of process, no more rendering issues with any background or background change.

raveit65:
Looks like the issue is landed in fedora 25, but abrt will give me stacktraces.
I will attach them later.

raveit65:
Hmm, using '--with-in-process-applets=none' seems to fix the issue for applets in mate-panel package.
I noticed that system-monitor applet is also affected, of course it crashes very often.
But man pages told me that using 'none' is the default.
```
  --with-in-process-applets=APPLETS
                          comma-separated list of applets to compile
                          in-process (possible values: none, clock, fish,
                          notification-area, wncklet, all) [default=none]
```
Looks like the default does not work.

lukefromdc:
Building with no applets at all in-process is what stopped the rendering
bugs for me. Just added system monitor applet to my panel to test for 
a while, so far seeing no issues but by git master version of GTK3 is a
bit more recent too,.

monsta:
I wonder if it affects gnome-panel - IIRC it has in-process applets by default these days.

lukefromdc:
Gnome-panel has problems too-but I was able to fix them entirely in
my theme. I will not try a build of mate-panel with in-process applets
as see if it works there too. 

For gnome-panel, gnome-panel-menu-bar had to be explicltly set transparent
[...]

Issue #556, opened 2017-04-09
Wrong behavior for drawer applet


edomato:
![drawer](https://cloud.githubusercontent.com/assets/1550373/24839089/e0da0988-1d2a-11e7-8a89-4e08074af8f3.gif)

As you can see in the animated gif, when you open a drawer and close it, if you move the cursor in and out of the panel, the icon for the drawer changes to the first application in it when it you always display the drawer icon.

lukefromdc:
What version of mate-applets and mate-panel, is this GTK2 or GTK3, and if GTK 3 which version. Also over what distro? I can't duplicate this on Debian Unstable with GTK 3.22.11 and MATE 1.19 compiled locally from git master

edomato:
Ok, sorry for the lack of information :-(

The distribution is also Debian Unstable with the latest packages from that repository.

mate-applets 1.16.0-1
mate-panel 1.16.0-2
libgtk-3-0:amd64 3.22.11-1

DavideRossi:
I have the same behavior with Mint's latest version (18.2).
Not only the icon displayed is that of the first application but, at times, if you click in the drawer area the first application is started (and the drawer is not opened).

mate-applets 1.18.1-1
mate-panel 1.18.3-2
libgtk-3-0 3.18.9

MobiusTesseract:
Just went from Linux Mint 18.1 to 18.2, don't know what my previous versions were, but now I am on:
mate-applets 1.18.1-1+sonya
mate-panel 1.18.3-2+sonya
libgtk-3-0 3.18.9-1ubuntu3.3

I have the drawer on the bottom panel and never had a problem before, it just appeared. Since for a bottom mounted panel the first element of the drawer is the close button (instead of last) that is what appears instead and mine is big (since the drawer is wide the way I use it) and covers other elements as well. Unfortunately this is a deal-breaking bug for me if it's going to hide what is around it...

![image](https://user-images.githubusercontent.com/6500579/28324474-24fd30a6-6be4-11e7-8f06-192f1afdf10c.png)
![image](https://user-images.githubusercontent.com/6500579/28324555-61ee7c2c-6be4-11e7-8c5d-d0d106d8df2d.png)

mp107:
Went from Linux Mint MATE 18.1 to 18.2 and the issue came up also for me.

GIF:
![GIF](http://obrazki.elektroda.pl/8334091300_1501599408.gif)

Video (YouTube):
[...]

PR #622, opened 2017-07-18
Fix custom backgrounds on in-process applets with GTK 3.22

lukefromdc:
Fix https://github.com/mate-desktop/mate-panel/issues/468
which is a nasty failure to redraw foregrounds on in-process applets when custom backgrounds are used , thus adding them to the background. 

Note that the handles for applets must  be themed with the 
MatePanelAppletFrameDBus>MatePanelAppletFrameDBus
selector as the single instance of MatePanelAppletFrameDBus is also used on applet windows and will otherwise pick up the handle theming. This is hidden on out of process applets but shows through on in-process applets. This theming change also permits the handles to be rendered with custom backgrounds, which was not possible before.

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -763,6 +763,9 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
 
         if (cd->calendar_popup)
                 gtk_widget_destroy (cd->calendar_popup);
+/* FIXME this stops segfaults on removing the clock applet when it is*/
+/* built in-process but leaves part of it in RAM until the panel is restarted */
+#ifndef CLOCK_INPROCESS
         cd->calendar_popup = NULL;
 
         g_free (cd->timeformat);
@@ -790,6 +793,7 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
         }
 
         g_free (cd);
+#endif
 }
 
 static gboolean
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -791,7 +791,10 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
                 g_object_unref (cd->builder);
                 cd->builder = NULL;
         }
-#ifndef CLOCK_INPROCESS
+#ifdef CLOCK_INPROCESS
+        cd = NULL;
+        g_free (cd);
+#else
         g_free (cd);
 #endif
 }
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -501,6 +501,8 @@ get_updated_timeformat (ClockData *cd)
 static void
 update_timeformat (ClockData *cd)
 {
+        if (cd->settings == NULL)
+                return;
         if (cd->timeformat)
                 g_free (cd->timeformat);
         cd->timeformat = get_updated_timeformat (cd);
@@ -701,6 +703,8 @@ update_tooltip (ClockData * cd)
 static void
 refresh_clock (ClockData *cd)
 {
+        if (cd->settings == NULL)
+                return;
         unfix_size (cd);
         update_clock (cd);
 }
@@ -708,6 +712,9 @@ refresh_clock (ClockData *cd)
 static void
 refresh_clock_timeout(ClockData *cd)
 {
+        if (cd->settings == NULL)
+                return;
+
         unfix_size (cd);
 
         update_timeformat (cd);
@@ -765,10 +772,8 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
                 gtk_widget_destroy (cd->calendar_popup);
         cd->calendar_popup = NULL;
 
-        cd->timeformat = NULL;
         g_free (cd->timeformat);
 
-        cd->custom_format = NULL;
         g_free (cd->custom_format);
 
         free_locations (cd);
@@ -791,9 +796,8 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
                 g_object_unref (cd->builder);
                 cd->builder = NULL;
         }
-#ifndef CLOCK_INPROCESS
+
         g_free (cd);
-#endif
 }
 
 static gboolean
@@ -1993,6 +1997,9 @@ static void
 update_weather_bool_value_and_toggle_from_gsettings (ClockData *cd, gchar *key,
                                                  gboolean *value_loc, const char *widget_name)
 {
+        if (cd->settings == NULL)
+                return;
+
         GtkWidget *widget;
         gboolean value;
 
@@ -2072,6 +2079,9 @@ location_set_current_cb (ClockLocation *loc,
 static void
 locations_changed (ClockData *cd)
 {
+        if (cd->settings == NULL)
+                return;
+
         GList *l;
         ClockLocation *loc;
         glong id;
@@ -2275,6 +2285,9 @@ temperature_unit_changed (GSettings    *settings,
                           gchar        *key,
                           ClockData    *cd)
 {
+        if (cd->settings == NULL)
+                return;
+
         cd->temperature_unit = g_settings_get_enum (settings, key);
         if (cd->temperature_unit > 0)
         {
@@ -2293,6 +2306,9 @@ speed_unit_changed (GSettings    *settings,
                     gchar        *key,
                     ClockData    *cd)
 {
+        if (cd->settings == NULL)
+                return;
+
         cd->speed_unit = g_settings_get_enum (settings, key);
         if (cd->speed_unit > 0)
         {
@@ -2311,6 +2327,9 @@ custom_format_changed (GSettings    *settings,
                        gchar        *key,
                        ClockData    *clock)
 {
+        if (clock->settings == NULL)
+                return;
+
         gchar *value;
         value = g_settings_get_string (settings, key);
 
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -791,10 +791,7 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
                 g_object_unref (cd->builder);
                 cd->builder = NULL;
         }
-#ifdef CLOCK_INPROCESS
-        cd = NULL;
-        g_free (cd);
-#else
+#ifndef CLOCK_INPROCESS
         g_free (cd);
 #endif
 }
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -763,12 +763,12 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
 
         if (cd->calendar_popup)
                 gtk_widget_destroy (cd->calendar_popup);
-/* FIXME this stops segfaults on removing the clock applet when it is*/
-/* built in-process but leaves part of it in RAM until the panel is restarted */
-#ifndef CLOCK_INPROCESS
         cd->calendar_popup = NULL;
 
+        cd->timeformat = NULL;
         g_free (cd->timeformat);
+
+        cd->custom_format = NULL;
         g_free (cd->custom_format);
 
         free_locations (cd);
@@ -791,7 +791,7 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
                 g_object_unref (cd->builder);
                 cd->builder = NULL;
         }
-
+#ifndef CLOCK_INPROCESS
         g_free (cd);
 #endif
 }
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -995,6 +995,9 @@ static void display_fortune_dialog(FishApplet* fish)
 
 static void name_changed_notify(GSettings* settings, gchar* key, FishApplet* fish)
 {
+	if (fish->source_id == 0)
+		return;
+
 	char *value;
 
 	value = g_settings_get_string (settings, key);
@@ -1020,6 +1023,9 @@ static void name_changed_notify(GSettings* settings, gchar* key, FishApplet* fis
 
 static void image_changed_notify(GSettings* settings, gchar* key, FishApplet* fish)
 {
+	if (fish->source_id == 0)
+		return;
+
 	char *value;
 
 	value = g_settings_get_string (settings, key);
@@ -1771,9 +1777,6 @@ static gboolean fishy_factory(MatePanelApplet* applet, const char* iid, gpointer
 static void fish_applet_dispose (GObject *object)
 {
 	FishApplet* fish = (FishApplet*) object;
-
-	if (fish->source_id == 0)
-		return;
 
 	if (fish->timeout)
 	{
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -1772,6 +1772,9 @@ static void fish_applet_dispose (GObject *object)
 {
 	FishApplet* fish = (FishApplet*) object;
 
+	if (fish->source_id == 0)
+		return;
+
 	if (fish->timeout)
 	{
 		g_source_remove (fish->timeout);
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -591,9 +591,6 @@ static void set_ally_name_desc(GtkWidget* widget, FishApplet* fish)
 	AtkObject  *obj;
 	char       *desc, *name;
 
-	/*Return if the fish has been destroyed*/
-	if (!GTK_IS_WIDGET(widget))
-		return;
 	obj = gtk_widget_get_accessible (widget);
 	/* Return immediately if GAIL is not loaded */
 	if (!GTK_IS_ACCESSIBLE (obj))
@@ -1360,10 +1357,6 @@ static void update_pixmap(FishApplet* fish)
 	cairo_matrix_t matrix;
 	cairo_pattern_t *pattern;
 
-	/*Return if the fish has been destroyed*/
-	if (!GTK_IS_WIDGET(widget))
-		return;
-
 	gtk_widget_get_allocation (widget, &allocation);
 
 	if (!gtk_widget_get_realized (widget) ||
@@ -1635,10 +1628,6 @@ static void set_tooltip(FishApplet* fish)
 {
 	const char *desc_format = _("%s the Fish, the fortune teller");
 	char       *desc;
-
-	/*Return if the fish has been destroyed*/
-	if (fish)
-		return;
 
 	desc = g_markup_printf_escaped (desc_format, fish->name);
 	gtk_widget_set_tooltip_markup (GTK_WIDGET (fish), desc);
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -591,6 +591,9 @@ static void set_ally_name_desc(GtkWidget* widget, FishApplet* fish)
 	AtkObject  *obj;
 	char       *desc, *name;
 
+	/*Return if the fish has been destroyed*/
+	if (!GTK_IS_WIDGET(widget))
+		return;
 	obj = gtk_widget_get_accessible (widget);
 	/* Return immediately if GAIL is not loaded */
 	if (!GTK_IS_ACCESSIBLE (obj))
@@ -1357,6 +1360,10 @@ static void update_pixmap(FishApplet* fish)
 	cairo_matrix_t matrix;
 	cairo_pattern_t *pattern;
 
+	/*Return if the fish has been destroyed*/
+	if (!GTK_IS_WIDGET(widget))
+		return;
+
 	gtk_widget_get_allocation (widget, &allocation);
 
 	if (!gtk_widget_get_realized (widget) ||
@@ -1628,6 +1635,10 @@ static void set_tooltip(FishApplet* fish)
 {
 	const char *desc_format = _("%s the Fish, the fortune teller");
 	char       *desc;
+
+	/*Return if the fish has been destroyed*/
+	if (fish)
+		return;
 
 	desc = g_markup_printf_escaped (desc_format, fish->name);
 	gtk_widget_set_tooltip_markup (GTK_WIDGET (fish), desc);
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -153,8 +153,11 @@ static void destroy_tasklist(GtkWidget* widget, TasklistData* tasklist)
 
 	if (tasklist->properties_dialog)
 		gtk_widget_destroy(tasklist->properties_dialog);
-
+/* FIXME: this stops segfaults when removing in-process applet
+/* but leaves part of it in RAM until the panel is restarted */
+#ifndef WNCKLET_INPROCESS
 	g_free(tasklist);
+#endif
 }
 
 /* TODO: this is sad, should be used a function to retrieve  applications from
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -63,6 +63,7 @@ static void call_system_monitor(GtkAction* action, TasklistData* tasklist);
 static void display_properties_dialog(GtkAction* action, TasklistData* tasklist);
 static void display_help_dialog(GtkAction* action, TasklistData* tasklist);
 static void display_about_dialog(GtkAction* action, TasklistData* tasklist);
+static void destroy_tasklist(GtkWidget* widget, TasklistData* tasklist);
 
 static void tasklist_update(TasklistData* tasklist)
 {
@@ -145,19 +146,6 @@ static void applet_change_pixel_size(MatePanelApplet* applet, gint size, Tasklis
 	tasklist->size = size;
 
 	tasklist_update(tasklist);
-}
-
-static void destroy_tasklist(GtkWidget* widget, TasklistData* tasklist)
-{
-	g_object_unref(tasklist->settings);
-
-	if (tasklist->properties_dialog)
-		gtk_widget_destroy(tasklist->properties_dialog);
-/* FIXME: this stops segfaults when removing in-process applet
-/* but leaves part of it in RAM until the panel is restarted */
-#ifndef WNCKLET_INPROCESS
-	g_free(tasklist);
-#endif
 }
 
 /* TODO: this is sad, should be used a function to retrieve  applications from
@@ -696,4 +684,32 @@ static void display_properties_dialog(GtkAction* action, TasklistData* tasklist)
 	gtk_window_set_resizable(GTK_WINDOW(tasklist->properties_dialog), FALSE);
 	gtk_window_set_screen(GTK_WINDOW(tasklist->properties_dialog), gtk_widget_get_screen(tasklist->applet));
 	gtk_window_present(GTK_WINDOW(tasklist->properties_dialog));
+}
+
+static void destroy_tasklist(GtkWidget* widget, TasklistData* tasklist)
+{
+
+	g_signal_handlers_disconnect_by_func(G_OBJECT(tasklist->applet),
+						G_CALLBACK(applet_change_orient), tasklist);
+	g_signal_handlers_disconnect_by_func(G_OBJECT(tasklist->applet),
+						G_CALLBACK(applet_change_pixel_size), tasklist);
+	g_signal_handlers_disconnect_by_func(G_OBJECT(tasklist->applet),
+						G_CALLBACK(applet_change_background), tasklist);
+	g_signal_handlers_disconnect_by_func (tasklist->settings,
+					  G_CALLBACK (display_all_workspaces_changed),
+					  tasklist);
+	g_signal_handlers_disconnect_by_func (tasklist->settings,
+					  G_CALLBACK (group_windows_changed),
+					  tasklist);
+	g_signal_handlers_disconnect_by_func (tasklist->settings,
+					  G_CALLBACK (move_unminimized_windows_changed),
+					  tasklist);
+
+	g_object_unref(tasklist->settings);
+
+	if (tasklist->properties_dialog)
+		gtk_widget_destroy(tasklist->properties_dialog);
+
+	g_free(tasklist);
+
 }
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -325,8 +325,11 @@ static void destroy_pager(GtkWidget* widget, PagerData* pager)
 
 	if (pager->properties_dialog)
 		gtk_widget_destroy(pager->properties_dialog);
-
+/* FIXME: this stops segfaults when removing in-process applet
+/* but leaves part of it in RAM until the panel is restarted */
+#ifndef WNCKLET_INPROCESS
 	g_free(pager);
+#endif
 }
 
 static const GtkActionEntry pager_menu_actions[] = {
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -86,6 +86,7 @@ typedef struct {
 static void display_properties_dialog(GtkAction* action, PagerData* pager);
 static void display_help_dialog(GtkAction* action, PagerData* pager);
 static void display_about_dialog(GtkAction* action, PagerData* pager);
+static void destroy_pager(GtkWidget* widget, PagerData* pager);
 
 static void pager_update(PagerData* pager)
 {
@@ -317,19 +318,6 @@ static gboolean applet_scroll(MatePanelApplet* applet, GdkEventScroll* event, Pa
 	wnck_workspace_activate(wnck_screen_get_workspace(pager->screen, index), event->time);
 
 	return TRUE;
-}
-
-static void destroy_pager(GtkWidget* widget, PagerData* pager)
-{
-	g_object_unref (pager->settings);
-
-	if (pager->properties_dialog)
-		gtk_widget_destroy(pager->properties_dialog);
-/* FIXME: this stops segfaults when removing in-process applet
-/* but leaves part of it in RAM until the panel is restarted */
-#ifndef WNCKLET_INPROCESS
-	g_free(pager);
-#endif
 }
 
 static const GtkActionEntry pager_menu_actions[] = {
@@ -969,4 +957,25 @@ static void display_properties_dialog(GtkAction* action, PagerData* pager)
 	gtk_window_set_icon_name(GTK_WINDOW(pager->properties_dialog), WORKSPACE_SWITCHER_ICON);
 	gtk_window_set_screen(GTK_WINDOW(pager->properties_dialog), gtk_widget_get_screen(pager->applet));
 	gtk_window_present(GTK_WINDOW(pager->properties_dialog));
+}
+
+static void destroy_pager(GtkWidget* widget, PagerData* pager)
+{
+	g_signal_handlers_disconnect_by_func (pager->settings,
+					  G_CALLBACK (num_rows_changed),
+					  pager);
+	g_signal_handlers_disconnect_by_func (pager->settings,
+					  G_CALLBACK (display_workspace_names_changed),
+					  pager);
+	g_signal_handlers_disconnect_by_func (pager->settings,
+					  G_CALLBACK (all_workspaces_changed),
+					  pager);
+	g_signal_handlers_disconnect_by_func (pager->settings,
+					  G_CALLBACK (wrap_workspaces_changed),
+					  pager);
+	g_object_unref (pager->settings);
+
+	if (pager->properties_dialog)
+		gtk_widget_destroy(pager->properties_dialog);
+	g_free(pager);
 }
--- a/libmate-panel-applet/Makefile.am
+++ b/libmate-panel-applet/Makefile.am
@@ -24,7 +24,8 @@ libmate_panel_applet_4_la_SOURCES =		\
 	mate-panel-applet-gsettings.c		\
 	mate-panel-applet-gsettings.h       \
 	panel-plug.c			\
-	panel-plug-private.h
+	panel-plug-private.h    \
+	panel-applet-private.h
 
 libmate_panel_applet_4_la_LIBADD  = \
 	$(LIBMATE_PANEL_APPLET_LIBS) \
--- a/libmate-panel-applet/mate-panel-applet-factory.c
+++ b/libmate-panel-applet/mate-panel-applet-factory.c
@@ -31,6 +31,9 @@ struct _MatePanelAppletFactory {
 	gboolean   out_of_process;
 	GType      applet_type;
 	GClosure  *closure;
+
+	GHashTable      *applets;
+	guint            next_uid;
 };
 
 struct _MatePanelAppletFactoryClass {
@@ -42,14 +45,28 @@ struct _MatePanelAppletFactoryClass {
 
 G_DEFINE_TYPE (MatePanelAppletFactory, mate_panel_applet_factory, G_TYPE_OBJECT)
 
+static GHashTable *factories = NULL;
+
 static void
 mate_panel_applet_factory_finalize (GObject *object)
 {
 	MatePanelAppletFactory *factory = MATE_PANEL_APPLET_FACTORY (object);
 
+	g_hash_table_remove (factories, factory->factory_id);
+
+	if (g_hash_table_size (factories) == 0) {
+		g_hash_table_unref (factories);
+		factories = NULL;
+	}
+
 	if (factory->factory_id) {
 		g_free (factory->factory_id);
 		factory->factory_id = NULL;
+	}
+
+	if (factory->applets) {
+		g_hash_table_unref (factory->applets);
+		factory->applets = NULL;
 	}
 
 	if (factory->closure) {
@@ -63,6 +80,8 @@ mate_panel_applet_factory_finalize (GObject *object)
 static void
 mate_panel_applet_factory_init (MatePanelAppletFactory *factory)
 {
+	factory->applets = g_hash_table_new (NULL, NULL);
+	factory->next_uid = 1;
 }
 
 static void
@@ -77,6 +96,12 @@ static void
 mate_panel_applet_factory_applet_removed (MatePanelAppletFactory *factory,
 				     GObject            *applet)
 {
+	guint uid;
+
+	uid = GPOINTER_TO_UINT (g_object_get_data (applet, "uid"));
+
+	g_hash_table_remove (factory->applets, GUINT_TO_POINTER (uid));
+
 	factory->n_applets--;
 	if (factory->n_applets == 0)
 		g_object_unref (factory);
@@ -95,6 +120,11 @@ mate_panel_applet_factory_new (const gchar *factory_id,
 	factory->out_of_process = out_of_process;
 	factory->applet_type = applet_type;
 	factory->closure = g_closure_ref (closure);
+	if (factories == NULL) {
+		factories = g_hash_table_new (g_str_hash, g_str_equal);
+	}
+
+	g_hash_table_insert (factories, factory->factory_id, factory);
 
 	return factory;
 }
@@ -148,6 +178,7 @@ mate_panel_applet_factory_get_applet (MatePanelAppletFactory    *factory,
 	GVariant    *props;
 	GdkScreen   *screen;
 	guint32      xid;
+	guint32      uid;
 	const gchar *object_path;
 
 	g_variant_get (parameters, "(&si@a{sv})", &applet_id, &screen_num, &props);
@@ -169,7 +200,10 @@ mate_panel_applet_factory_get_applet (MatePanelAppletFactory    *factory,
 		gdk_screen_get_default ();
 
 	xid = mate_panel_applet_get_xid (MATE_PANEL_APPLET (applet), screen);
+	uid = factory->next_uid++;
 	object_path = mate_panel_applet_get_object_path (MATE_PANEL_APPLET (applet));
+	g_hash_table_insert (factory->applets, GUINT_TO_POINTER (uid), applet);
+	g_object_set_data (applet, "uid", GUINT_TO_POINTER (uid));
 
 	g_dbus_method_invocation_return_value (invocation,
 					       g_variant_new ("(ou)", object_path, xid));
--- a/libmate-panel-applet/mate-panel-applet-factory.c
+++ b/libmate-panel-applet/mate-panel-applet-factory.c
@@ -241,9 +241,9 @@ static const gchar introspection_xml[] =
 	        "<arg name='screen' type='i' direction='in'/>"
 	        "<arg name='props' type='a{sv}' direction='in'/>"
 	        "<arg name='applet' type='o' direction='out'/>"
-            "<arg name='out-of-process' type='b' direction='out'/>"
+	        "<arg name='out-of-process' type='b' direction='out'/>"
 	        "<arg name='xid' type='u' direction='out'/>"
-            "<arg name='uid' type='u' direction='out'/>"
+	        "<arg name='uid' type='u' direction='out'/>"
 	      "</method>"
 	    "</interface>"
 	  "</node>";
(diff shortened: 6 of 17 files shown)


lukefromdc:
Still one issue with this: the invest applet won't open, manually running its python executable (invest-applet) gets this error:

```
 File "/usr/libexec/invest-applet", line 99, in <module>
    MatePanelApplet.Applet.factory_main(
AttributeError: type object 'Applet' has no attribute 'factory_main'
```

Some kind of introspection issue, as replacing MatePanelApplet-4.0.typelib with one from a build of master permits invest to open normally. All other applets work fine, background rendering works fine with system, image, or color backgrounds, in process or out

lukefromdc:
Also works with or without compositing in the window manager with no visible hiccups from that in or out of process.

lukefromdc:
Tested this with GTK 3.14 through 3.22, fixed an issue with 3.14/3.16 that turned up. Retests with GTK 3.16 and 3.22 gave consistant behavior.

raveit65:
Do we need to rebuild mate-applets?

raveit65:
See resulting diff in browser, indents are a bit worse at some places.
https://patch-diff.githubusercontent.com/raw/mate-desktop/mate-panel/pull/622.diff
Ie. in an xml file for introspection, maybe this causes the issue with invest-applet?

raveit65:
hmm, the panel crashes every time if i try to remove an applet now.
It doesn't matter if it is an in-process applet or a normal one from mate-applets.
And yes, i rebuilt mate-applets too.
Tested in a fedora rawhide VM.
I don't see those crashes with released 1.19.2 tarball from today.
stacktrace:
```
Core was generated by `mate-panel'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00000041af865eb3 in mate_panel_applet_locked_change_notify (settings=0x41b1933a60, key=0x7fc3c036cffa "locked", applet=0x7fc3a001b7e0) at applet.c:186
186		if (applet == NULL || !GTK_IS_WIDGET (applet))
[Current thread is 1 (Thread 0x7fc3c34a6f00 (LWP 1662))]

Thread 1 (Thread 0x7fc3c34a6f00 (LWP 1662)):
#0  0x00000041af865eb3 in mate_panel_applet_locked_change_notify (settings=0x41b1933a60, key=0x7fc3c036cffa "locked", applet=0x7fc3a001b7e0) at applet.c:186
        __inst = 0x7fc3a001b7e0
        __t = 282143398912
        __r = <optimized out>
#1  0x00007fc3bffe457d in g_closure_invoke (closure=0x41b18db980, return_value=0x0, n_param_values=2, param_values=0x7ffefdfb4080, invocation_hint=0x7ffefdfb4000) at gclosure.c:804
        marshal = 0x7fc3bffe7030 <g_cclosure_marshal_VOID__STRING>
        marshal_data = 0x0
        in_marshal = 0
        real_closure = 0x41b18db960
        __func__ = "g_closure_invoke"
#2  0x00007fc3bfff7e6e in signal_emit_unlocked_R (node=node@entry=0x41b1135000, detail=detail@entry=773, instance=instance@entry=0x41b1933a60, emission_return=emission_return@entry=0x0, instance_and_params=instance_and_params@entry=0x7ffefdfb4080) at gsignal.c:3635
        tmp = <optimized out>
        handler = 0x41b13a9340
        accumulator = 0x0
        emission = {next = 0x7ffefdfb4760, instance = 0x41b1933a60, ihint = {signal_id = 169, detail = 773, run_type = G_SIGNAL_RUN_FIRST}, state = EMISSION_RUN, chain_type = 4}
        class_closure = 0x41b1135c20
        handler_list = 0x41b13a9340
        return_accu = 0x0
        accu = {g_type = 0, data = {{v_int = 0, v_uint = 0, v_long = 0, v_ulong = 0, v_int64 = 0, v_uint64 = 0, v_float = 0, v_double = 0, v_pointer = 0x0}, {v_int = 0, v_uint = 0, v_long = 0, v_ulong = 0, v_int64 = 0, v_uint64 = 0, v_float = 0, v_double = 0, v_pointer = 0x0}}}
        signal_id = 169
        max_sequential_handler_number = 21817
        return_value_altered = 0
#3  0x00007fc3c0000985 in g_signal_emit_valist (instance=0x41b1933a60, signal_id=<optimized out>, detail=773, var_args=var_args@entry=0x7ffefdfb4260) at gsignal.c:3391
        instance_and_params = 0x7ffefdfb4080
        signal_return_type = 4
        param_values = 0x7ffefdfb4098
        node = <optimized out>
        i = <optimized out>
        n_params = 1
        __func__ = "g_signal_emit_valist"
#4  0x00007fc3c00012ef in g_signal_emit (instance=instance@entry=0x41b1933a60, signal_id=<optimized out>, detail=<optimized out>) at gsignal.c:3447
        var_args = {{gp_offset = 32, fp_offset = 48, overflow_arg_area = 0x7ffefdfb4340, reg_save_area = 0x7ffefdfb4280}}
#5  0x00007fc3c033fa22 in g_settings_real_change_event (settings=0x41b1933a60, keys=0x41b189d790, n_keys=<optimized out>) at gsettings.c:386
        key = 0x7fc3c036cffa "locked"
        i = 3
#6  0x00007fc3b9dd5bde in ffi_call_unix64 () at ../src/x86/unix64.S:76
No locals.
#7  0x00007fc3b9dd554f in ffi_call (cif=cif@entry=0x7ffefdfb45b0, fn=fn@entry=0x7fc3c033f9c0 <g_settings_real_change_event>, rvalue=<optimized out>, avalue=avalue@entry=0x7ffefdfb44a0) at ../src/x86/ffi64.c:525
        classes = {X86_64_INTEGER_CLASS, 32707, 4261103024, 32766}
        stack = <optimized out>
        argp = <optimized out>
        arg_types = <optimized out>
        gprcount = 4
        ssecount = <optimized out>
        ngpr = 1
        nsse = 0
        i = <optimized out>
        avn = <optimized out>
        ret_in_memory = <optimized out>
        reg_args = <optimized out>
#8  0x00007fc3bffe516d in g_cclosure_marshal_generic_va (closure=0x41b11296e0, return_value=0x7ffefdfb4790, instance=<optimized out>, args_list=<optimized out>, marshal_data=<optimized out>, n_params=2, param_types=0x41b1126250) at gclosure.c:1604
        rtype = 0x7fc3b9dd5f10 <ffi_type_sint32>
        rvalue = 0x7ffefdfb4500
        n_args = 4
        atypes = 0x7ffefdfb44d0
        args = 0x7ffefdfb44a0
        storage = 0x7ffefdfb4480
        i = <optimized out>
        cif = {abi = FFI_UNIX64, nargs = 4, arg_types = 0x7ffefdfb44d0, rtype = 0x7fc3b9dd5f10 <ffi_type_sint32>, bytes = 0, flags = 10}
        cc = 0x41b11296e0
        enum_tmpval = <optimized out>
        tmpval_used = 0
        args_copy = {{gp_offset = 40, fp_offset = 48, overflow_arg_area = 0x7ffefdfb4940, reg_save_area = 0x7ffefdfb4880}}
#9  0x00007fc3bffe47b6 in _g_closure_invoke_va (closure=0x41b11296e0, return_value=0x7ffefdfb4790, instance=0x41b1933a60, args=0x7ffefdfb4860, n_params=2, param_types=0x41b1126250) at gclosure.c:867
        marshal = 0x7fc3bffe2a10 <g_type_class_meta_marshalv>
        marshal_data = 0xa0
        in_marshal = 0
        real_closure = 0x41b11296c0
        __func__ = "_g_closure_invoke_va"
#10 0x00007fc3c0000649 in g_signal_emit_valist (instance=0x41b1933a60, signal_id=<optimized out>, detail=0, var_args=var_args@entry=0x7ffefdfb4860) at gsignal.c:3300
        return_accu = 0x7ffefdfb4790
        accu = {g_type = 20, data = {{v_int = 0, v_uint = 0, v_long = 0, v_ulong = 0, v_int64 = 0, v_uint64 = 0, v_float = 0, v_double = 0, v_pointer = 0x0}, {v_int = 0, v_uint = 0, v_long = 0, v_ulong = 0, v_int64 = 0, v_uint64 = 0, v_float = 0, v_double = 0, v_pointer = 0x0}}}
        accumulator = 0x41b1133930
        emission = {next = 0x0, instance = 0x41b1933a60, ihint = {signal_id = 170, detail = 0, run_type = G_SIGNAL_RUN_LAST}, state = EMISSION_RUN, chain_type = 282143710784}
        signal_id = <optimized out>
        instance_type = 282143710784
        emission_return = {g_type = 20, data = {{v_int = 0, v_uint = 0, v_long = 0, v_ulong = 0, v_int64 = 0, v_uint64 = 0, v_float = 0, v_double = 0, v_pointer = 0x0}, {v_int = 0, v_uint = 0, v_long = 0, v_ulong = 0, v_int64 = 0, v_uint64 = 0, v_float = 0, v_double = 0, v_pointer = 0x0}}}
        rtype = 20
        static_scope = 0
        fastpath_handler = <optimized out>
        closure = <optimized out>
        run_type = <optimized out>
        l = <optimized out>
        fastpath = 1
        instance_and_params = <optimized out>
        signal_return_type = <optimized out>
        param_values = <optimized out>
        node = <optimized out>
        i = <optimized out>
        n_params = <optimized out>
        __func__ = "g_signal_emit_valist"
#11 0x00007fc3c00012ef in g_signal_emit (instance=instance@entry=0x41b1933a60, signal_id=<optimized out>, detail=detail@entry=0) at gsignal.c:3447
        var_args = {{gp_offset = 24, fp_offset = 48, overflow_arg_area = 0x7ffefdfb4940, reg_save_area = 0x7ffefdfb4880}}
#12 0x00007fc3c0340e0f in settings_backend_path_changed (target=<optimized out>, backend=<optimized out>, path=0x7fc3a002d950 "/org/mate/panel/objects/object-25/", origin_tag=<optimized out>) at gsettings.c:461
        settings = 0x41b1933a60
        ignore_this = 32707
#13 0x00007fc3c033b26a in g_settings_backend_invoke_closure (user_data=user_data@entry=0x41b18d1700) at gsettingsbackend.c:267
        closure = 0x41b18d1700
#14 0x00007fc3bfd07f67 in g_idle_dispatch (source=0x7fc3a4010cd0, callback=0x7fc3c033b250 <g_settings_backend_invoke_closure>, user_data=0x41b18d1700) at gmain.c:5500
        again = <optimized out>
#15 0x00007fc3bfd0b587 in g_main_dispatch (context=0x41b10e7580) at gmain.c:3148
        dispatch = 0x7fc3bfd07f50 <g_idle_dispatch>
        prev_source = 0x0
        was_in_call = 0
        user_data = 0x41b18d1700
        callback = 0x7fc3c033b250 <g_settings_backend_invoke_closure>
        cb_funcs = 0x7fc3bffd3280 <g_source_callback_funcs>
        cb_data = 0x41b1337f90
        need_destroy = <optimized out>
        source = 0x7fc3a4010cd0
        current = 0x41b1129d40
        i = 75
#16 g_main_context_dispatch (context=context@entry=0x41b10e7580) at gmain.c:3813
No locals.
#17 0x00007fc3bfd0b928 in g_main_context_iterate (context=0x41b10e7580, block=block@entry=1, dispatch=dispatch@entry=1, self=<optimized out>) at gmain.c:3886
        max_priority = 0
        timeout = 0
        some_ready = 1
        nfds = <optimized out>
        allocated_nfds = 6
        fds = 0x41b16b2cb0
#18 0x00007fc3bfd0bc42 in g_main_loop_run (loop=0x41b11e3000) at gmain.c:4082
        __func__ = "g_main_loop_run"
#19 0x00007fc3c2372065 in gtk_main () at gtkmain.c:1312
        loop = 0x41b11e3000
#20 0x00000041af859699 in main (argc=<optimized out>, argv=<optimized out>) at main.c:168
        desktopfile = <optimized out>
        context = <optimized out>
        error = 0x0
```
Complete stacktrace at
https://www.dropbox.com/s/19kmps1n91pn41b/panel-backtrace?dl=0

lukefromdc:
I also get routine crashes on removing applets-but get them with master as well. I'll take a look at that, and also found an issue with custom image backgrounds on out of process applets sometimes using the wrong part of an image as well as the introspection issue.

One indent issue that you found came from where GNOME used a mix of tabs and spaces, I'll have to clean that up, either match it or change it for that whole stanza.

I have simply not been able to find a simpler way to fix in-process applets, they work for GNOME because they got the plug and socket out of in-process applets. Since they now have all applets in-process, they don't use the plugs at all, clearing one obstacle for wayland use.

Not ready yet, but I suspect we are going to have to deal both with in-process applets and with removing the plug and socket setup for wayland:

https://developer.gnome.org/gtk3/stable/GtkPlug.html
"The communication between a GtkSocket and a GtkPlug follows the XEmbed Protocol. This protocol has also been implemented in other toolkits, e.g. Qt, allowing the same level of integration when embedding a Qt widget in GTK+ or vice versa.

The GtkPlug and GtkSocket widgets are only available when GTK+ is compiled for the X11 platform and GDK_WINDOWING_X11 is defined. They can only be used on a GdkX11Display. To use GtkPlug and GtkSocket, you need to include the gtk/gtkx.h header."

lukefromdc:
I had to modify GNOME's code in one way by using an out_process dummy variable to keep existing applets happy. GNOME did not worry about changing the interface for all applets but for us its far simpler to avoid that kind of mess.

lukefromdc:
There is a GNOME commit concerning crashes on applet removal I will try, I had that flagged last night for possible use but got bogged down in other stuff

lukefromdc:
applying
https://github.com/GNOME/gnome-panel/commit/0c3844c0e5f1bee5885d2c7aeaa49763e6708f6d
fixed the crash on removing out of process applets, though I still get a crash and segfault on removing an in-process applet. Used to get crashes on all applet removals, all the time.

Also thanks to albert muktupavels for finding the problem with introspection, now the invest applet works

lukefromdc:
No need to rebuild applets, I've got them all working from a build against current git master

lukefromdc:
Got the issue with in-process applets getting the wrong part a custom panel image bg fixed by reverting the "simplify background handling" GNOME commit that turned out not to be necessary for later commits to apply.  With that gone, out of process applets being moved on a custom background update their bg continuously and thus match the panel beneath them, same as on master and same as in-process applets. You can also drag them between panels with the same behavior as on master.

All the applets and their backgrounds now open clean and render both bg and fg properly both in and out of process, every applet tested and clock, wncklet, fish, and tray in-process. . The issue with the drawer getting one of its icons on the panel and then refusing to open came up during this test, that is pre-existing issue  https://github.com/mate-desktop/mate-panel/issues/556

There is still the issue that removing an in-process applet crashes the panel, but out of process applets can be added and removed freely, something I've not been able to do in months, maybe years.

Themes still need the previously mentioned change to avoid showing applet handles in the wrong locations when a custom background is in use. My way limits them to the theme bg, same as before. Might be possible to force a fixed handle image all the way to the left so it will show even with a custom background, something I've never seen work before.

raveit65:
> I also get routine crashes on removing applets-but get them with master as well.

opps my fault, master wasn't build with in-process

raveit65:
I see a new build warning
```
In file included from test-dbus-applet.c:4:0:
test-dbus-applet.c: In function 'main':
mate-panel-applet.h:177:9: warning: 'out_process' may be used uninitialized in this function [-Wmaybe-uninitialized]
  retval = mate_panel_applet_factory_main (factory_id,out_process, type, callback, data); \
         ^
mate-panel-applet.h:152:11: note: 'out_process' was declared here
  gboolean out_process; \
           ^
test-dbus-applet.c:206:1: note: in expansion of macro 'MATE_PANEL_APPLET_OUT_PROCESS_FACTORY'
 MATE_PANEL_APPLET_OUT_PROCESS_FACTORY ("TestAppletFactory",
 ^
  CCLD     test-dbus-applet
make[3]: Leaving directory '/media/Programme/Linux/Mate-Desktop/git-version/github-matedesktop/mate-panel-dist/libmate-panel-applet'
```

raveit65:
I just tested it and out-of-process applets don't crash any more during removal.
Stacktrace  when removing in-process applets are the same like before.
https://www.dropbox.com/s/xyxmiaqf7xsj9t1/backtrace-panel-1.txt?dl=0
And i got a new one which is generate by ' mate-panel --replace --reset' command.
https://www.dropbox.com/s/qaf5p9cge12l374/backtrace-panel-2.txt?dl=0
Honestly, this needs really tested by more people than me......

lukefromdc:
I agree with the need for more testing, by more people on more distros. While not as complex as last year's GtkApplication work on Caja, this is still a big change and needs testing in a variety of scenarios.  I will hold this open, run it as my daily use panel, and keep it up to date until we can get enough testing done.

There are 5 supported GTK versions, the composited and noncomposited case, system backgrounds, color backgrounds with and without alpha, and image backgrounds with and without alpha. That alone is 50 possible configurations for just one panel layout. That much I can handle as it goes in less than an hour and I've already done it once. 

Then there are many distros, with Debian, Ubuntu, Mint, and Fedora/Red Hat as probably the most widely used. Next there are the many different use cases, and for these we need more testers on more distros.

The GtkApplication work in Caja still generates occasional reports of a new bug that would have been easily found and fixed prior to release if anyone had actually tested for it. This occurs because we have too few team members to test all conditions, and too few people trying out development versions.

In the old days, GNOME had the advantage that development versions of GNOME 2 went into Ubuntu alpha versions, so they got tested at least there and presumably also in rolling releases. That meant realtime testing and bug reports.

One thing I can do though is try to find bug reports on GNOME panel that were reported after they did this work, and find any relevant to this and see if those issues have been imported.

lukefromdc:
The variable in the build warning is a dummy, I will give it a default to stop the build warning. Also digging into a runtime underallocation warning from wncklet that exists only in-process, probably inherited from master but not sure about that,

lukefromdc:
I just found that these warnings from wnck-tasklist (window list) are the same with this PR and Master when built with the window list in-process:

```
(mate-panel:24710): Gtk-WARNING **: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner GtkToggleButton)

(mate-panel:24710): Gtk-WARNING **: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner GtkToggleButton)

(mate-panel:24710): Gtk-WARNING **: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner GtkToggleButton)

(mate-panel:24710): Gtk-WARNING **: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner GtkToggleButton)

(mate-panel:24710): Gtk-WARNING **: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner GtkToggleButton)

(mate-panel:24710): Gtk-WARNING **: Negative content width -1 (allocation 1, extents 1x1) while allocating gadget (node button, owner GtkToggleButton)
```
Thus they do not come from this PR but I am still looking for their source

raveit65:
I will update my daily installation with this PR for testing changes compiled with out-of-process.

lukefromdc:
Found out segfaults on removing in-process applets are applet specific. Only the workspace switcher, tasklist, and clock were having this problem. Found a workaround, not the best but functional. First panel I've had with no crashes on removing applets in a long time.

This workaround does not remove all of the applet from RAM when it is taken off the panel, but the ram in question is cleared when the panel is restarted. No more segfaults, I suspect a double-free somewhere was involved. Still some warnings on removing in-process applets but the panel stays up.
This is from removing the fish (simplest):

```
(mate-panel:26885): Gtk-CRITICAL **: gtk_widget_get_allocation: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:26885): Gtk-CRITICAL **: gtk_widget_get_realized: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:26885): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkWidget'

(mate-panel:26885): Gtk-CRITICAL **: gtk_widget_set_tooltip_markup: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:26885): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkWidget'

(mate-panel:26885): Gtk-CRITICAL **: gtk_widget_get_accessible: assertion 'GTK_IS_WIDGET (widget)' failed
```
It appears one or more GtkWidgets get destroyed before the process is all the way done with them. Not sure if they are new or longstanding though, will have to check. The workaround had no effect on the warnings, just stopped the segfaults and kept the panel running.

raveit65:
I can confirm that removing the workspace-switcher doesn't crash the panel any more.
But all other applets from m-p package crashed the panel here and triggered a abrt alarm in fedora rawhide VM.

lukefromdc:
OK, that means we now have difference in behavior and I don't get the crashes you have, making it harder to work on. You are getting crashes from removing in-process applets again? They were not supposed to get any changes whatsoever from the last two commits. When was your last test where out of process applets could be cleanly removed?

lukefromdc:
@raveit65 , are you using glib 2.53.4 by any chance? I found last week that it made both removing any applet and also moving launchers and action button applets with the left click drag segfault, even with mate-panel from master.

Using glib 2.53.4, mate-panel from master, and the connect to server applet, moving it twice crashed the panel with a segfault on the second move. Also with mate-panel from master and any version of glib I get a segfault on removing any applet, every time.

raveit65:
Yes, i think glib 2.53.4 is the problem in fedora rawhide.
I can't build any more m-p with that version of glib, see my paste.
https://paste.fedoraproject.org/paste/TkpsNslpQ9bYZ3cEMox7cA
with glib2-2.53.3-1.fc27.x86_64 building is fine.
But running glib2-2.53.3-1.fc27.x86_64 does not help to stop the crashes.
I think fedora rawhide is currently complete unstable and not the right area to test something.
Ok , i will test all with f26 on bare metal later and report it here.

raveit65:
I forgot to answer, building applets with out-of process switch gives me no crashes on bare metal f26.

lukefromdc:
I just did a test build from master, and yes, got the same segfaults on removing the clock build in-process. Oddly I did not get the background rendering issue, and the plugs showed in GtkInspector.  Must be a recent update to something underlying that did that. 
Here's what I got from GTK on removing the clock with master and built in process:

```
(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_parent: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): GLib-GObject-CRITICAL **: g_object_get_data: assertion 'G_IS_OBJECT (object)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_queue_resize: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_state_flags: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_style_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_style_context_get: assertion 'GTK_IS_STYLE_CONTEXT (context)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_pango_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Pango-CRITICAL **: pango_context_get_language: assertion 'context != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_context_get_metrics: assertion 'PANGO_IS_CONTEXT (context)' failed

(mate-panel:9884): Pango-CRITICAL **: pango_font_metrics_get_ascent: assertion 'metrics != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_font_metrics_get_descent: assertion 'metrics != NULL' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_style_context_get_padding: assertion 'GTK_IS_STYLE_CONTEXT (context)' failed

(mate-panel:9884): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkLabel'

(mate-panel:9884): Gtk-CRITICAL **: gtk_label_set_markup: assertion 'GTK_IS_LABEL (label)' failed

(mate-panel:9884): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkLabel'

(mate-panel:9884): Gtk-CRITICAL **: gtk_label_get_text: assertion 'GTK_IS_LABEL (label)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_pango_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_new: assertion 'context != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_set_alignment: assertion 'layout != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_set_text: assertion 'layout != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_get_pixel_extents: assertion 'PANGO_IS_LAYOUT (layout)' failed

(mate-panel:9884): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_state_flags: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_style_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_style_context_get_padding: assertion 'GTK_IS_STYLE_CONTEXT (context)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_allocation: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkLabel'

(mate-panel:9884): Gtk-CRITICAL **: gtk_label_get_angle: assertion 'GTK_IS_LABEL (label)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_queue_resize: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_tooltip_text: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_set_tooltip_text: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_state_flags: assertion 'GTK_IS_WIDGET (widget)' failed
```

lukefromdc:
I just did a test build from master, and yes, got the same segfaults on removing the clock build in-process. 

For some reason https://github.com/mate-desktop/mate-panel/issues/468 which is the redrawing issue did NOT appear, on several test builds from master or other branches than this one with all possible applets in-process. Wasn't GTK 3.22.16 that made the difference as I tried rolling that back, same results. Must have been some underlying update in Debian, can't figure out what it could be though libatk just got an update about a week ago.

Issues with master were the crashes on removing in-process applets, and flickering most tasklist buttons when the mouse moved on or off any one of them. 

We still need to get rid of the plugs for in-process applets if we ever want to run on wayland, however.

Here's what I got from GTK on removing the clock with master and built in process:

```
(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_parent: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): GLib-GObject-CRITICAL **: g_object_get_data: assertion 'G_IS_OBJECT (object)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_queue_resize: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_state_flags: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_style_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_style_context_get: assertion 'GTK_IS_STYLE_CONTEXT (context)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_pango_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Pango-CRITICAL **: pango_context_get_language: assertion 'context != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_context_get_metrics: assertion 'PANGO_IS_CONTEXT (context)' failed

(mate-panel:9884): Pango-CRITICAL **: pango_font_metrics_get_ascent: assertion 'metrics != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_font_metrics_get_descent: assertion 'metrics != NULL' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_style_context_get_padding: assertion 'GTK_IS_STYLE_CONTEXT (context)' failed

(mate-panel:9884): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkLabel'

(mate-panel:9884): Gtk-CRITICAL **: gtk_label_set_markup: assertion 'GTK_IS_LABEL (label)' failed

(mate-panel:9884): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkLabel'

(mate-panel:9884): Gtk-CRITICAL **: gtk_label_get_text: assertion 'GTK_IS_LABEL (label)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_pango_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_new: assertion 'context != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_set_alignment: assertion 'layout != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_set_text: assertion 'layout != NULL' failed

(mate-panel:9884): Pango-CRITICAL **: pango_layout_get_pixel_extents: assertion 'PANGO_IS_LAYOUT (layout)' failed

(mate-panel:9884): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_state_flags: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_style_context: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_style_context_get_padding: assertion 'GTK_IS_STYLE_CONTEXT (context)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_allocation: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkLabel'

(mate-panel:9884): Gtk-CRITICAL **: gtk_label_get_angle: assertion 'GTK_IS_LABEL (label)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_queue_resize: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_tooltip_text: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_set_tooltip_text: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:9884): Gtk-CRITICAL **: gtk_widget_get_state_flags: assertion 'GTK_IS_WIDGET (widget)' failed
```

raveit65:
omg, we can say every widget crashes with glib. 2.53.4
I recall that every year glib dev versions had enabled a debug switch or something.
I noticed that in previous dev phase too that applications crashes for nothing, and all is fine if glib is stable again.
Of course my build issue with gsettings keys is new this year.
Maybe we need to adjust something here.

Any ways, i've some good news.
With f26 and stable glib on baremetal i have no crashes anymore if i remove an applet (in-process), menubar or launchers.
So i would say you did a good job ;-)
...and only the RAM problem should be solved.

raveit65:
I forgot , only the fish applet has a little issue.
I need to click on the applet after adding them to panel for displaying them correct.

lukefromdc:
I just confirmed the issue with the fish built in-process is inherited from master and not specific to this PR either. 

Also found there is a memory leak on removing the in-process clock or fish in master(the two that don't crash the panel on removal) though repeated removal and addtion would segfault and stop the test. Still saw that after adding either once, then removing it, the panel was left with several hundred kb of RAM remaining in use. Thus the bulk of this is not coming from the changes in the PR to the wncklet or clock either, in fact the amount of ram involved looked almost the same.

Essentially testing this PR is revealing underlying problems in master, with luck we can clean all of them up at once.

lukefromdc:
Found the gnome-panel commit that fixes the fish opening incorrectly when built in-process.

https://github.com/GNOME/gnome-panel/commit/f0f91805a1e39543487cfe67dfc47129e45e76f3

raveit65:
For better understanding, which applets are affected from the switch built-in/out-process?
Does this affect the menubar and the drawer too for example?
Or only what is in the applet folder in tarball?
Btw. i never use the built-in-process configure switch before.
So, if i post here that something doesn't work i don't know if this is an issue with master or the PR.

raveit65:
I can confirm that last commit fixes the draw issue with fish applet.

raveit65:
Sadly i was able to crash the panel again if i remove the fish applet, on stable f26 with stable glib2-2.52.3-1.fc26.x86_64
In the stackstrace it looks like a problem with gsettings.
```
Core was generated by `mate-panel'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __strcmp_ssse3 () at ../sysdeps/x86_64/strcmp.S:173
173		movlpd	(%rdi), %xmm1
[Current thread is 1 (Thread 0x7f29f04c2a80 (LWP 30162))]

Thread 1 (Thread 0x7f29f04c2a80 (LWP 30162)):
#0  0x00007f29ebe91ef6 in __strcmp_ssse3 () at ../sysdeps/x86_64/strcmp.S:173
#1  0x00007f29cabd9f54 in command_changed_notify (settings=<optimized out>, key=<optimized out>, fish=0x559272888210) at fish.c:1064
        value = 0x5592738675f0 "fortune"
#5  0x00007f29ed0c6b0f in <emit signal changed:command on instance 0x7f29cc01a700 [GSettings]> (instance=instance@entry=0x7f29cc01a700, signal_id=<optimized out>, detail=<optimized out>) at gsignal.c:3447
        var_args = {{gp_offset = 32, fp_offset = 48, overflow_arg_area = 0x7ffec23fa330, reg_save_area = 0x7ffec23fa270}}
```
full stacktrace at https://www.dropbox.com/s/rwwfrk4y58phyz4/backtrace-fish-applet?dl=0
In xsession-errors i get always those warnings if i remove the fish-applet without crashing the panel.
```
(mate-panel:30162): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkImage'

(mate-panel:30162): Gtk-CRITICAL **: gtk_image_set_from_pixbuf: assertion 'GTK_IS_IMAGE (image)' failed

(mate-panel:30162): Gtk-CRITICAL **: gtk_widget_get_allocation: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:30162): Gtk-CRITICAL **: gtk_widget_get_realized: assertion 'GTK_IS_WIDGET (widget)' failed

(mate-panel:30162): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkEntry'

(mate-panel:30162): Gtk-CRITICAL **: gtk_entry_get_text: assertion 'GTK_IS_ENTRY (entry)' failed

(mate-panel:30647): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```

raveit65:
I am still wondering about this grip inside the workspace-switcher (in-process).
workspace-switcher-in-process
This is with blue-submarine theme and it looks like what is used for the grip from na-tray or wncklist-applet.
https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/Blue-Submarine/gtk-3.0/mate-applications.css#L462
MatePanelAppletFrameDBus {}
Displaying that here inside the applet is a bit wrong, imo.

lukefromdc:
No crash on removing the fish here, probably from different libraries. Will take a look at that.

Keeping the grip image from showing in in-process applets requires an edit to themes because the whole applet uses MatePanelAppletFrameDBus. When the PanelPlug is used the portion over most of the applet background is covered, without the plug (in-process) it is not and cannot be.  

Fortunately the handle can be themed by MatePanelAppletFrameDBus>MatePanelAppletFrameDBus 
as it comes up as a second instance of it.. That's how I kept a background color off the clock and window-list in UbuntuStudio_Legacy with https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy/commit/3242876bcc6267111aecc63fc02a7242b936e98c . 
Here's the relevent code as appled to Blue-Submarine:
```

/* the grid left from wnckpager and wncktasklist */
MatePanelAppletFrameDBus>MatePanelAppletFrameDBus {
    background-image: -gtk-scaled(url("assets/panel-grid.svg"));
    background-repeat: no-repeat;
    background-position: left;
}

```

And here are my results:

panel_bluesub_edited

This works fine with both in-process and out of process applets. The grip will not be shown on the switcher at all as it does not have an external grip to put it on.

lukefromdc:
The applets that can be built in-process are the clock, fish, tray, and wncklet. The wncklet includes the window-list, window-menu, workspace-switcher, and show-desktop applets

lukefromdc:
About the fish crash: I am running it without fortunes installed, looks like I need to test with that in use

raveit65:
Yeah, setting the grip on MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { }
works like it should.
Gtk3inspector did confuse me as it's doesn't shown the double selector name.
With gtk3inspector i see only MatePanelAppletFrameDBus { }
Another offtopic question is why i don't see the grip for wncklist and na-try if using transparency panel bg , meanwhile i see the grip on workspace-switcher with using transparency.
But that's also a issue with master.
offtopic off

The crash on fish is a very rare case and i can't reproduce them anymore.
For me gsettings can be in general a bit buggy, especially the implementation for our panel.
I have several reports at redhad bugzilla with gsettings problems.
For me are the warnings more worse if i remove the fish applet.

lukefromdc:
Branch rebased after https://github.com/mate-desktop/mate-panel/commit/33c4bd47a7210cd9e695ea1a2cbf26abbc3b7bed ( don't use deprecated gtk_show_uri )

raveit65:
I noticed some weirdness with wnck-list applet if build with in-process.
Mostly after session start it doesn't display open windows. I use the fedora layout with
show-desktop, wncklist and workspace-switcher from left-to-right at bottom panel.
The applet is there as i can right-click on the grip.
weird:
panel-wncklist-inprocess-bad
good after restarting the panel several times:
panel-wncklist-inprocess-good
I rechecked it with building the panel with out-of-process applets from the same tarball. Here i don't see that after session start.

Tarball is created from this branch here + https://github.com/mate-desktop/mate-panel/pull/600 + https://github.com/mate-desktop/mate-panel/pull/624 + https://github.com/mate-desktop/mate-panel/pull/633
But i saw that before with a build without 633.

lukefromdc:
Github is acting up so I have to reply by e-mail. Does this problem occur
if you build wnck-applet in-process from master?   I've never seen the
open windows fail to appear at any point in developing this. 

Also, the original redraw problem is now gone in Debian Unstable, no
idea why but must be some underlying library. Thus, removing the plugs
from in-process applets now concerns two scenarios: installing new versions
of MATE on an older system with in-process applets, and any future plans
to run on Wayland. For that all applets would have to be ported to in-process
and the plugs removed as GNOME has done, as xembed is used in the 
socket/plug setup.

raveit65:
> Does this problem occur if you build wnck-applet in-process from master?

Good idea to test this with master.

lukefromdc:
In both this PR and other PRs on other packages I've been finding a lot of old bugs in master versions while testing options not prevously tested.

raveit65:
You was right, problem still exists with master.

raveit65:
>   .... and any future plans to run on Wayland. For that all applets would have to be ported to in-process and the plugs removed as GNOME has done, as xembed is used in the socket/plug setup.

Now i understand why all applets in gnome ported to in-process, i saw that also in external sensor-applet.
https://sourceforge.net/p/sensors-applet/code/commit_browser
Does that mean that all applets run in the same process as the panel?

lukefromdc:
That's exactly what it means as far as I know. For out of process applets the plug and socket architecture wrapped around Xembed is used to get windows of one process (however small) into the windows of another process, that being the panel itself. 

While one might think out of process would give better performance (as it would for say, an effect in a video editor), in the case of a panel the whole combination is too small to need multiple cores, multiple threads could keep cores awake. and more importantly, out of process applets duplicate structures that are already present inside the panel. 

On the old, slow single-core machines, GNOME originally advised that out of process was more stable (as an applet could not crash the whole panel) and in-process better performing:

https://developer.gnome.org/panel-applet/stable/getting-started.in-out-process.html

lukefromdc:
Rebased after 
https://github.com/mate-desktop/mate-panel/commit/e3e2e8775f5509ce2bb60dedb28347b29404fd47 and
https://github.com/mate-desktop/mate-panel/commit/bdee659c5338a846c941dc716f2df34c01d3c60d

lukefromdc:
Redraw bug with master is BACK with Debian Unstable after the Aug 2 round up updates, as mysteriously as it disappeared. There were gcc updates included, so I wonder if compiler changes are varying the built binary just enough to make the redraw issue come and go?

Removing the plugs (with the PR) still fixes it, just as before. This just got more important in the short run again. Still no progress on the warnings and memory leak with removing in-process applets, but I have confirmed that is also an issue with master and the warnings seem to be the same.

lukefromdc:
This was in fact never solved in Debian Unstable. Rather, I simply had been testing with a system background that looks the same as the custom background that I had used for testing for years(and recently switched to in the GKT theme). The mess in master with in-process applets is as bad as it ever was with custom backgrounds, I just had forgotten that only custom backgrounds have a problem and was on the system background

lukefromdc:
Looks like all those warnings on removing applets may have to be hunted down one at a time, just got all but one of them on the fish. Somehow destroying an in-process applet seems to allow the panel to attempt to load it or do something with it (and all its widgets) one more time, and this is consistant across all applets. Warnings are same with master if applets are built in-process.

If I could find the offending function in the overall applet code it might be possible to have that check to see if the applet has been removed before doing anything else, but it should be doable in the applets themselves, the fish here serving as a prototype.

raveit65:
You can try to use **export G_DEBUG=fatal-criticals** when you run the applets from terminal with **gdb** and load them via mate-panel-test-applets.
 Btw. i will test your latest commit.

raveit65:
Indeed, removing the fish-applet gives only one warning now.
```
(mate-panel:21528): GLib-GObject-WARNING **: invalid unclassed pointer in cast to 'GtkWidget'
```

lukefromdc:
I could not find that last warning because it does not give a function name I can search for in the source code. If I could find it I could suppress it. Finding a way to stop these functions from being called at all-in ANY applet built in-process, from the main applet code would be better if it is possible.

I think this would always be an issue, except that an out-of-process applet means a terminated process entirely, so my guess is the OS is cleaning it up and otherwise we'd always get this.

raveit65:
If you set *export G_DEBUG=fatal-criticals* in a terminal and run the applet and/or the panel itself with *gdb*, it will crash with the warning.
Than you can run 'thread apply all bt full' inside gdb to create the stacktrace.
For using gdb see http://fedoraproject.org/wiki/StackTraces#gdb

lukefromdc:
I tried that, and the warning did not trigger a crash. Backtrace I got without a crash didn't show anything familiar

raveit65:
I can confirm that addin/removing the fish applet causes any warnings no more, but adding the fish again failed with.
```
Failed to register object /org/mate/panel/applet/FishAppletFactory: An object is already exported for the interface org.mate.panel.applet.AppletFactory at /org/mate/panel/applet/FishAppletFactory

(mate-panel:30989): Gtk-CRITICAL **: gtk_container_add: assertion 'GTK_IS_WIDGET (widget)' failed
```
After restarting the panel i see the fish in panel.

lukefromdc:
I got that too, will find that tomorrow. Also found an issue with the fish on a vertical panel, on starting the panel. Looks fine but gives a warning.,

Just managed to clean up all the clock warnings too, and that one can be freely removed and added again, with the cd structure freed and no sefaults or warnings.

raveit65:
Removing the clock-applet doesn#t cause any warning any more with horz/vrt. panel, good work.
But i noticed that adding locations can (not always) crashes the panel.
I never noticed that with out-of-process applet.
But i need to test this first with master and i have a stacktrace.

raveit65:
Ok, problem with adding locations exists also with master , build-in and build-out process.
But stacktraces are mention only gtkwidgets or weather locations data, and they are complete different.
And it's more a problem with the add-location-gui.
If interested i can provide stacktraces here, if not ignore it if you can't reproduce it.

lukefromdc:
I've seen the panel crashes with adding locations before. A single PR for fixing out of process backgrounds is not the place to try to fix every crash from master as it gets too unwieldy and if squashed too hard to bisect later if new bugs come in. 

The only way that related to the in-process work is if a fix for it one way does not work the other way. Stacktrace would be different, as an out of process clock should not be able to take down the whole panel if it crashes at the expense of having to create more of its own structure.

Dealing with crash on adding locations should be a separate PR, since it happens both in process and out and did not come from this PR

lukefromdc:
This warning from the fish on a vertical panel and not rotated is inherited from master:
`
** (mate-panel:3229): CRITICAL **: fish_applet_draw: assertion 'fish->surface != NULL' failed
`

lukefromdc:
Adding back the fish is now fixed, and the tray does not cause any warnings on removal or adding back. That just leaves the wncklet and this is done. Any other bugfixes I should do against master and rebase this against, though there is the concern with so many other open PR's. If that is an issue I could keep them all in this branch, and just now squash them into the rest at the final merge. On the other hand, we might want to cherrypick any fixes for other bugs prior to MATE 1.20's release.

raveit65:
Confirmed that adding the fish applet again is working now, and all warnings are gone.
Btw. i am not against to merge this PR to master if remaining warnings for window-list and workspace-switcher are solved.
The main issue that the panel crashed if removing applets is solved, remaining minor issue can be fixed later at master.
And i think that most ppl and distro-maintainer use the default build option at the moment, which is build-out-of-process.
I don't ask any more for a review by other team members, as it looks like they want to fix possible issues in stable releases of their distros ;-)

lukefromdc:
Tonight's commit cleans up freeing the window list and workspace selector, and stops the warnings on removing them. No segfaults or warnings in removing them when built in-process now. Will check later to ensure none of these create issues with out of process builds, no reason they should but this must be checked.

raveit65:
Great, all warnings are gone with adding/removing clock, fish, tray, and wncklet applets to the panel.
I see only one new one (for me) with wncklist applet, but this isn't related to this PR as i can reproduce it also with a tarball generated from master if applets are build with in-process.
Un-expanding the panel with the checkbox in preferences stops displaying the wncklist-applet.
Well, any remaining issues can be fixed later after merge, as they aren't related to this PR and redraw issues and warnings are fixed.
...great job :-)

raveit65:
I think the *revert* commits can be squashed?

lukefromdc:
Yes, they can be. Will rebase, squash those, and merge

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/025a11fdaa76111afc8e03faf6ac36b17028e585
through
https://github.com/mate-desktop/mate-panel/commit/9f43dd5c7c37b84960d06a98a4738ee40d8bfd50

lukefromdc:
Leaving the branch in case anyone finds a reason to revert. Barring that, Ubuntu and Debian will be able to go back to building with in-process applets as of MATE 1.20, and later porting of mate-applets to in-process will be required for Wayland in the future. I can start on that process, using the same option to build either way as here

PR #627, opened 2017-07-29
don't use deprecated gtk_show_uri

lukefromdc:
Tested on Debian Unstable. Uses NULL as the window, seems to work fine. Help tested from clock, window list, tray, fish, and panel context menu. Tested with applets both in and out of process and help always worked and never saw a redraw problem afterwards.

### DIFF
--- a/applets/clock/clock-utils.c
+++ b/applets/clock/clock-utils.c
@@ -72,10 +72,13 @@ clock_utils_display_help (GtkWidget  *widget,
 		uri = g_strdup_printf ("help:%s/%s", doc_id, link_id);
 	else
 		uri = g_strdup_printf ("help:%s", doc_id);
-
+#if GTK_CHECK_VERSION (3, 22, 0)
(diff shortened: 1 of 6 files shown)


raveit65:
Thanks, seems that using NULL is the solution here. All help links are working and i see no redraw issues with the applets after calling the help link, Theme switching works well too.
I just tested this with Out-of-process applets.
You added some code from build-in-process PR for the fish,
https://github.com/mate-desktop/mate-panel/pull/627/commits/36c481cfba82f6b20f4c2b51a4d7322d00aebea9#diff-3d36d818353cdf7315a51c4624ac23c7R1337
, for that reason the commit doesn't apply on in-process branch.
I guess this was a late night mistake :-)
I will clean the commit local so that i am able to test this with applets compiled in-process.
Maybe you can fix some indents too if you rebase this PR, please.

lukefromdc:
Fish changes and indent cleaned, branch force-updated

PR #632, opened 2017-07-30
add redmond-no-indicators.layout

ThanosApostolou:
Add support for redmond layout without the indicators, for the distributions which don't support indicators (arch, fedora, etc...). I intent to do a pull request in mate-tweak in order to make it show it when no indicators are installed.

I don't know if any other changes in the code are needed, or if it gets automatically copied from the folder when mate-panel is built.

### DIFF
--- a/data/redmond-no-indicators.layout
+++ b/data/redmond-no-indicators.layout
@@ -1,0 +1,73 @@
+[Toplevel bottom]
+expand=true
+orientation=bottom
+size=24
+
+[Object mate-menu]
+locked=true
+position=0
+toplevel-id=bottom
+applet-iid=MateMenuAppletFactory::MateMenuApplet
+object-type=applet
+
+[Object showdesktopapplet]
+locked=true
+position=10
+toplevel-id=bottom
+applet-iid=WnckletFactory::ShowDesktopApplet
+object-type=applet
+
+[Object firefox]
+locked=true
+position=20
+launcher-location=/usr/share/applications/firefox.desktop
+toplevel-id=bottom
+object-type=launcher
+menu-path=applications:/
+
+[Object thunderbird]
+locked=true
+position=30
+launcher-location=/usr/share/applications/thunderbird.desktop
+toplevel-id=bottom
+object-type=launcher
+menu-path=applications:/
+
+[Object welcome]
+locked=true
+position=40
+launcher-location=/usr/share/applications/ubuntu-mate-welcome.desktop
+toplevel-id=bottom
+object-type=launcher
+menu-path=applications:/
+
+[Object window-list]
+object-type=applet
+applet-iid=WnckletFactory::WindowListApplet
+toplevel-id=bottom
+position=50
+locked=true
+
+[Object drivemountapplet]
+object-type=applet
+applet-iid=DriveMountAppletFactory::DriveMountApplet
+toplevel-id=bottom
+position=20
+panel-right-stick=true
+locked=true
+
+[Object notification-area]
+object-type=applet
+applet-iid=NotificationAreaAppletFactory::NotificationArea
+toplevel-id=bottom
+position=10
+panel-right-stick=true
+locked=true
+
+[Object clock]
+object-type=applet
+applet-iid=ClockAppletFactory::ClockApplet
+toplevel-id=bottom
+position=0
+panel-right-stick=true
+locked=true


lukefromdc:
How do I test these layouts? If I create a new user I always get the default mate-panel layout of one on top and one on the bottom. Also, a non-indicator layout probably depends on 
https://github.com/mate-desktop/mate-panel/pull/600 (Add a gsettings key to enable/disable SNI Support)
given that current mate-panel 1.18 and 1.19 always support indicators

ThanosApostolou:
Well, I don't have much experience so I'm not really sure how this is supposed to work. The original problem was that in arch based distributions **mate-tweak** was used together with the panel layouts from **ubuntu-mate-settings**. However, @flexiondotorg told me that now most of them are require several indicators which don't exist in arch based distros https://github.com/ubuntu-mate/mate-tweak/issues/27 (indicator-application, indicators-messages, indicator-sound, etc...). So, I thought to add redmond panel layout without the need of those indicators.

So, to test it you need to add it at **/usr/share/mate-panel/layouts** and then add the extra lines from here: https://github.com/ubuntu-mate/mate-tweak/pull/28/files at **/usr/bin/mate-tweak,** so that it is shown under **mate-tweak**

Maybe, I haven't understood completely how mate panel layouts and mate-tweak work. So, if someone can help/expalin me if I'm doing anything wrong.

lukefromdc:
I have no idea what determines which of the layouts is used when mate-panel is started for the first time by a new user or with an empty user directory

raveit65:
> How do I test these layouts? If I create a new user I always get the default mate-panel layout of one on top and one on the bottom.

Selecting layouts is done via gsettings key, no need to use mate-tweak ;-)
[rave@mother ~]$ gsettings get org.mate.panel default-layout
'fedora'

raveit65:
> Add support for redmond layout without the indicators, for the distributions which don't support indicators (arch, fedora, etc...). 

I don't see any indicator applet in our layouts.
If you see that in ubuntu-mate (downstream), than this is here (upstream) the wrong address.

Mate-menu (ubuntu) isn't an project from us, so adding this to a layout can make problems as not all distros ships this applet.

I am not sure about drivemount applet, as it isn't in mate-panel package.
So this layout will fail if mate-applets package isn't installed.
I prefer only to add a layout which use applets from mate-panel package.
Adding layouts with other applets can be done by upstream distro maintainers to be sure all applets are installed.
....imo.

ThanosApostolou:
Hmmm... I didn't think about mate-applets package. However, I see that, for example, the **linuxmint.layout** depends on **applet-iid=MintMenuAppletFactory::MintMenuApplet** and gives an error if I set this panel layout through dconf like you described above. But, I don't believe that's a problem, because no one really changes panel layouts like this. End users change panel layouts through mate-tweak and we can add a check there to make sure that the required applet is in the system, in order to make the panel layout visible (as it's done for linuxmint.layout, opensuse.layout, etc...).

raveit65:
Mate-tweak is from ubuntu and not used by every standard installation in other distros.
I am not against to add another layout, but isn't ubuntu-mate not the right place to add a layout without indicators?
Or do we have other distros which use a layout with indicators as default?

@flexiondotorg @monsta 
thoughts?

monsta:
I don't get the point. We don't have any layouts with indicators here.

flexiondotorg:
I would suggest the layouts being proposed are kept outside of the MATE project. The layouts for Ubuntu MATE and Solus MATE are distro maintainer branding modifications. It would seem sensible for Manjaro to do the same. It also also for more flexibility with regards to release management, you are not tied to the MATE Desktop release cycle ☺️

ThanosApostolou:
Well, mate-tweak is quite popular among distributions, except Fedora (I believe you should add it to Fedora too, because it's needed for simple MATE tasks like turning on/off the desktop icons without messing with dconf-editor). My concern is that mate-tweak shows very few panel layouts (before the panel layouts from ubuntu-mate-settings were used at manjaro's build of mate-tweak, but now most of them require indicators in order to be visible under mate-tweak). So, if the distributions want to include some basic layouts without indicators, then this must be done:

1. In the AUR package for arch
2. In the manjaro specific package
3. In Revenge OS specifically (which has it's own mate-tweak build)
4. In Opensuse's package...

So, I just thought to add some simple panel layouts, which will be distribution independent, so that the work is done once and not multiple times. I chose the name **redmond-no-indicators.layout** because ubuntu-mate-settings has the name **redmond.layout**, so there would be a conflict. If it's ok with you, I would like to add 2-3 more (which won't depend at indicators or anything distribution specific).

ThanosApostolou:
@flexiondotorg Aaaa ok feel free to close this then. I still need to do the PULL requests to ubuntu-mate/mate-tweak so that the panel layouts are recognized by mate-tweak right?

raveit65:
> Well, mate-tweak is quite popular among distributions, except Fedora (I believe you should add it to Fedora too, because it's needed for simple MATE tasks like turning on/off the desktop icons without messing with dconf-editor).

I have really a work overload and i am really amused if other ppl talk about my time management ;-)
For me dconf-editor is the gui to do that.

ThanosApostolou:
@raveit65 Yeah I feel you. I tried to become a Fedora packager, but I gave up because I felt that I didn't have the time to achieve that (I kinda got scared too with all the switching to flatpak for GUI apps conversation, which make me feel that there's not point in learning rpm packaging in depth). If I feel more confident in the future I will try to help where I can with some simple packages (mate-tweak and brisk-menu are really missed from fedora's mate spin).

I will try to add these the layouts which I need locally on manjaro and I will wait for Wimpy to tell me if I need to add the mate-tweak changes (which are needed so that mate-tweak recognizes the new layouts) upstream, or if I need to make patches for the manjaro build.

raveit65:
>  If I feel more confident in the future I will try to help where I can with some simple packages (mate-tweak and brisk-menu are really missed from fedora's mate spin).

You're welcome, i am happy to review your package reviews.

raveit65:
Btw. i don't see that i ever will switch to flatpack.
Why?
only why it is new?
and some people are bored about good all working stuff ?
:-)

flexiondotorg:
Please do file a PR to MATE Tweak to add the support you require. And also ask anyone with their own builds to do the stuff. I'm happy to maintain something that works for all ☺️

PR #633, opened 2017-07-31
avoid deprecated gdk_display_get_screen

sc0w:


raveit65:
For the record , i am just testing this, no regressions in a quick test so far.

PR #634, opened 2017-08-01
avoid deprecated gdk_screen_make_display_name

sc0w:


raveit65:
> The issue with the fish opening as a line when added to the panel was greatly reduced (maybe not by this PR though?)

This was fixed by a commit from you :-)
https://github.com/mate-desktop/mate-panel/commit/0549a98fffbacecfe61d7f646744fce25ce96d75

lukefromdc:
That was only fixed in the dev-in-process-bgfix branch (https://github.com/mate-desktop/mate-panel/pull/622), though if you want I can cherrypick that to master and then rebase the in-process branch

raveit65:
Ahh, normal i build master always with out-process configure flag, so i forgot that the problem exists in master too.
Well, maybe it is useful to cherry-pick harmless single commits from dev-in-process-bgfix branch to reduce whole the PR.

lukefromdc:
Just cherrypicked the fish fix with https://github.com/mate-desktop/mate-panel/commit/7d98669e567f164efbfefc63ac2a94a78ac5798d

PR #636, opened 2017-08-03
avoid deprecated gdk_screen_get_n_monitors

sc0w:
Use `gdk_display_get_n_monitors` instead

since gtk 3.22


raveit65:
Thank you.

PR #640, opened 2017-08-07
don't use excessive string duplication

sc0w:
@monsta you did the same in https://github.com/mate-desktop/caja-extensions/commit/5fa187fd69c9a352bae38cbced4120dc942fc7a6

please test

### DIFF
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -815,7 +815,7 @@ static void display_fortune_dialog(FishApplet* fish)
 	char       **argv;
 	GdkDisplay  *display;
 	GdkScreen   *screen;
-	char        *display_name;
+	const char  *display_name;
 
 	/* if there is still a pipe, close it */
 	if (fish->source_id)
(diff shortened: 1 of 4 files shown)


sc0w:
I see the warning:

> panel-run-dialog.c:412:5: warning: passing argument 6 of ‘g_spawn_async’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
>      display_name,

I am not sure, so, I decide to close this PR

monsta:
Hmm, `g_spawn_async` doesn't accept const pointer for that argument.

raveit65:
@monsta 
Should i merge existing similar PRs like they are?
With
**display_name = g_strdup (gdk_display_get_name (display));**

monsta:
It doesn't hurt, just a bit more memory allocation.

lukefromdc:
Using a few extra bytes of RAM is preferable to a few extra warnings or especially actual bugs that users would notice.

PR #624, opened 2017-07-18
Forces a window scale of 1 for the panel. Closes #621

flexiondotorg:
This pull request forces a specific window scale for `mate-panel`, instead of using the default or user configured scale.

Tested on a Dell XPS 15 with 3840 x 2160 and `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` exported to the Xsession. mate-panel dimensions are now correct.

PR #642, opened 2017-08-11
Add missing include.

flexiondotorg:
Fixes the build warning discussed here: 

  * https://github.com/mate-desktop/mate-panel/pull/624#issuecomment-321738661

### DIFF
--- a/mate-panel/main.c
+++ b/mate-panel/main.c
@@ -13,6 +13,7 @@
 #include <sys/wait.h>
 
 #include <glib/gi18n.h>
+#include <gtk/gtkx.h>
 
 #include <libegg/eggdesktopfile.h>
 #include <libegg/eggsmclient.h>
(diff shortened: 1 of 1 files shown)


sc0w:
@monsta said in https://github.com/mate-desktop/mate-panel/pull/624#issuecomment-321763141

#include <g**d**k/g**d**kx.h>

sc0w:
I have tested, and the warning is gone with `#include <gtk/gtkx.h>` or `#include <gdk/gdkx.h>`

but, according to the documentation, I think it needs `#include <gdk/gdkx.h>`

monsta:
Yes, https://developer.gnome.org/gdk3/stable/gdk3-X-Window-System-Interaction.html#gdk3-X-Window-System-Interaction.includes

lukefromdc:
I tested this too, and it removed the build warning. Not sure, but my guess is the only reason the code worked at all was the inclusion of one of these in a file which is itself included. That must be what that "nested extern" means, and I can see how these could make a real mess out of maintaining a package.

PR #645, opened 2017-08-14
avoid deprecated gdk_screen_get_monitor... functions:

sc0w:
avoid deprecated:

gdk_screen_get_monitor_geometry
gdk_screen_get_monitor_at_window
gdk_screen_get_monitor_at_point

Issue #447, opened 2016-06-08
Removing any applet crashes mate-panel, at least in GTK3


lukefromdc:
With mate-panel from current GIT master buitl with GTK 3.21, applets add and run fine and I get no applet crashes, but removing any applet crashes the panel. In process or out, either way the panel crashes.

I got this error when removing the inhibit applet from the laptop's panel as a test:
traps: mate-panel[4355] general protection ip:42e22f sp:7ffe398a99c0 error:0 in mate-panel[400000+8b000]

Similar errors from removing any other applet, on any system with mate from current GIT master
##

raveit65:
I can reproduce this only in fedora rawhide (mate gtk3) tested with 1.14.1 release,
but not with f24.
Abrt gaves me a core_backtrace:
```
{   "signal": 11
,   "executable": "/usr/bin/mate-panel"
,   "only_crash_thread": true
,   "stacktrace":
      [ {   "crash_thread": true
        ,   "frames":
              [ {   "address": 93883586263423
                ,   "build_id": "982839219a40ac1131ce4a9a44bc3f834db52d6b"
                ,   "build_id_offset": 190847
                ,   "function_name": "mate_panel_applet_locked_change_notify"
                ,   "file_name": "/usr/bin/mate-panel"
                }
              , {   "address": 140087278436292
                ,   "build_id": "af0d4a312054e41039eb733581ccb819011a633d"
                ,   "build_id_offset": 73668
                ,   "function_name": "g_cclosure_marshal_VOID__STRINGv"
                ,   "file_name": "/usr/lib64/libgobject-2.0.so.0.4901.0"
                }
              , {   "address": 140087278425988
                ,   "build_id": "af0d4a312054e41039eb733581ccb819011a633d"
                ,   "build_id_offset": 63364
                ,   "function_name": "_g_closure_invoke_va"
                ,   "file_name": "/usr/lib64/libgobject-2.0.so.0.4901.0"
                }
              , {   "address": 140087278537945
                ,   "build_id": "af0d4a312054e41039eb733581ccb819011a633d"
                ,   "build_id_offset": 175321
                ,   "function_name": "g_signal_emit_valist"
                ,   "file_name": "/usr/lib64/libgobject-2.0.so.0.4901.0"
                }
              , {   "address": 140087278539583
                ,   "build_id": "af0d4a312054e41039eb733581ccb819011a633d"
                ,   "build_id_offset": 176959
                ,   "function_name": "g_signal_emit"
[...]

Issue #414, opened 2016-02-28
GTK+-3.20: panel crashed if removing an applet


raveit65:
This is reproducible with every applet.
```
Thread 1 (Thread 0x7f68eae77a80 (LWP 24494)):
#0  0x000055a596b33bdf in mate_panel_applet_locked_change_notify (settings=0x55a597512300 [GSettings], key=0x7f68e79b733a "locked", applet=0x55a5972001f0) at applet.c:186
        __inst = 0x55a5972001f0
        __t = 94169690598512
        __r = <optimized out>
#1  0x00007f68e763dfac in g_cclosure_marshal_VOID__STRINGv (closure=0x55a5971c8630, return_value=<optimized out>, instance=<optimized out>, args=<optimized out>, marshal_data=<optimized out>, n_params=<optimized out>, param_types=0x55a596f8a1e0) at gmarshal.c:1794
        cc = 0x55a5971c8630
        data1 = <optimized out>
        data2 = <optimized out>
        callback = <optimized out>
        arg0 = 0x7f68e79b733a
```
full stacktrace at
https://dl.dropboxusercontent.com/u/49862637/Mate-desktop/Bugs/backtrace-applet-remove

lukefromdc:
For some reason this does not occur on my system. I am using gtk3.19.10 from GIT master current as of Feb 26 at the moment. Wonder if this is gtk3.19 version specific?

raveit65:
It looks like this is unreleated to gtk+ toolkit version.
I see this crash also in fedora 24 Mate gtk+-2 environment test with tarball from #/419 PR.
Need to test this more.

raveit65:
@lukefromdc 
which glib2 version are you using?

lukefromdc:
I've got 2.46.2-3 from Debian Unstable

raveit65:
It is an glib (probably debugging issue),
Fedora's bug reporting tool pointed me to an old an closed report from mate-panel.
https://bugzilla.redhat.com/show_bug.cgi?id=1068968
our upstream report:
https://github.com/mate-desktop/mate-panel/issues/166
After installing stable glib2-2.48.0 the issue is gone in gtk2 and gtk3 environment.
Same with crashing of some applications if closing, ie. caja-actions.
I'm so so bored about their debbuging stuff during during development.
Every year the same.
Ok, i will leave the report 1 or 2 days open to be sure.

Issue #166, opened 2014-02-13
mate-panel 1.7.2 crashes when deleting an applet


dmashal:
Seems to be same issue as  https://github.com/mate-desktop/mate-panel/issues/125

Please let me know what data you need to troubleshoot the problem.

PR #648, opened 2017-08-21
Fix crashes on moving/removing applets with glib 2.53.4 or later

lukefromdc:
*On destroying applet disconnect signals for functions that otherwise segfaults with glib 2.53.4 or later.
*Fish: it's not always enough for a function to return immediately if applet destroyed anymore
*panel-action-button: suppress a warning

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -501,8 +501,6 @@ get_updated_timeformat (ClockData *cd)
 static void
 update_timeformat (ClockData *cd)
 {
-        if (cd->settings == NULL)
-                return;
         if (cd->timeformat)
                 g_free (cd->timeformat);
         cd->timeformat = get_updated_timeformat (cd);
@@ -703,8 +701,6 @@ update_tooltip (ClockData * cd)
 static void
 refresh_clock (ClockData *cd)
 {
-        if (cd->settings == NULL)
-                return;
         unfix_size (cd);
         update_clock (cd);
 }
@@ -712,9 +708,6 @@ refresh_clock (ClockData *cd)
 static void
 refresh_clock_timeout(ClockData *cd)
 {
-        if (cd->settings == NULL)
-                return;
-
         unfix_size (cd);
 
         update_timeformat (cd);
@@ -756,6 +749,9 @@ free_locations (ClockData *cd)
 static void
 destroy_clock (GtkWidget * widget, ClockData *cd)
 {
+        if (cd->settings)
+                g_signal_handlers_disconnect_by_data( cd->settings, cd);
+
         if (cd->settings)
                 g_object_unref (cd->settings);
         cd->settings = NULL;
@@ -2002,9 +1998,6 @@ static void
 update_weather_bool_value_and_toggle_from_gsettings (ClockData *cd, gchar *key,
                                                  gboolean *value_loc, const char *widget_name)
 {
-        if (cd->settings == NULL)
-                return;
-
         GtkWidget *widget;
         gboolean value;
 
@@ -2084,9 +2077,6 @@ location_set_current_cb (ClockLocation *loc,
 static void
 locations_changed (ClockData *cd)
 {
-        if (cd->settings == NULL)
-                return;
-
         GList *l;
         ClockLocation *loc;
         glong id;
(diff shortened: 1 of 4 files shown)


lukefromdc:
Rebased after
https://github.com/mate-desktop/mate-panel/pull/645

raveit65:
start to testing this with unstable/stable  or f26/f27.

lukefromdc:
The tests I use are this: I have the clock, fish etc in process, and I use the "connect to server" applet as a left-click movable test applet. With Master and unstable glib, removing any applet in-process or out crashes the panel again, and moviong the "connect to server" applet works once, then segfaults the second time you move it. Sometimes you get a bunch of warnings, and the segfaults mostly relate to mate_panel_applet_locked_change_notify and within it to  

```
	if (applet == NULL || !GTK_IS_WIDGET (applet))
		return;
```

It can't be fixed there, the handlers invoking functions calling this all have to be properly disconnected. The fish had another problem, where handlers that were set to return immediately if the fish had been destroyed and normally worked fine caused segfaults with glib 2.53.4 or later. The wncklet already is set up the correct way, the clock for some reason works fine with unstable glib even with the handlers called and returning.

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/648/commits/2dbcb02ab02f99d423f99aea92f3e234475f3bb0
applies the same fix glib 2.53.4 or later required for the fish to the clock. Although having the handlers return when called after destroy was not generating warnings or segfaults, it could with a future version of glib the way this is going, plus this is better code. No warnings when running from the terminal and adding, then removing the clock either way

raveit65:
Panel is built with out-of-process applets and using glib2-2.53.5-1.fc27.x86_64 with fedora 27.
Without the PR removing any applet can crash the panel. But i see no problem s with moving an applet to another panel.
With the PR those crashes are gone. :-)
I tested this with removing several applets.

With stable fedora 26 and stable glib2-2.52.3-1.fc26.x86_64 i see no different or regressions with using that fixes in a quick test.

lukefromdc:
I'm going to run through the GTK versions 3.14 through GTK 3.22 before merge, just in case

lukefromdc:
I just tested this with every supported GTK version 3.14 through GTK 3.22, with all possible and with no applets built in-process. In every case applets could be moved or removed without triggering any warnings. All of this with glib 2.53.6, and no other program is giving me trouble with that.

lukefromdc:
Leaving the branch for the moment in case anyone has to revert

monsta:
Wow... disconnecting the signal that leads to calling `mate_panel_applet_locked_change_notify` actually fixes very old random crash, which happens when already-freed memory is passed into one of the arguments of that function. It was reported in #125, #128, #166, #414, #447.

monsta:
Also there's a second backtrace in #128 with crash in `locations_changed`, which apparently has the same cause. It should be fixed by disconnecting signals in https://github.com/mate-desktop/mate-panel/commit/2dbcb02ab02f99d423f99aea92f3e234475f3bb0, but I don't see that commit in 1.18...

alexarnaud:
Maybe @raveit65  could release a version 1.18.5.1 to ship the commit if it affects multiple users.

Best regards.

raveit65:
@monsta 
please push **everything** what is missing to 1.18 and i will do a new release.
... some reports are from fedora.

monsta:
Yeah, I'll pick that commit and check what else could be missing.

monsta:
@lukefromdc: https://github.com/mate-desktop/mate-panel/commit/9f43dd5c7c37b84960d06a98a4738ee40d8bfd50 is about disconnecting signal handlers as well, so it might be good for out-of-process applets too. So it could be used in 1.18. Maybe we can optimize it too: use `g_signal_handlers_disconnect_by_data` as in https://github.com/mate-desktop/mate-panel/commit/2dbcb02ab02f99d423f99aea92f3e234475f3bb0.

lukefromdc:
I haven't had any trouble with it in either in-process or out-or-process applets (the stuff in mate-applets is still out-of-process), so I don't see any reason not to. As for optimizing, I don't havce enough experience with g_signal_handlers_(foo) to evaluate that one way or the other. I was focussed on getting it to work, and did everything by experiment and A/B/A testing

monsta:
Ok, please check https://github.com/mate-desktop/mate-panel/pull/680.

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/680 confirmed good as applied to master as of Nov 14

monsta:
Ok, #680 is merged, and new fixes are picked into 1.18.

Issue #653, opened 2017-09-03
menu buttons don't respond (open/close menu) consistently to clicks


whitwuv:
#### Expected behaviour
If a menu is closed and I click it's menu button, that menu should always open; if a menu is open and I click it's menu button, that menu should always close.

#### Actual behaviour
When a menu is open and I click it's menu button, it doesn't always close the menu; when a menu is closed and I click it's menu button, it doesn't always open.

I encounter this issue several times per day, since upgrading from Fedora 24 to Fedora 26.

#### Steps to reproduce the behaviour
For the issue where, when the menu is open, if the menu button is clicked, the menu doesn't always close:
place your mouse over a menu button and start clicking as fast as you can; you should notice that the menu spends most of the time open and isn't closing on every 2nd click

try the same in Fedora 24, and you will notice the menu button responds immediately to every click, opening or closing the menu, respectively

when does this issue come up? when user opens a menu but changes their mind, so tries to dismiss the menu by clicking the menu button again.

For the issue where, when the menu is closed, if the menu button is clicked, the menu doesn't always open (sometimes i'll have to click several times before it will open):
press your mouse cursor up against the very top edge of the screen when clicking the menu button (this is usually where my mouse falls when I click (since mouse has to go up there to unhide the panel (although autohide doesn't have to be enabled to reproduce))); start at the left of one of the menu buttons and click ... if the menu opens, click again to close it, then move the cursor a tiny amount to the right, still pressed up against the top of the screen and click again ... repeat as necessary until you find one of the "dead zones" where clicking doesn't do anything. for me, for example (using TraditionalOk theme (although the issue affects e.g. BlueMenta as well) and Bluecurve-inverse mouse theme, with Appearance Preferences->Fonts set to use "Subpixel smoothing (LCDs)", for the "Places" menu, if I line the left edge of the mouse cursor arrow up with the left edge of the "l", then move the cursor one pixel to the right, there's a dead spot. there are several more that seem to remain in the same general areas.

I encounter this issue daily with normal use.

this issue didn't occur in Fedora 24.

#### MATE general version
1.18.0

#### Package version
1.18.4

#### Linux Distribution
Fedora 26

lukefromdc:
I was unable to duplicate this with mate-panel 1.19.3 from an August 23 git pull on Debian Unstable with GTK 3.22.19 and glib 2.53.6 locally built.

raveit65:
[...]

PR #654, opened 2017-09-04
GTK 3.22: Avoid deprecated gtk_menu_popup

lukefromdc:
Port to gtk_menu_popup_at_widget and when that cannot be used gtk_menu_popup_at_pointer.

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -780,6 +780,7 @@ container_has_focusable_child (GtkContainer *container)
 	return retval;
 }
 
+#if !GTK_CHECK_VERSION (3, 22, 0)
 static void
 mate_panel_applet_position_menu (GtkMenu   *menu,
 			    int       *x,
@@ -862,11 +863,16 @@ mate_panel_applet_position_menu (GtkMenu   *menu,
 	*y = menu_y;
 	*push_in = FALSE;
 }
+#endif
 
 static void
 mate_panel_applet_menu_popup (MatePanelApplet *applet,
-			 guint        button,
-			 guint32      time)
+#if GTK_CHECK_VERSION (3, 22, 0)
+                              GdkEvent *event)
+#else
+                              guint        button,
+                              guint32      time)
+#endif
 {
 	GtkWidget *menu;
 
@@ -884,11 +890,19 @@ mate_panel_applet_menu_popup (MatePanelApplet *applet,
 	context = gtk_widget_get_style_context (GTK_WIDGET(toplevel));
 	gtk_style_context_add_class(context,"gnome-panel-menu-bar");
 	gtk_style_context_add_class(context,"mate-panel-menu-bar");
+#if GTK_CHECK_VERSION (3, 22, 0)
+	gtk_menu_popup_at_widget (GTK_MENU (menu),
+	                          GTK_WIDGET (applet),
+	                          GDK_GRAVITY_SOUTH_WEST,
+	                          GDK_GRAVITY_NORTH_WEST,
+	                          event);
+#else
 	gtk_menu_popup (GTK_MENU (menu),
 			NULL, NULL,
 			(GtkMenuPositionFunc) mate_panel_applet_position_menu,
 			applet,
 			button, time);
+#endif
 }
 
 static gboolean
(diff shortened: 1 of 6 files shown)


lukefromdc:
One behavior change resulted: the context menu for the main menubar now follows the pointer where previously it aligned with the left hand edge of the entire menubar. Trying to get the widget in mate-panel-applet.c for this caused the context menu for the main menubar and only the main menubar to come up far out of position, well to the right. Apparently the widget here is the entire panel, so pointer has to be used unless we want to get fancy and find the rectangle occupied by the main menubar and attach the context menu to that,

lukefromdc:
An attempt to use gtk_menu_popup_at_widget for the panel main menubar right-click menu did not work unless the menu was all the way to the left. It worked then and only then, otherwise the context menu was rendered too far right (or down for vertical panel), and the farther the applet was from the left or top the worse the problem.  When it did work, aligning the context menu with the left edge of the menubar was less natural than positioning it to the pointer, due to the width of the menubar. 

When right-clicking "system" the behavior with gtk_menu_popup_at_pointer puts the context menu close to the original location,  while aligning it with the far left edge of the menubar looked odd. Anyway, the experimental code exists on another branch at
https://github.com/mate-desktop/mate-panel/tree/dev-menu-popup-experimental
I excluded it from this PR as it not only is not ready but did not look as good when it did work. The branch is there though if anyone wants to play with it

raveit65:
First, the deprecation fix doesn't solve https://github.com/mate-desktop/mate-panel/issues/653
Any way, we should fix the deprecation.
I didn't test the experimental branch yet. but using gtk_menu_popup_at_pointer for the menubar looks OK for me.
Only if the menubar is at the right edge of the monitor, than the right-click-context-menu pops up on the left side of the pointer, which looks a bit weird if you click on the application menu.
The application-menu isn't at the monitor edge in this case. It seems to be a new behaviour with the new functions that the context-menu pops up on the other side of the pointer if there isn't enough space.
But i think this is a rare case and i don't know a layout for it. I never heard about a *RTL* layout.

Can we try to use gtk_menu_popup_at_pointer also for the work-space-switcher, if it's possible to use it only for this wnck-applet?
I am using 8 workspaces in one row and and if you click on the right side of the applet with the pointer, the context-menu pops up on the left side, which looks weird.

raveit65:
Ok, i just tested experimental branch.
For me using gtk_menu_popup_at_pointer for the menubar context-menu looks better.
Only for the rare case that the menubar is on the right side of the monitor it seems a better alternative.

lukefromdc:
Rebased after https://github.com/mate-desktop/mate-panel/commit/8a158fe4fba20e89a055e304357db68111a45422

lukefromdc:
I just found that using gtk_menu_popup_at_pointer for the wncklet/switcher does not work, because that is mate-panel-applet.c and then does not work for the tray. The tray menu only pops up on the handle, and gtk_menu_popup_at_pointer apparently calls gtk_menu_popup_at_rect within GTK and I get this error when trying to open the tray context menu:

`(mate-panel:14881): Gtk-CRITICAL **: gtk_menu_popup_at_rect: assertion 'GDK_IS_WINDOW (rect_window)' failed`

That same tray context menu works fine with gtk_menu_popup_at_widget, at the expense that a long switcher is going to align the menu with the left edge

lukefromdc:
gtk_menu_popup_at_pointer also does not work for the window list context menu, returning the exact same error.It works on the switcher quite nicely, but fails whenever it is called with the pointer on a handle.

 Apparently gtk_menu_popup_at_pointer invoked the same invalid window issue for gtk_menu_popup_at_rect within GTK anytime it's called on a pointer over a handle in mate-panel

raveit65:
Hmm, maybe we leave that for the future if we ever will switch to wayland or gtk+4 ?
No need to delete the branch in that case.
Still wondering how this will handle with gnome-panel as switch to gtk+-4 and x-wayland will be earlier....

raveit65:
At least all this new stuff isn't mark as stable in gtk docs.

lukefromdc:
We've already got several other deprecations that cannot be easily fixed, such as the deal with reparenting a moved applet and the actiongroup stuff that removes icons from menus if it is "fixed." We didn't get any advantage from fixing this one, and mate-panel cannot now run in Wayland due to direct dependencies on X and bugs in xwayland

lukefromdc:
Maybe we just wait for whatever gnome-paned does and copy it?

lukefromdc:
Closing this. leaving branch: we get no advantages and a visual issue in at least one use case

raveit65:
yeah, let's wait, it isn't urgent.

raveit65:
For the record, i think we can't use gtk_menu_popup_at_pointer for the panel as we support keyboard navigation.
I tried to use *gtk_menu_popup_at_pointer* for volume-applet as GtkStatusIcon isn't a widget.
So i couldn't use *gtk_menu_popup_at_widget* here.
The result was that the right-click menu pops up where the pointer is, possible at another place of the monitor. So this breaks keyboard navigation.
For me this new gtk functions isn't really smart outside of gnome-(s)hell.

lukefromdc:
This, in turn means that a port to Wayland that supports keyboard navigation will need gtk_menu_popup_at_rect and a location function to calculate where that rectangle should be.  The existing (deprecated) function apparently has issues in wayland with positioning.

PR #655, opened 2017-09-05
Add option to context menu to reset the panel

ikem-krueger:
It can happen that a panel configuration is so messed up, that only a reset helps.

Only a few users know how to do that.

This patch adds a context menu entry "Reset Panel", which enables a user to do that.

Internally the function behind "mate-panel --reset" is called.

### DIFF
--- a/mate-panel/panel-context-menu.c
+++ b/mate-panel/panel-context-menu.c
@@ -47,6 +47,7 @@
 #include "panel-lockdown.h"
 #include "panel-addto.h"
 #include "panel-icon-names.h"
+#include "panel-reset.h"
 
 static void
 panel_context_menu_show_help (GtkWidget *w,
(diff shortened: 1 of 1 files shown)


lukefromdc:
I can confirm this works as intended

lukefromdc:
Ideally this would have a confirmation dialog, but this is better without one than users being unable to reset their panel at all. It's much eaiser to rearrange an accidently reset panel than to get out of a borked configuration, assuming the panel will start at all. Merging this

lukefromdc:
I was not succesful in trying to add a popup confirmation dialog for this, which would be the best solution. Other options are to move it in the menu, perhaps to between "delete this panel" which does have a confirmation dialog and the separator below it, or to revert this entirely.

Simplest fix is just to move it in the menu, though as the separator is of course not a meny entry it would be possible to hit this trying to hit "new panel." If someone can get a confirmatiion dialog to work that is better than anything else.

lukefromdc:
Should I revert this for now?

lukefromdc:
Try https://github.com/mate-desktop/mate-panel/pull/664 , it moves "reset panel" to a safer location, between "delete this panel" which has a confirmation dialog and "new panel" with separators above and below the "delete this panel" and "reset panel" pair of menu items.

raveit65:
I noticed that "delete this panel" option only has a confirmation dialog if the panel has a item, but not with an empty panel.

ikem-krueger:
The code for that dialog is in „panel.c“. It is cluttered across four different functions. Still need to find out how they all work together and how to make use of it.

lukefromdc:
I finally managed to get a confirmation dialog working:
https://github.com/mate-desktop/mate-panel/pull/665
Closed https://github.com/mate-desktop/mate-panel/pull/664 in preference to doing this right, with the dialog

PR #665, opened 2017-09-29
panel-context-menu: Add confirmation dialog to panel reset option

lukefromdc:
Fix for https://github.com/mate-desktop/mate-panel/pull/655#issuecomment-332534233 , which is the danger of accidently resetting a panel and trashing  a user created layout


raveit65:
damned, i was 1 min to llate with my review ...... :-)
works like a charm , ready to go.

PR #669, opened 2017-10-12
status-notifier: hide mate-polkit icon if passive

lukefromdc:
Hide the mate-polkit statusnotifier icon (and only that icon) if it is set passive.  Works with https://github.com/mate-desktop/mate-polkit/pull/38 in mate-polkit to fix https://github.com/mate-desktop/mate-polkit/issues/35 (status notifier icon keeps showing after dropping elevated privileges). Keep the other passive indicators showing as they are too often needed.


lukefromdc:
Reviewers please test this and https://github.com/mate-desktop/mate-polkit/pull/38 together, this could not be fixed in a single package

PR #671, opened 2017-10-13
GTK 3.22/status-notifer Fix menu position if size changed since last shown

lukefromdc:
Mostly fixes serious mispositioning (on popup) of network-manager applet (indicator mode) menus when connections were changed on last showing of menu and applet is on bottom panel and especially on the bottom right. A slight low positioning on a bottom panel can remain when a connection is activated and a few menu items are added, thus expanding the menu while it is still showing.

### DIFF
--- a/applets/notification_area/status-notifier/sn-dbus-menu.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu.c
@@ -111,7 +111,6 @@ layout_update_item (SnDBusMenu *menu,
     {
       sn_dbus_menu_item_update_props (item, props);
     }
-
   return item->submenu;
 }
 
@@ -179,8 +178,11 @@ get_layout_cb (GObject      *source_object,
 
   g_hash_table_remove_all (menu->items);
   layout_parse (menu, layout, GTK_MENU (menu));
-  /*Reposition to accomodate size change if necessary*/
+#if GTK_CHECK_VERSION (3, 22, 0)
+  /* Reposition menu to accomodate any size changes   */
+  /* Menu size never changes with GTK 3.20 or earlier */
   gtk_menu_reposition(GTK_MENU(menu));
+#endif
   g_variant_unref (layout);
 }
 
--- a/applets/notification_area/status-notifier/sn-dbus-menu.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu.c
@@ -179,6 +179,8 @@ get_layout_cb (GObject      *source_object,
 
   g_hash_table_remove_all (menu->items);
   layout_parse (menu, layout, GTK_MENU (menu));
+  /*Reposition to accomodate size change if necessary*/
+  gtk_menu_reposition(GTK_MENU(menu));
   g_variant_unref (layout);
 }
 
--- a/applets/notification_area/status-notifier/sn-item.c
+++ b/applets/notification_area/status-notifier/sn-item.c
@@ -277,6 +277,8 @@ sn_item_button_press_event (GtkWidget      *widget,
           gtk_menu_popup (priv->menu, NULL, NULL,
                           sn_item_popup_menu_position_func, widget,
                           event->button, event->time);
+          /*Fix positioning if size changed since last shown*/
+          gtk_menu_reposition(priv->menu);
 #endif
         }
       else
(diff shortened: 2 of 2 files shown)


lukefromdc:
This is within a GTK 3.22 only function, no need to test on other GTK versions. Tested with tray in top, bottom,  left, and right side panels

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/671/commits/42c22dccba296dae429a2bd60adfcba615308932 repositions the menu when the size changes while it is showing, the first commit ensures it gets the correct position on initial popup if the size changed since it was last shown. 

The second commit reduces the likelihood of a status notifier menu on the bottom panel for something like GNOME's nm-applet in indicator mode ending up too low when it repeatedly resizes while showing, though it is still possible in some cases of repeated resizing and moving to miss the last move and end up low. With only the first commit this happened every time networking with wireless was disabled and re-enabled. 

With neither commit nm-applet (indicator) menus on bottom panels consistantly came up in the wrong positions entirely after enabling or disabling networking, as they were using their last known top positions it seemed. Thus they looked OK on and only on top or side panels. This fixes most of the trouble on bottom panels but not yet all of it.

lukefromdc:
https://github.com/mate-desktop/mate-panel/commit/6c6b662131c17f6e19835c32151ad17068118f11

Applies the first fix to GTK 3.20 and earlier(turns out its needed), and excludes the second fix from GTK 3.20 and earlier(where if causes the menu to flash). With GTK 3.20 or earlier, the menu sent by nm-applet never resizes once it has been sent. If the icon is clicked on while the connection is still being set up, the menu comes up short of full size and must be scrolled, whether as an indicator or as a tray applet.

As an indicator and with GTK 3.22, this is fixed and the menu changes size as items get added and removed.  https://github.com/mate-desktop/mate-panel/pull/671/commits/ae740f3ffe10aff6505921f8b78a5ff90a94ab33 reduces the instances oif it rendering low, but does not stop them all due to what appears to be either libappindicator or more likely GTK 3.22 itself being able to add an item late, and none of the code here seems to get called again in that case.

The indicator/GTK 3.22 case replaces deprecated gtk_menu_popup and its GtkMenuPositionFunc with gtk_menu_popup_at_widget , and that seems to be what permits the menus to be dynamically resized. This is at the expense that when on the bottom panel they must be moved each time they are resized, as GTK adds or subtracts space from the bottom. GNOME seems to assume menus will always be popdown from the top and never popup from the bottom. The right fix would be for GtkMenu to include a direction property so this could be set by the application. GdkGravity only determines where the menu it attached to the parent widget and in this case changes to it seem to do nothing at all.

Since gtk_menu_popup is used in both network-manager-gnome and in gnome-panel, the non-resized menu remains an issue in gnome-panel and any panel with nm-applet not run in indicator mode. 

Interestingly,  failure to add menu height for menu items used to be a problem in libappindicator itself, Ubuntu patched it but only in GTK 2 some years ago:
https://bugs.launchpad.net/glipper/+bug/1203888
Now in GTK 3.22 and only GTK 3.22 it's possible to fix most but not quite all of it in mate-panel without touching libappindicator's code.

The sni support recently added to mate-panel offers three interesting possiblities:
1: Working around exactly this sort of bug in tray menus from non-MATE apps
2: Getting control of menu theming from non-MATE apps like nm-applet(the reason I started playing with this). 
3: Wayland-porting tray applets to libappindicator allows not using Xembed, thus clearing the way for a future Wayland port.

lukefromdc:
I tested with all supported GTK versions before merging. Note that to use nm-applet (network-manager-gnome) to duplicate the dynamic menu resize and move issues requires a slow to enable connection, may only be an issue with wifi connections.

All supported GTK 3 versions prior to 3.22 gave the shortened menus on adding items, and testing this revealed that my theme was not supporting scrollbars in menus(rarely seen when things work right, nor arrows in menus with GTK 3.20 or later. Once I fixed my theme or changed to another theme, all of these GTK 3 versions worked without issue. 

On 3.22, the shortened menus are avoided entirely by the replacement for deprecated gtk_menu_popup, and that's where calling gtk_menu_reposition on the now dynamically resized menu does its job. On a bottom panel, menus by default still resize from the top, as though they were on a top panel, the reposition function usually catches and fixes that. Limitations in GTK itself seem to preclude doing any better.

PR #679, opened 2017-11-10
Fix some simple warnings found by Cppcheck

lukefromdc:
Static code analysis using Cppcheck found these simple warnings (considered style warning) as well as a huge number of variables that could be reduced in scope and have been left alone as that would be a major job.

### DIFF
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -473,7 +473,7 @@ format_time (struct tm   *now,
                 }
         }
 
-        if (strftime (buf, sizeof (buf), format, now) <= 0) {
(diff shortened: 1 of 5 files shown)


lukefromdc:
Merged, plus https://github.com/mate-desktop/mate-panel/commit/78d0e5f89c82c42448a1294d32fdd98227caec50 which comments out a variable left unused by https://github.com/mate-desktop/mate-panel/pull/679/commits/64cdec0796d325d07ffc9ec4b56110b621b2947b 

These are commented out rather than all the way removed just in case some kind of problem ever crops up, but that should be impossible as the commented out block could never execute and the commented out variable was used nowhere else.

PR #681, opened 2017-11-17
Clock applet: Fix world map being stretched.

clefebvre:
This commit fixes https://github.com/mate-desktop/mate-panel/issues/594
but in a different way than https://github.com/mate-desktop/mate-panel/commit/03bf5cecd39646e0e7c19af097d5fc64b05c40d5#diff-6d04c3d43db9cf8cc304993f1b61c79e

With the original fix, the menu was wider and thus the world map
was being stretched.

To make more space for the label this commit changes
the button text to a button icon. The action is still quite explicit
thanks to the button tooltip.

The map is no longer stretched and we no longer have space issues.

### DIFF
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -123,7 +123,7 @@ calendar_window_create_calendar (CalendarWindow *calwin)
 	struct tm                  tm1;
 
 	calendar = gtk_calendar_new ();
-	gtk_widget_set_size_request(GTK_WIDGET(calendar), 330, 100);
+	gtk_widget_set_size_request(GTK_WIDGET(calendar), 260, 100);
 	options = gtk_calendar_get_display_options (GTK_CALENDAR (calendar));
 	if (calwin->priv->show_weeks)
 		options |= GTK_CALENDAR_SHOW_WEEK_NUMBERS;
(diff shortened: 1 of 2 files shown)


clefebvre:
https://i.imgur.com/mf4PyaQ.png

lukefromdc:
I just tested this myself, it work though I found one possible negative for users: the icon for selecting a new timezone is much less clear in its meaning than the previous button was until the tooltip comes up. I would suggest editing the clock help file in m-u-g if we do this.

lukefromdc:
This commit is also valuble to themers as without such a large minimum calendar size it is easier to handle smaller fonts

raveit65:
as mentioned by @monsta at irc channel we have already a problem with using this PR.
```
<monsta> clem_l: dammit... not enough :(
<monsta> https://imgur.com/a/pXLFX
<monsta> it's my debian testing VM, TraditionalGreen theme, mate-faenza icons
<clem_l> it makes it larger?
<monsta> it jumps as before - see 2nd image
```

lukefromdc:
OK, that means it needs more work

monsta:
Side note about these [screenshots](https://imgur.com/a/pXLFX) from IRC: there was some change in Debian Testing a few months ago, which made all fonts larger while using the same size (10) and dpi (96). It looks like the size is 11-12, but it's still 10 as before. I have no idea what they changed.

clefebvre:
Hi,

I'll close this PR now. It doesn't "ALWAYS" fix the issue and whichever icon we choose, it's never going to be ideal.

I recommend removing the hover buttons altogether and placing a Set button in the settings dialog instead. Much simpler and much better design imo.

PR #683, opened 2017-11-23
Keep the right icons on bottom still displaying right when the resolution change larger. #682

### LINKED ISSUE #682: the right icons on bottom will offset when the resolution change larger
#### Expected behaviour
 the right icons on bottom will still stay on right positon when the resolution change larger

#### Actual behaviour
the right icons on bottom will offset when the resolution change larger

1.Set the resolution to 600*800, reboot the system.
2.Set the resolution to 1024*768 after the system started.
3.The workspace-switch will be offset rather than on the far right. 

#### MATE general version
1.18.4

#### Linux Distribution
Fedora-Mate 25

zhangwen2:
this commit fixes #682

When the resolution changes, at the same time refresh the location of the icon on the right.

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -2915,6 +2915,11 @@ panel_toplevel_move_resize_window (PanelToplevel *toplevel,
 {
 	GtkWidget *widget;
 
+	GList *list;
+	const char *id;
+	int position;
+	gboolean stick;
+
 	widget = GTK_WIDGET (toplevel);
 
 	g_assert (gtk_widget_get_realized (widget));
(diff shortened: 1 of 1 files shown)


monsta:
Seems to work with a single applet right-sticked... clock for example.

mate-panel/panel-toplevel.c
```diff
@@ -2933,6 +2938,29 @@ panel_toplevel_move_resize_window (PanelToplevel *toplevel,
...
+
+			AppletInfo *info;
+			info = mate_panel_applet_get_by_id (id);
+
+			stick = g_settings_get_boolean (info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY);
+
+			if(stick)
+			{
+				position = g_settings_get_int (info->settings, PANEL_OBJECT_POSITION_KEY);
+				ad->pos = toplevel->priv->geometry.width - position;
```
monsta: Hmm what would happen when the position setting is non-zero?

monsta:
Looks like I found another bug. Setting `panel-right-stick` to true and `position` to something non-zero results in applet placed to the wrong side of the screen.

wrong-placement

The screenshot shows notification area at position 1025 while screen width is 1280. The right-stick setting makes it move to the left instead.

It happens both with and without this PR, but it gets in the way of testing it. 🙂

monsta:
Hmm or is it a feature to count offsets from the right side instead...?

rilian-la-te:
It is, AFAIK))) It just move applets to RTL pack)))

monsta:
It's amazing, after several years of working on MATE, I still discover new features... 😄

mate-panel/panel-toplevel.c
monsta: Nevermind, now I know how positions work with right-stick enabled

monsta:
So I completely forgot to test this with vertical panel. That's where this introduces a regression... https://github.com/mate-desktop/mate-panel/issues/745

monsta:
Ok, https://github.com/mate-desktop/mate-panel/pull/1004 should fix it... hopefully I didn't introduce some other regression.

Issue #620, opened 2017-07-16
mate-panel not hided automatically if we clic on an icon


aaaaadrien:
I use a mate-panel at the bottom of my screen. I enabled the option which hide automatically the panel.
I added on this panel some icons (launchers) of programs that I use the most.
When the bar is hided, if i put my mouse in the bottom of my screen, the panel is showing, and if i remove my mouse, it hides. But If i put my mouse in the bottom of my screen, i click on a launcher to start an application, and i remove my mouse of the panel, the panel doesn't hide. 

#### MATE general version
1.18

#### Package version
1.18.2

#### Linux Distribution
Gentoo

vfscanf:
I experience this bug as well and have found the problem. In mate-panel/panel-toplevel.c (Line 3822) is the following code:

`
	if (!toplevel->priv->auto_hide ||
	    panel_toplevel_contains_pointer (toplevel) ||
	    panel_toplevel_get_autohide_disabled (toplevel))
	  return;
`

Inside the panel_toplevel_queue_auto_hide it checks if the mouse pointer is really outside the panel. This works most of the time, however if you move the mouse quickly the mouse coordinates reported lag behind and the check fails. Removing the panel_toplevel_contains_pointer (toplevel) check fixes the problem, but I'm not sure if it breaks some other part of the code. I'm willing to produce a Pull request but I would like to get feedback from one of the developers first.

meequz:
Came here to report the same issue. I have even recorded a video
[panel-hide-bug.ogv.zip](https://github.com/mate-desktop/mate-panel/files/1513778/panel-hide-bug.ogv.zip)

vfscanf:
I've made a pull request with my little fix. You can build and test it. I've already used it for a few days and didn't detect any issues. Feedback is welcome.

raveit65:
Hmm, i tried to reproduce the issue with fedora 26 and MATE-1.19 dev version, but i don't see the issue here with top or bottom panel.
After i started an application with a launcher the panel hides.
I followed steps to reproduce, did i miss something?

vfscanf:
You have to setup an autohide panel with only application launchers on it. Then click on an application launcher and move your mouse quickly outside the panel (while the application is starting). If you leave the mouse on the panel while the application is starting it doesn't work.

I will install Fedora 26 later and try to reproduce the issue there.

raveit65:
Can you try f27 which use Mate 1.19, normal f26 use 1.18?

[...]

PR #685, opened 2017-11-29
panel-toplevel: Remove pointer boundary check in panel_toplevel_queue…

vfscanf:
Attempt at fixing issue #620 . The pointer boundary check inside this function seems unnecessary so we remove it.


raveit65:
First i'd like to reproduce the issue, see https://github.com/mate-desktop/mate-panel/issues/620#issuecomment-349055945

Issue #695, opened 2017-12-22
Telegram (>= 1.2) has 1 pixel wide systray icon on Mate (1.18)


Maxxie:
#### Expected behaviour
Telegram's systray icon should be visible.

#### Actual behaviour
Since version 1.2, Telegram's systray icon stopped being visible and became a seemingly invisible 1x1 icon in the systray. It's still clickable though.

#### MATE general version
1.18.0

#### Package version
1.18.6

#### Linux Distribution
Debian buster/sid

lukefromdc:
Confirmed with mate-panel from git master (1.19.3 current as of Dec 21, 2017) on Debian Unstable.

Although I have no accounts for actually using Telegram, I was able to get far enough into startup to turn the tray icon on and off. Got the clickable but invisible 1px line, and these errors in terminal:

Cannot connect to server socket err = No such file or directory
Cannot connect to server request channel
jack server is not running or cannot be started
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock
JackShmReadWritePtr::~JackShmReadWritePtr - Init not done for 4294967295, skipping unlock

None of those would impact the icon, but when I tried running mate-panel from terminal while running Telegram and turning the icon on and off, turning it on or turning it off generated one of these warnings each time:

(mate-panel:19063): GLib-GObject-CRITICAL **: 15:35:57.632: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

Don't know if this is related or not

Maxxie:
Those warnings shouldn't be related. I think the same warnings used to appear in earlier versions when the icon was showing fine.
Also I forgot to add that both the official telegram build and the one from the debian repos seem to have this problem.

lukefromdc:
Try testing both with and without sni support. Open dconf-editor, try setting org>mate>panel>general>enable-sni-support to false and restarting the panel (killall mate-panel or mate-panel --replace) and see what you get. Then try setting it back to true and restarting the panel again.  

I strongly suspect this is another QT issue and will affect other QT apps. There is another QT bug affecting the tray
https://github.com/mate-desktop/mate-panel/issues/629
but that one is an excessive CPU use bug

Maxxie:
Thank you for your help, disabling SNI seems to work.
What is SNI exactly? Can't seem to find much about it...

lukefromdc:
[...]

PR #696, opened 2017-12-25
status-notifier: fix failure to show icon for missing icon

lukefromdc:
Fix 1px wide invisible icon in cases such as
https://github.com/mate-desktop/mate-panel/issues/695
Show the "image-missing" icon if the named icon cannot be found or loaded

### DIFF
--- a/applets/notification_area/status-notifier/sn-item-v0.c
+++ b/applets/notification_area/status-notifier/sn-item-v0.c
@@ -252,8 +252,11 @@ update (SnItemV0 *v0)
   if (v0->icon_name != NULL && v0->icon_name[0] != '\0')
     {
       GdkPixbuf *pixbuf;
-
       pixbuf = get_icon_by_name (v0->icon_name, icon_size);
+      if (!pixbuf){
+          /*deal with missing icon or failure to load icon*/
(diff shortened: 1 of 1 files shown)


lukefromdc:
I could not find the code where Telegram actually sets the icon name, but I did find that the installed icon is findable by temporarily hardcoding "telegram" in place of "image-missing" and then running Telegram with https://github.com/mate-desktop/mate-panel/issues/695 applied to mate-panel. Thus, it's not passing NULL or failing to pass any icon name, rather something other then "telegram" is coming up. 

Also, telegram uses org.kde.StatusNotifierWatcher to set up the status notifier, and it looks like Avidemux-gtk is doing somthing different, because it only puts an icon into /usr/share/pixmaps and nowhere else. Thus I can no longer rule out a broader QT to mate-panel issue

raveit65:
Btw. the status icon from dnfdragora, a package-manager for fedora, has the sam problem.
With the PR this application displays the missing-image icon too.
And those app use the hicolor dir too.
https://github.com/manatools/dnfdragora
dnfdragora is the main package-manager for the MATE-Compiz spin from fedora.
I disabled SNI in fedora for this reason.

PR #697, opened 2017-12-26
status-notifier: load icons specified by path

lukefromdc:
Fix  https://github.com/mate-desktop/mate-panel/issues/695 and similar failures to load icons when the application calls for a path and not an icon name

### DIFF
--- a/applets/notification_area/status-notifier/sn-item-v0.c
+++ b/applets/notification_area/status-notifier/sn-item-v0.c
@@ -256,6 +256,10 @@ update (SnItemV0 *v0)
       if (!pixbuf){
           /*try to find icons specified by path and filename*/
           pixbuf = gdk_pixbuf_new_from_file(v0->icon_name, NULL);
+          if (pixbuf){
+              /*An icon specified by path and filename may be the wrong size for the tray */
+              pixbuf=gdk_pixbuf_scale_simple(pixbuf, icon_size-2, icon_size-2,GDK_INTERP_BILINEAR);
+          }
       }
       if (!pixbuf){
           /*deal with missing icon or failure to load icon*/
--- a/applets/notification_area/status-notifier/sn-item-v0.c
+++ b/applets/notification_area/status-notifier/sn-item-v0.c
@@ -254,6 +254,10 @@ update (SnItemV0 *v0)
       GdkPixbuf *pixbuf;
       pixbuf = get_icon_by_name (v0->icon_name, icon_size);
       if (!pixbuf){
+          /*try to find icons specified by path and filename*/
+          pixbuf = gdk_pixbuf_new_from_file(v0->icon_name, NULL);
+      }
+      if (!pixbuf){
           /*deal with missing icon or failure to load icon*/
           pixbuf = get_icon_by_name ("image-missing", icon_size);
       }


lukefromdc:
I just checked this to make sure that it did not break the case of an entirely missing icon by starting Telegram, deleting the icon image in ~/.local and restarting the panel, then turning the status icon on and off in Telegram. With the icon file present it was shown, if finding it was deliberately broken by deleting or renaming the icon the "image-missing" icon was shown, showing that this code did not prevent the following block from executing correctly if the path is bad or (presumably ) the icon was missing and not a path to begin with.

lukefromdc:
Rebased after
https://github.com/mate-desktop/mate-panel/commit/4a4552148355790a35d5e9dd148778bd588d11fd

raveit65:
This fixes telegram-desktop but i see a new issue with the na-tray-applet from dnfdragora.
Now we have a icon-size problem.
normal panel size:
na-tray-1
very huge panel:
na-tray-2

Without the PR dnfdragora displays the missing-icon image if SNI is used and with that commit.
https://github.com/mate-desktop/mate-panel/commit/2d2009c19d0d8b6a4e2eeb3a62144be747e93d94
Icons for this application are stored here.
/usr/share/icons/hicolor/128x128/apps/dnfdragora.png
/usr/share/icons/hicolor/16x16/apps/dnfdragora.png
/usr/share/icons/hicolor/256x256/apps/dnfdragora.png
/usr/share/icons/hicolor/32x32/apps/dnfdragora.png
/usr/share/icons/hicolor/48x48/apps/dnfdragora.png
/usr/share/icons/hicolor/64x64/apps/dnfdragora.png

Deleting the 128x128 icon without updating the icon-cache gives me this error message.
```
[rave@mother ~]$ dnfdragora-updater 
Skipped exception: <[Errno 2] No such file or directory: './dnfdragora.yaml'> 
Traceback (most recent call last):
  File "/usr/bin/dnfdragora-updater", line 25, in <module>
    dnfdragora.updater.Updater().main()
  File "/usr/lib/python3.6/site-packages/dnfdragora/updater.py", line 61, in __init__
    self.__icon  = Image.open(icon_path)
  File "/usr/lib64/python3.6/site-packages/PIL/Image.py", line 2410, in open
    fp = builtins.open(filename, "rb")
FileNotFoundError: [Errno 2] No such file or directory: '/usr/share/icons/hicolor/128x128/apps/dnfdragora.png'
```
So it's clear that the 128x128 icon is used here for some reason?
Let's see what fedora dnfdragora maintainer thinks about.
@besser82
thoughts?

lukefromdc:
. You have found what appears to be an application directly calling for a 128x128 icon by path, not by icon name. 

If the package specifies a 128x128 icon by path, the only possible fix is taking whatever pixbuf has been loaded and rescaling it to whatever size the panel has been set to.  This would have to be done every time, barring some way of testing the pixbuf for size. 

I will see if that's possible but this is getting complex fast, all because 3ed party packages are abusing the icon naming conventions. Testing will require a replaced icon for Telegram, which I have been using as a test mule as I can't install a whole pile of programs I don't use and this one isn't even in repo.

lukefromdc:
Got the icon scaling to work very simply, using
https://developer.gnome.org/gdk-pixbuf/stable/gdk-pixbuf-Scaling.html#gdk-pixbuf-scale-simple
and included a check to prevent this line from executing unless a pixbuf has been loaded and was specified by filename and path. If the icon name was used, this code block is never reached, if the icon name is null or the icon is not found, the scaling function is skipped.

Tested by replacing Telegram's icon with a same-name icon in 128x128 in the same path in .local, verified the issue, and that this fixed it. I then changed panel height up and down, and the Telegram icon scaled with all the others, something not possible with the previous version of this. 

This should also fix the issue in Dragora, which I cannot directly test. Please retest and verify that it does.

raveit65:
I guess next time users will blame us why icons selected by name don't work like this ;-)
Btw. DNF is the replacement for YUM package-manager, for this reason you will never see a gui application for dnf in debian.

monsta:
@lukefromdc: thanks, I've picked all the recent fixes to 1.18 branch too.

PR #698, opened 2017-12-27
Fix build warning: ‘toplevel_configure_event’ defined but not used

sc0w:


lukefromdc:
Works in GTK 3.14 too (oldest we support), should be good to go

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/4a4552148355790a35d5e9dd148778bd588d11fd

raveit65:
> Works in GTK 3.14 too (oldest we support)

I think with 1.20 next year we can re-think about the min supported version, as there is no major distro any more which will use 1.20.
Rhel-7.4 was switched to gtk+-3.22 last summer.

monsta:
Sure, for 1.20 we can safely migrate to GTK+ 3.22 (unless some of supported distros have 3.20). Also bump minimum GLib version to the one that's minimum for GTK+ 3.22 itself.

monsta:
@lukefromdc: BTW, it's possible to rebase+merge via the merge button. Just push on the down arrow to the right of it, and choose the needed option. It won't create an additional merge commit.

github-merge-options

raveit65:
Why not using command line?

monsta:
I use the command line for PRs which don't need to be rebased or squashed before merging.
The button does two things: closes PR as "Merged" (you won't have to close it) and remembers the link to PR (you will see it if you look at the commit).

raveit65:
> The button does two things: closes PR as "Merged" (you won't have to close it) and remembers the link to PR (you will see it if you look at the commit).

Where is the link written? In commit body or header?
If it's written in commit header i don't like this.....

monsta:
I think it's somewhere in Github's database, because it can't be seen when you look at `git log`.

Issue #431, opened 2016-04-30
mate-panel crashes at startup


bol-van:
Ubuntu 16.04 MATE.  mate-panel crashes once after logon if compiz window manager is used instead of marco. Then it restarts. Tested on physical hardware (nvidia) and vmware.
crash is very stable. same crash was reported many times on the web

segfault at 6 ip 00007f155df0f7e5 sp 00007ffc1328bd18 error 4 in libgobject-2.0.so.0.4800.0[7f155dedb000+52000]
##

monsta:
Is there a report on Launchpad about it? I'm asking because it might contain a backtrace collected via apport (which would be helpful to us).

bol-van:
Since this bug is very easy to reproduce I thought it would not be necessary.
But if you need it I post it.

[_usr_bin_mate-panel.1000.crash.zip](https://github.com/mate-desktop/mate-panel/files/243544/_usr_bin_mate-panel.1000.crash.zip)

monsta:
Well, I don't see a backtrace in that archive. I can't reproduce it either in my VirtualBox.

bol-van:
This archive is standard apport report. It can be unpacked by "apport-unpack" tools. Unpacked directory has CoreDump file.  It can be opened by "gdb /usr/bin/mate-panel <path>/CoreDump". Then "bt" command in gdb.
Here is your backtrace :

```
#0  0x00007f0300a277e5 in g_type_check_instance_is_fundamentally_a () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#1  0x00007f0300a07c17 in g_object_unref () from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#2  0x00000000004685bd in ?? ()
#3  0x00007f0300cbc447 in g_simple_async_result_complete () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#4  0x00000000004697d0 in ?? ()
#5  0x00007f0300ccdb33 in ?? () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#6  0x00007f0300cce1de in ?? () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#7  0x00007f0300d1ca2f in ?? () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#8  0x00007f0300ccdb33 in ?? () from /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
[...]

Issue #214, opened 2014-06-21
container_child_background_set: Process /usr/bin/mate-panel was killed by signal 11 (SIGSEGV) 


raveit65:
orginal report at fedora bugzilla.
https://bugzilla.redhat.com/show_bug.cgi?id=1023604
```
 satellit@bendbroadband.com 2013-10-25 18:20:51 EDT

Description of problem:
DVD install of TC5 Beta x86_64 to HD of Mate with updates block unchecked in software sources
doing test and added 2nd user and logged in then got error.

Version-Release number of selected component:
mate-panel-1.6.1-4.fc20

Additional info:
reporter:       libreport-2.1.8
backtrace_rating: 4
cmdline:        mate-panel
crash_function: container_child_background_set
executable:     /usr/bin/mate-panel
kernel:         3.11.5-302.fc20.x86_64
runlevel:       unknown
type:           CCpp
uid:            1000

Truncated backtrace:
Thread no. 1 (8 frames)
 #1 container_child_background_set at mate-panel-applet-frame-dbus.c:240
 #2 g_simple_async_result_complete at gsimpleasyncresult.c:777
 #3 set_applet_property_cb at mate-panel-applet-container.c:556
 #4 g_simple_async_result_complete at gsimpleasyncresult.c:777
 #5 g_dbus_connection_call_done at gdbusconnection.c:5484
 #6 g_simple_async_result_complete at gsimpleasyncresult.c:777
 #7 complete_in_idle_cb at gsimpleasyncresult.c:789
 #12 gtk_main at gtkmain.c:1257
```
Note, there are about 10 users who are affected by this issue, latetest from 2014-06-20.
So i decided to forward it to github.
If you've questions to users pls asked at fedora bugzilla.
More Infos and descriptions you'll also find there.

monsta:
@NiceandGently: here's the missing link to the PR - https://github.com/mate-desktop/mate-panel/pull/261 :)

NiceandGently:
@stefano-k 
Can you please review this for 1.10.
I use the patch for several month in fedora and downstream report is closed since 2014-12-01.

monsta:
@stefano-k: and the commit that fixes it (https://github.com/monnerat/mate-panel/commit/d02ce625ff7643df609f8911066c6a3304223d33) is very small, so maybe it should be applied to 1.8 branch as well.

monsta:
@monnerat: check #701, it should fix the crash.

[...]

PR #701, opened 2018-01-04
rework the process of cancelling the background change operation

monsta:
don't mess with additional references to a GCancellable, instead
keep just a const pointer to the internal operation data and use
it as a key when it's needed to cancel the operation.

fixes https://github.com/mate-desktop/mate-panel/issues/214
fixes https://github.com/mate-desktop/mate-panel/issues/431

### DIFF
--- a/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
+++ b/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
@@ -235,8 +235,7 @@ container_child_background_set (GObject      *source_object,
 
 	mate_panel_applet_container_child_set_finish (container, res, NULL);
 
-	if (frame->priv->bg_operation)
-		frame->priv->bg_operation = NULL;
+	frame->priv->bg_operation = NULL;
 }
 
 static void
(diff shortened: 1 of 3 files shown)


monsta:
The crash with the backtrace from #214 and #431 shouldn't happen anymore, but as it's not reliably reproducible (I couldn't reproduce it at all for example), please just check that the panel still works fine during normal use. E.g. panel launch on session start, restart after killall, changing the background via properties dialog, etc.

@raveit65: this fixes rhbz [1023604](https://bugzilla.redhat.com/show_bug.cgi?id=1023604) and [1238820](https://bugzilla.redhat.com/show_bug.cgi?id=1238820) (both have the same backtrace), so you can make some scratch builds for people to test. 🙂 

I'll probably make some builds for Ubuntu users to test as well.

mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
```diff
@@ -235,33 +235,32 @@ container_child_background_set (GObject      *source_object,
 
 	mate_panel_applet_container_child_set_finish (container, res, NULL);
 
-	if (frame->priv->bg_cancellable)
-		g_object_unref (frame->priv->bg_cancellable);
-
-	frame->priv->bg_cancellable = NULL;
+	if (frame->priv->bg_operation)
+		frame->priv->bg_operation = NULL;
```
joakim-tjernlund: This seems a bit off, should that be if(frame->priv->bg_cancellable) ?
Same thing a few lines down.

joakim-tjernlund: Na, it should not. But the if should be removed I guess

monsta: Yeah, `bg_cancellable` is gone. Indeed, I can just assign NULL to `bg_operation`, should be the same.

lukefromdc:
Worked fine again on retest

Issue #687, opened 2017-12-05
FIX Workspace Switcher: increase the height of workspace list or allow resizing the window


allanlaal:
currently the window only fits 2 workspaces and it is not possible to resize it

increase the list height or make it have automatic height based on the amount of workspaces. so it would start scrolling when its reaching the height of the screen

![screenshot at 2016-11-12 07 53 34](https://cloud.githubusercontent.com/assets/740826/20236009/6a094a86-a8ad-11e6-90be-ddb254ea070e.png)

raveit65:
Do you really think someone will work on it if you don't file out the template?

allanlaal:
I'm using MATE 1.12.1 on a Ubuntu Mate 16.04 and I am unable to test this in a newer release at this time, has this already been fixed?

muesli4:
https://github.com/mate-desktop/mate-panel/pull/702

info-cppsp:
@monsta 
If you would close this, as it seems to be fixed by the PR. Thanks.

monsta:
Yes, this is fixed by now.

allanlaal:
resize is fixed, but workspace list default height is still too small

PR #702, opened 2018-01-12
Update workspace switcher properties dialog

muesli4:
Removed deprecated ui elements to make it fully compatible with GTK+ 3
and additionally made the dialog resizeable.


muesli4:
Fixes https://github.com/mate-desktop/mate-panel/issues/687.

Also makes some minor cosmetic changes (but nothing big, only for consistency). (Can be tested by replacing just the ui file.)

lukefromdc:
workspace-switcher_ui

muesli4:
Thanks for testing it. I fixed it (by adding the missing group). (I amended to the last commit.)

sc0w:
beforelaterpr

without the PR (left) / with the PR (right)

and now resizable, looks ok for me

PR #703, opened 2018-01-14
get focus when displaying "applet broken" dialog

sc0w:
how to test:

for example: `killall mate-sensors-applet` and see the focus without/with the PR

applet_broken

### DIFF
--- a/mate-panel/panel-applet-frame.c
+++ b/mate-panel/panel-applet-frame.c
@@ -827,6 +827,10 @@ _mate_panel_applet_frame_applet_broken (MatePanelAppletFrame *frame)
 	gtk_window_set_title (GTK_WINDOW (dialog), _("Error"));
 
 	gtk_widget_show (dialog);
+
+	gtk_window_present_with_time (GTK_WINDOW (dialog),
+				      gdk_x11_get_server_time (gtk_widget_get_window (GTK_WIDGET (dialog))));
+
 	g_free (dialog_txt);
 }
 


mate-panel/panel-applet-frame.c
```diff
@@ -827,6 +827,10 @@ _mate_panel_applet_frame_applet_broken (MatePanelAppletFrame *frame)
 	gtk_window_set_title (GTK_WINDOW (dialog), _("Error"));
 
 	gtk_widget_show (dialog);
```
muktupavels: I think that `gtk_widget_show ()` is not needed if you use gtk_window_present...

sc0w: @muktupavels no, you are wrong

I have tested without gtk_widget_show and the dialog don't get the focus

alexarnaud:
Le 16/01/2018 à 12:40, raveit65 a écrit :
> As it is bugfix, small fix and no translation strings are involved, it 
> isn't affected by freeze for 1.20 and can be merged, imo

Of course, this fix will be an accessibility improvement.

Best regards.
-- 
Alex ARNAUD

mate-panel/panel-applet-frame.c
monsta: Hmm, I didn't think about it when fixing similar issue in https://github.com/mate-desktop/mate-polkit/commit/ed37b52fbfbce5d57d0549fac1a9dca2502f3372 and https://github.com/mate-desktop/mate-session-manager/commit/9e8d8ad24ad1025634b2558bc572937e55c64a2c. Didn't test without `gtk_widget_show` too. I guess it does no harm.

PR #705, opened 2018-01-17
Fix rendering of radio buttons in status notifier

lukefromdc:
Ensure only checked radio button is shown as checked, not all the radio buttons. Issue is apparently caused by a bug in gtk_radio_menu_item_new and also effects gnome-panel and Cinnamon. We can fix it here so radio buttons are actually usable in the status notifier.

### DIFF
--- a/applets/notification_area/status-notifier/sn-dbus-menu-item.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu-item.c
@@ -180,7 +180,11 @@ sn_dbus_menu_item_new (GVariant *props)
         }
       else if (g_strcmp0 (item->toggle_type, "radio") == 0)
         {
-          item->item = gtk_radio_menu_item_new (NULL);
+          item->item = gtk_check_menu_item_new ();
+          gtk_check_menu_item_set_draw_as_radio (GTK_CHECK_MENU_ITEM(item->item),TRUE);
+          AtkObject *atk_obj;
+          atk_obj = gtk_widget_get_accessible (item->item);
+          atk_object_set_role (atk_obj,ATK_ROLE_RADIO_MENU_ITEM);
         }
       else
         {


lukefromdc:
This is the current problem with radio buttons in the status notifier, caused by what appears to be a bug in GTK.

before_pr_1-17-2018

This is after the fix which avoids the buggy code:

after_pr_1-17-2018

muktupavels:
At least mention status notifier item that can be used to reproduce problem?

lukefromdc:
I was using fusion-icon, though the same bug will also appear in Ubuntu's indicator-cpufreq or anything else using radio buttons.

lukefromdc:
before_pr_1-17-2018
after_pr_1-17-2018

muktupavels:
How this is bug in gtk+? And more importantly - why not fix gtk+ if that is gtk+ bug?

But I have bad news for you, it is not bug in gtk+:

> A radio menu item is a check menu item that belongs to a group. At each instant exactly one of the radio menu items from a group is selected.

Your proposed fix might look like fix and to speak truth in this case probably will also work good enough, but it is not correct! Correct fix would be to get group of previous item with `gtk_radio_menu_item_get_group` and pass it to next radio item.

lukefromdc:
The string "group" is not found anywhere in all of sn-dbus-menu-item-c and I suppose following 
gtk_radio_menu_item_get_group
with
gtk_radio_menu_item_set_group
for the same item would change nothing.  Presumably the original code worked before something in GTK changed, or it would not have been inherited from gnome-panel's status notifier in the first place.  

It is beyond me to rewrite this entire file to deal with this

lukefromdc:
More details: The only instance of "radio" in that entire file is where I changed it. When I temporarily replaced  gtk_radio_menu_item_new with  gtk_check_menu_item_new the result worked, so adding 
 gtk_check_menu_item_set_draw_as_radio gave fully working results.

I didn't get an unwanted ability to check more than one item at once, and was able to avoid any build warnings from the new code. Whether this is a correct fix or a kludge, the fact it is works and right now radio items in the status notifier do NOT work, they show all options checked at once.

Complicating matters is that the status notifier applet itself has to draw whatever menu items are called for by the application creating an indicator, rather than being coded into the applet itself like everywhere else.

lukefromdc:
One more point: We at MATE cannot fix a bug in GTK and have to work around whatever changes come from GNOME in GTK3.  Thus, if something is a bug in GTK3 we can either wait for a fix that may or may not ever come, or if a valid workaround exists (and won't create a new problem if and when FGTK gets a fix) we can use it.  

If the issue in GTK is not considered a bug, but rather a difference in how the code is meant to be used (say, between versions) we definately cannot expect GNOME to revert the change or add a workaround to support older code.

muktupavels:
Please stop blaming gtk+ if you can not show version / commit that introduced regression. It is not GTK+ bug if you can't read documentation and/or understand how radio menu items are supposed to work.

Again:

> A radio menu item is a check menu item that belongs to a group. At each instant exactly one of the radio menu items from a group is selected.

`gtk_radio_menu_item_new` is used with `NULL` effectively creating each radio item as separate group. And according to documentation each group has exactly one item **selected**. So what you show in screenshots are correct behaviour from gtk+ side. 

That code was written by me, apparently I have written some very buggy code. Quick look in d-feet shows that there is no extra info for radio items so code probably should check if previous menu item of menu also is radio item, get group `GSList` with `gtk_radio_menu_item_get_group` and pass this group to `gtk_radio_menu_item_new`.

lukefromdc:
It would seem to me that the group cannot be known until the applet gets the information from the application creating an indicator.  The test of whether this is a valid fix or not would be to find something creating a menu with two or more groups of radio items. I don't think this would render correctly with my code (only ONE would be shown), nor with checking each previous item to see if it too was a radio item. That would work if there was a check item between the radio items, otherwise not.

The real problem thus appears to be that supporting radio items at all in a menu whose layout cannot be known until an external application calls the application creating it is by no means easy to do, in this case I suspect it might require a total rewrite. 

For quite a while this has been broken in gnome-panel, broken in Cinnamon, and with the addition of sni support broken in MATE as well. In all cases the radio buttons will actually function if the user can just figure out which ones to fix.

My fix or kludge will handle the common case of an indicator with one or more menus, no one menu containing multiple separate groups of radio items. I propose we use the partial fix and add a "FIXME" notation that more work is needed against the more complex cases. I will add that note to the PR

muktupavels:
I was thinking about something like this:
https://pastebin.com/0RZTQDjj

But as I said in this case your solution should work. And in fact original libdbusmenu does same thing. Still think that radio items should be properly grouped...

lukefromdc:
I just tested your patch, had to also make a change in  sn-dbus-menu-item.h .

Unfortunately it acted rather badly. In indicator-cupfreq the radio items are effectively one group, as only one can be selected at a time. What I got was one of the top ones selected while two of the bottom ones flickered on and off, and enough freezes that I had to kill the indicator. Even worse with fusion-icon, where only one radio item was shown checked, but trying to change it nearly froze my entire session, had to kill fusion-icon from a VT to regain control.

lukefromdc:
I rebooted to clear an indicator-applet problem in RAM and retested,this time making sure to start the panel first, then the two test applets using radio buttons. Nothing went unresponsive this time, and no errors showed up in xsession-errors. Fusion-icon worked fine, but I  still had trouble with indicator-cpufreq flickering between two radio buttons among the menuitems below the separator. 

In indicator-cpufreq, there are two sets of radio items and a separator in between, so in truth they are one group. In my test the top "group" came up with first radio item as checked default (should not be checked unless set to top speed in previous run of same applet or by another applet. Could be set and would hold checked item but was not actually appled(this however may be a problem in the indicator applet itself with recent kernels). The bottom "group" flickered between whatever it was set to and the default setting.

muktupavels:
I did not say that patch is ready. Probably should block `activate` signal / `activate_cb` callback...

Also how using `check` menu items instead of `radio` affects accessibility?

alexarnaud:
Le 18/01/2018 à 13:58, Alberts Muktupāvels a écrit :
> Also how using |check| menu items instead of |radio| affects accessibility?

It shouldn't affect accessibility, radio buttons and checkboxes are 
correctly presented by ATK/AT-SPI if the code respect the GTK requirements.
I don't know much more because I'm not familiar with GTK coding.

Best regards.
-- 
Alex ARNAUD

muktupavels:
From ATK side, is radio menu item same as check menu item? How this can be tested?

alexarnaud:
Le 18/01/2018 à 16:01, Alberts Muktupāvels a écrit :
>  From ATK side, is radio menu item same as check menu item? How this can 
> be tested?

It can be tested with the Orca screen reader. If you need help to make 
it working on your system, we could discuss on IRC.

Best regards.
-- 
Alex ARNAUD

raveit65:
This problem with fusion-icon was the last bug which forced me to disable SNI support for Fedora. Fusion-icon is part of the Mate-Compiz livecd.
Thanks for fixing that.
But honestly i don't know which fix is better :-), Up to you to decide this.
I will test @muktupavels fix too.
As i am on fedora i can't test ubuntus indicator apps. ;-)

alexarnaud:
Maybe @flexiondotorg could test it. It will be great if SNI support could be enabled by default on Ubuntu. 

Best regards.

raveit65:
@muktupavels 
Your solution works too and i was warned.
After reinstall of mate-panel and restarting the panel i had a hard repeating flashing desktop, i was forced to kill X-session.
After session restart the flashing desktop was gone and menu of fusion-icon was fine.
Honestly, i prefer @lukefromdc solution for the moment ;-)

muktupavels:
> In indicator-cpufreq, there are two sets of radio items and a separator in between, so in truth they are one group.

Is not that bug in indicator-cpufreq? I would say that it should not use one group for all menu items. There should be one group for frequencies and second group for governors! No idea if that changes anything related to dbus menu...

Anyway that indicator seems no longer maintained:
https://code.launchpad.net/indicator-cpufreq

@raveit65 I already told that patch was not ready, I posted that just to show idea. The problem is that activate signal is emitted. That could be fixed with something like this:
https://pastebin.com/cuQzDtJ0

Proposed fix will cause Orca to tell that it is check menu item not radio menu (that can be changed).

Radio menu grouping is used to make sure that previous item is unchecked when new item is selected, but this does not matter in this case because menu is managed by indicator and not by applet...

lukefromdc:
Albert's latest change mostly works, but the (apparently broken anway) indicator-cpufreq applet  becomes nonresponsive when one of the radios is clicked in the top half of the menu and then one from the bottom half or vice versa. This  causes the panel (and only the panel) to lock up until it is restarted.  Indicator-cpufreq needs these to be one group, because all the speeds in the top group select the "userspace" governor, which is exclusive of other governors like "performance" or "ondemand."

This comes from the fact that we are not getting the grouping from the indicator itself, I do not know if the spec even supports doing that. Using anything not a radio (like that separator) to make the distinction will sometimes work, and sometimes not. Indicator-cpufreq is broken anyway with current kernels, as I can make the menus work right but it no longer connects sucessfully to the kernel governors, just as our own cpufreq-applet now has to be built with libcpupower to work.

Fusion-icon works fine with it though, as each menu is entirely separate and there 

So the question becomes this: which of these approaches breaks the fewest indicators, and if using Albert's approach can we ensure that no broken indicator can lock up the panel or the tray?  If Albert's approach breaks fewer indicators, we should debug it and use it. If mine (all radios in one indicator menu always treated as one group) breaks fewer indicators than that's the one we should use

lukefromdc:
Should I go ahead and merge the PR as I wrote it? 

A note concerning accessability: Accerciser cannot be used to look at the menu items, because clicking in its window to expand the view for the menu's window closes the menu.  Last time I tried to install Orca it installed OK but I was unable to get the sound to actually work on my setup.

Whether gtk_check_menu_item_set_draw_as_radio also has it spoken as a radio item or not should not kill usability for someone relying on sound. Either way they would select the item they want to turn on, and all others would turn off. Presumably that's how they would be spoken as well. As things stand in master with each radio button in its own group, all of them would presumably be spoken as selected, just as they are rendered. Thus they are now surely broken for all users, and this PR should make them usable for all users. The only question is whether telling GTK to draw a check item as a radio also makes Orca speak it as a radio item, and if it does not is that a big enough usabilty issue to require releasing 1.20 with this not fixed at all for any users.

lukefromdc:
I was not able to get Orca to actually speak, as speech-dispatcher won't speak on my system if tested directly. All other audio badly distorted too if its running, possibly a pulseaudio issue? None the less, it turns out Orca can send debug information to a text file. Very verbose but I got this text showing the radio buttons read as check buttons, but with accurate information as to whether or not they are checked. Test was with fusion-icon, and Marco previously set as the window manager. The relevent "speech output" lines are these:

SPEECH OUTPUT: 'Select Window Manager menu.'{'established': False}
SPEECH OUTPUT: ' Marco check menu item checked.'{'established': False}
SPEECH OUTPUT: ' Compiz check menu item not checked.'{'established': False}

Keep in mind, the indicator itself seems to ensure only one is ever checked, same as a properly grouped radio button

lukefromdc:
With Albert's code I get this:

SPEECH OUTPUT: 'Fusion Icon push button'{'established': False}
SPEECH OUTPUT: 'Select Window Manager menu.'{'established': False}
SPEECH OUTPUT: ' Marco not selected radio menu item'{'established': False}
SPEECH OUTPUT: ' Compiz selected radio menu item'{'established': False}

This tells me I should simply have all radio items join a group named "group" to get the results I get in terms of robustness with my code but the accessability results from Albert's code

lukefromdc:
Unfortunately, I have found that it is almost impossible to set up a group for radio buttons that forces all of them in each menu into one group, because it seems like no matter what I do I either get radios from one indicator interacting with ANOTHER indicator (global variable in sn-dbus-menu.c) , problems with the top menu in indicator-cpufreq and rare interaction between indicator-cpufreq and fusion-icon(global variable in sn-dbus-menu-item.c) , or every radio button in its own group and checked(everything else except using  gtk_check_menu_item_set_draw_as_radio).

Where Albert's patch seems to go wrong in indicator-cpufreq is that trying to forcibly set two options at once (from the two separate groups) is freezing the indicator and (with the in-process tray) the entire panel.

Thus the only thing that works right now is what I first came up with, and accepting that Orca will call them "check menu item" rather than "radio menu item" while correctly identifying that one and only one is checked.

lukefromdc:
I just managed to get Orca to call the radio buttons drawn from checkbuttons radio buttons with the last commit, got these results in an Orca debug file:

With default accessability:

00:03:28 - SPEECH OUTPUT: ' Metacity check menu item not checked.'{'established': False}
00:03:28 - SPEECH DISPATCHER: Speaking ' Metacity check menu item not checked.' Orca rate 5.0, pitch 5.0, volume 5.0, language en; SD rate 2, pitch 10, volume 100, language en
TOTAL PROCESSING TIME: 0.0258


With atk_object_set_role:

00:05:44 - SPEECH OUTPUT: ' Marco.'{'established': False}
00:05:44 - SPEECH DISPATCHER: Speaking ' Marco.' Orca rate 5.0, pitch 5.0, volume 5.0, language en; SD rate 2, pitch 10, volume 100, language en
00:05:44 - SPEECH OUTPUT: 'not selected radio button'{'established': False}
00:05:44 - SPEECH DISPATCHER: Speaking 'not selected radio button' Orca rate 5.0, pitch 5.0, volume 5.0, language en; SD rate 2, pitch 10, volume 100, language en
TOTAL PROCESSING TIME: 0.0311

While I still cannot get sound due to something in speech-dispatcher, to the user this should sound like "Marco not selected radio button" or  "Marco, not selected radio button" , not sure which

muktupavels:
No idea why it could freeze, my quick test at that time showed that it worked, but my indicator has only two items - `Performance` and `Powersave`... And `fusion-icon` does not work so my testing is / was limited.

Just retested, apparently activate signal is also emitted for radio menu item that was unselected as part of selecting new item. I can imagine that it can cause problem....

Also just googled indicator-cpufreq images to see how that indicator menu should look. And now I see that I was 100% wrong thinking there should be two groups. So unless required info is sent over dbus there is no way to properly group radio items. And in the end orca does not care if items are grouped or not. 

So trying to do "right" thing are causing more problems then needed. Menu handling is done server side, but our applet is just client and all we need to do is to make sure that we display items based on info we have.

I would remove FIXME commit, because in the end there is nothing to fix. Otherwise your commits does not match style in that file, no idea if you care about code style...

applets/notification_area/status-notifier/sn-dbus-menu-item.c
```diff
@@ -186,6 +186,9 @@ sn_dbus_menu_item_new (GVariant *props)
           /*Note that the layout of the menu cannot be known in advance here                         */
           item->item = gtk_check_menu_item_new ();
           gtk_check_menu_item_set_draw_as_radio (GTK_CHECK_MENU_ITEM(item->item),TRUE);
+          AtkObject *atk_obj;
+          atk_obj = gtk_widget_get_accessible (item->item);
+          atk_object_set_role (atk_obj,ATK_ROLE_RADIO_BUTTON);
```
muktupavels: it is not button, but menu item... `ATK_ROLE_RADIO_MENU_ITEM`

lukefromdc:
Nice catch, now we get:

SPEECH OUTPUT: ' IceWM not selected radio menu item'{'established': False}

SPEECH OUTPUT: ' Marco not selected radio menu item'{'established': False}

SPEECH OUTPUT: ' Compiz selected radio menu item'{'established': False}

lukefromdc:
Are there any further issues for merging this or should I go ahead now that Albert's change to the ATK role spoken has been applied?

muktupavels:
Remove FIXME and squash remaining commits into one...

lukefromdc:
OK, will do this and then merge as a single commit

PR #708, opened 2018-01-22
bump minimum required libwnck version to 3.4.6

monsta:
It could be bumped to something more recent, but 3.4.6 is enough to get rid of `WNCK_CHECK_VERSION` blocks.


lukefromdc:
Rebased against latest change in Master, waiting for more reviews

Issue #686, opened 2017-11-30
 SEGFAULT when StatusNotifier is Used.


oblongau:
#### Expected behaviour
No segfault. Notification area to be updated, presumably to show calibre icon.

#### Actual behaviour

Popup saying, "Notification Area" has quit unexpectedly. If you reload a panel object, it will automatically be added back to the panel."
The network connection/speaker etc. icons are then not showing. If I click "Reload" then they reappear. There is no indication there that calibre is running, but it is actually properly running in the background.

dmesg has this entry:
`notification-ar[17348]: segfault at 22 ip 00007f90aed14268 sp 00007ffeda1f5bc0 error 6 in libglib-2.0.so.0.4800.2[7f90aecf4000+10f000]`

/var/log/syslog has:

`
Dec  1 07:27:09 bob-mint org.mate.panel.applet.NotificationAreaAppletFactory[1712]: (notification-area-applet:17348): status-notifier-watcher-WARNING **: Status Notifier Item with bus name 'org.kde.StatusNotifierItem-18318-1' and object path '/StatusNotifierItem' is already registered

Dec  1 07:27:09 bob-mint org.mate.panel.applet.NotificationAreaAppletFactory[1712]: (notification-area-applet:17348): Gtk-WARNING **: gtk_widget_size_allocate(): attempt to allocate widget with width -1 and height 15

Dec  1 07:27:09 bob-mint org.mate.panel.applet.NotificationAreaAppletFactory[1712]: (notification-area-applet:17348): GLib-GIO-WARNING **: Received property Version with type i does not match expected type u in the expected interface

Dec  1 07:27:09 bob-mint org.mate.panel.applet.NotificationAreaAppletFactory[1712]: (notification-area-applet:17348): GLib-CRITICAL **: the GVariant format string '(i@a{sv}@av)' has a type of '(ia{sv}av)' but the given value has a type of '(ia{sv}a(ia{sv}av))'

Dec  1 07:27:09 bob-mint org.mate.panel.applet.NotificationAreaAppletFactory[1712]: (notification-area-applet:17348): GLib-CRITICAL **: g_variant_get: assertion 'valid_format_string (format_string, TRUE, value)' failed

Dec  1 07:27:09 bob-mint kernel: [44554.934631] notification-ar[17348]: segfault at 22 ip 00007f90aed14268 sp 00007ffeda1f5bc0 error 6 in libglib-2.0.so.0.4800.2[7f90aecf4000+10f000]

[...]

PR #709, opened 2018-01-22
status-notifier: fix crashes on wrong GVariant format

monsta:
Should address both crashes mentioned in https://github.com/mate-desktop/mate-panel/issues/686

### DIFF
--- a/applets/notification_area/status-notifier/sn-dbus-menu.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu.c
@@ -126,6 +126,15 @@ layout_parse (SnDBusMenu *menu,
   GVariantIter iter;
   GVariant *child;
 
+  if (!g_variant_is_of_type (layout, G_VARIANT_TYPE ("(ia{sv}av)")))
+    {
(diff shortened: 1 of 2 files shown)


monsta:
Yes, it's normal, it doesn't show menu because status-notifier won't accept the invalid format for it.

lukefromdc:
I didn't expect an invalid menu to function, this PR did exactly what I expected from it, seems ready to go

lukefromdc:
https://github.com/mate-desktop/mate-panel/issues?utf8=%E2%9C%93&q=status-notifier

Now shows all the status-notifier issues as fixed/closed
and https://github.com/mate-desktop/mate-panel/issues?utf8=%E2%9C%93&q=sni
shows all the issue for SNI closed except the references to it in a Wayland support suggestion issue and an issue with a conventional tray icon (guake) that does not offer an SNI mode.

Thus, I think we have now squashed all of the reported bugs with sni-support, so distros should be able to enable it by default. First appeared in 1.18, debugged for 1.20

PR #710, opened 2018-01-22
Add support for HiDPI displays

### LINKED ISSUE #668: System tray drawing issues on HiDPI display
#### Expected behaviour
- system tray contents look the same as on LoDPI displays (except for scaling)
 
#### Actual behaviour
- 'half' copies of Icons are often appearing on the left of the tray
- language indicator text is not scaled up correctly 

Image:
tray

Run MATE with system tray active in panel. Set resolution in Appearance->Fonts in  to 150 dpi. Set panel height to 32 pixels.

#### MATE general version
1.18

#### Package version
mate-panel-1.18.4-1-x86_64.pkg.tar.xz

#### Linux Distribution
Arch

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/50153116-system-tray-drawing-issues-on-hidpi-display?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

vkareh:
When setting `GDK_SCALE=2` and `GDK_DPI_SCALE=0.5` as env variables, mate-panel will scale properly now.

### Known issue:
When changing the size of the panel (Right-click > Properties), the pixel value shown in the spinner is still the unscaled one, but it gets scaled up in the code. This is due to how the panel would scale if left to its own devices: it allocates double space for all internal GtkWidget componets, making everything mis-aligned, and a lot of things end up off-screen.

To solve this, instead of correcting each individual geometric component (of which there are a lot), we down-scale the values of the screen/monitor geometry so that the panel behaves as though it's dealing with a regular display. Everything ends up aligning properly this way.

This *may* cause issues with multiple monitors, which should be a very easy fix, but I don't currently have the capabilities of testing this.

----
This is part of multiple pull requests:
* mate-desktop/mate-settings-daemon/pull/208
* mate-desktop/mate-control-center/pull/325
* mate-desktop/mate-panel/pull/710
* ubuntu-mate/mate-hud/pull/18
* ubuntu-mate/mate-tweak/pull/37

Fixes #668

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,3 +1,30 @@
+### mate-panel 1.19.4
+
+  * Translations update
+  * WidthOfScreen and HeightOfScreen implementation
+  * panel-toplevel: Remove pointer boundary check in panel_toplevel_queue_auto_hide()
+  * remove ancient gnome-panel leftovers
+  * Keep the right icons on bottom still displaying right when the resolution change larger
+  * wncklet: disconnect signals by data on destroy
+  * clock: disconnect one more signal handler on destroy
+  * panel-applet-frame-dbus-c: comment out newly unused variable
+  * Clock: don't check variables that can never be < 0 for < 0
+  * panel-applet-frame-dbus.c: comment out never used code block
+  * panel-toplevel: remove duplicate break statement
+  * status-notifier: remove duplicate check found by cppcheck
+  * status-notifier: hide mate-polkit icon if passive
+  * Apply position fix on popup to GTK 3.20 and earlier
+  * GTK 3.22 status notifier: reposition menu when size changes
+  * GTK 3.22/status-notifer Fix menu position if size changed since last shown
+  * Stop panel collapse left on adding certain out-of-process applets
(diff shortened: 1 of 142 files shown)


mrnhmath:
Wow! I wonder if this is too late for 1.20? It's a huge feature.

raveit65:
I am just testing all  PRs with one none-hdpi monitor (1920x1200).
This PR is really huge...

lukefromdc:
We can fix bugs in hidpi cases post-release if we have to, so long as users for whom mate worked before get no regressions.

lukefromdc:
Squashed and merged with 
https://github.com/mate-desktop/mate-panel/commit/7f99492c269f541c81e4f36742ecc3f0b4ebac04
since the m-s-d and m-c-c portions are now merged

PR #713, opened 2018-01-26
Down-scale out-of-process applets on HiDPI displays

vkareh:
Out-of-process applets end up scaled up doubly. This divides the preferred width/height by the scale factor to ensure they are back at the correct preferred size.

For this change I had to move the GTK_VERSION guards on the `get_preferred_width/height` methods. The methods are now exposed, but the focus style class padding is still guarded.

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1020,29 +1020,44 @@ mate_panel_applet_popup_menu (GtkWidget *widget)
 	return TRUE;
 }
 
-#if !GTK_CHECK_VERSION (3, 18, 0)
 static void
 mate_panel_applet_get_preferred_width (GtkWidget *widget,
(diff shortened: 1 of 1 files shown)


lukefromdc:
Others on the team should have a look at this too, of course

lukefromdc:
The GTK version guards will soon become unnecessary, as MATE 1.20 is going to be GTK3.22-only. We will be able to drop all the old code blocks at that point, so no need to test with anything but GTK 3.22

flexiondotorg:
Tested on my XPS 15 and this resolves the issue with out-process applet allocation too much space on the panel. I see no regressions when using a LoDPI display.

monsta:
@vkareh: I hope the code is still correct after https://github.com/mate-desktop/mate-panel/pull/723 (where all `GTK_CHECK_VERSION` blocks are removed).

PR #718, opened 2018-01-30
README - Add dependency list

info-cppsp:


lukefromdc:
The package names given are specific to Ubuntu MATE: can a more general description also be included for those not on Debian based distros, possibly by giving the general package (maybe source) name, then the specific name for Ubuntu/Debian/Mint? We have to allow for all other distros here

info-cppsp:
I understand what you mean, but to make this list I had to try and error for an hour.
There are many other distros...
I only know that you need at least this much.

I have it on the new mp TODO to update the readme more fully, maybe then I'll go and grab some distros and try them.
Or if anyone else has some free time.

I couldn't find the new mint 19 alpha. That is what I wanted to use actually.

Still it is at least a start.

lukefromdc:
Waiting to see what other team members think. We could just list Ubuntu-Mate as an example, noting that the source packages stay the same in all distros

info-cppsp:
OK. It was not my intention to single out Ubuntu... ;-)

raveit65:
Why not using build deps from here , they are tested.
debian:
https://github.com/mate-desktop/debian-packages/tree/master/mate-panel/debian
For the other side of linux world (rpm based distros) feel free to look at my fedora packages under Build requires:
http://pkgs.fedoraproject.org/cgit/rpms/mate-panel.git/tree/mate-panel.spec#n42

info-cppsp:
Build dependencies
==================

On Ubuntu MATE 18.04 (daily 2018-01-30)
build-essential
mate-common
libglib2.0-dev
yelp-tools
gobject-introspection
libgtk-3-dev
libmate-desktop-dev
libmate-menu-dev
libdbus-glib-1-dev
libwnck-3-dev
librsvg2-dev
libmateweather-dev

These are the packages I had to install to compile mate-panel.
Others may be needed.

For Debian, see
https://github.com/mate-desktop/debian-packages/blob/master/mate-panel/debian/control#L10
For Fedora, see
http://pkgs.fedoraproject.org/cgit/rpms/mate-panel.git/tree/mate-panel.spec#n42

-----
how about this?

raveit65:
Hmm, i don't know.
Do build dependencies from @flexiondotorg  ubuntu-mate-18.04 really differ to our debian packages?
And if you don't write package names for rpm based distros here (why?), because package names can be different to debian/ubuntu world, than please remove the line hint from link. 
```
#n42
```
This can be changed if i update a package.

Or maybe better don't add the fedora link or rpm dependencies.
I don't care very much and people can look at distro website directly.
And i am to lazy for updating the readme if something changed ;-)

info-cppsp:
edited.
the links are good, bc even if they are updated, the user will be able to find the deps.
anyway, this is just a start, I can't possibly know or list all distros and their versions + dep list.
that is why I have written, this is what's needed for ub, very likely most will be needed on other systems as well.

muktupavels:
That is why you don't list distro specific package names. Open configure.ac and you will see package names that are required or optional. At same time you will also see required versions at least for few packages.

And let developer "find" distro specific package names. On ubuntu it is typically lib[packge-name]-dev and might include also API version. I would assume that developers knows how things work on his distro.

info-cppsp:
that is nice and all, just not helpful.
After this at least the debian derivative distro users can copy and paste the dep list instead of wasting any time finding out what they'll need.

muktupavels:
Debian based distros can do this:
`sudo apt build-dep mate-panel`

raveit65:
Merging this as it doesn't hurt and helps ubuntu users.

PR #720, opened 2018-01-30
Fix project cleanup

info-cppsp:
Add new 'make mp-clean' command and add some files, that were left out before.


info-cppsp:
@raveit65 
Could you please review and merge this?

raveit65:
I would really like to see a discussion about if we should add something like this to our project repos.
Personal i never compile in a local git repo folder where i commit, i always use a copy for this.
@monsta @lukefromdc @flexiondotorg @sc0w  ???

info-cppsp:
I would like to note, that before the `make maintainer-clean` was used to achieve the same effect.
However according to the [autotools documentation](https://www.gnu.org/software/automake/manual/html_node/Clean.html) it is not recommended to use maintainer-clean to remove everything.
At least that is how I understand the 'However maintainer-clean should not delete anything that needs to exist in order to run ‘./configure && make’.' part...

So it is not new functionality. I just fixed it.

info-cppsp:
Just to be sure, please check if m4/d-type.m4 is needed or not.
I remove the whole m4 directory in the PR, but I realized reading a documentation that it may be there for a reason.

lukefromdc:
I always use copied repos, as it is a lot of hand deletion to cleanup after a build in the same directory. So much so that if I do that by mistake I normally trash the whole repo and restore from a known good tarball, then pull to the restored directory. I never build intentionally in my local git repo.

info-cppsp:
You never have to do that again. Simply use my command and get rid of the mess in a few seconds.
No more copying needed.

I find it so much faster, I'm never going back to copying.
I build in the local repo, bc GitKraken shows modified files and with like 2 clicks I have started a new PR. If I'm programming, I like to spend my time with coding, not with anything else.
Especially not with file and folder management, especially if I can automate / avoid it.

((Like when doing web dev, I installed mercurial on a dedicated server, just so that I could push updates, without having to mess with ftp. (Of course there are better tools for this...)))

So for me this is a must-have. So much so, that I am using a tmpmaster branch locally, with this commit already merged. ;-)

raveit65:
@lukefromdc @sc0w @monsta @flexiondotorg 
No idea if we really need this?
Should we close PR?

lukefromdc:
My vote is to close it. Adding new stuff to autotools now may be a lot of work for gain that won't be kept in any future Meson/Nijna conversion

monsta:
The part regarding `d-type.m4` could probably be useful, as a small cleanup.

raveit65:
Ok, i close PR and you can pick up what is useful.

PR #721, opened 2018-01-30
Run dialog accessibility

### LINKED ISSUE #676: run dialog focus on list
#### Expected behaviour
I expect easy way to go to list of appliactions like in xfce4-appfinder. In xfce4-appfinder when you type part of app name and press Enter you go to list of applications and can navigate throught it with arrows. More suitable way is to focus on the apps list by tab (and only then on other buttons, flag etc.).

#### Actual behaviour
I see no easy way to go from focused execution line straight for applications list (if it shown). I need to press Tab 5 times for it.

#### MATE general version
1.18.0

#### Linux Distribution
Ubuntu Mate 16.04

info-cppsp:
fix #676

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -1881,6 +1881,35 @@ panel_run_dialog_setup_pixmap (PanelRunDialog *dialog,
 			  dialog);
 }
 
+/* this runs after entry_event() */
+static gboolean
+key_press_event (GtkWidget    *run_dialog,
+				GdkEventKey    *event,
+				PanelRunDialog *dialog)
+{
+
+	/* allow only key presses */
+	if (event->type != GDK_KEY_PRESS)
+		return FALSE;
+
+	/* If the program list is enabled and open and the user presses the F6 key
+	 * the focus should jump between GtkComboBoxText and the program list  */
+	if (panel_profile_get_enable_program_list () && panel_profile_get_show_program_list () && event->keyval == GDK_KEY_F6) {
+
+		/* jump to the program list from anywhere */
+		if (!gtk_widget_has_focus (dialog->program_list)) {
+			gtk_widget_grab_focus (dialog->program_list);
+		/* on the program list, jump to the entry box */
+		} else {
+			gtk_widget_grab_focus (dialog->combobox);
+		}
+
+		return TRUE;
+	}
+
+	return FALSE;
+}
+
 static PanelRunDialog *
 panel_run_dialog_new (GdkScreen  *screen,
 		      GtkBuilder *gui,
@@ -1908,6 +1937,10 @@ panel_run_dialog_new (GdkScreen  *screen,
 	panel_run_dialog_setup_file_button   (dialog, gui);
 	panel_run_dialog_setup_program_list  (dialog, gui);
 	panel_run_dialog_setup_list_expander (dialog, gui);
+
+    /* key press event signal for the whole dialog */
+	g_signal_connect (dialog->run_dialog, "key-press-event",
+			  G_CALLBACK (key_press_event), dialog);
 
 	panel_run_dialog_set_default_icon    (dialog, FALSE);
 


alexarnaud:
Hello @info-cppsp 

What's about the history ? If I press up arrow I switch to the previous history and if I press down arrow I go to the next one.

Another approach to avoid this should be to use the "F6" shortcut. It is used for that in LibreOffice and Thunderbird. It allows to go to the most important part of a window.

I'm a visual-impaired person and I primary use keyboard.

Best regards.

info-cppsp:
history? I am not sure what you mean.
But the button could be anything.
I just thought, the down button would be the most intuitive here.
Can't use tab as that is for autocomplete.

alexarnaud:
@info-cppsp on the run dialog there is a history of the command and application you've typed. You could cycle through it with up/down arrow as it is the case on the mate-terminal with bash.

info-cppsp:
Right.
This just shows, how much I'm using the run dialog.
However I feel, that not many people would think of using F6.
Anything better?

info-cppsp:
@bergentroll
what do you think?

bergentroll:
I believe it's better to use Tab key and just change widgets order in tab cycle.

info-cppsp:
Yes, that was my initial idea, but as I have written tab is used for autocomplete.
Or may be used if autocomplete is enabled.

alexarnaud:
Le 31/01/2018 à 10:58, info-cppsp a écrit :
> However I feel, that not many people would think of using F6.
> Anything better?

It's the convention on other software. For example on Thunderbird I 
could tab between the folder to a message but I'll pass over over 
elements. If I press F6 I switch directly between folder pane and 
message pane.

It's the same stuff on LibreOffice, you switch between pane with F6 and 
cycle through panes with tab.

Best regards.
-- 
Alex ARNAUD

bergentroll:
Does autocompletion work in the run dialog? I never saw this.
F6 is a way too, buy I agree, that it's not obvious. It needs a remark in help at least.

info-cppsp:
Well, I saw it in the code, haven't tried / used it.
On my dev setup it doesn't work, even if it is enabled in gsettings...
Tab goes through the window elements, like normal...
another issue
meh.

@raveit65 
Could you please ask the team to share their opinion?

(I can set F6 if all agree, no problem.)

alexarnaud:
Le 31/01/2018 à 13:13, Anton Karmanov a écrit :
> Does autocompletion work in the run dialog?

On Debian Jessie with Mate 1.8 GTK2 and Debian Sid with Mate 1.18 (GTK3) 
it works on my test case.

Steps to reproduce:
1) Open run dialog with alt+F2
2) Type mate-te
3) Press tab

Result: mate-terminal is auto-completed.

Best regards.

raveit65:
I do not agree to change current behaviour which is arrow down/up.
You will always make someone unhappy if you change defaults.
Be careful with that.
Especially , a few days before release...
Missing accessibility  support itself is a bug fix and should be include in 1.20 or a point release.

Btw. we will look for a solution for making you part of team discussions at github mate-groups when  @flexiondotorg comes back.
I don't have the rights to manage that.

Also @alexarnaud should be better integrated, imo.

alexarnaud:
Le 31/01/2018 à 14:57, raveit65 a écrit :
> Btw. we will look for a solution for making you part of team discussions 
> at github mate-groups when @flexiondotorg 
> <https://github.com/flexiondotorg> comes back.
> I don't have the rights to manage that.

Maybe the team discussion should be public as a devel mailing list is.
It could be useful if I could be integrated to the team discussion from 
a user perspective and visual-impaired perspective.

Best regards.

raveit65:
> Maybe the team discussion should be public as a devel mailing list is. It could be useful if I could be integrated to the team discussion from a user perspective and visual-impaired perspective.

sigh,
it is very useful to have an area in every project to discuss things with public ;-)
It's Ok if you don't want, np.

lukefromdc:
I would not want to lose the BASH style up/down cycling through previously used commands, it's something I use a lot

raveit65:
Great, i have forgotten again how to enable orca on fedora box.
Can someone other test this PR?
@alexarnaud 
@bergentroll

info-cppsp:
OK. locally I have changed the down key to F6.
There is the label 'List of known applications.'
I could write sg like 'List of known applications. F6 to jump here.' or whatever.
Is this a good idea? What should I write?

lukefromdc:
How about "Show List of Known Applications: F6" like in a menu listing of a keyboard shortcut?

alexarnaud:
Le 31/01/2018 à 21:05, raveit65 a écrit :
> Great, i have forgotten again how to enable orca on fedora box.
> Can someone other test this PR?
> <https://github.com/alexarnaud>

I could do indeed but maybe I should wait the final implementation with 
another shortcut.

Best regards.

raveit65:
I got orca runing and i hear orca speaking when i use the arrow up/down keys for scrolling the history...
.... without the PR.  😕

alexarnaud:
Le 01/02/2018 à 11:52, raveit65 a écrit :
> I got orca runing and i hear orca speaking when i use the arrow up/down 
> keys for scrolling the history...
> .... without the PR. 😕

As I understand the issue that @info-cppsp tries to solve is the way to 
switch between search box and applications list with keyboard. I agree 
with him that the current behavior is not efficient, not really a 
accessibility issue but a usability issue with keyboard so it impacts 
keyboard-only users (blind people).

Best regards.

raveit65:
For the record, this is tested with using merged https://github.com/mate-desktop/mate-panel/commit/f20f4b9e45c1ac33e651fdd77ccf8a530c894d73
and 'show-program-list' gsettings key is enabled and program-list is open.

raveit65:
Same with 'show-program-list' gsettings key is false.
Using the arrow up/down keys are working well with orca.
So what does the PR do?

raveit65:
> As I understand the issue that @info-cppsp tries to solve is the way to switch between search box and applications list with keyboard.

Please test, this works for me with normal tab keyboard navigation.

alexarnaud:
Le 01/02/2018 à 12:16, raveit65 a écrit :
>     As I understand the issue that @info-cppsp
>     <https://github.com/info-cppsp> tries to solve is the way to switch
>     between search box and applications list with keyboard.
> 
> Please test, this works for me with normal tab keyboard navigation.

Yes, it works but you've to press tab 4 or 5 times, right ? I found that 
less efficient than just pressing F6.

Best regards.

alexarnaud:
Le 01/02/2018 à 12:46, raveit65 a écrit :
> To avoid more confusion in this already confused Pr.
> What should be be the exactly function of F6 ?
> selecting the program-list directly?
> Or replacing arrow up/down keys for navigation inside program-list?
> Last one we don't want to change as it is a default.

Switch between the search box (text field) and the applications list.

Best regards.

raveit65:
> Switch between the search box (text field) and the applications list.

OK, this is now clear and i agree if this is possible.
But my origin question you didn't answer.
But what does the current code of this PR if orca is working?

info-cppsp:
Just force pushed a new version.
The behavior is the following:
If the program list is enabled and open and the user presses the F6 key - while inside the GtkComboBoxText (entry field) - the focus should jump to the list.
There is no switching between GtkComboBoxText and program list.
It is a one way trip from GtkComboBoxText to program list.
This solves the original issue.
(Also: nothing happens if the focus is not in the GtkComboBoxText)

If you think it's a good idea, I can try to add code to enable the switching.

alexarnaud:
Le 05/02/2018 à 15:54, info-cppsp a écrit :
> Just force pushed a new version.
> The behavior is the following:
> If the program list is enabled and open and the user presses the F6 key 
> - while inside the GtkComboBoxText (entry field) - the focus should jump 
> to the list.
> There is no switching between GtkComboBoxText and program list.
> It is a one way trip from GtkComboBoxText to program list.
> This solves the original issue.
> (Also: nothing happens if the focus is not in the GtkComboBoxText)
> 
> If you think it's a good idea, I can try to add code to enable the 
> switching.

Indeed, it could be great if we could have the switching with F6 between 
the search entry and the application list.

Best regards.
-- 
Alex ARNAUD

info-cppsp:
added new commit
this version switches between program list and text box with F6
jumps to program list regardless where the current focus is

raveit65:
This works.
@info-cppsp 
Can you please squash both commits and rename the commit to something significantly?
eg:
run-dialog: use F6 to switch to program list

alexarnaud:
Le 14/03/2018 à 16:19, Victor Kareh a écrit :
> I know I'm late to this conversation, but have we given some thought to 
> /also/ supporting |Ctrl+L|? @alexarnaud <https://github.com/alexarnaud> 
> - do you have any thoughts on this?

What ctrl+l is intended to do?

Best regards,
Alex.

vkareh:
@alexarnaud - for me `Ctrl+L` usually does the same as `F6` - in browsers it focuses on the address bar. Same in caja. Also `Ctrl+K` usually focuses on the search input field (again on browsers, I've also seen it on Slack). I've also seen these bindings on other random software. Just wanted to check with you if this is standard behavior, but based on your question it sounds as though it's not :stuck_out_tongue:

alexarnaud:
Le 14/03/2018 à 16:32, Victor Kareh a écrit :
> bindings on other random software. Just wanted to check with you if this 
> is standard behavior, but based on your question it sounds as though 
> it's not 😛

I'm not aware of this but GTK/GNOME guidelines are a little bit 
differents than Windows one.

After verifying here: 
https://developer.gnome.org/hig/stable/keyboard-input.html.en

ctrl+l seems only intended for location.

Best regards,
Alex.

vkareh:
Makes sense. I use `Ctrl+L` extensively for both firefox and caja, but you're right: it seems like an application-specific shortcut.

In that case, I'd say this pull request seems good, @raveit65.

alexarnaud:
Le 14/03/2018 à 19:09, lukefromdc a écrit :
> I can confirm that F6 works, tabbing through still works, history still 
> works. Ideally changes like this would get added to the panel help so as 
> to be discoverable by new users. That's getting rather far behind I 
> suspect but would be a PR to M-U-G I'm pretty sure (help only covers 
> clock and fish in this repo)

Right point. I'll do that when this PR will be merged.

Best regards,
Alex.

info-cppsp:
thank you!

info-cppsp:
:confused: 
???
I have done that.

raveit65:
Sorry , i didn't saw that.
Thank you.

PR #726, opened 2018-02-02
panel-stock-icons.c: fix failed assertation  from deprecated code

lukefromdc:
Fix three  "gtk_icon_factory_add: assertion 'icon_set != NULL' failed" warnings. Note that the gtk_icon_factory stuff is all deprecated:
https://developer.gnome.org/gtk3/stable/gtk3-Themeable-Stock-Images.html#gtk-icon-factory-add

### DIFF
--- a/mate-panel/panel-stock-icons.c
+++ b/mate-panel/panel-stock-icons.c
@@ -53,82 +53,9 @@ GtkIconSize panel_add_to_icon_get_size(void)
 	return panel_add_to_icon_size;
 }
 
-typedef struct {
-	char *stock_id;
-	char *icon;
-} PanelStockIcon;
-
-static PanelStockIcon stock_icons [] = {
-	{ PANEL_STOCK_FORCE_QUIT, PANEL_ICON_FORCE_QUIT }
-};
-
-static void
-panel_init_stock_icons (GtkIconFactory *factory)
-{
-	GtkIconSource *source;
-	int            i;
-
-
-	source = gtk_icon_source_new ();
-
-	for (i = 0; i < G_N_ELEMENTS (stock_icons); i++) {
-		GtkIconSet *set;
-
-		gtk_icon_source_set_icon_name (source, stock_icons [i].icon);
-
-		set = gtk_icon_set_new ();
-		gtk_icon_set_add_source (set, source);
-
-		gtk_icon_factory_add (factory, stock_icons [i].stock_id, set);
-		gtk_icon_set_unref (set);
-	}
-
-	gtk_icon_source_free (source);
-
-}
-
-typedef struct {
-	char *stock_id;
-	char *stock_icon_id;
-	char *label;
-} PanelStockItem;
-
-static PanelStockItem stock_items [] = {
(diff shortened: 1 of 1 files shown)


lukefromdc:
These are the runtime warnings from mate-panel fixed by this PR:

```
(mate-panel:21500): Gtk-CRITICAL **: 15:18:06.980: gtk_icon_factory_add: assertion 'icon_set != NULL' failed

(mate-panel:21500): Gtk-CRITICAL **: 15:18:06.980: gtk_icon_factory_add: assertion 'icon_set != NULL' failed

(mate-panel:21500): Gtk-CRITICAL **: 15:18:06.980: gtk_icon_factory_add: assertion 'icon_set != NULL' failed
```
This is with Debian Unstable up to date as of Feb 1, 2018, gtk 3.22.26 and glib 2.0-0 2.55.1 both locally compiled.

Here are some test results, looking at icons that were named in the old structures, and the run dialog which is one of the few places any of this is directly invoked: The applet "don't delete" icon on applet loading error works fine, as do icons in run dialog. These icons and others also seem to change fine with theme changes, same as without the PR

The entire panel_init_stock_icons_and_items function cannot be removed or replaced with an empty function, because the rest of it sets up the icon sizes. Without it, icons in the run dialog appication list balloon to huge sizes. Apparently with current GTK 3.22 at least the old and deprecated icon factory code is no longer needed here.

Found all of this while testing https://github.com/mate-desktop/mate-applets/pull/305 and naturally wondered if something had changed in an applet. Warnings occurred even in a panel with all applets removed, so went digging for gtk_icon_factory_add and found this.

raveit65:
I see this warnings too im my ~/.xsession-errors.
So is this a regression?

muktupavels:
What it fixes? I see only code removal.

Did you use `gdb` to find place where this warning is generated? `gtk_icon_factory_add` is called in two places in code you remove. `gtk_icon_set_new` according to documentation always returns valid object so only one place remains. `gtk_icon_factory_lookup_default` is documented that it can return `NULL`, but code does not check for that.

Most likely `NULL` is returned because no such stock ids exist. My guess is that it is regression from:
https://github.com/mate-desktop/mate-panel/commit/1f72d441289df987dd7fbb528089f71e719e3c46

raveit65:
I can confirm that the PR removes the warnings from ~/.xsession-errors.

sc0w:
@muktupavels 

> My guess is that it is regression from: 1f72d44

yes, I opened an alternative fix: https://github.com/mate-desktop/mate-panel/pull/729

sc0w:
any way, If the code removed with this PR is useless, I agree with this PR, I am not sure

raveit65:
No this code isn't useless!
Please check icons and translations.
w/o PR
force-quit-dialog-good
with PR
force-quit-dialog-bad

sc0w:
ok, @raveit65 please test https://github.com/mate-desktop/mate-panel/pull/729

lukefromdc:
Closing in favor of https://github.com/mate-desktop/mate-panel/pull/729

PR #730, opened 2018-02-05
Added gir for mate_panel_applet_set_size_hints()

### LINKED ISSUE #727: Problem with Python introspection and MatePanelApplet.set_size_hints
I'm not sure if this is the right place or not to report this, but I'm having a problem with MatePanelApplet python introspection. I'm trying to call MatePanelApplet.set_size_hints() but keep getting "TypeError: Must be number, not list" exceptions.

I've looked at the code in applets/wncklet/window_list.c and it seems that I'm passing the right parameters in the right order. I've even tried changing them around but no joy - I get the same TypeError no matter what I do.

This happens on both mate-panel 1.18.4 (Ubuntu Mate 17.10) and mate-panel 1.18.7 (Arch), but seems to have been around at least a year or so. I found [this](https://ubuntu-mate.community/t/cant-get-python-matepanelapplet-set-size-hints-to-work/11377) while googling for a solution.

To reproduce the error, I've adapted the the Mate University python applet and attached the code

[university-python-applet.tar.gz](https://github.com/mate-desktop/mate-panel/files/1691547/university-python-applet.tar.gz)

info-cppsp:
fix #727

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -273,6 +273,13 @@ mate_panel_applet_size_hints_changed (MatePanelApplet *applet,
 	return FALSE;
 }
 
+/**
+ * mate_panel_applet_set_size_hints: 
+ * @applet: applet
+ * @size_hints: (array length=n_elements): List of integers
+ * @n_elements: Length of @size_hints
+ * @base_size: base_size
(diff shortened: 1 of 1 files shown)


info-cppsp:
See issue.
Added writing annotations to the TODO.

PR #739, opened 2018-02-18
Fix hide buttons make panel bigger

### LINKED ISSUE #733: Hide buttons make panel bigger
on ub 18.04
if hide buttons are enabled, when the mp is like 20px thick, it gets much wider, so that the buttons can fit.
screenshot at 2018-02-10 12-59-15
screenshot at 2018-02-10 12-59-36

info-cppsp:
fix #733

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -1275,14 +1275,14 @@ static void panel_toplevel_update_hide_buttons_size(GtkWidget* button, gboolean 
 
 	GtkStyleContext *context;
 	context = gtk_widget_get_style_context(button);
-	gtk_style_context_add_class(context, "panel_toplevel_hide_button");
+	gtk_style_context_add_class(context, "panel-button");
 
 	GtkCssProvider *css_provider;
 	css_provider = gtk_css_provider_get_default ();
 
 	/* set custom css */
 	if (set_size) {
-		gtk_css_provider_load_from_data (css_provider, ".panel_toplevel_hide_button {padding: 0px;}", -1, NULL);
+		gtk_css_provider_load_from_data (css_provider, ".panel-button {padding: 0px;min-height: 15px;}", -1, NULL);
 
 	/* get defaults from theme */
 	} else {
@@ -1295,7 +1295,7 @@ static void panel_toplevel_update_hide_buttons_size(GtkWidget* button, gboolean 
 		g_free (gtk_theme_name);
 	}
 
-	gtk_style_context_add_provider (context, GTK_STYLE_PROVIDER (css_provider), GTK_STYLE_PROVIDER_PRIORITY_SETTINGS);
+	gtk_style_context_add_provider (context, GTK_STYLE_PROVIDER (css_provider), GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
 
 }
 
@@ -1349,13 +1349,13 @@ static void panel_toplevel_update_hide_buttons(PanelToplevel* toplevel)
 				      NULL);
 
 		gtk_widget_set_size_request (toplevel->priv->hide_button_top,
-					     -1, arrow_size);
+					     -1, toplevel->priv->size);
 		gtk_widget_set_size_request (toplevel->priv->hide_button_bottom,
-					     -1, arrow_size);
+					     -1, toplevel->priv->size);
 		gtk_widget_set_size_request (toplevel->priv->hide_button_left,
-					     arrow_size, -1);
+					     toplevel->priv->size, -1);
 		gtk_widget_set_size_request (toplevel->priv->hide_button_right,
-					     arrow_size, -1);
+					     toplevel->priv->size, -1);
 
 	} else {
 		gtk_widget_hide (gtk_bin_get_child (GTK_BIN (toplevel->priv->hide_button_top)));
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -1179,42 +1179,18 @@ panel_toplevel_add_hide_button (PanelToplevel *toplevel,
 				int            left,
 				int            top)
 {
+
 	GtkWidget *button;
 	AtkObject *obj;
 	GtkWidget *arrow;
-	int        arrow_size;
 
 	button = gtk_button_new ();
 	obj = gtk_widget_get_accessible (button);
 	atk_object_set_name (obj, _("Hide Panel"));
 	gtk_widget_set_can_default (button, FALSE);
 
-	gtk_widget_style_get (GTK_WIDGET (toplevel),
-			      "arrow-size", &arrow_size,
-			      NULL);
-
-	switch (arrow_type) {
-	case GTK_ARROW_UP:
-		gtk_widget_set_size_request (button, -1, arrow_size);
-		break;
-	case GTK_ARROW_DOWN:
-		gtk_widget_set_size_request (button, -1, arrow_size);
-		break;
-	case GTK_ARROW_LEFT:
-		gtk_widget_set_size_request (button, arrow_size, -1);
-		break;
-	case GTK_ARROW_RIGHT:
-		gtk_widget_set_size_request (button, arrow_size, -1);
-		break;
-	default:
-		g_assert_not_reached ();
-		break;
-	}
-
 	arrow = gtk_image_new ();
 	set_arrow_type (GTK_IMAGE (arrow), arrow_type);
-	gtk_widget_set_margin_start(GTK_WIDGET(arrow), 0);
-	gtk_widget_set_margin_end(GTK_WIDGET(arrow), 0);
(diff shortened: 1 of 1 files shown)


lukefromdc:
Might be the way you did the cssprovider, which is different from anything else we've done in MATE. I will play with this a bit

info-cppsp:
commit2
updated my PR
now on ub 18.04 with all themes
the minimum panel size, where the panel height doesn't jump with enabling the buttons is 18px.
so if panel is 17px, enabling the buttons makes the panel grow to 18px.
this is not that bad...

changing the size of the mp kinda distorts the buttons, but everything is correct if you toggle the buttons or if the mp is restarted or --replace

so far I couldn't make the buttons grow as a square according to the mp size change

-----
commit3
updated PR
removed old useless code
added min-width - needed for panels on the sides
added code to change arrow size

info-cppsp:
"The button is now a small stripe if the arrow is disabled, especially with a small panel size."
yeah I know. it should be a nice square...

info-cppsp:
the 2 traditional themes on ub 18.04 make the panel grow with 15px
the 2 high contrast themes with 14px
works with 13px though
```c
gtk_css_provider_load_from_data (css_provider, ".panel-button {min-height: 13px; min-width: 13px; padding: 0px;}", -1, NULL);
```

info-cppsp:
commit 4
this should make the arrows the right size, so that there is no jumping when you enable the buttons or the arrows, regardless of theme on ub 18.04 and panel size

note: still doesn't work right when resizing the panel - you have to disable / enable the buttons to have it working correctly

raveit65:
Ok, button isn't really a square, but i can live with it ;)
Same with the need of enable/disable the buttons with a new panel size, i don't think the panel-size will be change very often.

info-cppsp:
OK, commit 5
buttons are square and change with panel size change
w00t

alexarnaud:
Le 25/02/2018 à 12:32, info-cppsp a écrit :
> "The button is now a small stripe if the arrow is disabled, especially 
> with a small panel size."
> yeah I know. it should be a nice square...

Sorry to pollute the thread but could it be possible for you to use the 
quote function of your mail client or Github ? I'm visual-impaired and I 
read mail with a screen reader and I can't know that you quote another 
person except if you use quote feature. I know you've quoted because I 
read the two mails but It's really difficult for me.

Best regards,
Alex.

raveit65:
> commit 5
> buttons are square and change with panel size change

Yes, confirmed

lukefromdc:
Here's what happened on changing theme from high contrast to green submarine:

theme_not_applied

info-cppsp:
@lukefromdc 
I can't reproduce this.
I know that there is a side effect, bc of the theme variant.
I don't know how to get whether or not the dark variant is set for a theme.
This (??) may cause the button on theme change w/o panel restart look different.

As per some tests, all of this is gone after a VM restart.

I have no idea how to fix this, as the panel_toplevel_update_hide_buttons () is called several times for a theme change.

-----
OK, I am going to push another commit.
The idea is to take the theme's css, add my own, and set this new css for the gtk style.
The only problem is that gtk_css_provider_to_string () throws 6 million 
`(mate-panel:8415): GLib-CRITICAL **: 17:29:38.202: g_base64_encode_step: assertion 'in != NULL' failed`
This can make a panel update on 'changing theme' take up minutes depending on theme...
https://www.bountysource.com/issues/51312063-glib-critical-g_base64_encode_step-assertion-in-null-failed
Note that this seems to fix the buttons though...

and all this because you can't set the size of something in gtk+ 3
...

lukefromdc:
I can confirm the mess with the warnings, and a theme change from Adwaita to my own theme left the top part of the Adwaita theme's button borders showing on the panel hide buttons. Also big-time CPU use while changing themes, with mate-panel as the CPU user and consuming up to a third of all the capacity of an 8-thread AMD bulldozer proc for a couple seconds on theme change.

lukefromdc:
Was not trying to approve this round of changes, only commenting. Will see if I can silence these warnings

lukefromdc:
Looking at the code, this is an incredibly complex way (in what the machine has to do, not the coding in this PR) to deal with forcing down the minimum size of the buttons. I will now try each varient of this to see where the redraw issue came in and try to find another fix

lukefromdc:
Redraw issue when changing away from the high contrast theme came in with the 3ed commit,
https://github.com/mate-desktop/mate-panel/pull/739/commits/f7ddeca362cfa613b793e8914b7765d29fda64f4

 redraws from high contrast at
https://github.com/mate-desktop/mate-panel/pull/739/commits/5c2800ea9081e82e8c79c44babc5586b15ffd30b

worked, so it's something to do with the arrow size code

lukefromdc:
Nope: even there, high contrast sometimes fails to draw the arrow in the first place. If the panel is then restarted from that theme, the redraw problem is back. Will now try my original code to see if it is present there too

lukefromdc:
I just found that this redraw issue came when the panel failed to render the arrows in high contrast and was then restarted. The early parts of the code loaded either the custom CSS or the theme but not both, so I tried loading first the theme, then the custom CSS with this:

```
            /*load theme settings*/
	GtkSettings * settings;
	settings = gtk_settings_get_default ();
	gchar *gtk_theme_name;
	g_object_get (settings, "gtk-theme-name", &gtk_theme_name, NULL);
	css_provider = gtk_css_provider_get_named (gtk_theme_name, NULL);

	g_free (gtk_theme_name);

        	/* set custom css for size*/
	if (set_size) {
		gtk_css_provider_load_from_data (css_provider,
                                        ".panel_toplevel_hide_button {\n"
                                        "min-height: 17px; \n"
                                        "}", -1, NULL);
        }

	gtk_style_context_add_provider (context, GTK_STYLE_PROVIDER (css_provider), GTK_STYLE_PROVIDER_PRIORITY_SETTINGS);

```
No redraw problem here, as the arrows always showed up in high contrast but high contrast was forced to 19 or 20px. Will now try this change against Version 5 of the PR and see what happens

lukefromdc:
https://github.com/info-cppsp/mate-panel/pull/1 will fix both the redraw and warning issues, I set it up as a PR to this PR which then could be rebased and merged. Alternately I could rebase it myself against master for a diredt merge. What I did was to revert the last attempt at a redraw fix, then apply my change loading the theme css first and unconditionally, then loading the custom css. No more problems with the arrow not showing up on a switch to high contrast with a small panel, then no redraw issue if then restarting the panel (e.g. the user's next session) and then changing themes again.

info-cppsp:
https://developer.gnome.org/gtk3/stable/GtkCssProvider.html#gtk-css-provider-load-from-data
"Loads data into css_provider, and by doing so **clears any previously loaded information**."

I believed the documentation...
Thanks, I'll check this out later.

lukefromdc:
That means I've now got something that works here but does so for an undocumented reason. 'll take a further look at it

info-cppsp:
I have checked your PR with all themes on ub 18.04 and changing, while the panel is running, works as expected.

So we have finally arrived at a solution for a very trivial problem...

lukefromdc:
OK, that's two of us finding
https://github.com/info-cppsp/mate-panel/pull/1
with this PR works, my guess is that loading data into a cssprovider must only in practice clear the previously loaded data. If the theme css is not loaded first and unconditionally, we get the redraw problem and fail to render the arrows with high contrast. I tried loading two different cssproviders altogether, that did not work at all as the second one failed to override the first-even if loaded with a higher priority.

If nobody else finds a problem, I propose to document in the code that this is not behavior documented in GTK, then merge this with my fix applied

info-cppsp:
06 also had init ptrs to NULL
so I just took your changes and added a new commit 07
I have tested this and it seems to work.

mate-panel/panel-toplevel.c
```diff
@@ -1269,11 +1245,77 @@ static void panel_toplevel_update_buttons_showing(PanelToplevel* toplevel)
...
+	g_object_get (settings, "gtk-theme-name", &gtk_theme_name, NULL);
+	css_provider = gtk_css_provider_get_named (gtk_theme_name, NULL);
+	g_free (gtk_theme_name);
+
+
+	/* set custom css by adding our custom code to the default theme css
+	 *
+	 * NOTE that contriary to the documentation:
+	 * https://developer.gnome.org/gtk3/stable/GtkCssProvider.html#gtk-css-provider-load-from-data
+	 * the previously loaded theme is NOT cleared from the css_provider. (reason unknown)
```
muktupavels: What? Did quick test and it works... Also did you know that mentioned function returns boolean? You could use that to check if it executed successfully... Also that function has GError that you could use to get error message in case it fails.

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/fa5cde36593be06eabfb2f752f38c26b80ea269c

PR #740, opened 2018-02-18
Fix hide buttons making panel bigger

lukefromdc:
Fix /github.com/mate-desktop/mate-panel/issues/733#issuecomment-366538876
Based on code by https://github.com/info-cppsp


info-cppsp:
Please note, that this is an incomplete solution, instead, please help with #739 
Thank you.

lukefromdc:
This was nothing but initial code to get things going and should be closed

Issue #737, opened 2018-02-14
[HiDPI] can't move default panel with window-scaling-factor=2


clayfreeman:
#### Expected behaviour

Default panel should remain on the original display and should be able to move freely between multiple displays.

#### Actual behaviour

Default panel becomes stuck on secondary display after editing `window-scaling-factor`.

#### Steps to reproduce the behaviour

1. Set `window-scaling-factor`=`2` with `dconf-editor`.
2. Observe that the default panel has moved to the secondary display.
3. Attempt to move the default panel back to the primary display.

#### MATE general version

`1.20.0`

#### Package version

`master` branch using Gentoo eBuild `>=mate-base/mate-9999`

#### Linux Distribution

Gentoo Linux

#### Link to downstream report of your Distribution

lukefromdc:
I was unable to duplicate this with 1.20.0 on Debian Unstable, one 1920x1080 monitor and one 1024x768 secondary monitor. I had to add a second panel so I would have a spot not taken by an applet to click on for the panel-properties dialog, put it on the bottom with my other panel, then set window scaling to 2. No problem moving it top or bottom. I've never had panels on or moved to the secondary display at all

clayfreeman:
Were you able to move panels across displays, even though they were never moved in the first place?

My default panel was forcibly rendered on the second display, even though `dconf-editor` claimed that it was on display `0`.

What other information should I provide so that this can be replicated successfully? I was using two Dell P2415Q displays running 3840x2160@60Hz.

vkareh:
I experience this issue using a 3840x2160 laptop with a 1920x1080 monitor (using xrandr to scale the resolution and reposition the second screen). At the moment I'm getting a segfault when resetting the panel layout, but as soon as I manage to fix that I can start working on this issue.

For reference: https://github.com/mate-desktop/mate-settings-daemon/issues/213

lukefromdc:
I have never moved a panel off the primary display, so I went looking in dconf-editor for a setting to do so. My panel is on the bottom, so the setting is org>mate>panel>toplevels.bottom>screen and by default is 0. Resetting this to another value with another monitor up and running was entirely ignored. That might be the root of the problem here

lukefromdc:
One more point: any chance this is WM dependent? I normally use compiz on the desktop and marco on the laptop.

vkareh:
[...]

PR #744, opened 2018-02-24
Scale monitor geometries down to device pixels

vkareh:
This allows for supporting multiple monitors in HiDPI.

_Caveat_: I do _not_ own an external HiDPI monitor, so I use `xrandr` in the command line to set the scale and positioning of the second monitor.

@flexiondotorg - I think you're the only one here with both a HiDPI laptop _and_ a HiDPI external monitor. Could you test this, please?

### DIFF
--- a/mate-panel/panel-multiscreen.c
+++ b/mate-panel/panel-multiscreen.c
@@ -219,21 +219,8 @@ panel_multiscreen_get_gdk_monitors_for_screen (GdkScreen     *screen,
 	num_monitors = gdk_display_get_n_monitors (display);
 	geometries = g_new (GdkRectangle, num_monitors);
 
-	for (i = 0; i < num_monitors; i++) {
-		GdkMonitor *monitor;
-		int         scale;
-
-		monitor = gdk_display_get_monitor (display, i);
-		scale = gdk_monitor_get_scale_factor (monitor);
-
-		gdk_monitor_get_geometry (monitor, &(geometries[i]));
-
-		/* Scale geometries down to device pixels to support HiDPI displays */
-		geometries[i].x /= scale;
-		geometries[i].y /= scale;
-		geometries[i].width /= scale;
-		geometries[i].height /= scale;
-	}
+	for (i = 0; i < num_monitors; i++)
+		gdk_monitor_get_geometry (gdk_display_get_monitor (display, i), &(geometries[i]));
 
 	*monitors_ret = num_monitors;
 	*geometries_ret = geometries;
--- a/mate-panel/panel-multiscreen.c
+++ b/mate-panel/panel-multiscreen.c
@@ -88,11 +88,14 @@ panel_multiscreen_get_randr_monitors_for_screen (GdkScreen     *screen,
 						 GdkRectangle **geometries_ret)
 {
 #ifdef HAVE_RANDR
+	GdkDisplay         *display;
+	GdkMonitor         *monitor;
 	Display            *xdisplay;
 	Window              xroot;
 	XRRScreenResources *resources;
 	RROutput            primary;
 	GArray             *geometries;
+	int                 scale;
 	int                 i;
 
 	if (!have_randr)
@@ -138,6 +141,11 @@ panel_multiscreen_get_randr_monitors_for_screen (GdkScreen     *screen,
 		return FALSE;
 
 	primary = XRRGetOutputPrimary (xdisplay, xroot);
+	display = gdk_screen_get_display (screen);
+	monitor = gdk_display_get_primary_monitor (display);
+
+	/* Use scale factor to bring geometries down to device pixels to support HiDPI displays */
+	scale = gdk_monitor_get_scale_factor (monitor);
 
 	geometries = g_array_sized_new (FALSE, FALSE,
 					sizeof (GdkRectangle),
@@ -157,10 +165,10 @@ panel_multiscreen_get_randr_monitors_for_screen (GdkScreen     *screen,
 			crtc = XRRGetCrtcInfo (xdisplay, resources,
 					       output->crtc);
 
-			rect.x	    = crtc->x;
-			rect.y	    = crtc->y;
-			rect.width  = crtc->width;
-			rect.height = crtc->height;
+			rect.x      = crtc->x / scale;
+			rect.y      = crtc->y / scale;
+			rect.width  = crtc->width / scale;
+			rect.height = crtc->height / scale;
 
 			XRRFreeCrtcInfo (crtc);
 
@@ -211,8 +219,21 @@ panel_multiscreen_get_gdk_monitors_for_screen (GdkScreen     *screen,
 	num_monitors = gdk_display_get_n_monitors (display);
 	geometries = g_new (GdkRectangle, num_monitors);
 
-	for (i = 0; i < num_monitors; i++)
-		gdk_monitor_get_geometry (gdk_display_get_monitor (display, i), &(geometries[i]));
+	for (i = 0; i < num_monitors; i++) {
+		GdkMonitor *monitor;
+		int         scale;
+
+		monitor = gdk_display_get_monitor (display, i);
+		scale = gdk_monitor_get_scale_factor (monitor);
+
+		gdk_monitor_get_geometry (monitor, &(geometries[i]));
+
+		/* Scale geometries down to device pixels to support HiDPI displays */
+		geometries[i].x /= scale;
+		geometries[i].y /= scale;
+		geometries[i].width /= scale;
+		geometries[i].height /= scale;
+	}
 
 	*monitors_ret = num_monitors;
 	*geometries_ret = geometries;
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -1708,7 +1708,6 @@ panel_run_dialog_setup_entry (PanelRunDialog *dialog,
 	GdkScreen             *screen;
 	int                    width_request;
 	GtkWidget             *entry;
-	gint                   scale;
 
 	dialog->combobox = PANEL_GTK_BUILDER_GET (gui, "comboboxentry");
 
@@ -1721,10 +1720,9 @@ panel_run_dialog_setup_entry (PanelRunDialog *dialog,
 		(GTK_COMBO_BOX (dialog->combobox), 0);
 
 	screen = gtk_window_get_screen (GTK_WINDOW (dialog->run_dialog));
-	scale = gtk_widget_get_scale_factor (GTK_WIDGET (dialog->run_dialog));
 
         /* 1/4 the width of the first monitor should be a good value */
-	width_request = panel_multiscreen_width (screen, 0) / (4 * scale);
+	width_request = panel_multiscreen_width (screen, 0) / 4;
 	g_object_set (G_OBJECT (dialog->combobox),
 		      "width_request", width_request,
 		      NULL);
--- a/mate-panel/panel-struts.c
+++ b/mate-panel/panel-struts.c
@@ -75,17 +75,10 @@ panel_struts_get_monitor_geometry (GdkScreen *screen,
 				   int       *width,
 				   int       *height)
 {
-        GdkDisplay *display;
-        int scale;
-
-        /* Use scale factor to bring strut dimensions up to application pixels to support HiDPI displays */
-        display = gdk_screen_get_display (screen);
-        scale = gdk_monitor_get_scale_factor (gdk_display_get_monitor (display, monitor));
-
-        *x      = panel_multiscreen_x      (screen, monitor) * scale;
-        *y      = panel_multiscreen_y      (screen, monitor) * scale;
-        *width  = panel_multiscreen_width  (screen, monitor) * scale;
-        *height = panel_multiscreen_height (screen, monitor) * scale;
+        *x      = panel_multiscreen_x      (screen, monitor);
+        *y      = panel_multiscreen_y      (screen, monitor);
+        *width  = panel_multiscreen_width  (screen, monitor);
+        *height = panel_multiscreen_height (screen, monitor);
 }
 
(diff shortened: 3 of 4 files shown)


vkareh:
Fixes #737 
Fixes mate-desktop/mate-settings-daemon/issues/213

flexiondotorg:
I do not have an external HiDPI capable monitor. I have 2x 27" 1440p monitors. Using that workstation, if I set `window-scaling-factor` to `2` with `dconf-editor` to force scaling, create a new panel it is not possible to locate the new panel on the secondary display.

vkareh:
@flexiondotorg - okay, I'll try to work on a fix for that. Are you able to move existing panels between the monitors though? It used to be that panels got somehow constrained to the secondary display - is that fixed for you with this patch?

flexiondotorg:
All panels are constrained to the primary display when scaling is enabled. 

If I create a panel on each display while scaling is auto and then switch to forced scaling, all panels are displayed on the primary display only. Switching back to auto-scaling all panels are restored to the correct displays.

vkareh:
@flexiondotorg - based on your description, are these the steps?
1. Set scale = 0 (using HiDPI laptop)
2. Create a panel on laptop screen, another on monitor
3. Set scale = 2 - here panels get constrained to laptop only
4. Set scale = 0 - panels are back to screens in which they were created

Do those look like appropriate reproduction steps? (Don't have a monitor at the moment, but want to try this out a bit later when I do)

raveit65:
@flexiondotorg What setting do you use for moving the extra panel to external monitor?
I use this (path in dconf-editor)
 /org/mate/panel/toplevels/topevel-0/monitor/  1

Here monitor 1 is the external monitor.
For some reasons the other normal 2 panels doesn't have the monitor entry.
/org/mate/panel/toplevels/top/
/org/mate/panel/toplevels/bottom/
I assume they use monitor 0 and are the primary monitor
Enable scale=2 will leave the extra panel on the external monitor here.

Sorry, i have no idea to show you this relocated schema via command line.

Sorry, i have no idea

vkareh:
I just `Alt+Click+Drag` the panel from one monitor to another. I'll have to try it with gsettings

raveit65:
> I just `Alt+Click+Drag` the panel from one monitor to another.

This key doesn't do anything here O_o
What magic did you do to enable this?

vkareh:
huh, what's your mouse button modifier key? in dconf: `/org/mate/marco/general mouse-button-modifier` - I have mine set to `<Alt>`, which seems to be the default

vkareh:
@raveit65 - the standard top/bottom (in `/org/mate/panel/toplevels/`) panels have a "monitor" entry for me. It has the following description: "Xinerama monitor where the panel is displayed". Changing it there works well for me too using this patch

vkareh:
Okay, I was able to reproduce the unreachable panel thing when creating a new panel. The core top/bottom ones seem to work well on my side, but creating a new panel made it show up only on the secondary monitor, whenever it was in the primary monitor, I got a placeholder area where the panel would've been in the external monitor, but nothing in the primary - on using gsettings to move it to the secondary monitor, it showed up there just fine. The issue only applied to a third panel in right or left positions. It works well using top or bottom on either monitor...

vkareh:
Okay, I think I found an issue. I was scaling the monitor geometries whether they were reported by xrandr or by Gdk. Scaling the ones from Gdk seemed unnecessary, so I undid that change - I'll keep testing it just in case.

Also found what may be the core issue. Seems like moving panels to the side, the position in which they should end up is scaled incorrectly, which means that the center of the panel ends up being offset by the full height of the screen (I use my external monitor above my laptop) and so the panel ends up _below_ the laptop monitor, which is invisible. I'll have to explore how to fix this, but seems like a positioning (x, y) issue - achievable for sure.

clayfreeman:
As I mentioned in another thread - there is a possibly related bug:

> However, applications seem to ignore that the panel is there (or at least take up more space than they should) when maximized.

info-cppsp:
@vkareh 
note that there are a lot of issues concerning panel placement and 2 monitors, and they may not be related to hidpi.

raveit65:
I tested it again with new commit.
Now the third vertical panel overlaps the other panels (top+bottom) if i hover with the mouse in case of scale=2.  O_o


> As I mentioned in another thread - there is a possibly related bug:

>> However, applications seem to ignore that the panel is there (or at least take up more space than they should) when maximized.

Confirmed, but this happens for me with scale=0 too.
But only with the third exra vertical panel on internal and external monitor, and if the vertical panel is inside the Cinerama.
With normal standard (top+bottom) panels i don't see that.
So this is an old issue not related to HiDPI support.
Did we open Pandora's Box with starting HIDIPI support so late before 1.20 release?

offtopic on
Yes mouse-modifier (alt) is enabled here, but i can't drag a panel to another monitor.
Maybe a limitation of using vga cable for external monitor?
And do we use the same DE ?  :)
panel-bottom-dconf
panel-top-dconf
panel-vertical-dconf
offtopic off

vkareh:
@raveit65:
screenshot at 2018-02-27 06-18-02

Seems like we're not using the same DE after all! :astonished: 

Now, since you're in Fedora, you're in compiling with `--disable-schemas-compile` and  `--with-x` - Ubuntu doesn't use those options... I wonder if it's related?

What does the `--with-x` option do?

vkareh:
@clayfreeman - yeah, there is a known issue where applets that are in the division between both monitors overlap maximized apps. This happens in any monitor configuration (top-bottom, right-left) regardless of scale or resolution.

vkareh:
Okay, I finally got ahold of a secondary monitor (not HiDPI), and I've been playing with different configurations. Time for some...

Test results!
====

First, some notes:
* primary:   HiDPI - 3840x2160 (laptop)
* secondary: LoDPI - 1680x1050
* definition: _strut_: Space that is same size as panel, but pushes maximized windows away so that they don't overlap it. Well-behaved struts will always be the same position and size of the panel to which they belong.
* I only tested the snap points (left-aligned, top-aligned, etc) and not things like "secondary on top, halfway through the width of primary" or anything like that, since snap points are easier to reason about for now.
* Third panel overlapping seems to be inconsistent in different scenarios, so I'm ignoring it for now
* All tests are using scale=2
* Configuraion options are (apart from the path ones): `--disable-silent-rules --disable-static --enable-introspection --enable-gtk-doc --with-in-process-applets=all` (the last one is not standard in Ubuntu)

primary : secondary
----
#### top-aligned
* panel on top of secondary: panel strut is on primary monitor
* panel on bottom of secondary: maximized windows on primary take only left half of monitor

#### bottom-aligned
* panels are not visible on secondary monitor

secondary : primary
----
#### top-aligned
* panel on top of primary: panel strut is on secondary

#### bottom-aligned
* panel on bottom of primary: panel strut is on secondary
* panel on right of secondary: panel shows up on right of primary

primary
secondary
----
#### left-aligned
* panel on left of primary: panel shows up on bottom half on primary, strut takes up left side of both monitors
* panel on right of primary: panel shows up on bottom half of primary

#### right-aligned
* panel on right of primary: panel shows up on bottom half of primary, strut takes up right side of both monitors
* panel on left of primary: panel shows up on bottom half of primary

secondary
primary
----
#### left-aligned
* panel on left of secondary: panel is invisible, strut shows on primary monitor
* panel on right of secondary: panel is invisible, maximized windows on primary take only top half of monitor

#### right-aligned
* panel on right of secondary: panel is invisible, strut shows on primary monitor
* panel on left of secondary: panel is invisible, maximized windows on primary take only top half of monitor

clayfreeman:
After reading a bit more about these struts, I can confirm that the struts for my default panel are on my secondary (left) monitor, even after applying the patch in this PR.

screenshot at 2018-02-27 08-10-49

info-cppsp:
no wonder that there are so many issues about multimon setups...
thanks for the testing!

vkareh:
@flexiondotorg, @raveit65, @clayfreeman - update! I added some extra scaling logic for the struts. This _should_ solve the struts being placed on the wrong monitor, as well as panels overlapping each other.

raveit65:
> Seems like we're not using the same DE after all! :astonished: 
>Now, since you're in Fedora, you're in compiling with `--disable-schemas-compile` and  `--with-x` - Ubuntu doesn't use those options... I wonder if it's related?
>What does the `--with-x` option do?

We're using the same DE \o/
1. all that keys are only added to the relocated schema if you set them in preferences, i forgot how relocated schemas are working.....
2. I learned that if the panel is expand the alt+drag click is blocked.
This is working after i un-expand the bottom and top panel :)
3. After that the monitor key is in the relocated schema.

offtopic on
Using --disable-schemas-compile is rpm specfic. After compiling, all files will be installed to the build_root folder with dirs during rpm creation.
Than those dirs/files were copied in the resulting rpm (compressed).
So, here it isn't needed to compile the schemas. This will be done with installation of the rpm with a macro of the dnf package-manager.
offtopic off

raveit65:
Ok, problem with overlapping panels are gone, looks OK.

clayfreeman:
Struts are now correctly positioned. However, I have a segregated window list for each monitor that works until I move a window past a certain point. See the attached screenshots.

## Correct
screenshot at 2018-02-27 23-37-15

## Incorrect
screenshot at 2018-02-27 23-37-30

clayfreeman:
It seems like the window list changes when the center of the window is past the center of the monitor.

raveit65:
Does this work with scale=1  ?

vkareh:
@clayfreeman - very interesting. I'll take a look at that. I suspect that the issue is inside the window list code (which has its own set of issues with HiDPI too anyway).

I'll play with this today, thanks for testing :)

clayfreeman:
@raveit65 The split window list works as expected with `window-scaling-factor=1`.

vkareh:
Yeah I can reproduce this... there's nothing obvious in the code that would trigger this behavior. Possibly an issue in the wnck library

vkareh:
@clayfreeman - what monitor configuration and resolutions are you using? Are you using any scaling through xrandr? If I set `--scale 1x1` on my external monitor, then window list issue goes away. This makes me think it's an xrandr issue and not really within the scope of the mate-panel (maybe?)

clayfreeman:
I have two Dell P2415Q displays at 3840x2160 each with my primary display being my right display. To my knowledge the only configuration that I’ve touched is the scaling factor setting. I am not forcing a scale with xrandr.

clayfreeman:
After running `xrandr --output DisplayPort-{0,1} --scale 1x1`, I observed no changes in interface scale or window list behavior with `window-scaling-factor=2`.

If I changed `window-scaling-factor` to `0` or `1`, my display positions got goofed up real bad.

vkareh:
Okay, yeah, this all sounds reasonable given your screen resolutions.

I have a few more things to try before I start looking for blame in another repo, but so far I'm not getting very far with the window list :/

Is this the last remaining issue here?

clayfreeman:
I believe so. Thank you very much for your help with this. I can finally use the desktop environment that I want to use on my HiDPI displays!

The only other issues that I've noticed are mouse cursor and window button size issues, but those might be related to the theme that I'm using or another MATE package. Either way, it's not a huge deal. I will file issues though if you know what packages might be causing the issues (maybe Marco?).

clayfreeman:
@vkareh You may try to find out which screen/display an application is on when it is moved via some debug logging. The fault might not be within the applet, but it might be due to some of the hackiness about how Marco scales things.

vkareh:
@clayfreeman - thanks for your patience and support in testing all this! :)

The window buttons are marco, the cursor may be mate-control-center or mate-settings-daemon, depending on what's causing it. Marco currently does not support proper HiDPI, that's the last one on the list (already have bug filed for that :P)

I will work on the window list separate from this issue so that users can have a fixed panel structure in the meantime. Thanks again.

vkareh:
@flexiondotorg - any chance you can test this and confirm that it fixes the multi monitor panels on your side?

clayfreeman:
@vkareh Sounds good! Again, thank you for your tremendous help!

raveit65:
@vkareh 
Anythng left to do with this PR, btw. is it fine to use this as patch?
I 've f28 beta freeze at 2018-03-05.

vkareh:
@raveit65 - I think we're good. It fixes the issue for both me and the reporter, and doesn't cause any regressions as far as I can tell.

For the patch, yeah. This is necessary for 1.20, so patch is good! :)

flexiondotorg:
Sorry for the late follow up on this issue, but I also confirm it works correctly when I force HiDPI scaling on mt dual 27" 2560x1440 displays. Nice work @vkareh

simbiryatin:
Hi guys,
Sorry that 2 years later, I raise this ticket, but right now I have the same issue as in the screenshot: https://github.com/mate-desktop/mate-panel/pull/744#issuecomment-369128206

Also, I read this thread: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910345

So, I understand that this fix was not included in GTK + 3.24 (and any above). Is there any chance to fix this?

Thanks

raveit65:
Fix is in >= mate-panel.1.21. see git history.

simbiryatin:
Hi @raveit65 ,
> Fix is in >= mate-panel.1.21. see git history. 
Is that you for me?  How can I upgrade this?  I have Mint 20 Ulyana. Thanks. 
P.S. Sorry, noob in the thread

raveit65:
Sorry, i don't know nothing about this distribution, maybe file out a report against linuxmint and ask them?

simbiryatin:
> Sorry, i don't know nothing about this distribution, maybe file out a report against linuxmint and ask them?

@raveit65 I checked, my version of panels is 1.24. But this not work. It seems to me that the main reason for the problem is that you have a version (GTK) check in your code (pseudocode): 
if(get_version(GTK)==323)
instead of 
if (get_version(GTK)>=323)

PR #747, opened 2018-03-02
Fix out-of-process applets flickering on HiDPI

vkareh:
This seems to stabilize out-of-process applets on my side so that they stop flickering. Some visual glitches might still be present, but the flickering should be gone.

Fixes mate-desktop/mate-applets/issues/308
Fixes mate-desktop/mate-settings-daemon/issues/213
Fixes mate-desktop/mate-settings-daemon/issues/214

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1026,10 +1026,8 @@ mate_panel_applet_get_preferred_width (GtkWidget *widget,
 		 * they are back at their own intended size.
 		 */
 		scale = gtk_widget_get_scale_factor (widget);
-		if (scale) {
-			*minimum_width /= scale;
-			*natural_width /= scale;
-		}
+		*minimum_width /= scale;
+		*natural_width /= scale;
 	}
 }
 
@@ -1050,10 +1048,8 @@ mate_panel_applet_get_preferred_height (GtkWidget *widget,
 		 * they are back at their own intended size.
 		 */
 		scale = gtk_widget_get_scale_factor (widget);
-		if (scale) {
-			*minimum_height /= scale;
-			*natural_height /= scale;
-		}
+		*minimum_height /= scale;
+		*natural_height /= scale;
 	}
 }
 
@@ -1062,6 +1058,9 @@ mate_panel_applet_get_request_mode (GtkWidget *widget)
 {
 	MatePanelApplet *applet = MATE_PANEL_APPLET (widget);
 	MatePanelAppletOrient orientation;
+
+	if (applet->priv->out_of_process)
+		return GTK_SIZE_REQUEST_CONSTANT_SIZE;
 
 	orientation = mate_panel_applet_get_orient (applet);
 	if (orientation == MATE_PANEL_APPLET_ORIENT_UP ||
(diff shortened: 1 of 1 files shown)


lukefromdc:
Speaking only for myself here, recalling core-team for review as this really needs to be tried on a large enough monitor, and by someone who has seen the issue in the first place. Code itself looks like a relatively simple fix

flexiondotorg:
OOops. Tested this on a non-HiDPI laptop. It does not correct the size of the na-tray applets I mentioned.

muktupavels:
`GTK_SIZE_REQUEST_CONSTANT_SIZE` part does not seem right thing to do... And commit does not include any explanation why is that necessary.

vkareh:
@muktupavels - to be honest, that bit there is just trial and error. That's the only thing that consistently stops the flickering applets.

But I still haven't figured out why they are flickering. I wish I knew. :/

raveit65:
Further testing shows me that only the fish and cpu-freq applets have graphical glitches if you add them after setting windows-scalling=2.
If applets are added with windows-scalling=0/1 than more have glitches when you switch to windows-scalling=2.
But this is with 1920x1080 px in a VM, maybe glitches are related to missing space with a non-hidipi monitor.

lukefromdc:
@raveit65, I assume your results in the last comment were with master and not with this PR?

raveit65:
No, my results are with PR and build --with-in-process-applets=none in a VM with a non-hidpi monitor.
But the flickering issue is gone.

lukefromdc:
OK, that with this PR and all applets out of process I can confirm the fish renders is contents intermittantly with this when you change the window scaling to 2 and then add it. Having done so,  the cpufreq applet one momentarily rendered half-height and empty, then quicky started rendering normally. Another instance of it was already running on the bottom panel, which may have altered my results.

vkareh:
In my experience it seems that the applets rendering at half size is not related to this patch. This patch only fixes the flickering on the na-tray widget. I should update this patch so that only the na-tray applet renders using constant size.

I don't think we should merge this just yet, though. If I manage to figure out what's causing the half-size rendering, then fixing that should also fix the flickering _without_ having to use this patch - I just haven't found what the issue is...

raveit65:
For clarification, i never said that possible glitches are coming from this PR.
I pretty sure i will see the same with master.

lah7:
I was having a flickering issue on a 4K display on Arch, and can confirm this PR fixes the flickering. :+1:

lukefromdc:
Fish glitch confirmed same in master with in-process build. Cpufreq-applet started normally with window-scaling=2 but was momentary glitch anyway with this PR.

lukefromdc:
Also, I don't see anything in the code here that should create new issues, as long as 

`scale = gtk_widget_get_scale_factor (widget);

`does not return NULL, which it should not as this function returns 1 or a higher integer assuming it does not break entirely.

lukefromdc:
Cpufreq-applet issue appears to be just a momentary half-size render until the applet size updates in about a second, so this should not be a blocker

lukefromdc:
Just confirmed two more things: very severe flicker in master with an overcrowded panel (scaling up a loaded panel on a 1080X1920 display), and that the window list upon being added renders half-size both with this PR and with master. Will open a separate issue for that one if someone has not already.

It thus appears that this PR is both needed and causes no new issues beyond maybe a slight delay in the Cpufreq-applet expanding to full size, and as that is intermittant (usually the applet just waits to show up at all until redrawn) I cannot confirm that that does not take place in master as well. Whatever makes the cpufreq applet redraw and display normally might be the fix for the window list as well.

lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/ff04f332136272a14c588da3f2d582ba0ff01be1

lukefromdc:
Opened 
https://github.com/mate-desktop/mate-panel/issues/754
for the window-list issue. Same behavior before and after this commit

lukefromdc:
Just created
https://github.com/mate-desktop/mate-panel/pull/755
to fix https://github.com/mate-desktop/mate-panel/issues/754
it turned out to be very simple

PR #755, opened 2018-03-20
Fix rendering of out-of-process window list with hidpi

### LINKED ISSUE #754: (Hidpi) Window list displays half-height at window-scaling = 2
#### Expected behaviour
Window list displayed full height of panel when setting window-scaling = 2 and adding applet to panel

#### Actual behaviour
If and only ifwncklet is built out-of-process, upon adding the window list to a panel with window-scaling = 2 , it will be displayed at half-height until the panel is restarted, another applet in wncklet is added, or possibly something else forces a redraw

Build mate-panel with wncklet out of process and install. Create a new, empty panel to ensureyou have space (I tested on a 2080x1920 display) Restart the panel, open dconf-editor and go to org>mate>desktop>interface and set window-scaling-factor =2. Add the window list to the panel. It stays at half panel height and not just for a few seconds either. Can be forced to redraw by adding the workspace-switcher

#### MATE general version
1.20.0 from git master

#### Package version
locally compiled

#### Linux Distribution
Debian Unstable

None as this is a local build of all of MATE

lukefromdc:
Fix #754 
Don't show the applet until it's size has been determined including any hidpi scaling

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -427,8 +427,6 @@ gboolean window_list_applet_fill(MatePanelApplet* applet)
 	g_signal_connect(G_OBJECT(tasklist->tasklist), "destroy", G_CALLBACK(destroy_tasklist), tasklist);
 
 	g_signal_connect(G_OBJECT(tasklist->applet), "size_allocate", G_CALLBACK(applet_size_allocate), tasklist);
-	tasklist_update(tasklist);
-	gtk_widget_show(tasklist->tasklist);
 
 	gtk_container_add(GTK_CONTAINER(tasklist->applet), tasklist->tasklist);
 
@@ -484,6 +482,8 @@ gboolean window_list_applet_fill(MatePanelApplet* applet)
 
 	g_object_unref(action_group);
 
+	tasklist_update(tasklist);
+	gtk_widget_show(tasklist->tasklist);
 	gtk_widget_show(tasklist->applet);
 
 	return TRUE;
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -526,13 +526,13 @@ gboolean workspace_switcher_applet_fill(MatePanelApplet* applet)
 	g_signal_connect(G_OBJECT(pager->pager), "scroll-event", G_CALLBACK(applet_scroll), pager);
 
 	gtk_container_add(GTK_CONTAINER(pager->applet), pager->pager);
-	gtk_widget_show(pager->pager);
 
 	g_signal_connect(G_OBJECT(pager->applet), "realize", G_CALLBACK(applet_realized), pager);
 	g_signal_connect(G_OBJECT(pager->applet), "unrealize", G_CALLBACK(applet_unrealized), pager);
 	g_signal_connect(G_OBJECT(pager->applet), "change_orient", G_CALLBACK(applet_change_orient), pager);
 	g_signal_connect(G_OBJECT(pager->applet), "change_background", G_CALLBACK(applet_change_background), pager);
 
+	gtk_widget_show(pager->pager);
 	gtk_widget_show(pager->applet);
 
 	mate_panel_applet_set_background_widget(MATE_PANEL_APPLET(pager->applet), GTK_WIDGET(pager->applet));


vkareh:
Holy crap, thanks for this!! I will test as soon as I have access to my laptop tonight :D

raveit65:
Not sure if this is the issue but it seems that the problem is still there with this PR.
Same for clock applet.
After a while something trigger a redraw and all looks nice.
And sometimes i see the issue and sometimes not after switching to windows-scale=2.
Test with older i5 proc (2 cores) and integrated graphic (intel) on my notebook, maybe the graphic hardware is to slow for HIDPI?
windows-list
clock-applet
Note: for triggering the issue for the clock-applet you need to disable a location.

vkareh:
I see the same issue as @raveit65, but the ocurrence of it is drastically reduced, without this PR it happens almost every other time, with this PR it happens only ocassionally, and only when the window list is the last applet to load in the panel. I'm testing this in Fedora 28, as the issue is not present for me in Ubuntu MATE.

One thing I noticed is that doing the same for the workspace-switcher also fixes it for me:
```diff
diff --git a/applets/wncklet/workspace-switcher.c b/applets/wncklet/workspace-switcher.c
index df7add9..e882bd0 100644
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -526,15 +526,12 @@ gboolean workspace_switcher_applet_fill(MatePanelApplet* applet)
        g_signal_connect(G_OBJECT(pager->pager), "scroll-event", G_CALLBACK(applet_scroll), pager);
 
        gtk_container_add(GTK_CONTAINER(pager->applet), pager->pager);
-       gtk_widget_show(pager->pager);
 
        g_signal_connect(G_OBJECT(pager->applet), "realize", G_CALLBACK(applet_realized), pager);
        g_signal_connect(G_OBJECT(pager->applet), "unrealize", G_CALLBACK(applet_unrealized), pager);
        g_signal_connect(G_OBJECT(pager->applet), "change_orient", G_CALLBACK(applet_change_orient), pager);
        g_signal_connect(G_OBJECT(pager->applet), "change_background", G_CALLBACK(applet_change_background), pager);
 
-       gtk_widget_show(pager->applet);
-
        mate_panel_applet_set_background_widget(MATE_PANEL_APPLET(pager->applet), GTK_WIDGET(pager->applet));
 
        action_group = gtk_action_group_new("WorkspaceSwitcher Applet Actions");
@@ -554,6 +551,9 @@ gboolean workspace_switcher_applet_fill(MatePanelApplet* applet)
 
        g_object_unref(action_group);
 
+       gtk_widget_show(pager->pager);
+       gtk_widget_show(pager->applet);
+
        return TRUE;
 }
```
In fact, I cannot reproduce the issue for workspace-switcher after doing this!

lukefromdc:
@raveit65, what your image shows looks different from anything I saw, you have the applet simultaniously rendering at half and full height while I had it rendering at half-height only.  Wonder if this is a WM difference? My test was with my compiz-reloaded branch, and I got either half-height rendering without this PR or full height rendering with it, never both at once.

Maybe what we really need is another gtk_widget_queue_draw or gtk_widget_queue_resize somewhere?

vkareh:
@lukefromdc - whenever mine fails it looks similar to @raveit65's. I use marco.

lukefromdc:
PR rebased after
https://github.com/mate-desktop/mate-panel/commit/0b15ea1f6d218cc2a850bef3f0cb98d303c63bc3

vkareh:
@lukefromdc - do you see the issue with the workspace-switcher app? If so, does my inline diff fix it for you?

lukefromdc:
The workspace-switcher always renders full height in my setup(in process or out), possibly due to a faster proc. I probably should add your fix to the PR though

lukefromdc:
I just found that in a second test today after merging the tray fix and with the change for the switcher, the half-height rendering is back. We must be looking at a race conditiion here, the real fix is going to have to be to force a a redraw somewhere

raveit65:
> what your image shows looks different from anything I saw, you have the applet simultaniously rendering at half and full height while I had it rendering at half-height only. Wonder if this is a WM difference?

All images are with marco WM.
I am +1 for merging it, as it improves the situation and changes are minimal

vkareh:
Agreed - after rebasing, still works on my side - thanks for adding the workspace-switcher fix, it does improve things for me :)

flexiondotorg:
I've not experienced this issue on Ubuntu MATE 18.04 daily, but this patch plus #756 do not introduce any regressions.

lukefromdc:
Will merge this, found my inconsistant results came from where I was adding the window list to the panel. If there was enough space for the buttons (as on a real hidpi display), they rendered normally.

Issue #742, opened 2018-02-24
[HiDPI] Gaps between icons in Notification Area enlarged with window scaling


MingcongBai:
#### Expected behaviour

Gaps between icons in notification area to remain the same regardless of window scaling settings.

#### Actual behaviour

With `window-scaling-factor=2`, gaps between icons are visibly enlarged, inconsistent with original layout. See the following two screenshots for comparison.

1366x768 @ 1x

![2018-02-23 23-25-42](https://user-images.githubusercontent.com/5006263/36626049-e5396ac4-18f0-11e8-83f9-a6cce8f6436a.png)

2560x1440 @ 2x

![screenshot at 2018-02-23 20-43-47](https://user-images.githubusercontent.com/5006263/36626054-f08bb1fc-18f0-11e8-8def-82f00cb3d6b5.png)

#### Steps to reproduce the behaviour

Set `window-scaling-factor=2`.

#### MATE general version

MATE 1.20.0.

#### Package version

mate-* 1.20.0.

#### Linux Distribution

AOSC OS, package tree: https://github.com/AOSC-Dev/aosc-os-abbs/tree/staging/extra-mate.

#### Link to downstream report of your Distribution

N/A

JosephMcc:
I'm seeing this exact issue with Mate-panel 1.20 under Ubunut 18.04. The spacing seems to be huge regardless of the theme being used. I'm running a 2x scaling factor on a 3840x2160 resolution.

lukefromdc:
I saw this is well in an unrelated test last night using window-scaling=2

lukefromdc:
closed by https://github.com/mate-desktop/mate-panel/pull/756

PR #756, opened 2018-03-21
hidpi: fix tray icon spacing with window-scaling > 1

lukefromdc:
Fix too wide spacing of tray icons when window-scaling-factor = 2 or more. Tested both in and out of process, and in both vertical and horizontal panels.

### DIFF
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -120,6 +120,8 @@ na_tray_child_get_preferred_width (GtkWidget *widget,
                                    gint      *minimal_width,
                                   gint      *natural_width)
 {
+  gint scale;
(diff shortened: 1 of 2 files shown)


lukefromdc:
Note that in testing this by moving panels from top to right, I found an unrelated issue where a panel with and only with at least one in-process applet collapses left/to the top when shifting from top/bottom to left/right orientation. Restarting the panel, adding or removing any applet, or resizing it at least two px then restores normal rendering. This is present in master and has nothing to do with this PR so don't worry about it yet if you see it in testing.

flexiondotorg:
@lukefromdc Am I right in saying that #742 is fixed by this pull request?

lukefromdc:
Yes it is, so we can close that now

PR #759, opened 2018-03-27
Updateable icons in the panels.

Isomorph70:
The icon files are now monitored and if they change in the file system, the new version is loaded and displayed in the panel. Making it possible to make simple information displays in the panels, simply by rewriting the icon file.

PS. also swapped some functions to clean up the code.

### DIFF
--- a/mate-panel/button-widget.c
+++ b/mate-panel/button-widget.c
@@ -4,6 +4,7 @@
 
 #include <glib/gi18n.h>
 #include <gdk/gdkkeysyms.h>
+#include <gio/gio.h>
 #include <gtk/gtk.h>
 #include <gdk-pixbuf/gdk-pixbuf.h>
 
@@ -30,15 +31,12 @@ struct _ButtonWidgetPrivate {
 	PanelOrientation  orientation;
 
 	int               size;
-
+	GFileMonitor *icon_monitor;
 	guint             activatable   : 1;
 	guint             ignore_leave  : 1;
 	guint             arrow         : 1;
 	guint             dnd_highlight : 1;
 };
-
-static void button_widget_icon_theme_changed (ButtonWidget *button);
-static void button_widget_reload_pixbuf (ButtonWidget *button);
 
 enum {
 	PROP_0,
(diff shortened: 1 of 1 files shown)


raveit65:
Can you please do a pull request to master branch?
1.10 is pretty EOL.

lukefromdc:
The file worked in master, which given what it was based on must have been pure chance. Thus this could be cherrypicked to a new branch based on master rather easily

raveit65:
Cherry-picking will fail because of removing gtk2 bits.
https://github.com/mate-desktop/mate-panel/pull/759/commits/04db8f34a2a5308aba2f1fdf6ee395f8c2e8b6fa#diff-91454453959d3a3d2ea52621511e4ae5L238

Isomorph70:
@lukefromdc to test it 'add to panel' a 'Custom Application launcher' , copy a svg icon to somewhere you can modify it. Then add it to the launcher as an icon. Now when you modify and save the svg file, the icon should update.

lukefromdc:
That will have to wait until tonight, and will require complete overwrite of the file as I do not know how to edit a .svg file, only make a new one in Inkscape. This is something I really have no idea how to use.

Isomorph70:
Download the chains.svg from [here](https://ipfs.io/ipfs/QmPsR2EPDVv2cbEY3SZWZ2JPD8R4VGURQoFXEYqcCTLGdy)

in terminal do
`cp chains.svg display.svg`
Then add the display.svg as an icon in the launcher , and maybe /usr/bin/eom /PATH/display.svg as a command. 

When the icon is displayed in the panel, do the following in the terminal. 

`cat chains.svg | sed 's/red/yellow/' >display.svg`
This overwrite the display.svg file with a new version of chains.svg where the color red is edited to yellow. As you do that the icon should update and display the new version.

muktupavels:
Why?

lukefromdc:
This does work, but I agree its not something we'd be likely to actually use. What are real-world likely use cases for this? I don't see rewritten .svg files being used as a notification device

lukefromdc:
Certainly we are not merging anything to 1.10, that is an old and no longer supported branch

info-cppsp:
after mate-panel --replace, or restarting the computer, the new svg file is going to be loaded anyway, right?
I also don't see any point in this feature.
It is way easier to just change the icon in the GUI.

flexiondotorg:
I'm unclear what the use case or benefit for this is. Closing.

Isomorph70:
@muktupavels @info-cppsp @flexiondotorg let me repeat for you 
> Making it possible to make simple information displays in the panels, simply by rewriting the icon file.

This make it possible for a (power) user to write scripts that display information in the panels. 
I have used if to make a task tracking app, using only a few scripts...

Or you could say it is a visual version of the [Command applet](https://github.com/mate-desktop/mate-applets/tree/master/commandl)

muktupavels:
Why should panel monitor file changes for all users when only few if not only you would actually use this functionality? If you want to see something added / accepted by maintainers you have to convince that it will be useful for **most** users not only to few...

Right now I would say that you are better of to create simple applet with functionality you need.

_I am **not** MATE developer, feel free to ignore anything I write..._

alexarnaud:
Le 28/03/2018 à 15:58, Svend Haugaard Sørensen a écrit :
> This make it possible for a (power) user to write scripts that display 
> information in the panels.
> I have used if to make a task tracking app, using only a few scripts...

You could probably right an applet to accomplish this task as it seems 
it's not a general use case.

Best regards,
Alex.

info-cppsp:
https://github.com/info-cppsp/mate-university/tree/cappletud
here I have a simple applet, which puts out an icon.
hopefully it works, it is a work in progress
you could take it and rewrite it pretty easily, so that it refreshes the icon every 2 seconds...

lukefromdc:
Splitting this out into an applet is the right approach. In MATE we have to balance features against things like total RAM consumption and time to execute, as many use MATE on slower systems or performance-critical systems. If the panel was unitary with all applets part of the same panel binary (as in many simple panels), nobody would be able to run any applet or the bare panel without loading them all, even if they were to exit 0 when not in use. The modular panel/applet system saves RAM and CPU cycles in every use case save someone loading all possible applets (as in a test of them) in which case it is probably slightly heavier.

Isomorph70:
@muktupavels @alexarnaud ,doing a applet is a good point. And that is how it started., I read the code for the 'Custom Application Launcher', but as I investigated how it load in the icon file, I discovered that it delegate it to the button-widget, and since I needed the file type to init the monitor, it seems like the natural place to add the code. 

Also button-widget have some similar code for updating the icon on theme change. 

Later I also found out the the drawer application, was more useful for what I wanted to do, and since I make the changes in the button-widget I could just use it without any programing.   
 
I also thought about adding a switch, so you could set in properties, if you want a updateable icon or a ordinary.

Isomorph70:
Here is my first [application ](https://github.com/Isomorph70/gochains) using the extension.

PR #762, opened 2018-04-09
avoid gtk_dialog_add_buttons with stock ids

sc0w:


raveit65:
I have no idea where are all the dialogs, but it seems OK.

Issue #758, opened 2018-03-26
image border of applet doesn't appear with GNOME Theme [30$]


alexarnaud:
#### Expected behaviour
Image border should be displayed even if with GNOME themes

#### Actual behaviour
No image border appears. It makes removing applet such as the notification area really difficult for a user, especially a visual-impaired users because there is no locations displayed on the screen to perform the removal. You need to be aware of the existence of such image border to perform the deletion.

#### Steps to reproduce the behaviour
1) Change the theme to HighContrast
2) Add notification area to a panel
3) Try to remove the notification area

#### MATE general version
1.20

#### Package version
1.20.0-3

#### Linux Distribution
Debian Sid

alexarnaud:
@raveit65 FYI, I've added a bounty for this bug as I found this a really annoying issue.

Best regards.

raveit65:
Maybe this image can be added with a GtkCssProvider?
Same what we do in themes
https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/BlueMenta/gtk-3.0/mate-applications.css#L345
or
https://github.com/mate-desktop/mate-themes/blob/master/desktop-themes/BlackMATE/gtk-3.0/mate-applications.css#L341
@lukefromdc 
Do you think this is possible?

lukefromdc:
I would not want to see borders forced onto applets so that themes could not override them, as this would make some panel themes ugly. This should be done in themes only and should not be hardcoded into any applet. 

Already in my theme I have to ensure panel buttons (notably show desktop) do not show borders. Thus this would require a way for users to turn it off in dconf-settings.

lukefromdc:
I just looked at the linked code, and see that is is the HANDLE, not the border of the tray applet and the window-list that is in question here. That is the part that is clicked on to move or bring up the preferences menu.

This part can be and should be hardcoded with a sane default that a theme can deliberately override, as would be necessary to set the color and pattern to match the theme. If the theme does not override it, the sane default is used. That's exactly how we get a panel background other than transparent in Adwaita and should be easy to do.

raveit65:
No, description is misleading. He means the grip area left from na-tray- and wncktasklist-applet.
The three points in BlackMate.

raveit65:
Yeah,  you posted faster than me.

lukefromdc:
Try 
https://github.com/mate-desktop/mate-panel/pull/766
[...]

PR #766, opened 2018-04-12
Theme: show applet handles on non-MATE themes

lukefromdc:
Fix https://github.com/mate-desktop/mate-panel/issues/758
Ensure that themes such as GNOME themes (e.g Adwaita) that do not explicitly support MATE show the drag handles for applets such as tray and window-list. Tuned for Adwaita and Adwaita-Dark

### DIFF
--- a/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
+++ b/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
@@ -262,6 +262,36 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
 
 		g_free (bg_str);
 	}
+    GtkCssProvider *provider;
+    provider = gtk_css_provider_new ();
+	gtk_css_provider_load_from_data (provider,
+		"MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { \n"
+        "border-style: solid; \n"
+        "border-width: 3px; \n"
+        "border-color: transparent; \n"
+        "background-repeat: no-repeat; \n"
+        "background-position: left; \n"
+        "background-image: linear-gradient(to bottom, \n"
+        "transparent, \n"
+        "transparent 20%, \n"
+        "alpha (#999999, 0.6) 21%, \n"
+        "alpha (#999999, 0.6) 29%, \n"
+        "transparent 30%, \n"
+        "transparent 45%, \n"
+        "alpha (#999999, 0.6) 46%, \n"
+        "alpha (#999999, 0.6) 54%, \n"
+        "transparent 55%, \n"
+        "transparent 70%, \n"
+        "alpha (#999999, 0.6) 71%, \n"
+        "alpha (#999999, 0.6) 79%, \n"
+        "transparent 80%, \n"
+        "transparent); \n"
+        "}",
+        -1, NULL);
+	gtk_style_context_add_provider (gtk_widget_get_style_context(GTK_WIDGET(frame)),
+		GTK_STYLE_PROVIDER (provider),
+		GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
+	g_object_unref (provider);
 }
 
 static void


raveit65:
Ahh, you made simply the handle from code itself visible with choosing the right color.
This works well with PR.
What do you think about to include image with the 3 points which we use in our themes?
This use symbolic-colors and we can be sure that it is visible in every theme.
But i am fine with your solution too if you don't like to ship an image.

lukefromdc:
If you can figure out how to get the image installed so the cssprovider can find it, and get it into the build system that's fine with me too. It was just beyond me to do it.

muktupavels:
[GResource](https://developer.gnome.org/gio/stable/GResource.html)?

lukefromdc:
That is something I do not know how to work with at this time, I don't think it is used anywhere in MATE

lukefromdc:
I am going to see if I can load a symbolic icon by name

lukefromdc:
How would I get the theme name to special-case GNOME's HighContrastInverse? A Startpage search (I do not use Google for anything for privacy reasons) returned no useful information;

muktupavels:
https://git.gnome.org/browse/gnome-panel/tree/gnome-panel/main.c#n61

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/766/commits/2056fca0ab5be89fe822592773db44e8508e86e9

makes the handle visible on all of the GNOME themes. I set the gradient image slightly lighter than the color I was seeing with the "3 holes" Menta image, so as to improve visability on Adwaita. The only panel it would be invisble on would be one with #999999 or something very close to it as the panel color, which does not match any GNOME theme and all the mate themes have their own image file. I played with borders but settled on transparent because the right side of the border is rendered not on the handle but instead is cut off by the rest of the applet. The background (around the three tiny rectangles) could be made lighter or darker to ensure contrast with absolutely ANY panel background but would not look as good in the GNOME themes. Is a trivial change to do that, and I tested both ways, and both in and out of process.

lukefromdc:
Here's my results so far:

Adwaita:
adwaita_drag_handle

Adwaita-Dark:
adwaita_dark_drag_handle

HighContrastInverse:
highcontrastinverse_drag_handle

Raleigh is a somewhat broken theme (or at least used to be) but at least the drag handle is visible with it:

raleigh_drag_handle

lukefromdc:
Rebased after
https://github.com/mate-desktop/mate-panel/commit/d3d057d4080e0dba6744a121d7eae5f50b92a278,
will squash for final merge when ready

alexarnaud:
Hello lukefromdc,

Thank you for fixing the bug. I can confirm that it works.

Could it be possible to increase the visibility of the image border with 
a color as visible as the volume tray icon on HighContrast (bold/white 
as I can see) ? As you know I'm visual-impaired and I find difficult to 
see the current image border. I assume other people using High Contrast 
has the same problem that I have.

Best regards,
Alex.

Le 14/04/2018 à 00:04, lukefromdc a écrit :

lukefromdc:
That's a bit complex: the border is transparent the MATE themes do not specify a border. If a non-transparent border is added, it will appear in the MATE themes as well  requiring changes to every one of them. Similarily, if the three rectangles are made lighter so as to show up more clearly in HighContrastInverse, they become invisible in Adwaita. If they are made darker so as to show up more clearly in light High Contrast themes, they become harder to see in HighContrastInverse. Lastly, it was important to minimuze the risk of selecting a color that would match someone's panel theme and turn invisible as a result. This was the best I can do.

Since the image used here is not and icons and the icon codepath is used for the handle, the symbolic icon  principle cannot be used here without a major rewrite.

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/779
special-cases the GNOME high contrast themes, as our normal theming for the applet handles turned out to be too subtle on them for visually impaired users (or possibly anyone in bright sunlight/shop/etc environments using these themes)
https://github.com/mate-desktop/mate-panel/issues/758#issuecomment-381467672

monsta:
This doesn't look so good with TraditionalGreen (which already had three dots)...

Before:
panel-before

After:
panel-after

sc0w:
@monsta yes, I can confirm the problem with Traditional OK, the grid is wrong

monsta:
I hope this can be fixed in a way that wouldn't tie some themes version to some panel version...

monsta:
BTW, is it possible to explicitly disable these lines/dots in a theme after these changes?

lukefromdc:
If the theme sets a background image, the three dashes should NOT show. If the theme sets a background color only,  setting background-image: none in the theme will disable it. That's what I did for UbuntuStudio_Legacy

lukefromdc:
Found the problem: Turns out the border is cutting our default image. We can either override it by adding

border-width: 0px;
border-style: none;

To the MatePanelAppletFrameDBus > MatePanelAppletFrameDBus
block of all themes,. or I can reduce the border width in the default. That has the disadvantage of looking less attractive in all non-MATE themes, but I can do it tonight if you want

lukefromdc:
More details: Our default image does not normally scale at all in mate-panel 1.20.0 as first released. I did not notice the cutting off of the left side of our image because it is subtle in most themes and I normally am using something else. The default iamge seemed to scale well and look right doing so, all the way to 110px on a horizontal panel

sc0w:
@monsta 

> BTW, is it possible to explicitly disable these lines/dots in a theme after these changes?

I like the lines/dots, it is needed to know where is it, to handle with right click / or holding left click to move it

sc0w:
The same problem with BlueMenta:

before:

withoutprpanel

later:

withprpanel

lukefromdc:
Disabling the default lines in a theme is very simple: set 
background-image: none 
or any other background image and it's gone. On close examination of the screenshots I can see what is wrong with the border: It is pushing the image over to the right rather than cutting off the left edge as I had thought it was doing. The result is the default image in mate themes (or any other theme with its own image) gets pushed over too-but already has the necessary space on the edges so it gets pushed right and cut off. I could remove the border or enable it only for GNOME themes. Experimenting now. What I really need is a way to define an image that itself does not go to any of four edges without having to load a file and with the ability to pick up fg and bg colors

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/782
Fixes the problem in the MATE themes. I got rid of the border entirely and found a way to do everything in just the background-image .

Turns out in CSS and at least in GTK 3.22 you can load two linear-gradients into an image, with the top image in the code rendering over (and thus able to partially conceal) the lower image. No idea how it will behave in older GTK3 versions, but I was able to get all GNOME themes (even Raleigh and win32 by a second special-case block) to look almost exactly like MATE theme for the handle except for using a bit more contrast to show up better.

About scaling: this scales well in a horizontal panel, we could just use it and drop the image file from the mate-themes to fix that problem. In a vertical panel, our MATE 3 holes image does not appear at all, while this does appear on the tray (but not the window list due to % stretching the vertical iamge away from the top) but does not scale on it. I set it's size so only one row of the grid appears in the horizontal panel, which also determined the spacing for a vertical panel to end up correct for about a 26px panel. It is cut by a smaller vertical panel (not widely used) and simply stays the same size on a larger vertical panel like our existing image does on a horizontal panel. 

The root of the vertical panel scaling issue is that the use of percentage on the top rendered grid prevented it from working correctly, as it would be not the width of the handle but rather the width of the entire applet determining it, with most of it being hidden. In the vertical panel the rendering is OK because the image coded in pixels is on top, though it is not as nice and even as in a horizontal panel at least it is there.

No issues with setting window-scaling=2 , that worked fine. There is only one thing that works perfectly though and that is setting only an alpha background color and using the old GNOME mesh image. That scales fines, works on all panel orientations, etc and is what I use in UbuntuStudio_Legacy by disabling the image and setting the color. Total mismatch to the normal MATE theme though, it comes from GTK itself and used to show by default on an unthemed panel with older GTK versions (and maybe requiring older panel code too).

monsta:
@sc0w: I know, I like this feature too, it's just that theme authors might prefer the look without them.

monsta:
Ok, manual setting of background color indeed removes these dots, at least with TraditionalGreen theme.

PR #769, opened 2018-04-13
panel-run-dialog.ui: avoid deprecated:

sc0w:
avoid deprecated:

GtkVBox
GtkHbox
GtkImage:stock
GtkButton:use-stock

### DIFF
--- a/mate-panel/panel-run-dialog.ui
+++ b/mate-panel/panel-run-dialog.ui
@@ -1,210 +1,48 @@
-<?xml version="1.0"?>
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.22.0 -->
 <interface>
-  <requires lib="gtk+" version="2.16"/>
-  <!-- interface-naming-policy toplevel-contextual -->
+  <requires lib="gtk+" version="3.22"/>
+  <object class="GtkImage" id="image1">
+    <property name="visible">True</property>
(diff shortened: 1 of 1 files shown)


lukefromdc:
Rebased and merged with
https://github.com/mate-desktop/mate-panel/commit/4d84a7e7881cfd3b02ef36b443661f5f333efbf7

lukefromdc:
Whoops: had to revert this for now! The panel run dialog looked exactly the same, and you could go through the program list and all-but the caret to enter text would not show up and a program name could not by type. Made a new branch containing the existing commit, pushed it as panel-run-dialog.ui again, and reverted in master until it is fixed.

sc0w:
I don't know how to show the run dialog, how to do it step by step?

btw, I see you reverted the wrong commit:

> lukefromdc Revert "panel-test-applets.ui: avoid deprecated:

lukefromdc:
Just fixed that,reverted the revert and then reverted the right commit. Sorry about the churn, I am tired and got a bit frantic. Run dialog now works again in master, not going to try to squash and get rid of those as that breaks local repos. To show the run dialog, the alt-F2 key combination will bring it up, and also there is a "run application" applet for the panel

sc0w:
ok, fixed in https://github.com/mate-desktop/mate-panel/pull/776

@lukefromdc please test

PR #779, opened 2018-04-16
Accessibility: white/black applet handles in GNOME high contrast themes

lukefromdc:
Special case the GNOME HighContrast and HighContrastInverse themes to give max contrast on tray and window list applet handles. Follow-on to https://github.com/mate-desktop/mate-panel/commit/2e0a74d9458480966f2a87d1d66a58552a873300

### DIFF
--- a/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
+++ b/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
@@ -245,6 +245,8 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
 	MatePanelAppletFrameDBus *dbus_frame = MATE_PANEL_APPLET_FRAME_DBUS (frame);
 	MatePanelAppletFrameDBusPrivate *priv = dbus_frame->priv;
 	char *bg_str;
+    gchar *theme_name;
+    GtkSettings *settings;
 
 	bg_str = _mate_panel_applet_frame_get_background_string (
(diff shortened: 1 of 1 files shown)


raveit65:
No idea to enable the HighContrastInverse from gtk+ in mate session, as they don't ship a index.theme file for it or a folder under /usr/share/themes.
So, a test with our theme where i disabled the icon source for testing should be sufficient.

muktupavels:
`gsettings set org.gnome.desktop.interface gtk-theme HighContrastInverse`

raveit65:
> gsettings set org.gnome.desktop.interface gtk-theme HighContrastInverse

Dhow!
For those who forgot to use dconf (like me) :)
Let us use gtk+ theme version of HighContrastInverse
https://github.com/mate-desktop/mate-themes/pull/210

@lukefromdc 
Don't forgot to claim the bounty.

alexarnaud:
@lukefromdc This doesn't seem to work correctly on Mate 1.24. Do you think you could look to it?

lukefromdc:
Not at the moment due to being effectively a refugee from coronavirus
stay at home orders.

On 4/22/2020 at 7:23 AM, "Alex ARNAUD"  wrote:
	@lukefromdc This doesn't seem to work correctly on Mate 1.24. Do you
think you could look to it?

lukefromdc:
I just found that in mate built from git master post-1.24, I do in fact get the expected white and black applet handles for the window-list and tray applets in both high contrast and high contrast inverse themes

PR #781, opened 2018-04-18
Apply default applet handle theme in GTK 3.18 and earlier builds in mate-panel 1.18

lukefromdc:
This reverts commit 3a8b1f66d2393a35f40dbcc211d3bf045d563a92.
Apply default applet handle theme in GTK 3.18 and earlier builds

### DIFF
--- a/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
+++ b/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
@@ -245,10 +245,9 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
 	MatePanelAppletFrameDBus *dbus_frame = MATE_PANEL_APPLET_FRAME_DBUS (frame);
 	MatePanelAppletFrameDBusPrivate *priv = dbus_frame->priv;
 	char *bg_str;
-#if GTK_CHECK_VERSION (3, 20, 0)
     gchar *theme_name;
     GtkSettings *settings;
-#endif
+
 	bg_str = _mate_panel_applet_frame_get_background_string (
 			frame, PANEL_WIDGET (gtk_widget_get_parent (GTK_WIDGET (frame))), type);
 
@@ -265,7 +264,6 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
 
 		g_free (bg_str);
 	}
-#if GTK_CHECK_VERSION (3, 20, 0)
     GtkCssProvider *provider;
     provider = gtk_css_provider_new ();
 
@@ -327,7 +325,6 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
                                     GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
     g_object_unref (provider);
     g_free (theme_name);
-#endif
 }
 
 static void
--- a/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
+++ b/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c
@@ -274,7 +274,7 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
     if (g_strcmp0 (theme_name, "HighContrast") == 0 ||
                    g_strcmp0 (theme_name, "HighContrastInverse") == 0){
         gtk_css_provider_load_from_data (provider,
-            "MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { \n"
+            "MatePanelAppletFrameDBus { \n"
             "border-style: solid; \n"
             "border-width: 3px; \n"
             "border-color: @theme_bg_color; \n"
@@ -296,7 +296,7 @@ mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
     }
     else{
 	    gtk_css_provider_load_from_data (provider,
-            "MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { \n"
+            "MatePanelAppletFrameDBus  { \n"
             "border-style: solid; \n"
             "border-width: 3px; \n"
             "border-color: transparent; \n"


monsta:
This sort of works with GTK+ 3.18... adds three dashes which look the same in every theme that doesn't draw dots or anything there. I think they're too close to the first taskbar item, it's visible in themes which draw outlines over inactive taskbar items. Another thing is that dashes aren't all scaled the same way when the panel height changes. I think all that applies to the original PR too...

Looks like Github broke uploading attachments... oh well.

Adwaita, 24px:
https://i.imgur.com/OJ6DvH0.png

Adwaita, 25px:
https://i.imgur.com/KrC3CmP.png

monsta:
Nothing changed in ContrastHigh and ContrastHighInverse themes though.

monsta:
With GTK+ 3.14, this doesn't work... no any changes in any themes.

lukefromdc:
Should this be closed, or applied to GTK 3.18 exclusively? I can do either one tonight

raveit65:
I think we need the simple selector for < 3.20
MatePanelAppletFrameDBus { }
This was working in 3.14 branch with Menta themes
https://github.com/mate-desktop/mate-themes/blob/gtk3.14/desktop-themes/BlueMenta/gtk-3.0/mate-applications.css#L416
@monsta Can please adjust this line https://github.com/mate-desktop/mate-panel/blob/7dcc2149608493ce8bc87ee189967854e9fd0033/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c#L277
for testing?

 "MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { \n"
to
 "MatePanelAppletFrameDBus { \n"

raveit65:
Do we need the the latest changes from PR in master?
Because this looks really weird with our themes.
menta
submarine
Tested with f26 VM with gtk+-3.22.x
I used 1.18.8 tarball +
https://github.com/mate-desktop/mate-panel/commit/a0bf261eeb8e500c757ad26b05477ece277af840
https://github.com/mate-desktop/mate-panel/commit/ab46918da1b7df56560cbdd7823e80497c6dee78
https://github.com/mate-desktop/mate-panel/commit/45026ae2442dee55d9dd8bbd4964cfab24c22c9e
Maybe i did something wrong.

And i meant to use a condition for the selector for < 3.20.

And do we really need this fix for 1.18 ?
I am really in doubt that debian stable will ever except an update, because of their policies.
Why spending time for something which will never  used by debian stable?

raveit65:
Ok, i found out why i had such a weird result.
We have to use the same selector which we use for our themes, after i edited Submarine theme to use the simple selector *MatePanelAppletFrameDBus { }*,
It looks like this (handle from theme is cut)
submarine-cuted
So, we need the latest changes from your PR for master to fix the cutting.
Plus a GTK_VERSION_CHECK to match what we use in themes with < 3.20

Or we need to re-think if the double selector is really needed
*MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { }*
Honestly, i never saw this with gtk3-inspector.
Here i see only the simple selector.
panel-gtk3-inspector
panel-gtk3-inspector-2

Personal i would prefer to use the right theme selector in master/1.20/1.18 and fix our themes to use the same.

lukefromdc:
What I just found by a quick test on my own theme is this: a background IMAGE will with GTK 3.22 be restricted to the handle with the simple selector, but a background COLOR will appear across the entire applet. This is with the applets in question built in-process.

lukefromdc:
Also, it's worth keepin in mind that anything applied to 
MatePanelAppletFrameDBus
is also SUPPOSED to be applied to 
Also, it's worth keepin in mind that anything applied to 
MatePanelAppletFrameDBus>Also, it's worth keepin in mind that anything applied to 
MatePanelAppletFrameDBus, 
so I will test just the simple selector and explicitly clearing it from the double selector. The right-side issue shown in some tests could be fixed by abandoning the effort to support this on vertical panels if necessary, leaving that for a proper fix later along with fixing it wheh out themes are used

lukefromdc:
I just found I was testing on the wrong PR, in GTK 3.22. I agree this may be too complex to now support given Master and 3.20 now only support GTK 3.22

lukefromdc:
Using the simple selector did not work in GTK 3.22 with in-process applets. The earlier changes to support the background in GTK 3.20 were also applied to GTK 3.18 builds as they greatly reduced an applet crash problem at that time. The double selector was made necessary by that work and thus should be needed with GTK 3.18 and later but probably not with GTK 3.16 and GTK 3.14 .

raveit65:
> The double selector was made necessary by that work and thus should be needed with GTK 3.18 and later but probably not with GTK 3.16 and GTK 3.14 .

Np, i agree with using them for 3.22/20. But can we use for PR in 3.18 a GTK_VERSION_CHECK < 3.20 for using the simple selector with < 3.20 ?
Otherwise we have to change the selector in mate-themes for gtk+-3.18 branch and release new tarballs. This is something which i want to avoid as this branch in mate-themes is closed and EOL since MATE-1.20 release.

lukefromdc:
If you have it working with GTK 3.18, feel free to use it. If the 3.18 themes are using the simple selector this would probably have to use it too. Check both in and out of process builds, but that does not really matter unless one or more distros is building mate-panel 1.18 with in-process applets and with GTK 3.18 specifically. GTK 3.14/3.16 should have no problem as they don't invoke the changes to the panel background drawing code that originally created problems in GTK 3.20.

lukefromdc:
If you cannot get the simple selector to work for GTK 3.18, the other option is just to go back to lmiting this whole featuree to GTK 3.20 and later builds, that still takes care of Debian Stretch at all.

raveit65:
I do not have an installation or VM with gtk+-3.18 for testing.
The result of mixing simple and double selector is that both will displayed, see my screenshots from 1.18 PR tested with gtk+-3.20.
menta
submarine
I expect the same weirdness with gtk+-3.18 and our themes.

raveit65:
> If you cannot get the simple selector to work for GTK 3.18, the other option is just to go back to lmiting this whole featuree to GTK 3.20 and later builds, that still takes care of Debian Stretch at all.

Yes, please.
I don't really want to release new mate-themes tarballs for closed branches.

lukefromdc:
Closing, as we would need multiple testers to ensure it worked right, and we sure as fsck don't want to release untested or poorly tested updates to old versions

lukefromdc:
Leaving the branch in case someone who can easily test with GTK 3.18 wants to take this up

monsta:
Yeah, this is complicated, and requires a lot more testing for older GTK+ versions... and I don't want to break Mint releases which still use these versions. I agree with keeping it limited to 3.20/22, at least for now.

PR #782, opened 2018-04-19
Theme: consistant theming for applet drag handles in all themes

lukefromdc:
*Do NOT use border, as it is not overridden in MATE themes and pushes the image out of place
*Use two gradient images to make a 3x3 grid resembling MATE theme 3 holes image that works in both vertical and horizontal panels
*Note that while drag handle now shows in vertical panel, it only scales in horizontal panel as px and not % must be used in horizontal direction

### DIFF
--- a/data/mate-panel.css
+++ b/data/mate-panel.css
@@ -1,0 +1,10 @@
+
+MatePanelAppletFrameDBus > MatePanelAppletFrameDBus {
+	border-style: none;
+	background-repeat: no-repeat;
+	background-position: left;
+	color: @theme_fg_color;
+	background-image: -gtk-recolor(url("panel-grid-symbolic.svg"));
+	background-size: 12px 22px;
+}
+
--- a/mate-panel/main.c
+++ b/mate-panel/main.c
@@ -60,7 +60,8 @@ main (int argc, char **argv)
 {
 	char           *desktopfile;
 	GOptionContext *context;
-	GError         *error;
+	GError         *error, *error2;
+	GtkCssProvider *provider;
 
 	bindtextdomain (GETTEXT_PACKAGE, MATELOCALEDIR);
 	bind_textdomain_codeset (GETTEXT_PACKAGE, "UTF-8");
@@ -165,6 +166,23 @@ main (int argc, char **argv)
 	/* Do this at the end, to be sure that we're really ready when
 	 * connecting to the session manager */
 	panel_session_init ();
+
+    /*Load a css file from a path so the drag handle image can be loaded*/
+    error2 = NULL;
+    provider = gtk_css_provider_new ();
+	        gtk_css_provider_load_from_path (provider,
+				DATADIR "/mate-panel/" "mate-panel.css", &error2);
+
+    if (error2 != NULL) {
+        g_warning ("Can't parse mate-panel CSS custom description: %s\n", error2->message);
+        g_error_free (error2);
+    }
+    else {
+	    gtk_style_context_add_provider_for_screen (gdk_screen_get_default(),
+                                    GTK_STYLE_PROVIDER (provider),
+                                    GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
+    }
+     g_object_unref (provider);
 
 	gtk_main ();
 
--- a/data/Makefile.am
+++ b/data/Makefile.am
@@ -7,6 +7,11 @@ layout_DATA = \
 	opensuse.layout \
 	ubuntu.layout
 
+matepaneldir   = $(datadir)/mate-panel
+matepanel_DATA = \
+    mate-panel.css  \
+    panel-grid-symbolic.svg
+
 @INTLTOOL_XML_NOMERGE_RULE@
 
 gsettings_ENUM_NAMESPACE = org.mate.panel
--- a/data/panel-grid-symbolic.svg
+++ b/data/panel-grid-symbolic.svg
@@ -1,0 +1,173 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   sodipodi:docname="view-more-symbolic.svg"
+   height="22"
+   id="svg7384"
+   inkscape:version="0.91 r13725"
+   version="1.1"
+   width="12">
+  <metadata
+     id="metadata90">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title>Gnome Symbolic Icon Theme</dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <sodipodi:namedview
+     inkscape:bbox-nodes="false"
+     inkscape:bbox-paths="true"
+     bordercolor="#666666"
+     borderopacity="1"
+     inkscape:current-layer="layer12"
+     inkscape:cx="37.386562"
+     inkscape:cy="10"
+     gridtolerance="10"
+     inkscape:guide-bbox="true"
+     guidetolerance="10"
+     id="namedview88"
+     inkscape:object-nodes="false"
+     inkscape:object-paths="false"
+     objecttolerance="10"
+     pagecolor="#3a3b39"
+     inkscape:pageopacity="1"
+     inkscape:pageshadow="2"
+     showborder="true"
+     showgrid="false"
+     showguides="true"
+     inkscape:snap-bbox="true"
+     inkscape:snap-bbox-midpoints="false"
+     inkscape:snap-global="true"
+     inkscape:snap-grids="true"
+     inkscape:snap-nodes="true"
+     inkscape:snap-others="false"
+     inkscape:snap-to-guides="true"
+     inkscape:window-height="444"
+     inkscape:window-maximized="0"
+     inkscape:window-width="534"
+     inkscape:window-x="788"
+     inkscape:window-y="500"
+     inkscape:zoom="2.8284271">
+    <inkscape:grid
+       empspacing="2"
+       enabled="true"
+       id="grid4866"
+       originx="119.9998px"
+       originy="650px"
+       snapvisiblegridlinesonly="true"
+       spacingx="1px"
+       spacingy="1px"
+       type="xygrid"
+       visible="true" />
+    <inkscape:grid
+       color="#000000"
+       empcolor="#000000"
+       empopacity="0"
+       empspacing="4"
+       enabled="true"
+       id="grid5968"
+       opacity="0.1254902"
+       originx="119.9998px"
+       originy="650px"
+       snapvisiblegridlinesonly="true"
+       spacingx="0.5px"
+       spacingy="0.5px"
+       type="xygrid"
+       visible="true" />
+  </sodipodi:namedview>
+  <title
+     id="title9167">Gnome Symbolic Icon Theme</title>
+  <defs
+     id="defs7386" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer9"
+     inkscape:label="status"
+     style="display:inline"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer10"
+     inkscape:label="devices"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer11"
+     inkscape:label="apps"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer13"
+     inkscape:label="places"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer14"
+     inkscape:label="mimetypes"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer15"
+     inkscape:label="emblems"
+     style="display:inline"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="g71291"
+     inkscape:label="emotes"
+     style="display:inline"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="g4953"
+     inkscape:label="categories"
+     style="display:inline"
+     transform="translate(-121.0004,-861)" />
+  <g
+     inkscape:groupmode="layer"
+     id="layer12"
+     inkscape:label="actions"
+     style="display:inline"
+     transform="translate(-121.0004,-861)">
+    <rect
+       height="4"
+       id="rect20592"
+       rx="0.38461545"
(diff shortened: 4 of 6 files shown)


raveit65:
Let's hold this for a moment.
See my comment in PR for 1.18 branch.
https://github.com/mate-desktop/mate-panel/pull/781#issuecomment-382722379
I am not really sure about
*MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { }*

lukefromdc:
I just tested the switch to just using MatePanelAppletFrameDBus and got ugly results on in-process applets:

lukefromdc:
Force-pushed branch after fixing an indent error

lukefromdc:
Should I go ahead and merge this, having established the need for the double selector?

lukefromdc:
I did a retest with out of process applets, they looked the same as with in-process so no trouble on that front

raveit65:
Yeah, merge it.
I am guessing we need the same for 1.18 branch for gtk+-3.22/20, or not ?

monsta:
Hmm, dots now look properly aligned with Menta, TraditionalGreen, BlackMATE... but there are no dots/lines at all with Adwaita, Adwaita Dark, HighContrast, HighContrastInverse.

I'm using mate-themes from git master, so I have https://github.com/mate-desktop/mate-themes/commit/10886a79f01abf33f55de321b5ff96d9ada0bbd9 and https://github.com/mate-desktop/mate-themes/commit/cf0a970f26d30c5db7c0af25396eb6e2dc1924a0. I don't have any in-process applets. GTK+3 version is 3.22.30.

monsta:
Just tried git master once again to make sure... Both Adwaita themes have dashes, though not very good looking (same as on screenshots in https://github.com/mate-desktop/mate-panel/pull/781), and both HighContrast themes have large dots.

muktupavels:
Based on pull request title? Why not use symbolic image/icon for panel handle? And than use css to only control/adjust color?

Also why you don't use `background-size` to control gradient size? For example `background-size: 14px 22px` and then with `background-position` you can control where it appears... That why panel size would not change how it looks.

How often `mate_panel_applet_frame_dbus_change_background` is called? You every time create new css provider. You do understand that GTK does not automatically remove your previous providers, right? So if that function has called 100 times you have added 100 providers, possibly for different themes with same priority...

monsta:
TraditionalGreen (dots ok):
traditionalgreen

Adwaita:
adwaita

Adwaita Dark:
adwaita-dark

HighContrast:
highcontrast

HighContrastInverse:
highcontrastinverse

monsta:
> you have tested with mate-themes git master ?

Yes

@alexarnaud: can you please test it too?

alexarnaud:
Le 20/04/2018 à 12:29, monsta a écrit :
> Yes
> 
> @alexarnaud <https://github.com/alexarnaud>: can you please test it too?
> 

Yes, I can, the test you're requested to me is only on master, not 1.18, 
right?

Best regards,
Alex.

raveit65:
> (I don't have installed adwaita to test)

@sc0w 
Adwaita and adwaita-dark is always installed if gtk+ is installed.
m-a-p
Add a metacity theme to their index.them file to use a thumbnail. Eg.
/usr/share/themes/Adwaita-dark/index.theme
```
GtkTheme=Adwaita-dark
IconTheme=Adwaita
MetacityTheme=Adwaita-dark
CursorTheme=Adwaita
CursorSize=24
```

raveit65:
Ok, for gtk+ themes the Handle is there for na-tray applet but not for the windows-list-applet.
Sorry , i always looked at the na-tray applet for testing.

lukefromdc:
OK, I have one other choice and will open an alternate PR, removing the offending border and doing nothing else. I tried to load a GtkIcon but that cannot be done, it did not show up and I got runtime warnings that MatePanelAppletFrameDBus is a subclass of GtkBin and can only accept one widget, thus excluding any GtkImage etc.

The tray looks fine in my install. It is possible we may need to move the CSS provider to the initial start of the applet to prevent repeated calls as Albert said. I DID test using background-size, but that can clip the default MATE image as it is not removed in MATE themes. If the MATE themes cannot be adjusted to remove anything added in that CSSprovider (as the image itself already gets replaced), than this gets really complicated really fast.

lukefromdc:
Alternate PR https://github.com/mate-desktop/mate-panel/pull/784
fixes the image being cut in mate themes without changing anything else. Not as pretty in Adwaita, Adwaita-dark etc though no changes in the high contrast themes as they are special-cased. The mate themes now look right, and with the default image going all the way to the left there should never be a problem with the window list (though the window list looked the same as the tray in my previous PR as well).

lukefromdc:
@monsta , what are the A11y issues mentioned, and what change are you requesting? Also check out the alternate PR https://github.com/mate-desktop/mate-panel/pull/784 which is simpler but a bit ugly in Adwaita and Adwaita-Dark

lukefromdc:
Why not use symbolic image/icon for panel handle? Because you can't load two widgets into a GtkBin,of which MatePanelAppletFrameDBus is a subclass. Thus a rewrite of the whole widget and all of the themes would be needed to do something like that. What I need is a way to load an image file: easy enough if we could assume the image would unconditionally be in /usr/share/mate-panel, but that's now how build and install options work. I don't see any way in the world to get a path that contains a variable into a cssprovider  and loading a GtkImage does not work because of the GtkBin issue.

Meanwhile this is greatly complicated by competing needs: Raleigh and win32 do not use the @theme_bg_color and instead use @bg_color (etc) selectors so require one special case if the colors are to vary with themes. The high contrast themes require black or white if there is to in fact be high contrast. A border cuts the MATE themes. I have no idea in the world what the referenced A11y issue is, and thus no starting point to address it.

I could reduce to horizontal component of the image in this PR to replace the 3X3 grid(meant to help the vertical panel case) with just a 3 hole vertical image, maybe that would help some of the concerns raised here, or not?  It would shorten the code a bit, and vertical panel support in mate-themes will require a rework anyway for this item.

I know of exactly ONE simple approach: go all the way back to the original system,. where I only set a background color and nothing else. That uses either a solid color (if not transparent) or if an alpha is used the old GTK default mesh appearance. We lose the MATE 3 holes image entirely in GNOME themes but the handle stays visible, and we can still special-case the high contrast themes

muktupavels:
You can use symbolic icon / image with `background-image` and control its color with `color`...
You can create css provider once and reuse/change css in it when needed...
You can create new providers, but then remove old providers that are not needed...
You can get event/callback when theme changes...
You can use GResource to ship icon/image with/in mate-panel so it is always available...

I would special case only gtk+ builtin themes. Your own themes probably already works good, no? Then why complicate things and add css that could do bad things to existing themes? Other themes can fix their themes if they want support MATE.

lukefromdc:
As a test, I was able to load a .svg image file from an absolute path (cannot ship that way though) but was not able to recolor it. Even borrowing a symbolic icon and calling it grig-symbolic.svg it was always renderede in its original color

muktupavels:
```
background-image:-gtk-recolor(url('/usr/share/icons/Adwaita/scalable/actions/action-unavailable-symbolic.svg'));
color: #ff0000;
```

lukefromdc:
That works, and it turns out our existing panel-grid.svg can function as a symbolic image. Best of all, I was able to strip out all code for special-casing the high contrast themes, as they now get the foreground color. Works in all GNOME themes except in win32, where the image still appears but is very faint as @theme_fg_color is not used in win32. This will work with ANY theme that uses @theme_fg_color, and as the background is transparent does not write over the panel's existing background or need @theme_bg_color which is not present in Raleigh.

Here is the cssprovider used, it is much simpler and could be moved to startup:
```

    GtkCssProvider *provider;
    provider = gtk_css_provider_new ();
     {
            "MatePanelAppletFrameDBus > MatePanelAppletFrameDBus { \n"
            "border-style: none; \n"
            "background-repeat: no-repeat; \n"
            "background-position: left; \n"
            "color: @theme_fg_color;"
            "background-image: -gtk-recolor(url(\"/usr/share/mate-panel/grid-symbolic.svg\")) \n"
            "}",
            -1, NULL);
        }
	gtk_style_context_add_provider (gtk_widget_get_style_context(GTK_WIDGET(frame)),
                                    GTK_STYLE_PROVIDER (provider),
                                    GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
    g_object_unref (provider);

```

lukefromdc:
https://github.com/mate-desktop/mate-panel/tree/dev-applet-handle-wip
requires dropping  https://github.com/mate-desktop/mate-panel/files/1933891/panel-grid-symbolic.svg.gz into /usr/share/mate-panel to work (it is a WIP needing more work) but can be tested with that image file. What I now need is a way to install that image, either as a gresource or any other way that works now matter how mate-panel is configured as to installation directories

lukefromdc:
I finally got an image to work and install properly! I realized that a css file loaded from a path could refer to an image file by relative path, same as in a theme file. Putting mate-panel.css and panel-grid-symbolic into the data directory installed them to /usr/share/mate-panel with my build options, and they worked. I moved the CSS loading into mate_panel_applet_frame_dbus_init so it would be loaded once only.  MUCH simpler code and works better too.

On changing themes the color changes fine, as the same css file works with whatever colors the themes feed them. Only in the win32 theme is there an issue as that theme does not use @theme_fg_color,   so there the image still shows up, it just is not recolored and looks a lot like it does in Menta exept for being slightly lighter rather than slightly darker. Solid contrast in the high-contrast themes, as they set background and foreground colors opposite.

Now we get the mate-panel drag handle dots from the mate-themes on the drag handles in all the GNOME themes and any other themes not supporting the panel. For the mate-themes we can ship the image (it's the same file) for backward compatability or not, and if desired can recolor it to anything desired whether it is the default image or the image from the theme file, it should make no difference at all.

Rebased for the latest commit to master, please retest

raveit65:
https://git.gnome.org/browse/gnome-panel/log/?qt=grep&q=gresource
And there is a later commit where @muktupavels added the gnome-panel.css

raveit65:
Looks good with our themes.
But with adwaita or any other theme which doesn't support mate-panel.
windows-list-applet-handle

Changing mate-panel.css to 
*background-image: -gtk-scaled(url("panel-grid-symbolic.svg"))*
fixes the problem here.
windows-list-applet-handle-good
This is the same what we use in themes css.

PS: i will add the gresource stuff when all is done, if you agree.

raveit65:
More testing points out the *-gtk-recolor* term is needed. But we need to set up a fixed background-size for HIDPI.
```
MatePanelAppletFrameDBus > MatePanelAppletFrameDBus {
	border-style: none;
	background-repeat: no-repeat;
	background-position: left;
	color: @theme_fg_color;
	background-image: -gtk-recolor(url("panel-grid-symbolic.svg"));
	background-size: 12px 22px;
}
```
This works for me with all themes.
Eg. HCI theme.
hci

data/mate-panel.css
```diff
@@ -0,0 +1,9 @@
+
+MatePanelAppletFrameDBus > MatePanelAppletFrameDBus {
+	border-style: none;
+	background-repeat: no-repeat;
+	background-position: left;
+	color: @theme_fg_color;
+	background-image: -gtk-recolor(url("panel-grid-symbolic.svg"))
```
raveit65: small typo, missing semicolon.

lukefromdc:
I just fixed the mate-panel.css file, adding the missing semicolon and the fixed size. Works nicely here, force-pushed to avoid further complicating the PR

lukefromdc:
Also, switching this to gresource is fine, makes the code a bit lighter at runtime so a good idea

muktupavels:
I would add global css provider not create new provider for each applet with same css.

raveit65:
Can we also squash both commits  before merging?

lukefromdc:
OK, this is squashed

lukefromdc:
Should I go ahead and merge or are you adding the gresource conversion to the same PR,  for which we would need to wait

lukefromdc:
@monsta , please retest. Does this fix the issues you found?

lukefromdc:
Just fixed an indent error in mate-panel.css

lukefromdc:
I just switched the cssprovider over to loading once only in main.c for the whole panel, same as it does in Caja

raveit65:
I just tested it again and it works like expected.
I will merge it as it blocks my local repo with adding GResource to mat-panel.

mate-panel/main.c
```diff
@@ -60,7 +60,8 @@ main (int argc, char **argv)
 {
 	char           *desktopfile;
 	GOptionContext *context;
-	GError         *error;
+	GError         *error, *error2;
```
muktupavels: Why second GError?

monsta:
@lukefromdc: I mentioned a11y issues because dots were lost for both high contrast themes, therefore it became the same as if https://github.com/mate-desktop/mate-panel/pull/779 wasn't applied. Now it looks alright, I just built from master, and dots are shown in all themes.

alexarnaud:
@raveit65 Do you think we need more backports for 1.18 or 1.20 branches ?

Best regards,
Alex.

raveit65:
@alexarnaud 
already done for 1.20
backport and testing for 1.18 should be done by @monsta 
I don't update f26 with 1.18 any more as it is EOL in 1.5 months.

lukefromdc:
The second gerror was used because I was unsure whether re-using the first one would work or would be a potential source of problems. The css block comes after a g_error_free (error) statement for the first gerror variable.

muktupavels:
If `g_option_context_parse` fails, `mate-panel` exits, no? So how would you expect it to reach css part?

Even if it would not be fatal error and `mate-panel` would continue to run you would just have to reset it to `NULL` after freeing it.

But have you read documentation? You are supposed to connect to `parsing-error` signal to track errors in css loading.

monsta:
@raveit65: yeah, I'll backport it to 1.18, just need to limit it to GTK+ 3.20/22, same as done in https://github.com/mate-desktop/mate-panel/commit/3a8b1f66d2393a35f40dbcc211d3bf045d563a92.

raveit65:
By the way, should not the old malfunctioning handle code be removed?
https://github.com/mate-desktop/mate-panel/blob/master/mate-panel/panel-applet-frame.c#L100

alexarnaud:
@monsta could you please tell us here when you've pushed to 1.18? I'm waiting the backport to provide new mate-panel 1.18 to a visual-impaired user that uses HighContrast.

Thank you very much @lukefromdc @monsta @lukefromdc @muktupavels for your involvement in this area :) !

Best regards,
Alex.

monsta:
Sure, I'll post here.

monsta:
Backported to 1.18 in https://github.com/mate-desktop/mate-panel/commit/4437af31ff82629c18f53f6f4c38430547378af9.

lukefromdc:
It looks like the only thing that broke in the old code at and below
https://github.com/mate-desktop/mate-panel/blob/master/mate-panel/panel-applet-frame.c#L100
was gtk_render_handle, which will only display the old mesh background at least in GTK 3.22 if a background color is explicitly set. It is still needed because if it is removed, the new image won't display either. The rest of the code displays the background for the applet outside the handle, and removing it causes any theme that does set a bg color for the handle to render it over the entire applet and not just the handle. Thus it also cannot be removed.

Only the part within the 

```
	if (bg_pattern && (background->type == PANEL_BACK_IMAGE ||
	    (background->type == PANEL_BACK_COLOR && background->has_alpha)))
```

if block seems to have no effect and may be something we can remove. I will try to fix the custom background cases

lukefromdc:
On further experiment no success with getting the image into applet handles with a custom background, I think .mate-custom-panel-background is overriding it, and specifying MatePanelAppletFrameDBus.mate-custom-panel-background as well in the css file (the experimental single should cover the double selector case) does nothing. Probably the removal of backgrounds elsewhere to allow custom backgrounds to show is still catching it.

muktupavels:
>  It is still needed because if it is removed, the new image won't display either. 

Why not look what gtk is actually doing in `gtk_render_handle`? If you would have done that then you probably would have find that you should be able to replace it with `gtk_render_background` and `gtk_render_frame`.

lukefromdc:
gtk_render_handle is not actually deprecated, not sure why the original patterns stopped rendering without an explicit background color set, but I really don't have time to dig into that now that we no longer use it. No mate-theme and no GNOME theme sets a transparent or shaded background for the handle, the condition I've found makes it work. Themers can still do that if they want, and can override and remove the new default image. Thus, this is not broken, the old behavior is just undocumented..

Meanwhile, we've got a lot of other stuff to work on

PR #784, opened 2018-04-20
Applet-handle: remove border as themes don't override it

lukefromdc:
Remove border from default applet drag handle, it clips image used in mate-themes Simpler alternate to https://github.com/mate-desktop/mate-panel/pull/782 but not as pretty in GNOME themes


lukefromdc:
Not as pretty in Adwaita but does work, and does not clip the images used in mate-themes

lukefromdc:
Closing as a far better fix is in the works

Issue #786, opened 2018-04-21
Hide buttons on vertical panel are huge with mate-panel-1.20.1-1.fc27.x86_64


william8000:
#### Expected behaviour

Hide buttons as in earlier versions that had a limited depth.

#### Actual behaviour

The hide buttons are as deep as they are wide.
I have a vertical panel, and it is decently wide (around 150 pixels) to have space for the time and date to show without wrapping.
My laptop has a FullHD 1920x1080 display, so I can give the panel a lot of horizontal space.
The top and bottom hide buttons are now 150 pixels deep, so over 1/4 of my panel is hide buttons.

#### Steps to reproduce the behaviour

Make a vertical panel 150 pixels wide with hide buttons enabled.

#### MATE general version

1.20.1

#### Package version

mate-panel-1.20.1-1.fc27.x86_64

#### Linux Distribution

Fedora 27 x86_64

#### Link to downstream report of your Distribution

Is this caused by or fixed by https://github.com/mate-desktop/mate-panel/commit/fa5cde36593be06eabfb2f752f38c26b80ea269c ?

[Screen capture](https://plus.google.com/u/0/109904539145143933071/posts/7tCVjcaiMHi)

raveit65:
Confirmed. I am using also a vertical panel for sensor-applet with auto-hide feature. So i never noticed that.
The vertical size of the button for a vertical panel should be reduced.

raveit65:
Hmm, gtk3-inspector shows me that the button use the same height as panel size.
Eg. panel size is 70 that height-request is 70px for vertical panel.
In gtk3-inspector i am been able to limit height-request to 30, than it looks OK.

raveit65:
@info-cppsp 
Can you please take a look at this?

raveit65:
done.

lukefromdc:
Confirmed here too

william8000:
Thanks!

monsta:
In other words, these buttons are always square now. It's so on a horizontal panel too.

info-cppsp:
monsta is right. If you have a small panel, like I do, the other way around looks stupid. like 24 px high, but only 5 px wide... (it was like that at some point of developing the PR, that took weeks...)
let me think about this. probably the best way is going to be a gsetting for hide button width / height.
I would imagine that not many people are using 150 px wide panels, so let's not set things like that was the most important use case.

raveit65:
[...]

PR #787, opened 2018-04-21
limit size of panel-hide-button

raveit65:
fix for https://github.com/mate-desktop/mate-panel/issues/786

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -1287,9 +1287,9 @@ static void panel_toplevel_update_hide_buttons_size(GtkWidget* button, int panel
 	} else if (panel_size < 40) {
 		gtk_image_set_pixel_size (GTK_IMAGE (arrow), 16);
 	} else if (panel_size < 60) {
-		gtk_image_set_pixel_size (GTK_IMAGE (arrow), 24);
+		gtk_image_set_pixel_size (GTK_IMAGE (arrow), 22);
 	} else {
-		gtk_image_set_pixel_size (GTK_IMAGE (arrow), 32);
+		gtk_image_set_pixel_size (GTK_IMAGE (arrow), 24);
 	}
 
 }
@@ -1298,6 +1298,7 @@ static void panel_toplevel_update_hide_buttons(PanelToplevel* toplevel)
 {
 
 	int panel_size = toplevel->priv->size;
+	int hb_size = 0;
 
 	if (toplevel->priv->buttons_enabled) {
 		panel_toplevel_update_buttons_showing (toplevel);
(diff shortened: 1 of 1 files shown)


raveit65:
> ....should we address that too?

If you want to address this than open another report which can be fix by another PR.

info-cppsp:
Please don't get this exited, just because a random user reports something.
My PR works for the majority of users.
Please try to think of mate as a complex system and try not to add features just because someone asks for it. As far as I am concerned this isn't a bloatware factory.

Apart from that, I see the problem with the use case.
Let me think about it.
Share your ideas.

Hardcoding something that works, is still just hardcoding. - Which is bad...
What if the next guy comes along and says that the button should be even smaller?
Or bigger / wider, because it looks stupid like this?
What if someone likes squares, even if they are 150*150px ?

raveit65:
Why should someone wants a squared button of 150x150px?

raveit65:
I just updated PR.
- same size of buttons with or without using the arrow
- much more simplify
- size values are updated for both arrow and button to match more previous behaviour before the square button was commited.

@monsta @lukefromdc 
Please test again.

monsta:
> same size of buttons with or without using the arrow

Not exactly, but I just pushed a small adjustment for large panel size :slightly_smiling_face:

monsta:
Looks ok, it's now close to behavior before https://github.com/mate-desktop/mate-panel/commit/fa5cde36593be06eabfb2f752f38c26b80ea269c.

monsta:
Interesting, before https://github.com/mate-desktop/mate-panel/commit/fa5cde36593be06eabfb2f752f38c26b80ea269c toplevel had `arrow-size` property which was supposed to define arrow size. But that property was only read in two places, nobody ever set it, except for the initial (default) value. :slightly_smiling_face: 

```
#define DEFAULT_ARROW_SIZE        20
```
Actually, the property is still there, just not used anymore at all.

raveit65:
Hmm, for me the button looks to big with using
> 40 = 26
> 60 = 30
with a HIDIPI monitor.
But i don't really care as i use auto-hide panel in favor of this button ;)

lukefromdc:
Works fine on a retest, and gives same size hide buttons with or without the arrow shown. Yes, if I set window-scaling=2 they render wider, remembering that on a real HiDIP display the close pixel spacing would presumably narrow that back down again

raveit65:
Ok, than HIDIPI users needs to live with it. I will squash commits.

raveit65:
@monsta Is this ready to go and you want approve the PR?

monsta:
Sorry, I didn't think about HiDPI (as usual). Is the button much bigger now? It should be just a few more pixels, and as @lukefromdc mentioned, it should be less visible difference on a real HiDPI monitor.

From my side, this looks good to go.

raveit65:
@monsta 
vert-panel with 70px (your values)
vert-panel-70px-to-large     previous version  vert-panel-70px

well, not a real problem for me, but it looks to huge for my taste.

lukefromdc:
If we shrink the hide buttons for the hidpi case and don't use a selector switch, they will become tiny when the arrows (which should force a minimum size) are not shown on a non-hidpi monitor. Don't judge by the ballooned appearance of testing window-scaling=2 on a small monitor like what I have.

raveit65:
Can someone post a screenshot how a 70px vert. panel looks with buttons and this PR, please?

lukefromdc:
Here are some screenshots, I used GtkInspector to change the theme to Adwaita-Dark to put borders on the buttons, and to quickly toggle the window scaling. Some icon names on the desktop are redacted.

70px, window-
left_panel_24px2_adwaita_dark
scaling=1
left_panel_70px_adwaita_dark
70px., window-scaling=2
left_panel_24px2_adwaita_dark

24px, window-scaling=2

raveit65:
any way, i will leave it like it is.

PR #789, opened 2018-04-22
applet.c: avoid deprecated GtkImageMenuItem

sc0w:
avoid deprecated:

gtk_image_menu_item_new_with_mnemonic
gtk_image_menu_item_set_image

finalprapplet

**EDIT:** this PR needs to work with https://github.com/mate-desktop/mate-themes/pull/219

### DIFF
--- a/mate-panel/applet.c
+++ b/mate-panel/applet.c
@@ -16,6 +16,7 @@
 #include <gio/gio.h>
 
 #include <libpanel-util/panel-show.h>
+#include <libpanel-util/panel-gtk.h>
 
 #include "button-widget.h"
 #include "drawer.h"
@@ -40,6 +41,8 @@
 static GSList *registered_applets = NULL;
 static GSList *queued_position_saves = NULL;
 static guint   queued_position_source = 0;
+
+static GtkCheckMenuItem *checkbox_id = NULL;
 
 static void applet_menu_show (GtkWidget *w, AppletInfo *info);
 static void applet_menu_deactivate (GtkWidget *w, AppletInfo *info);
@@ -98,14 +101,21 @@ mate_panel_applet_toggle_locked (AppletInfo *info)
 }
 
 static void
-mate_panel_applet_lock (GtkCheckMenuItem *menuitem,
-		   AppletInfo       *info)
+checkbox_status (GtkCheckMenuItem *menuitem,
+		 AppletInfo       *info)
+{
+	checkbox_id = GTK_CHECK_MENU_ITEM (menuitem);
+}
+
+static void
+mate_panel_applet_lock (GtkMenuItem *menuitem,
+			AppletInfo  *info)
 {
 	gboolean locked;
 
 	locked = mate_panel_applet_toggle_locked (info);
 
-	gtk_check_menu_item_set_active (menuitem, locked);
+	gtk_check_menu_item_set_active (checkbox_id, locked);
 
 	if (info->move_item)
 		gtk_widget_set_sensitive (info->move_item, !locked);
@@ -382,15 +392,8 @@ setup_an_item (AppletUserMenu *menu,
 	       GtkWidget      *submenu,
 	       int             is_submenu)
 {
-	GtkWidget *image = NULL;
+	menu->menuitem = panel_image_menu_item_new_from_gicon (menu->gicon, menu->text);
 
-	menu->menuitem = gtk_image_menu_item_new_with_mnemonic (menu->text);
-	if (menu->gicon) {
-		image = gtk_image_new_from_gicon (menu->gicon,
-						  GTK_ICON_SIZE_MENU);
-		gtk_image_menu_item_set_image (GTK_IMAGE_MENU_ITEM (menu->menuitem),
-					       image);
-	}
 	gtk_widget_show (menu->menuitem);
 
 	g_signal_connect (G_OBJECT (menu->menuitem), "destroy",
@@ -503,6 +506,8 @@ mate_panel_applet_create_menu (AppletInfo *info)
 
 	menu = g_object_ref_sink (gtk_menu_new ());
 
+	gtk_menu_set_reserve_toggle_size (GTK_MENU (menu), FALSE);
+
 	/* connect the show & deactivate signal, so that we can "disallow" and
 	 * "re-allow" autohide when the menu is shown/deactivated.
 	 */
@@ -524,7 +529,6 @@ mate_panel_applet_create_menu (AppletInfo *info)
 	}
 
 	if (!panel_lockdown_get_locked_down ()) {
-		GtkWidget *image;
 		gboolean   locked;
 		gboolean   lockable;
 		gboolean   movable;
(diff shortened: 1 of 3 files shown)


mate-panel/applet.c
```diff
@@ -98,14 +101,21 @@ mate_panel_applet_toggle_locked (AppletInfo *info)
...
+static void
+mate_panel_applet_lock (GtkMenuItem *menuitem,
+			AppletInfo  *info)
 {
 	gboolean locked;
 
 	locked = mate_panel_applet_toggle_locked (info);
 
-	gtk_check_menu_item_set_active (menuitem, locked);
+	gtk_check_menu_item_set_active (checkbox_id, locked);
```
muktupavels: What are you trying to do here?

mate-panel/libpanel-util/panel-gtk.c
```diff
@@ -164,22 +164,65 @@ GtkWidget *
...
+panel_check_menu_item_new_with_mnemonic (GtkWidget   *widget_check,
+					 const gchar *label_name)
+{
+	GtkWidget *box = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 6);
+	GtkWidget *menuitem = gtk_menu_item_new ();
+ 
+	gtk_widget_set_margin_start (widget_check, 7);
+	gtk_widget_set_margin_end (widget_check, 1);
+
+	gtk_container_add (GTK_CONTAINER (box), widget_check);
```
muktupavels: GtkCheckMenuItem as child to GtkMenuItem? This feels wrong...

sc0w:
@muktupavels 

> What are you trying to do here?

> GtkCheckMenuItem as child to GtkMenuItem? This feels wrong...

Yes, I did it intentionally, with this PR, the checkbox works like a GtkMenuItem, and I did some changes to work the checkbox properly

This was needed to see the items with a good look without the deprecated GtkImageMenuItem

muktupavels:
GtkCheckMenuItem is GtkMenuItem, so I don't get why you need to put menu item inside menu item. How deprecated GtkImageMenuItem affects GtkCheckMenuItem?

sc0w:
@muktupavels 

If I use the GtkCheckMenuItem without changes, it affects the look of the other items with extra space in the left

You can test it by yourself :)

muktupavels:
No, I can not... I don't use MATE!

I think the problem is not in GtkCheckMenuItem, but in your replacement to GtkImageMenuItem. The problem is that GtkImageMenuItem is not same as GtkMenuItem with box + image & label...

GTK documentation says:

> This is **functionally** equivalent to:

Functionality is same, but visually it is not, because image is not placed in "reserved" area... And since you already use `gtk_menu_set_reserve_toggle_size` I guess you know that. Still the way you done it looks wrong to me. What will you do when you have multiple check items? Are you going to connect to more map signals? Add more static variables?

https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-set-reserve-toggle-size

> Sets whether the menu should reserve space for drawing toggles or icons, regardless of their actual presence.

Since you add check menu item this space is used even if you set it to false.

Thinking little bit more, I have question if this really works as you expect? what happens if you click on checkbox? Now you have two items with activate signal, you might have problem that click does not go to real menu item...

sc0w:
please, feel free to open an alternative PR if you can done better

raveit65:
This works and the icons and checkbox are there in menu.
Adding a little gap of some pixels left from the checkbox would be the greatest ;)
Than checkbox position would be the same as other icons.
applet-menu
But it is OK if this isn't possible.


Well,  coding skills aren't good enough evaluate if using 'GtkCheckMenuItem as child to GtkMenuItem' is fine.
....up to you to decide that.

sc0w:
@raveit65 yes, confirmed

In TOK it looks exactly like before (see capture in my first comment)

I don't know why, but seems this doesn't work in other themes

I handled the margin here:

https://github.com/mate-desktop/mate-panel/pull/789/files#diff-975106c17b2926072a451e6cab9edd70R217

I will investigate it

sc0w:
now, with latest changes, in combination with https://github.com/mate-desktop/mate-themes/pull/219 I hope all is fine

@lukefromdc @raveit65 please test again

lukefromdc:
With this panel PR and  BlackMATE with the PR applied, in Adwaita, in my UbuntuStudio_Legacy theme the bottom item (lock to panel) was very slightly to the right, though many may not notice that. With the panel PR only and in Menta, that checkbox and the label after it were rendered slightly left of all others. 

blackmate_pr_and_panel_pr

lukefromdc:
Screwed up with my comment about Menta: I only applied the panel PR manually to BlackMATE, giving the results you showed. Menta was to the left because the commit for it had not been applied, nothing more.

raveit65:
Why not adding this css settings to our new mate-panel.css, than it fixes all themes?
Or is it not possible to use one setting for all themes?
```
/* checkboxes in applet popup menu */
.mate-panel-menu-bar menu > menuitem > box > menuitem {
    padding: 0px 0px 0px 4px;
}
```
I mean only BlackMate and GreenLagunga use a bit different setting.
`padding: 0px 0px 0px 3px;`

raveit65:
Hmm, adding this to mate-panel.css doesn`t have any effect.
Ok, we have to do this in themes.
But the position of the text of this entry looks really wrong with Adwaita or HighContrast themes.

muktupavels:
Do you really want to trade deprecated warning for theme / visual problems?

It is just warning, you can remove/silence it with `G_GNUC_BEGIN_IGNORE_DEPRECATIONS` and `G_GNUC_END_IGNORE_DEPRECATIONS`. GtkImageMenuItem wont be removed from gtk 3...

Otherwise I think that better option is to create your own version of image menu item that works very similar if not same as original gtk image menu item...

Also did you tried to click on checkbox? Does event go to real menu item and applet is really locked?

raveit65:
> Also did you tried to click on checkbox? Does event go to real menu item and applet is really locked?

Of course, out of any question ....this works well.

raveit65:
I use latest changes from PR and the HighContrast themes from gtk+, not the old HCI theme from previous mate-themes releases. 
launcher-hci-theme
launcher-hc-theme

I tested i with non HIDPI resolution and windows-scaling=1 too and i got same result.

I do not care much about Adwaita as it is a gnome theme and their devs stated very often that is isn't their problem how their themes and applications looks in other desktop.
So why we should care about?

But it should look OK in HighContrast themes as we decided to use them.
I am still wondering what they do different in their themes for menuitems?

raveit65:
For got my previous post, the patch in RPM wasn't updated with your latest version.
Now it looks fine wit Adwaita and HighContrast themes :)
With Adwaita
launcher-adwaita-theme
It seems the width of the menu is smaller now?

sc0w:
With adwaita the alignment is wrong, and the menuitem is bigger (you can compare it with mate-system-monitor popup menu)

It can be fixed with:

```
.mate-panel-menu-bar menu > menuitem > box > menuitem {
    padding: 0px 0px 0px 4px;
}
```

but I don't know how to add this setting to mate-themes for adwaita

> It seems the width of the menu is smaller now?

yes, I think more width is useless

raveit65:
See my screenshot, for me it looks ok with adwaita

> but I don't know how to add this setting to mate-themes for adwaita

????
We can't setup other themes.

sc0w:
> See my screenshot, for me it looks ok with adwaita

yes, apparently is ok, but it is misaligned, it is moved 2 pixels on the right

anyway, I never use adwaita, if it is ok for you, it is ok for me :)

PR #790, opened 2018-04-23
Gresource

raveit65:
I will add mate-panel.css later to GResource. 
This needs to be add first before other PRs will break it.
Please test


lukefromdc:
Window-menu works fine too

lukefromdc:
Panel-properties dialog is good

lukefromdc:
mate-panel-test-applets works fine except for a segfault on testing the notification area which is also present in master and not related to this PR

raveit65:
I did a approach for mate-panel.css but it doesn#t work for the moment.
https://paste.fedoraproject.org/paste/TsQrglR5c~fi0oeE~NVHEw
No idea what's wrong there....

data/theme/Makefile.am
```diff
@@ -0,0 +1,8 @@
+NULL =
+
+EXTRA_DIST = \
+    mate-panel.css  \
+    panel-grid-symbolic.svg \
+	$(NULL)
```
muktupavels: spaces vs. tabs...

mate-panel/main.c
```diff
@@ -167,22 +171,17 @@ main (int argc, char **argv)
...
-	    gtk_style_context_add_provider_for_screen (gdk_screen_get_default(),
-                                    GTK_STYLE_PROVIDER (provider),
-                                    GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
-    }
-     g_object_unref (provider);
+    /*Load a css file from a GResource so the drag handle image can be loaded*/
+	screen = gdk_screen_get_default ();
+	css = gtk_css_provider_new ();
+	provider = GTK_STYLE_PROVIDER (css);
+	resource = g_strdup_printf ("/org/mate/panel/theme/mate-panel.css");
```
muktupavels: why `g_strdup_printf`? you can use const char to avoid copying string and freeing it.

mate-panel/main.c
```diff
@@ -167,22 +171,17 @@ main (int argc, char **argv)
 	 * connecting to the session manager */
 	panel_session_init ();
 
-    /*Load a css file from a path so the drag handle image can be loaded*/
-    error2 = NULL;
-    provider = gtk_css_provider_new ();
-	        gtk_css_provider_load_from_path (provider,
-				DATADIR "/mate-panel/" "mate-panel.css", &error2);
-
-    if (error2 != NULL) {
```
muktupavels: error handling removed!? is that intentional? you can use `parsing-error` signal if you want get errors when loading css.

data/theme/Makefile.am
raveit65: fixed.

lukefromdc:
Newest version works fine here, and the default applet drag handle theme works as before

lukefromdc:
I had added the error-handling code taken from Caja, but in practice an error would require that the css file be corrupted in some way or that the install fail in some way.

raveit65:
Hmm, if i use
resource = "/org/mate/panel/theme/mate-panel.css";
than the panel doesn't load any more.
So what is wrong with using  `g_strdup_printf` ?
I took it from https://git.gnome.org/browse/gnome-panel/tree/gnome-panel/main.c#n76

raveit65:
No idea to  use  `parsing-error` here....
Help is welcome.

muktupavels:
I use `g_strdup_printf` because in my case string includes theme name as minimum... In your case you have const string, so you should use it as const char. Change `gchar  *resource;` to `const gchar *resource;`,  `resource = g_strdup_printf ("/org/mate/panel/theme/mate-panel.css");` to `resource = "/org/mate/panel/theme/mate-panel.css";` and remove `g_free (resource);`.

mate-panel/main.c
```diff
@@ -167,22 +171,15 @@ main (int argc, char **argv)
...
-        g_warning ("Can't parse mate-panel CSS custom description: %s\n", error2->message);
-        g_error_free (error2);
-    }
-    else {
-	    gtk_style_context_add_provider_for_screen (gdk_screen_get_default(),
-                                    GTK_STYLE_PROVIDER (provider),
-                                    GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
-    }
-     g_object_unref (provider);
```
muktupavels: You still need to unref provider.

mate-panel/main.c
```diff
@@ -167,22 +171,15 @@ main (int argc, char **argv)
...
+    /*Load a css file from a GResource so the drag handle image can be loaded*/
```
muktupavels: spaces/tabs; space after `/*` and before `*/`

raveit65:
Just updated.

lukefromdc:
Just tested again, and the drag handle theming again works fine.

mate-panel/main.c
monsta: I see that error handling isn't present after every `gtk_builder_add_from_resource` as well...

libmate-panel-applet/mate-panel-applet.c
```diff
@@ -711,6 +711,46 @@ mate_panel_applet_setup_menu_from_file (MatePanelApplet    *applet,
...
+ *
+ * Sets up the context menu of @applet. @filename is a resource path to a menu
+ * XML file, containing a #GtkUIManager UI definition that describes how to
+ * display the menu items. @action_group contains the various #GtkAction that
+ * are referenced in @xml.
+ *
+ * See also the <link linkend="getting-started.context-menu">Context
+ * Menu</link> section.
+ *
+ * Since: 3.4
```
monsta: This is not gnome-panel, we don't have 3.4 :slightly_smiling_face:

mate-panel/main.c
raveit65: I see that too, but is that needed in our case?, read comment from @lukefromdc

raveit65: and please read that comment from me.
https://github.com/mate-desktop/mate-panel/pull/790#issuecomment-384038849
> No idea to use parsing-error here....
> Help is welcome.

libmate-panel-applet/mate-panel-applet.c
raveit65: fixed

raveit65:
new rebase

muktupavels:
```
static void
parsing_error_cb (GtkCssProvider *provider,
                  GtkCssSection  *section,
                  GError         *error,
                  gpointer        user_data)
{
  /* do something with error... */
}

g_signal_connect (provider, "parsing-error", G_CALLBACK (parsing_error_cb), NULL);
```

For more info:
https://developer.gnome.org/gtk3/stable/GtkCssProvider.html#GtkCssProvider-parsing-error

lukefromdc:
Is this ready to go?

lukefromdc:
I just pushed Albert's css error handling code as 
https://github.com/mate-desktop/mate-panel/pull/790/commits/83e97512f519be23c0d5036b5dfa2e082d214b73

It works well in a test here. Here's what I got with an intentionally corrupted css file for testing purposes:

A deliberately bad image file name gives this warning:

Gtk-WARNING **: 00:00:17.725: Could not load a pixbuf from /org/mate/panel/theme/panel-grid-symbolic.svs.
This may indicate that pixbuf loaders or the mime database could not be found.

A deliberately Missing semicolon after the image line gives this warning:

** (mate-panel:25559): WARNING **: 00:01:26.713: Can't parse mate-panel's CSS custom description: Junk at end of value for background-image

With valid CSS file and image panel starts up silently, with no text in terminal

raveit65:
This works great.
```
** (mate-panel:22942): WARNING **: 11:49:45.359: Can't parse mate-panel's CSS custom description: Junk at end of value for border-style
```
thanks @lukefromdc

mate-panel/main.c
monsta: So, what about error handling after `gtk_builder_add_from_resource`? Or it's not supposed to fail with GResource...?

mate-panel/panel.gresource.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<gresources>
+  <gresource prefix="/org/mate/panel">
+    <file compressed="true">panel-properties-dialog.ui</file>
+    <file compressed="true">panel-run-dialog.ui</file>
+    <file alias="anim/wanda.png">../applets/fish/wanda.png</file>
```
monsta: I think this one belongs to fish applet's resource. It's actually still installed to `/usr/share/mate-panel/pixmaps/` along with other images from that applet.

raveit65:
Why you didn't say that yesterday?
Now it is merge......feel free to edit it....

mate-panel/main.c
muktupavels: It can fail, but check documentation:
https://developer.gnome.org/gtk3/stable/GtkBuilder.html#gtk-builder-add-from-resource

> It’s not really reasonable to attempt to handle failures of this call. The only reasonable thing to do when an error is detected is to call g_error().

lukefromdc:
One thing I would NOT do here is call g_error() , as I would not want to stop the  panel load entirely due to the inability to load the applet drag handle visibiity image. A panel with invisible tray and window list drag handles is a lot more usable than a panel that crashes on startup.

The warning message is intended to let future developers know if the css file isn't working or the image cannot be loaded, so they can fix it.  The warning gives pretty much the same information you would get starting a program in terminal with a problem in the theme's css files, and can thus be used for future debugging. Much harder to deal with a silent failure, especially if at some point in the future we are doing a lot more with the css as is the case in Caja.

muktupavels:
No, it should be exactly what you would want to do... If you have problem with resources you want to see that as soon as possible and avoid pushing such code to master. For example if .ui file is missing you would probably get crash anyway, so why not abort when trying to load it?

CSS is different thing, warning can help avoid committing invalid css... Anyway both are developer errors so these are to help you not user. You should not push code with errors and expect other devs to fix your errors.

lukefromdc:
Install errors at the distro level are also possible, and if a distro screws up and pushed a bad package (I've seen this plenty of times), better to be robust and warn not crash. This is also consistant with what Caja does. There is a middle ground between ignoring an error entirely and crashing the whole program over it. Anyway, I don't see us deciding to crash the panel over this.

raveit65:
I don't want to stop loading the panel only because of a css error,
Keep in mind that you need to do wrong things in the css  before you compile the panel.
After installation the css code is in binary of the panel. Nobody can edit the file at this point.
At least we're responsible to ship the code with a correct css file.
So i don't think a user will ever see this warning.
If a user get this warning than our tarball is broken.

muktupavels:
g_error thing was for `gtk_builder_add_from_resource` not for css... `gtk_css_provider_load_from_resource` even does not have `GError` parameter.

mate-panel/panel.gresource.xml
raveit65: This image was added here in g-p code 
https://git.gnome.org/browse/gnome-panel/diff/gnome-panel/nothing.cP?h=gnome-3-4&id=9fca67f7bf2db30838ba6dcf86d49ca965ed31e5
```
+ resource = g_strdup_printf ("%s%cand%c.png", "/org/gnome/panel/anim/", 'w', 'a');
```
as replacement for
```
- return g_strdup_printf ("%s/%cand%c.png", ICONDIR, 'w', 'a');
```
But nothing.cP with this code line was removed by @monsta here
http://git.mate-desktop.org/mate-panel/diff/mate-panel/nothing.cP?h=1.18&id=fbd0f1d8a815347beea1b2c5df5804c88aa406c6
No idea why you did that and where you  have moved the code for icon image?
So, it looks like we can remove this icon source from gresource file, as it does nothing.

raveit65:
especially for @monsta 
https://github.com/mate-desktop/mate-panel/pull/792

mate-panel/main.c
monsta: I see, the same is told about `gtk_builder_add_from_file` which was in the code before. Not sure if showing a warning can be considered as error handling... There was a return after that though, without any attempts to continue working with possibly broken widgets.

mate-panel/panel.gresource.xml
monsta: Because that was ancient easter egg which wasn't even enabled for GTK+3, and nobody ported it. But I see that `wanda.png` is still used in `org.mate.panel.applet.fish` schema.

monsta: I missed this before PR was merged.

raveit65: history looks different here....
Let's say you missed it at the same time i pushed the button to merge it :)

PR #792, opened 2018-04-26
fix the fish

raveit65:


monsta:
So not all fish files were moved to GResource... only menu stuff was moved there. How do we decide what to move there and what to leave in `/usr/share`? For example, clock applet now has everything in GResource, menu stuff and pixmaps.

monsta:
Ah, I see. Fish pixmaps should be in a directory because you can select another animation in the prefs dialog, and that invokes a file opening dialog.

PR #794, opened 2018-04-27
support mate-menus api change

yetist:
Please review and see this [PR](https://github.com/mate-desktop/mate-menus/pull/59).

### DIFF
--- a/mate-panel/drawer.c
+++ b/mate-panel/drawer.c
@@ -568,7 +568,7 @@ load_drawer_applet (char          *toplevel_id,
 
 static void
 panel_drawer_prepare (const char  *drawer_id,
-                      const char  *custom_icon,
+                      GIcon       *custom_icon,
                       gboolean     use_custom_icon,
                       const char  *tooltip,
                       char       **attached_toplevel_id)
@@ -587,7 +587,10 @@ panel_drawer_prepare (const char  *drawer_id,
     g_settings_set_boolean (settings, PANEL_OBJECT_USE_CUSTOM_ICON_KEY, use_custom_icon);
 
     if (custom_icon) {
-        g_settings_set_string (settings, PANEL_OBJECT_CUSTOM_ICON_KEY, custom_icon);
+	gchar *icon_name;
+	icon_name = g_icon_to_string(custom_icon);
+	g_settings_set_string (settings, PANEL_OBJECT_CUSTOM_ICON_KEY, icon_name);
+	g_free(icon_name);
     }
 
     if (attached_toplevel_id) {
@@ -618,7 +621,7 @@ panel_drawer_prepare (const char  *drawer_id,
 void
 panel_drawer_create (PanelToplevel *toplevel,
                      int            position,
-                     const char    *custom_icon,
+                     GIcon         *custom_icon,
                      gboolean       use_custom_icon,
                      const char    *tooltip)
 {
@@ -636,7 +639,7 @@ panel_drawer_create (PanelToplevel *toplevel,
 char *
 panel_drawer_create_with_id (const char    *toplevel_id,
                              int            position,
-                             const char    *custom_icon,
+                             GIcon         *custom_icon,
                              gboolean       use_custom_icon,
                              const char    *tooltip)
 {
--- a/mate-panel/menu.c
+++ b/mate-panel/menu.c
@@ -265,28 +265,28 @@ static void
 add_drawers_from_alias (MateMenuTreeAlias *alias,
 			const char     *toplevel_id)
 {
-	MateMenuTreeItem *aliased_item;
+	gpointer item;
 
-	aliased_item = matemenu_tree_alias_get_item (alias);
-
-	switch (matemenu_tree_item_get_type (aliased_item)) {
+	switch (matemenu_tree_alias_get_aliased_item_type (alias)) {
 	case MATEMENU_TREE_ITEM_DIRECTORY:
-		add_drawers_from_dir (MATEMENU_TREE_DIRECTORY (aliased_item),
+		item = matemenu_tree_alias_get_directory (alias);
+		add_drawers_from_dir (item,
 				      G_MAXINT/2,
 				      toplevel_id);
+		matemenu_tree_item_unref (item);
 		break;
 
 	case MATEMENU_TREE_ITEM_ENTRY:
+		item = matemenu_tree_alias_get_aliased_entry (alias);
 		panel_launcher_create_with_id (toplevel_id,
 					       G_MAXINT/2,
-					       matemenu_tree_entry_get_desktop_file_path (MATEMENU_TREE_ENTRY (aliased_item)));
+					       matemenu_tree_entry_get_desktop_file_path (item));
+		matemenu_tree_item_unref (item);
 		break;
 
 	default:
 		break;
 	}
-
-	matemenu_tree_item_unref (aliased_item);
 }
 
 static void
@@ -295,10 +295,10 @@ add_drawers_from_dir (MateMenuTreeDirectory *directory,
 		      const char         *toplevel_id)
 {
 	const char *name;
-	const char *icon;
-	GSList     *items;
-	GSList     *l;
+	GIcon *icon;
 	char       *attached_toplevel_id;
+	MateMenuTreeIter *iter;
+	MateMenuTreeItemType type;
 
 	name = matemenu_tree_directory_get_name (directory);
 	icon = matemenu_tree_directory_get_icon (directory);
@@ -311,36 +311,36 @@ add_drawers_from_dir (MateMenuTreeDirectory *directory,
 	if (!attached_toplevel_id)
 		return;
 
-	items = matemenu_tree_directory_get_contents (directory);
-	for (l = items; l; l = l->next) {
-		MateMenuTreeItem *item = l->data;
+	iter = matemenu_tree_directory_iter (directory);
+	while ((type = matemenu_tree_iter_next (iter)) != MATEMENU_TREE_ITEM_INVALID) {
+		gpointer item;
+		switch (type) {
+			case MATEMENU_TREE_ITEM_ENTRY:
+				item = matemenu_tree_iter_get_entry (iter);
+				panel_launcher_create_with_id (attached_toplevel_id,
+						G_MAXINT/2,
+						matemenu_tree_entry_get_desktop_file_path (item));
+				matemenu_tree_item_unref (item);
+				break;
 
-		switch (matemenu_tree_item_get_type (item)) {
-		case MATEMENU_TREE_ITEM_ENTRY:
-			panel_launcher_create_with_id (attached_toplevel_id,
-						       G_MAXINT/2,
-						       matemenu_tree_entry_get_desktop_file_path (MATEMENU_TREE_ENTRY (item)));
-			break;
+			case MATEMENU_TREE_ITEM_DIRECTORY:
+				item = matemenu_tree_iter_get_directory (iter);
+				add_drawers_from_dir (item,
+						G_MAXINT/2,
+						attached_toplevel_id);
+				matemenu_tree_item_unref (item);
+				break;
 
-		case MATEMENU_TREE_ITEM_DIRECTORY:
-			add_drawers_from_dir (MATEMENU_TREE_DIRECTORY (item),
-					      G_MAXINT/2,
-					      attached_toplevel_id);
-			break;
-
-		case MATEMENU_TREE_ITEM_ALIAS:
-			add_drawers_from_alias (MATEMENU_TREE_ALIAS (item), attached_toplevel_id);
-			break;
-
-		default:
-			break;
+			case MATEMENU_TREE_ITEM_ALIAS:
+				item = matemenu_tree_iter_get_alias (iter);
+				add_drawers_from_alias (item, attached_toplevel_id);
+				matemenu_tree_item_unref (item);
+				break;
+			default:
+				break;
 		}
-
-		matemenu_tree_item_unref (item);
 	}
-
-	g_slist_free (items);
-
+	matemenu_tree_iter_unref (iter);
 	g_free (attached_toplevel_id);
 }
 
@@ -354,7 +354,7 @@ add_menudrawer_to_panel (GtkWidget      *menuitem,
 	PanelData         *pd;
 	int                insertion_pos;
 
-	directory = matemenu_tree_item_get_parent (MATEMENU_TREE_ITEM (entry));
+	directory = matemenu_tree_entry_get_parent (entry);
 
 	panel = menu_get_panel (menuitem);
 
@@ -380,7 +380,7 @@ add_menu_to_panel (GtkWidget      *menuitem,
 	char               *menu_path;
 	const char         *menu_filename;
 
-	directory = matemenu_tree_item_get_parent (MATEMENU_TREE_ITEM (entry));
+	directory = matemenu_tree_entry_get_parent (entry);
 	if (!directory) {
 		g_warning ("Cannot find the filename for the menu: no directory");
 		return;
@@ -393,8 +393,8 @@ add_menu_to_panel (GtkWidget      *menuitem,
 		return;
 	}
 
-	menu_filename = matemenu_tree_get_menu_file (tree);
-	matemenu_tree_unref (tree);
+	menu_filename = matemenu_tree_get_canonical_menu_path (tree);
+	g_object_unref (tree);
 	if (!menu_filename) {
 		matemenu_tree_item_unref (directory);
 		g_warning ("Cannot find the filename for the menu: no filename");
@@ -476,7 +476,7 @@ create_item_context_menu (GtkWidget   *item,
 	if (!entry)
 		return NULL;
 
-	directory = matemenu_tree_item_get_parent (MATEMENU_TREE_ITEM (entry));
+	directory = matemenu_tree_entry_get_parent (entry);
 	if (!directory)
 		return NULL;
 
@@ -485,8 +485,8 @@ create_item_context_menu (GtkWidget   *item,
 	if (!tree)
 		return NULL;
 
-	menu_filename = matemenu_tree_get_menu_file (tree);
-	matemenu_tree_unref (tree);
+	menu_filename = matemenu_tree_get_canonical_menu_path(tree);
+	g_object_unref (tree);
 	if (!menu_filename)
 		return NULL;
 
@@ -726,7 +726,7 @@ setup_menuitem_with_icon (GtkWidget   *menuitem,
 		icon = panel_gicon_from_icon_name (image_filename);
 
 	gtk_image_set_from_gicon (GTK_IMAGE(image), icon, icon_size);
-	g_object_unref (icon);
+	g_clear_object (&icon);
 
 	gtk_widget_show (image);
 
@@ -980,8 +980,8 @@ create_submenu_entry (GtkWidget          *menu,
 
 	setup_menuitem_with_icon (menuitem,
 				  panel_menu_icon_get_size (),
-				  NULL,
 				  matemenu_tree_directory_get_icon (directory),
+				  NULL,
 				  matemenu_tree_directory_get_name (directory));
 
 	gtk_menu_shell_append (GTK_MENU_SHELL (menu), menuitem);
@@ -1051,7 +1051,7 @@ create_menuitem (GtkWidget          *menu,
 				matemenu_tree_item_ref (entry),
 				(GDestroyNotify) matemenu_tree_item_unref);
 
-	if (alias_directory)
+	if (alias_directory) {
 		//FIXME: we should probably use this data when we do dnd or
 		//context menu for this menu item
 		g_object_set_data_full (G_OBJECT (menuitem),
@@ -1059,26 +1059,35 @@ create_menuitem (GtkWidget          *menu,
 					matemenu_tree_item_ref (alias_directory),
 					(GDestroyNotify) matemenu_tree_item_unref);
 
-	setup_menuitem_with_icon (menuitem,
-				  panel_menu_icon_get_size (),
-				  NULL,
-				  alias_directory ? matemenu_tree_directory_get_icon (alias_directory) :
-						    matemenu_tree_entry_get_icon (entry),
-				  alias_directory ? matemenu_tree_directory_get_name (alias_directory) :
-						    matemenu_tree_entry_get_display_name (entry));
+		setup_menuitem_with_icon (menuitem,
+				panel_menu_icon_get_size (),
+				matemenu_tree_directory_get_icon (alias_directory),
+				NULL,
+				matemenu_tree_directory_get_name (alias_directory));
+		if (matemenu_tree_directory_get_comment (alias_directory)) {
+			panel_util_set_tooltip_text (menuitem, matemenu_tree_directory_get_comment (alias_directory));
+		}
+	} else {
+		GDesktopAppInfo *ginfo;
+		const gchar* desc;
+		const gchar* gename;
 
-	if (alias_directory &&
-	    matemenu_tree_directory_get_comment (alias_directory))
-		panel_util_set_tooltip_text (menuitem,
-					     matemenu_tree_directory_get_comment (alias_directory));
-	else if	(!alias_directory &&
-		 matemenu_tree_entry_get_comment (entry))
-		panel_util_set_tooltip_text (menuitem,
-					     matemenu_tree_entry_get_comment (entry));
-	else if	(!alias_directory &&
-		 matemenu_tree_entry_get_generic_name (entry))
-		panel_util_set_tooltip_text (menuitem,
-					     matemenu_tree_entry_get_generic_name (entry));
+		ginfo = matemenu_tree_entry_get_app_info (entry);
+		desc= g_app_info_get_description(G_APP_INFO(ginfo));
+		gename = g_desktop_app_info_get_generic_name(ginfo);
+
+		setup_menuitem_with_icon (menuitem,
+				panel_menu_icon_get_size (),
+				g_app_info_get_icon(G_APP_INFO(ginfo)),
+				NULL,
+				g_app_info_get_name(G_APP_INFO(ginfo)));
+
+		if (desc != NULL) {
+			panel_util_set_tooltip_text (menuitem, desc);
+		} else if ( gename != NULL) {
+			panel_util_set_tooltip_text (menuitem, gename);
+		}
+	}
 
 	g_signal_connect_after (menuitem, "button_press_event",
 				G_CALLBACK (menuitem_button_press_event), NULL);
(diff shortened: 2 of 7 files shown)


lukefromdc:
Also note this REQUIRES the mate-menus commit: the panel will start with the old version of mate-menus but clicking the menu will crash it on an undefined symbol

lukefromdc:
The version of Mozo with the PR for this works fine from this panel version and mate-menus version

mate-panel/drawer.c
```diff
@@ -587,7 +587,10 @@ panel_drawer_prepare (const char  *drawer_id,
     g_settings_set_boolean (settings, PANEL_OBJECT_USE_CUSTOM_ICON_KEY, use_custom_icon);
 
     if (custom_icon) {
-        g_settings_set_string (settings, PANEL_OBJECT_CUSTOM_ICON_KEY, custom_icon);
```
monsta: Are these drawer changes necessary?

mate-panel/menu.c
```diff
@@ -1238,13 +1229,9 @@ create_applications_menu (const char *menu_file,
 	g_signal_connect (menu, "button_press_event",
 			  G_CALLBACK (menu_dummy_button_press_event), NULL);
 
-	matemenu_tree_add_monitor (tree,
-			       (MateMenuTreeChangedFunc) handle_matemenu_tree_changed,
-			       menu);
-	g_signal_connect (menu, "destroy",
```
monsta: Shouldn't we keep some cleanup function? Like disconnecting the signal handler when `menu` goes away.

mate-panel/drawer.c
```diff
@@ -636,7 +639,7 @@ panel_drawer_create (PanelToplevel *toplevel,
 char *
 panel_drawer_create_with_id (const char    *toplevel_id,
                              int            position,
-                             const char    *custom_icon,
+                             GIcon         *custom_icon,
```
monsta: Looks like `panel_drawer_create` above should be modified in the same way.

```C
drawer.c: In function ‘panel_drawer_create’:
drawer.c:632:31: warning: passing argument 2 of ‘panel_drawer_prepare’ from incompatible pointer type [-Wincompatible-pointer-types]
     panel_drawer_prepare (id, custom_icon, use_custom_icon, tooltip, NULL);
                               ^~~~~~~~~~~
drawer.c:570:1: note: expected ‘GIcon * {aka struct _GIcon *}’ but argument is of type ‘const char *’
 panel_drawer_prepare (const char  *drawer_id,
 ^~~~~~~~~~~~~~~~~~~~
```

lukefromdc:
A couple years ago I cleaned up a whole pile of warnings in Caja caused by signals that were not being disconnected

mate-panel/drawer.c
yetist: Yes, because matemenu_tree_directory_get_icon() now return GIcon object, not the icon_name string.

mate-panel/drawer.c
yetist: Thank you for review, It needs to be modified here.

mate-panel/menu.c
yetist: Ok, now added disconnect signal handle when menu destroy.

mate-panel/drawer.c
muktupavels: It feels wrong, now icon is not saved and most likely is / will be regression. Or if saving icon in settings has never been needed then I would suggest to remove it in separate commit before this API change...

mate-panel/drawer.c
```diff
@@ -587,7 +587,10 @@ panel_drawer_prepare (const char  *drawer_id,
     g_settings_set_boolean (settings, PANEL_OBJECT_USE_CUSTOM_ICON_KEY, use_custom_icon);
 
     if (custom_icon) {
-        g_settings_set_string (settings, PANEL_OBJECT_CUSTOM_ICON_KEY, custom_icon);
+	gchar *icon_name;
+	icon_name = g_icon_to_string(custom_icon);
+        g_settings_set_string (settings, PANEL_OBJECT_CUSTOM_ICON_KEY, icon_name);
+	g_free(icon_name);
```
muktupavels: Something wrong with indentation, also there should be space after function name like it is done with `g_settings_set_string`.

mate-panel/drawer.c
```diff
@@ -618,7 +621,7 @@ panel_drawer_prepare (const char  *drawer_id,
 void
 panel_drawer_create (PanelToplevel *toplevel,
                      int            position,
-                     const char    *custom_icon,
+		     GIcon         *custom_icon,
```
muktupavels: Indentation...

mate-panel/menu.c
```diff
@@ -726,7 +720,8 @@ setup_menuitem_with_icon (GtkWidget   *menuitem,
 		icon = panel_gicon_from_icon_name (image_filename);
 
 	gtk_image_set_from_gicon (GTK_IMAGE(image), icon, icon_size);
-	g_object_unref (icon);
+	if (icon != NULL)
+		g_object_unref (icon);
```
muktupavels: `g_clear_object (&icon);` might be better...

mate-panel/panel-addto.c
```diff
@@ -971,7 +985,7 @@ panel_addto_dialog_free (PanelAddtoDialog *dialog)
 	dialog->application_model = NULL;
 
 	if (dialog->menu_tree)
-		matemenu_tree_unref (dialog->menu_tree);
+		g_object_unref(dialog->menu_tree);
```
muktupavels: You could replace these 3 lines with `g_clear_object (&dialog->menu_tree);`

mate-panel/menu.c
```diff
@@ -720,8 +720,7 @@ setup_menuitem_with_icon (GtkWidget   *menuitem,
 		icon = panel_gicon_from_icon_name (image_filename);
 
 	gtk_image_set_from_gicon (GTK_IMAGE(image), icon, icon_size);
-	if (icon != NULL)
-		g_object_unref (icon);
+	g_clear_object(&icon);
```
muktupavels: It should be `g_clear_object (&icon);` - space after function name.

mate-panel/panel-addto.c
```diff
@@ -984,8 +984,7 @@ panel_addto_dialog_free (PanelAddtoDialog *dialog)
 		g_object_unref (dialog->application_model);
 	dialog->application_model = NULL;
 
-	if (dialog->menu_tree)
-		g_object_unref(dialog->menu_tree);
+	g_clear_object(&dialog->menu_tree);
```
muktupavels: `dialog->menu_tree = NULL;` is not needed, g_clear_object already does that. :)

mate-panel/drawer.c
muktupavels: Oh, I guess I missed something... I saw that it removes line, somehow missed that new code is added, sorry for that.

monsta: It's ok :slightly_smiling_face:

monsta:
Found one more place with wrong type warning:

```C
panel-menu-button.c: In function ‘panel_menu_button_get_icon’:
panel-menu-button.c:720:23: warning: passing argument 1 of ‘g_strdup’ from incompatible pointer type [-Wincompatible-pointer-types]
    retval = g_strdup (matemenu_tree_directory_get_icon (directory));
In file included from /usr/include/glib-2.0/glib.h:79:0,
                 from /usr/include/gtk-3.0/gdk/gdkconfig.h:13,
                 from /usr/include/gtk-3.0/gdk/gdk.h:30,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from button-widget.h:4,
                 from panel-menu-button.h:28,
                 from panel-menu-button.c:27:
/usr/include/glib-2.0/glib/gstrfuncs.h:217:22: note: expected ‘const gchar * {aka const char *}’ but argument is of type ‘GIcon * {aka struct _GIcon *}’
 gchar*               g_strdup        (const gchar *str) G_GNUC_MALLOC;
                      ^~~~~~~~
```

mate-panel/panel-run-dialog.c
```diff
@@ -766,79 +776,83 @@ static GSList *
...
 static GSList* get_all_applications(void)
 {
 	MateMenuTree* tree;
 	MateMenuTreeDirectory* root;
+	GError        *error = NULL;
 	GSList* retval;
 
-	tree = matemenu_tree_lookup("mate-applications.menu", MATEMENU_TREE_FLAGS_NONE);
-	matemenu_tree_set_sort_key(tree, MATEMENU_TREE_SORT_DISPLAY_NAME);
+	tree = matemenu_tree_new ("mate-applications.menu", MATEMENU_TREE_FLAGS_NONE);
```
monsta: Looks like `MATEMENU_TREE_SORT_DISPLAY_NAME` is lost here.

mate-panel/panel-addto.c
```diff
@@ -662,27 +666,37 @@ static void panel_addto_make_application_model(PanelAddtoDialog* dialog)
...
 
 		matemenu_tree_item_unref(root);
 	}
 
-	matemenu_tree_unref(tree);
+	g_object_unref(tree);
 
-	tree = matemenu_tree_lookup("mate-settings.menu", MATEMENU_TREE_FLAGS_NONE);
-	matemenu_tree_set_sort_key(tree, MATEMENU_TREE_SORT_DISPLAY_NAME);
+	if (! matemenu_tree_load_sync (tree, &error)) {
```
monsta: The `tree` had just been unref'ed, but we try to load from it. Isn't another `matemenu_tree_new` with "mate-settings.menu" as argument missing here?

yetist: Yes, I lost one line here.

mate-panel/panel-run-dialog.c
yetist: ok, I will add it.

yetist:
I found a new bug after this PR:
Right click on submenut item, select "Entire Menu", click "Add this as menu to panel",  it do nothing.

The following patch can solve this problem, do I need to include it  in the PR?

```
diff --git a/mate-panel/panel-menu-button.c b/mate-panel/panel-menu-button.c
index 2dca673..0f4af4b 100644
--- a/mate-panel/panel-menu-button.c
+++ b/mate-panel/panel-menu-button.c
@@ -128,20 +128,33 @@ panel_menu_filename_to_scheme (const char *filename)
 {
 	const char *retval;
 	int         i;
+	gchar *basename;
 
 	retval = NULL;
 
 	if (!filename)
 		return retval;
 
+	/* Note:
+	 * filename in root_items dose not have the parent directory,
+	 * if filename passed is absolute path, this function will return NULL.
+	 * so here use the basename for compare.
+	 */
+	if (g_path_is_absolute (filename)) {
+		basename = g_path_get_basename (filename);
+	} else {
+		basename = g_strdup(filename);
+	}
+
 	for (i = 0; i < G_N_ELEMENTS (root_items); i++) {
 		if (root_items [i].filename &&
-		    !strncmp (filename, root_items [i].filename,
+		    !strncmp (basename, root_items [i].filename,
 			      strlen (root_items [i].filename))) {
 			retval = root_items [i].scheme;
 			break;
 		}
 	}
+	g_free(basename);
 
 	return retval;
 }
` ``

File: mate-panel/panel-addto.c [hunk 12]
```diff
@@ -517,16 +517,20 @@ panel_addto_prepend_directory (GSList             **parent_list,
...
 	data->item_info.type          = PANEL_ADDTO_MENU;
 	data->item_info.name          = g_strdup (matemenu_tree_directory_get_name (directory));
 	data->item_info.description   = g_strdup (matemenu_tree_directory_get_comment (directory));
-	data->item_info.icon          = g_strdup (matemenu_tree_directory_get_icon (directory));
+	data->item_info.icon          = g_icon_to_string(gicon);
 	data->item_info.menu_filename = g_strdup (filename);
 	data->item_info.menu_path     = matemenu_tree_directory_make_path (directory, NULL);
 	data->item_info.static_data   = FALSE;
+	if (gicon != NULL)
+		g_object_unref(gicon);
```
monsta: Do we need this unref? It's not done in other places, and `matemenu_tree_directory_get_icon` has "transfer none" in the description.

monsta:
I confirm the new issue. But which change caused it?

mate-panel/panel-run-dialog.c
```diff
@@ -766,79 +776,83 @@ static GSList *
...
-			break;
+	MateMenuTreeIter *iter;
+	MateMenuTreeItemType type;
+
+	iter = matemenu_tree_directory_iter (directory);
+	while ((type = matemenu_tree_iter_next (iter)) != MATEMENU_TREE_ITEM_INVALID) {
+		gpointer item;
+		switch (type) {
+			case MATEMENU_TREE_ITEM_ENTRY:
+				item = matemenu_tree_iter_get_entry (iter);
```
monsta: In other places, similarly created `item` is later unref'ed via `matemenu_tree_item_unref`. Looks like it's missing here (in all three cases).

yetist:
menu.c file, old use matemenu_tree_get_menu_file, new use matemenu_tree_get_canonical_menu_path. xiaotian.wu 邮箱：xiaotian.wu@i-soft.com.cn 签名由 网易邮箱大师 定制 在2018年05月11日 19:42，monsta 写道： I confirm the new issue. But which change caused it? — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or mute the thread.

mate-panel/panel-run-dialog.c
```diff
@@ -888,20 +904,23 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
...
-				    COLUMN_EXEC,      matemenu_tree_entry_get_exec (entry),
+				    COLUMN_NAME,      g_app_info_get_display_name(G_APP_INFO(ginfo)),
+				    COLUMN_COMMENT,   g_app_info_get_description(G_APP_INFO(ginfo)),
+				    COLUMN_EXEC,      g_app_info_get_commandline(G_APP_INFO(ginfo)),
 				    COLUMN_PATH,      matemenu_tree_entry_get_desktop_file_path (entry),
 				    COLUMN_VISIBLE,   TRUE,
 				    -1);
-
-		g_object_unref (gicon);
+		if (gicon != NULL)
```
monsta: `g_app_info_get_icon` is "transfer none", which means the icon shouldn't be unref'ed.

mate-panel/panel-addto.c
```diff
@@ -547,15 +551,21 @@ panel_addto_prepend_entry (GSList         **parent_list,
...
 	data->item_info.type          = PANEL_ADDTO_LAUNCHER;
-	data->item_info.name          = g_strdup (matemenu_tree_entry_get_display_name (entry));
-	data->item_info.description   = g_strdup (matemenu_tree_entry_get_comment (entry));
-	data->item_info.icon          = g_strdup (matemenu_tree_entry_get_icon (entry));
+	data->item_info.name          = g_strdup (g_app_info_get_display_name(G_APP_INFO(ginfo)));
+	data->item_info.description   = g_strdup (g_app_info_get_description(G_APP_INFO(ginfo)));
+	data->item_info.icon          = g_icon_to_string(gicon);
 	data->item_info.launcher_path = g_strdup (matemenu_tree_entry_get_desktop_file_path (entry));
 	data->item_info.static_data   = FALSE;
+	g_object_unref(gicon);
```
monsta: Again, `g_app_info_get_icon` is "transfer none"...

mate-panel/menu.c
```diff
@@ -311,36 +305,36 @@ add_drawers_from_dir (MateMenuTreeDirectory *directory,
...
+				break;
+			default:
+				g_assert_not_reached ();
+				break;
 		}
-
-		matemenu_tree_item_unref (item);
 	}
-
-	g_slist_free (items);
```
monsta: Iterator created by `matemenu_tree_directory_iter` should be freed with `matemenu_tree_iter_unref`.

mate-panel/menu.c
```diff
@@ -1257,55 +1263,55 @@ populate_menu_from_directory (GtkWidget          *menu,
...
+				break;
+
+			default:
+				break;
 		}
-
-		matemenu_tree_item_unref (item);
 	}
 
-	g_slist_free (items);
```
monsta: Again, leak of iterator created by `matemenu_tree_directory_iter`...

mate-panel/panel-addto.c
```diff
@@ -565,63 +575,57 @@ panel_addto_prepend_alias (GSList         **parent_list,
...
 			break;
-
 		default:
 			break;
 		}
-
-		matemenu_tree_item_unref (l->data);
 	}
 
-	g_slist_free (items);
```
monsta: One more leak of iterator created by `matemenu_tree_directory_iter`.

mate-panel/panel-addto.c
yetist: Yes, gicon created by matemenu_tree_directory_get_icon dose not need unref.

mate-panel/panel-addto.c
monsta: I think `matemenu_tree_iter_unref` is needed

monsta:
Ok, so `panel_menu_button_create` gets the wrong path in its `filename` argument?

yetist:
@monsta 
> Ok, so panel_menu_button_create gets the wrong path in its filename argument?

Yes, please see [here](https://github.com/mate-desktop/mate-panel/pull/794/files#diff-13ae31aebe3f9407924db614e17be9d5R390), the master branch use ```menu_filename = matemenu_tree_get_menu_file (tree)```, this pr use ```menu_filename = matemenu_tree_get_canonical_menu_path (tree);```

The difference between matemenu_tree_get_menu_file and matemenu_tree_get_canonical_menu_path:

In matemenu_tree_get_menu_file, if tree type is  MATEMENU_TREE_BASENAME, [here](https://github.com/mate-desktop/mate-menus/pull/59/files#diff-4b3cf32d4c04d857cd5f92c2bff7bfe6L757) return basename of tree->canonical_path.

But in backport from gnome-menus, the matemenu_tree_get_canonical_menu_path just return tree->canonical_path [here](https://github.com/mate-desktop/mate-menus/pull/59/files#diff-4b3cf32d4c04d857cd5f92c2bff7bfe6R709)

monsta:
Ok, I see, so the patch for this is needed here.

mate-panel/panel-run-dialog.c
```diff
@@ -766,79 +776,82 @@ static GSList *
...
+	tree = matemenu_tree_new ("mate-applications.menu", MATEMENU_TREE_FLAGS_SORT_DISPLAY_NAME);
+	if (!matemenu_tree_load_sync (tree, &error)) {
+		g_printerr ("Failed to load tree: %s\n", error->message);
+		g_error_free(error);
+		g_object_unref(tree);
+		return NULL;
+	}
 
-	root = matemenu_tree_get_root_directory(tree);
+	g_print ("Loaded menu from %s\n", matemenu_tree_get_canonical_menu_path (tree));
```
monsta: Let's remove this debug print, otherwise it will add messages to `~/.xsession-errors` on every Run dialog call.

mate-panel/panel-run-dialog.c
monsta: Ok, the item is left with reference if it's added to list with `g_slist_append`. But what if it's a directory? Shouldn't we unref it after `get_all_applications_from_dir` call is finished?

monsta: Seems like the same applies to `get_all_applications_from_alias` call as well.

raveit65:
Looks like we have open questions here?

raveit65:
@monsta 
All good now?

monsta:
The patch for adding submenu to panel works, thanks.

I still suspect we might have some memory leaks during recursive function calls (https://github.com/mate-desktop/mate-panel/pull/794#discussion_r187905637).

mate-panel/panel-run-dialog.c
yetist: If call matemenu_tree_item_unref here, the running dialog(Alt-F2) will crash.

monsta: Not right here, but after recursive calls to `get_all_applications_from_dir` and `get_all_applications_from_alias`. They don't seem to add their `directory` and `alias` arguments to any list, so these can be unref'ed after the function call is finished.

muktupavels: Quickly looking `iter` is created before `while` loop and unrefed after it. But I guess you think that `item` should be unrefed, right?

At least in GNOME `gmenu_tree_iter_get_entry` has `Returns: (transfer full): An entry` and if same thing is in MATE then it looks like there are memory leak...

I think you must make sure that all items that are added to list should be freed! If somewhere some functions returns `transfer none` then you should ref it before adding to list.

And then when you are done with list, you should free it with `g_slist_free_full (all_applications, matemenu_tree_item_unref);`.

yetist: Please see [here](https://github.com/mate-desktop/mate-panel/pull/794/files#diff-065586010f2f4c80630c35dce7531240R923), already use g_slist_free_full to free them.
```
-	g_slist_free (all_applications);
+	g_slist_free_full (all_applications, matemenu_tree_item_unref);
```
If do that like @monsta said, call  `matemenu_tree_item_unref` here, Alt-F2 will crash at `g_slist_free_full(all_applications, matemenu_tree_item_unref)` line.

muktupavels: Oh, I missed that, but I am quite sure I saw `g_slist_free (all_applications);`... If it crash in `g_slist_free_full` then you have some item in list that is missing item_ref call (because it returned from function with transfer none) before it is added to list... Or it is unrefed when it should not... Or no idea... Maybe post stacktrace of crash?

monsta: I didn't mean the item that's added to the list via `g_slist_append`. I meant the one that is passed into `get_all_applications_from_dir` and `get_all_applications_from_alias`.

muktupavels: @monsta is right, you need to unref item that is returned from `matemenu_tree_iter_get_directory` and `matemenu_tree_iter_get_alias`.

yetist: @monsta You are right, the code is update, please review, thank you.

yetist:
In [line 924](https://github.com/mate-desktop/mate-panel/pull/794/commits/eaada4f21386ea54a1770757b1935c14cc2727e9#diff-065586010f2f4c80630c35dce7531240R924), already call matemenu_tree_item_unref for data in slist.

monsta:
I think second method looks better.

mate-panel/panel-run-dialog.c
monsta: Thanks, now the memory should be freed properly.

mate-panel/menu.c
```diff
@@ -265,28 +265,22 @@ static void
...
 {
-	MateMenuTreeItem *aliased_item;
-
-	aliased_item = matemenu_tree_alias_get_item (alias);
-
-	switch (matemenu_tree_item_get_type (aliased_item)) {
+	switch (matemenu_tree_alias_get_aliased_item_type (alias)) {
 	case MATEMENU_TREE_ITEM_DIRECTORY:
-		add_drawers_from_dir (MATEMENU_TREE_DIRECTORY (aliased_item),
+		add_drawers_from_dir (matemenu_tree_alias_get_directory (alias),
```
monsta: Hmm, `matemenu_tree_alias_get_directory` returns result of `matemenu_tree_item_ref`. Should we unref it after using?

mate-panel/menu.c
```diff
@@ -1124,34 +1127,29 @@ static void
...
-	MateMenuTreeItem *aliased_item;
-
-	aliased_item = matemenu_tree_alias_get_item (alias);
-
-	switch (matemenu_tree_item_get_type (aliased_item)) {
+	switch (matemenu_tree_alias_get_aliased_item_type (alias)) {
 	case MATEMENU_TREE_ITEM_DIRECTORY:
 		create_submenu (menu,
-				MATEMENU_TREE_DIRECTORY (aliased_item),
+				matemenu_tree_alias_get_directory (alias),
```
monsta: Same as in previous comment about `matemenu_tree_alias_get_directory`

monsta: Though it was already used below in the original code... so either it shouldn't be unref'ed here, or we already had a memory leak before this PR.

mate-panel/menu.c
```diff
@@ -265,28 +265,22 @@ static void
...
+		add_drawers_from_dir (matemenu_tree_alias_get_directory (alias),
 				      G_MAXINT/2,
 				      toplevel_id);
 		break;
 
 	case MATEMENU_TREE_ITEM_ENTRY:
 		panel_launcher_create_with_id (toplevel_id,
 					       G_MAXINT/2,
-					       matemenu_tree_entry_get_desktop_file_path (MATEMENU_TREE_ENTRY (aliased_item)));
+					       matemenu_tree_entry_get_desktop_file_path (matemenu_tree_alias_get_aliased_entry (alias)));
```
monsta: I guess the same applies to `matemenu_tree_alias_get_aliased_entry`. This one is even marked as "transfer full".

monsta: Right... it's used in another place where the entry is added to the list of ref'ed items.

mate-panel/menu.c
```diff
@@ -1124,34 +1127,29 @@ static void
...
 		create_submenu (menu,
-				MATEMENU_TREE_DIRECTORY (aliased_item),
+				matemenu_tree_alias_get_directory (alias),
 				matemenu_tree_alias_get_directory (alias));
 		break;
 
 	case MATEMENU_TREE_ITEM_ENTRY:
 		create_menuitem (menu,
-				 MATEMENU_TREE_ENTRY (aliased_item),
+				 matemenu_tree_alias_get_aliased_entry(alias),
```
monsta: More `matemenu_tree_alias_get_aliased_entry`

mate-panel/panel-addto.c
```diff
@@ -565,62 +572,57 @@ panel_addto_prepend_alias (GSList         **parent_list,
...
-
-	aliased_item = matemenu_tree_alias_get_item (alias);
-
-	switch (matemenu_tree_item_get_type (aliased_item)) {
+	switch (matemenu_tree_alias_get_aliased_item_type (alias)) {
 	case MATEMENU_TREE_ITEM_DIRECTORY:
 		panel_addto_prepend_directory (parent_list,
-					       MATEMENU_TREE_DIRECTORY (aliased_item),
-					       filename);
+				matemenu_tree_alias_get_directory(alias),
```
monsta: Here and a few lines below - same `matemenu_tree_alias_*` functions as I mentioned before...

raveit65:
@monsta 
Any open concerns?

Current commit list:

    use gicon instead icon name
    rewrite add_drawers_from_alias/dir func
    use new api from mate-menus
    swap gicon and icon name
    use new api, and keep it easy to understand
    rewrite funcs use new api
    use signal to replace monitor
    use new mate-menus api
    use new mate-menus api
    call load sync when received changed signal
    change panel_drawer_create argument from char to GIcon
    disconnect signal handle when menu destroy
    Fixed runtime warning when icon is NULL
    check is null before g_object_unref
    indent code
    use g_clear_object
    for code style
    sort application display name
    fixed type error
    add matemenu_tree_new for settings.menu
    remove assert and memory leak
    fix memory leak
    remove print info
    fixed add submenu to panel with new api
    fixed memory leak
    fixed memory leak
    fixed runtime warning when no icon in desktop file
    use new api, and keep it easy to understand
    swap gicon and icon name
    use new api from mate-menus
    rewrite add_drawers_from_alias/dir func
    use gicon instead icon name

@yetist 
When review is done can please squash/rebase the PR so that we don't have so much commits?
Also renaming some (not all) commits to something meaningfully would be nice.
So that we know what a commit does after some months if someone browse git history ;)

Thank you

monsta:
Ah, there's a new commit after squashing. Actually it was there before squashing, see https://github.com/mate-desktop/mate-panel/pull/794#issuecomment-388325367. Should we squash it now too?

monsta:
I think all the commits from that list are squashed now. Now we have a large commit + a small bugfix commit.

raveit65:
Ok, here we go.....
Thanks for your hard work.

Issue #797, opened 2018-04-28
Window List should display window titles when Panel is not expanded


darkoni:
#### Expected behaviour

Window List should display window titles when Panel is not expanded.
When Panel is expanded, window titles are shown ok.

#### Steps to reproduce the behaviour

1) Make Panel [x] Expand
2) Add Window List to Panel - window titles are displayed
3) Make Panel [ ] Expand - in Window list, window titles are not displayed, only space is reserved where window tiles should be displayed.

#### MATE general version

1.20.1.

#### Package version

Window List 1.20.1

#### Linux Distribution

Fedora 27
Linux 4.15.17-300.fc27.x86_64 #1 SMP Thu Apr 12 18:19:17 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

raveit65:
wtf, confirmed with master.

raveit65:
@darkoni 
I need to know the exact mate-panel package version, please post the result of
'rpm -qa mate-panel'
There is more than one 1.20.1 package version built for f27 repos.

raveit65:
I could track down the problem, it occurs if panel is build with
` --with-in-process-applets=all` configure flag.
If panel is build with  --with-in-process-applets=none than the windows-list applet is still visible like before.
I switch building panel for 1.20 in f27 to buiild in process because it is better for HIDIPI support.
Steps to reproduce:
1. build panel in process
2. create an empty panel
3. add the windows-list applet
4. un-expand the panel

@lukefromdc 
Can you please take a look into it?

@darkoni 
You can rebuild the rpm with `--with-in-process-applets=none` as quick workaround.

darkoni:
rpm -qa mate-panel
mate-panel-1.20.1-4.fc27.x86_64

raveit65:
well, i switched with 1.20.1-2.fc27.x86_64 to build in-process applets.
You can try downgrade to 1.20.1-1.fc27.x86_64 or just rebuild current version for the moment.

darkoni:
@raveit65 

i can not find  1.20.1-1.fc27.x86_64  so i downgraded to  1.19.3-2.fc27.x86_64

# dnf install mate-panel-1.19.3-2.fc27.x86_64

many thx

ps: i hope mate-panel-1.20.1-5.fc27.x86_64 or some other version will be ready in future,
if not, i see that i can easily downgrade to 1.19.3-2.fc27.x86_64

raveit65:
[...]

PR #798, opened 2018-04-28
window-list: fix rendering when in-process and panel not expanded

lukefromdc:
Fix https://github.com/mate-desktop/mate-panel/issues/797
Note that another issue (present in master) prevents removal of the clock, workspace-switcher, or window-list from an unexpanded panel, the offending panel (only that panel and not all panels) becoming unresponsive until restarted.


lukefromdc:
Hold this for now: I just found out that this causes the panel to expand to the left past it's origjnal position over other applets, and that locking it to the panel does not stop it. Need to find a fix

lukefromdc:
This simply does not work right, and the bug is very old (all the way back to 1.10), more research needed

Issue #799, opened 2018-04-28
clock, window-list, workspace-switcher(all  in-process) can't be removed from unexpanded panel until restart


lukefromdc:
#### Steps to reproduce the behaviour
1:Build mate-panel with the clock and wncklet in-process and install
2: Add a new panel
3: Add the clock, (window list if built with https://github.com/mate-desktop/mate-panel/issues/797), or workspace switcher to that panel
4: try to remove the applet.
5: panel freezes: If "add to panel" is still open, adding another applet will force a redraw, and the panel will render with that applet added and the removed applet gone. Otherwise restart the panel and the removed applet will now be gone.

Note that if they are all on the panel at once, they may or may not be removable, depending on order of removal and probably other conditions.

#### MATE general version
1.20/1.21

#### Package version
mate-panel 1.21.0 from git master as of 4-28-2018
found while testing
https://github.com/mate-desktop/mate-panel/issues/797

#### Linux Distribution
Debian Unstable

#### Link to downstream report of your Distribution
None as this is a local build of MATE as a whole

raveit65:
Related PR fixes the issue, tested with vertical and horizontal panel.

lukefromdc:
This is partially fixed, all but the LAST in-process applet can now be removed from a nonexpanded panel without restarting it. There are also  issues (old issues) some but not all of the time with removing out-of-process applets from a non-expanded panel as well. With in-process applets, so long as one remains on the panel others can be freely added and removed now.

raveit65:
@lukefromdc 
Can this be closed?

lukefromdc:
It's now possible to remove all applets from a non-expanded panel that contained multiple in-process applets, closing

PR #806, opened 2018-05-08
fix rendering and removal of in-process applets on unexpanded panel

lukefromdc:
Fix https://github.com/mate-desktop/mate-panel/issues/797 and https://github.com/mate-desktop/mate-panel/issues/799

Use  GTK_SIZE_REQUEST_CONSTANT_SIZE  rather than using height-for-width or width-for-height geometry management(https://wiki.gnome.org/HowDoI/CustomWidgets#What_virtual_functions_to_implement.3F)
as was recently done for on out-of-process applets to fix HiDPI wrong-size rendering. Note that disabling this only for the window list causes other in-process applets to disappear when the window list is present on a nonexpanded panel, and that adding and removing the other in-process applets is also made to work after a delay of about a second.

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1096,18 +1096,15 @@ mate_panel_applet_get_preferred_height (GtkWidget *widget,
 static GtkSizeRequestMode
 mate_panel_applet_get_request_mode (GtkWidget *widget)
 {
-	MatePanelApplet *applet = MATE_PANEL_APPLET (widget);
-	MatePanelAppletOrient orientation;
-
-	if (applet->priv->out_of_process)
-		return GTK_SIZE_REQUEST_CONSTANT_SIZE;
-
-	orientation = mate_panel_applet_get_orient (applet);
-	if (orientation == MATE_PANEL_APPLET_ORIENT_UP ||
-	    orientation == MATE_PANEL_APPLET_ORIENT_DOWN)
-		return GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH;
-
-	return GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT;
+	/*Do not use GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH
+	 *or GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT
+	 *to avoid problems with in-process applets
+	 *when the panel is not expanded
+	 *See https://github.com/mate-desktop/mate-panel/issues/797
+	 *and https://github.com/mate-desktop/mate-panel/issues/799
+	 *Out of process applets already use GTK_SIZE_REQUEST_CONSTANT_SIZE
+	 */
+	return GTK_SIZE_REQUEST_CONSTANT_SIZE;
 }
 
 static void


libmate-panel-applet/mate-panel-applet.c
```diff
@@ -1096,18 +1096,15 @@ mate_panel_applet_get_preferred_height (GtkWidget *widget,
...
-		return GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH;
-
-	return GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT;
+	/*Do not use GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH
+	 *or GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT because
+	 *they cause problems for in-process applets
+	 *when the panel is not expanded
+	 *See https://github.com/mate-desktop/mate-panel/issues/797
+	 *and https://github.com/mate-desktop/mate-panel/issues/799
+	 *Out of process applets have always used GTK_SIZE_REQUEST_CONSTANT_SIZE
```
muktupavels: Wrong! See https://github.com/mate-desktop/mate-panel/commit/ff04f332136272a14c588da3f2d582ba0ff01be1...

raveit65:
@muktupavels 
What is wrong?
Can you please a bit more unexcited in favour of given detailed explanation ;)

muktupavels:
> Out of process applets have always used GTK_SIZE_REQUEST_CONSTANT_SIZE

This text is wrong. Applets have never used `GTK_SIZE_REQUEST_CONSTANT_SIZE`, it is there only recently because of HiDPI fixes... No idea if that has introduced problem or not, just saying that text is misleading and wrong.

Also I would suggest to find out why applets have used non constant size, most likely there is or at least was reason to do so...

Also `gnome-panel` still use non constant size and I think @lukefromdc somewhere wrote that `gnome-panel` has no such problem that might indicate either that real bug is somewhere else or `gnome-panel` has other bugs/problems that hides this one.

So that comment or maybe just commit message probably should explain why `GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH` and/or `GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT` is no longer needed with GTK 3 and ideally also why it causes problems. If that is the case then I would just remove that function since `GTK_SIZE_REQUEST_CONSTANT_SIZE` is default so there is no need to override it.

lukefromdc:
I have no idea why it causes problems, only that removing height for width/width for height geometry management fixed the issue. I tried removing the entire function, only to get applets not rendering at all. Yes, gnome-panel still uses this, but something somewhere is different and I cannot find it. If people want a different comment or commit message, that can be force-pushed before merge

lukefromdc:
I found this by starting with the earlier code that fixed the issue but made other problems, then following what the code was doing, and was led back to this function. This was by experiment.

muktupavels:
> I have no idea why it causes problems

And that is main problem!!! You should invest time to find that... How sure are you that it fixes problem AND does not introduce new problems? What is your test cases to ensure you have no regressions?

raveit65:
@muktupavels 
any chance of an alternative fix by you?
Or do you only know what we shouldn't do?

Any, this fix works here, and i see no regression.

muktupavels:
@raveit65 I think you know that I don't use MATE. So I am not going to use/debug mate-panel to find problems. Also there might be no need for alternative fix, I am **not** saying it is wrong...

I simply would not accept some random fix that appears to fix problem without any explanation... From my own experience I have had commit that I have committed without any explanation (but I guess fixed some problem at that time) and after year or so I have no clue why I did that... Don't remember details, but I think that at that point I even think that it was wrong or something like that. So good commit message can help even you or other that might look at commit in future.

**How can you fix something if you don't know what is the problem?**

Really you think that the ones who wrote panel was fools and did not understand what they did and simply randomly choose to use non constant size for applets? Yes, reasons might not be valid now, but then tell why in commit message...

I will repeat that I only commented because:
> Out of process applets have always used GTK_SIZE_REQUEST_CONSTANT_SIZE

is not true. Imagine that you read something like that in future, comment says that, but truth is that applets never used that...

I would suggest you and anyone on team to ask for explanations why something fixes problem, where is actual problem and something like that. Simple run time test is not valid reason, I doubt that anyone of you really every time tests **all** possible options and/or configurations to ensure that there is no regressions...

lukefromdc:
Sometimes with a small team and limited assets there is the choice between using 
something found by expereiment that works, and not fixing the bug. In weeks of work
I have NOT been able to find the root cause. I cannot build old versions of gnome-pamel
due to build dependency changes, so I cannoot bisect for it.

lukefromdc:
If there is an issue with the comment being inaccurate I will change it, no problem
there and does not change the resulting executabe. Same for commit message.

Could use "out of process applets already using..." rather than "always used" since
Albert says this was not always the case.

Test cases were panel on left, top, and bottom, expanded and not. Will do custom bg
tests and all themes tonight. Already tested are with only in-process applets on the panel
and a mix of in and out of process applets, due to the issue with returning to the expanded
panel moving the window list all the way to the right and compressing it to just the handle

lukefromdc:
I saw somewhere that  widget_class->get_preferred_width_for_height and widget_class->get_preferred_height_for_width should both be called when  GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT or GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH are used, and both only go back to GTK 3.0 and thus would of course not have existed in the GNOME 2 days.    A file search found that  get_preferred_width_for_height is only used in sn-flat-button.c and nowhere else in mate-panel, and  I did not find any instances of it in any file in gnome-panel. 

These are normally used to allow a widget to become wider if it is not tall enough or vice-versa. Since this is not used in out-of-process applets, I don't see the point of using it on in-process applets. Both are constrained in one direction by panel width/height and in the other by available space. GNOME is using it in gnome-panel without problems, wonder why it does not work here?

The panel_applet_set_size_hints in gnome-panel's panel-applet.c and mate_panel_applet_set_size_hints  in mate-panel-applet.c are indentical except for our use of "mate-panel" in names, and this is one of the very few things turned on by the EXPAND_MAJOR flag that invokes the orignal bug.

lukefromdc:
Force-pushed updated comment text, as indeed the use of GTK_SIZE_REQUEST_CONSTANT_SIZE is recently added to fix out-of-process applet sizes in HiDPI panels. Now we find it fixes rendering of the in-process window list too

BTW, the original function enabled width for height/height for width geometry management unconditionally in and only in GTK3 builds(where it first became available):
```
mate_panel_applet_get_request_mode (GtkWidget *widget)
{
	MatePanelApplet *applet = MATE_PANEL_APPLET (widget);
	MatePanelAppletOrient orientation;

	orientation = mate_panel_applet_get_orient (applet);
	if (orientation == MATE_PANEL_APPLET_ORIENT_UP ||
	    orientation == MATE_PANEL_APPLET_ORIENT_DOWN)
		return GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH;

	return GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT;
}
```

and yes, the issue goes all the way back in my available GTK3 builds to mate-panel 1.10, the oldest I have.

lukefromdc:
Should I go ahead and merge this?

raveit65:
Yes, it fixes original reported issue.

insaner:
Hi, can I bother someone here to take a quick look at !1005? I figure it might be a simple fix, but I don't think I can do that myself at this point. Thanks in advance.
@raveit65 , @lukefromdc ?

lukefromdc:
That's a text handling issue of some type, I don't know enough about that to contribute to dealing with it effectively, though @cwendling says "Sounds like the applet is reading the string as containing mnemonics" and those start with underscores

PR #813, opened 2018-05-31
workspace-switcher: provide themed fallback color

### LINKED ISSUE #807: Workspace switcher: with HighContrast theme, no way to determine the current workspace
#### Expected behaviour
Current desktop should be distinguishable by a visual effect

#### Actual behaviour
There is no way to determine the currently selected workspace.

capture du 2018-05-09 14-33-40

1) Select the GNOME High Contrast theme
2) Open Pluma on the workspace 1, maximize the Pluma window
3) Open Mate terminal on the workspace 2, maximize the Mate terminal window

#### MATE general version
1.20.1

#### Package version
1.20.1-3

#### Linux Distribution
Debian Sid

vkareh:
When a theme does not specify a color for the highlighted workspace, we can look it up from the current theme and provide that as a fallback color.

Fixes #807

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -248,6 +248,27 @@ static void applet_change_background(MatePanelApplet* applet, MatePanelAppletBac
                 type == PANEL_NO_BACKGROUND ? GTK_SHADOW_NONE : GTK_SHADOW_IN);
 }
 
+static void applet_style_updated (MatePanelApplet *applet, GtkStyleContext *context)
+{
+	GtkCssProvider *provider;
+	GdkRGBA color;
+
+	provider = gtk_css_provider_new ();
+
+	/* Provide a fallback color for the highlighted workspace based on the current theme */
+	gtk_style_context_lookup_color (context, "theme_selected_bg_color", &color);
+	gchar *bg_css = g_strconcat (".wnck-pager:selected {\n"
+		                     "	background-color:", gdk_rgba_to_string (&color), ";\n"
+		                     "}", NULL);
+	gtk_css_provider_load_from_data (provider, bg_css, -1, NULL);
+	g_free (bg_css);
+
+	gtk_style_context_add_provider (context,
+					GTK_STYLE_PROVIDER (provider),
+					GTK_STYLE_PROVIDER_PRIORITY_FALLBACK);
+	g_object_unref (provider);
+}
+
 /* Replacement for the default scroll handler that also cares about the wrapping property.
  * Alternative: Add behaviour to libwnck (to the WnckPager widget).
  */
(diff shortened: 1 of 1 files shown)


lukefromdc:
Here is mate-panel with this PR, compiz, an overlapping window on an adjacent desktop, and HichContrast
workspace_switcher_w_pr_and_highcontrast

lukefromdc:
The other themes looked fine, the code itself seems to work fine.

vkareh:
Okay, I'll see if I can add a contrasting border color

vkareh:
@lukefromdc - I cannot add a border color, as HighContrast already adds a black border. The real issue is that they don't provide a contrasting color by default, which is ironic given the theme name. This sets the fallback to do just that, but the border gets lost in there... The proper solution is to change the GTK theme directly. I'll work on a patch for that, but this should improve accessibility in the meantime.

@alexarnaud - does this explanation above make sense? Is this an acceptable solution for now? (see @lukefromdc's screenshot above)

Also to note that the HighContrast theme uses the Adwaita icon set by default, rather than the HighContrast icon set... this should also be addressed, of course.

lukefromdc:
I'm not the one who should decide this, we need to know what works for those with visual impairments or working in bright sun with dim laptop screens

raveit65:
Isn't possible to add the border only for some specific themes and using GTK_STYLE_PROVIDER_PRIORITY_APPLICATION to override the wrong border-color from gtk+ themes?
See what @lukefromdc did for xrandr-applet.
https://github.com/mate-desktop/mate-settings-daemon/blob/master/plugins/xrandr/msd-xrandr-manager.c#L1848
and
https://github.com/mate-desktop/mate-settings-daemon/blob/master/plugins/xrandr/msd-xrandr-manager.c#L1868

Personal i would prefer to add a border only for themes where it is needed.
Our own themes don't need that.

lukefromdc:
@vkareh says HighContrast already sets a black border. We could override that in HighContrast and maybe HighContrastInverse only

vkareh:
Hmmm... At that point we could very well remove this entire chunk of code and have mate's own HighContrast add the missing styling

alexarnaud:
Le 31/05/2018 à 22:57, Victor Kareh a écrit :
> @alexarnaud <https://github.com/alexarnaud> - does this explanation 
> above make sense? Is this an acceptable solution for now? (see 
> @lukefromdc <https://github.com/lukefromdc>'s screenshot above)

It looks good to me.

> Also to note that the HighContrast theme uses the Adwaita icon set by 
> default, rather than the HighContrast icon set... this should also be 
> addressed, of course.

It has been recently solved: 
https://github.com/mate-desktop/mate-themes/commit/becf3ea2f1ce8b43bd628c9d9c7af9aa9e97fd1b

Best regards,
Alex.

lukefromdc:
If this is OK for  @alexarnaud than it's good enough for me. I prefer in accessability questions to defer to the judgement of those who are directly affected.

raveit65:
HC theme looks OK for me with fedora 28.
wnck-switcher-hc
Other gtk+ themes are fine too, and our themes use their own styling.

alexarnaud:
@raveit65 @vkareh Do you think we could backport this fix for 1.20 ?

Best regards,
Alex.

vkareh:
Yes, this fixes an existing bug, so we should definitely backport this

Issue #760, opened 2018-04-01
Submenu appears detached with about 200px below the usual position


dertom95:
#### Expected behaviour
selecting any panel's toplevel menu-item(Application,Places,System) will lead to a submenu popping up right below the toplevel menu.

#### Actual behaviour
The submenu appears detached about 150px below the expected position (also applies for tooltips)

#### Steps to reproduce the behaviour
Select any toplevel menu-item.
I installed 18.03 yesterday and it worked great (still does). Once I installed (via Software boutique) "Nexuiz" and "Super Tux Cart" the problem occured. Before it worked. I guess some dependency of one of those two is making the problem. I guess it is GTK2/3 related. I had same issue with eclipse on 16.04 and there was some fix to use gtk2 instead of gtk3(?) (Something like that)

![like this](https://www.dropbox.com/s/e654m2rpupe4mil/Untitled.png?dl=0)

#### MATE general version
Mate Desktop 1.20.1

PS: I hope this is the right repo at all...if not sry for that

lukefromdc:
Does not happen here with MATE 1.20 from git master, Debian Unstable, and neither Nexuiz and Super Tux Cart on my system.  Does this happen only after one of those apps has been run, or any time they are installed?

Yes, these menus are from the mate-panel package, my guess is something is corrupting the calculations for their vertical positions.

dertom95:
It happend right after installing those games....

lukefromdc:
Wonder if they are somehow specifying menu entries incorrectly, such that space is allocated but nothing displayed

dertom95:
Hmm,...I doubt this is a mate only problem. I just checked and eclipse have this problem as well (also gtk). I think somehow gtk2/3 are interfering each other? I just checked and this is the gtk-situation on my system atm:
``dpkg -l libgtk* | grep -e '^i' | grep -e 'libgtk-*[0-9]'
ii  libgtk-3-0:amd64             3.22.29-2ubuntu1  amd64        GTK+ graphical user interface library
ii  libgtk-3-bin                 3.22.29-2ubuntu1  amd64        programs for the GTK+ graphical user interface library
ii  libgtk-3-common              3.22.29-2ubuntu1  all          common files for the GTK+ graphical user interface library
ii  libgtk2-perl                 2:1.24992-1build1 amd64        Perl interface to the 2.x series of the Gimp Toolkit library
[...]

PR #816, opened 2018-06-08
struts: Create struts for panels on inside edges of monitors

### LINKED ISSUE #135: Window maximized wrong with seconday screen on top
I usually have a secondary screen on top of my primary.

When the seconday screen is active, maximizing a window on the primary does not take into account the bar (which is on top of the primary) anymore, and a part of the window is hidden behind the bar 
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/4836759-window-maximized-wrong-with-seconday-screen-on-top?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F841233&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #303: Windows maximize 'under' panels
In certain setups (e.g. dual monitor positioned vertically) I notice that the bottom panel of the top monitor and the top panel of the bottom 'cover' any maximized windows. 

It seems that the area that these panels take up are denoted as available space for windows, while it should not.
## 

<bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/14602735-windows-maximize-under-panels?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #803: Panel issues in dual screen layouts (overlapping window decoration, broken hiding behavior) [$115]
#### Expected behaviour

Unless using maximus, whatever the screen layout is, the title bar of a maximized window should be displayed.

#### Actual behaviour

In a dual screen environment and vertical layout (external screen on top of primary), maximized windows on the primary do not have a title bar.

More precisely, the top bar overlaps on the title bar, hiding it.

It can be verified by setting the top bar to a small size so that a tiny portion of the title bar actually shows up.

1. Plug in an external monitor
2. Set a vertical layout with the external monitor on top of the primary
3. Put a panel at the top of the primary screen
4. Maximize any window in the primary monitor
5. The title bar is hidden by the panel

Normal maximized window in single monitor layout:

capture du 2018-05-11 00-52-58

Maximized window with decoration overlapped by the panel in dual screen vertical layout:

capture du 2018-05-11 00-53-18

#### Linux Distribution

Ubuntu Mate 18.04

vkareh:
This is a hack to provide partial struts for panels that are positioned on the inside edge of a multi-monitor setup (e.g. top panel on bottom monitor, or left panel on right-side monitor).

Fixes #303
Partially fixes #803
Fixes mate-desktop/marco/issues/135

### DIFF
--- a/mate-panel/panel-struts.c
+++ b/mate-panel/panel-struts.c
@@ -250,13 +250,15 @@ panel_struts_allocate_struts (PanelToplevel *toplevel,
 void
 panel_struts_set_window_hint (PanelToplevel *toplevel)
 {
-	GtkWidget  *widget;
-	PanelStrut *strut;
-	int         strut_size;
-	int         monitor_x, monitor_y, monitor_width, monitor_height;
-	int         screen_width, screen_height;
-	int         leftmost, rightmost, topmost, bottommost;
-	int         scale;
+	GtkWidget       *widget;
+	PanelStrut      *strut;
+	int              strut_size;
+	int              monitor_x, monitor_y, monitor_width, monitor_height;
+	int              screen_width, screen_height;
+	int              leftmost, rightmost, topmost, bottommost;
+	int              scale;
+	PanelOrientation strut_orientation;
+	int              strut_start, strut_end;
 
 	widget = GTK_WIDGET (toplevel);
 
@@ -288,26 +290,69 @@ panel_struts_set_window_hint (PanelToplevel *toplevel)
                                                  &topmost,
                                                  &bottommost);
 
+	/* HACK: Here we purposefully break freedesktop wm-spec standard for _NET_WM_STRUT_PARTIAL
+	 * for multi-monitor setups. Since _NET_WM_STRUT_PARTIAL only allow struts along the edges
+	 * of the screen (rather than each monitor), panels positioned on the inside edge of a
+	 * monitor (e.g. top panel on the bottom-most monitor, left panel on the right-most
+	 * monitor, etc), cannot have proper struts. We "fix" this by setting a partial strut that
+	 * is orthogonal to the "correct" one, and pivot the dimensions for it.
+	 *
+	 * For example: a top panel defaults to having a strut that is 24px high, a start
+	 * coordinate of 0px, and an end coordinate of 1920px. If this panel is on the bottom-most
+	 * monitor, and thus no longer at the edge of the screen, we instead "fix" it here by
+	 * applying struts as though it was a *left* panel, that is 1920px wide, with a start
+	 * coordinate of 1080px (top monitor height) and an end coordinate of (1080px + 24px).
+	 *
+	 * This hack only works properly for expanded panels, or panels that start at either edge
+	 * of the monitor, since otherwise they require floating struts, which are not really a
+	 * thing. This last part we just ignore and let the strut exist snapped to the far edge of
+	 * the monitor, which only mildly affects edge snapping of non-maximized windows against
+	 * the floating panel.
+	 */
+	strut_start = strut->allocated_strut_start;
+	strut_end = strut->allocated_strut_end;
+	strut_orientation = strut->orientation;
+
 	switch (strut->orientation) {
 	case PANEL_ORIENTATION_TOP:
-		if (monitor_y > 0)
+		if (!topmost) {
+			strut_orientation = strut_start ? PANEL_ORIENTATION_RIGHT : PANEL_ORIENTATION_LEFT;
+			strut_start = monitor_y;
+			strut_end = (strut_size / scale) + strut_start - 1;
+			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+		} else if (monitor_y > 0) {
 			strut_size += monitor_y;
-		if (!topmost) strut_size = 0;
+		}
 		break;
 	case PANEL_ORIENTATION_BOTTOM:
-		if (monitor_y + monitor_height < screen_height)
+		if (!bottommost) {
+			strut_orientation = strut_start ? PANEL_ORIENTATION_RIGHT : PANEL_ORIENTATION_LEFT;
+			strut_end = monitor_height - 1;
+			strut_start = strut_end - (strut_size / scale) + 1;
+			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+		} else if (monitor_y + monitor_height < screen_height) {
 			strut_size += screen_height - (monitor_y + monitor_height);
-		if (!bottommost) strut_size = 0;
+		}
 		break;
 	case PANEL_ORIENTATION_LEFT:
-		if (leftmost && monitor_x > 0)
+		if (!leftmost) {
+			strut_orientation = strut_start ? PANEL_ORIENTATION_BOTTOM : PANEL_ORIENTATION_TOP;
+			strut_start = monitor_x;
+			strut_end = (strut_size / scale) + strut_start - 1;
+			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+		} else if (monitor_x > 0) {
 			strut_size += monitor_x;
-		if (!leftmost) strut_size = 0;
+		}
 		break;
 	case PANEL_ORIENTATION_RIGHT:
-		if (monitor_x + monitor_width < screen_width)
+		if (!rightmost) {
+			strut_orientation = strut_start ? PANEL_ORIENTATION_BOTTOM : PANEL_ORIENTATION_TOP;
+			strut_end = monitor_width - 1;
+			strut_start = strut_end - (strut_size / scale) + 1;
+			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+		} else if (monitor_x + monitor_width < screen_width) {
 			strut_size += screen_width - (monitor_x + monitor_width);
-		if (!rightmost) strut_size = 0;
+		}
 		break;
 	default:
 		g_assert_not_reached ();
@@ -315,10 +360,10 @@ panel_struts_set_window_hint (PanelToplevel *toplevel)
 	}
 
 	panel_xutils_set_strut (gtk_widget_get_window (widget),
-				strut->orientation,
+				strut_orientation,
 				strut_size,
-				strut->allocated_strut_start * scale,
-				strut->allocated_strut_end * scale);
+				strut_start * scale,
+				strut_end * scale);
 }
 
 void
--- a/mate-panel/panel-struts.c
+++ b/mate-panel/panel-struts.c
@@ -259,6 +259,8 @@ panel_struts_set_window_hint (PanelToplevel *toplevel)
 	int              scale;
 	PanelOrientation strut_orientation;
 	int              strut_start, strut_end;
+	const char      *wm_name;
+	gboolean         wm_compatible;
 
 	widget = GTK_WIDGET (toplevel);
 
@@ -308,48 +310,72 @@ panel_struts_set_window_hint (PanelToplevel *toplevel)
 	 * thing. This last part we just ignore and let the strut exist snapped to the far edge of
 	 * the monitor, which only mildly affects edge snapping of non-maximized windows against
 	 * the floating panel.
+	 *
+	 * This hack will be removed once _NET_WM_STRUT_AREA is implemented.
 	 */
 	strut_start = strut->allocated_strut_start;
 	strut_end = strut->allocated_strut_end;
 	strut_orientation = strut->orientation;
 
+	/* Only "Metacity" and "Metacity (Marco)" are compatible with this workaround, so we set a
+	 * flag to make sure that other window managers get the standard behavior.
+	 */
+	wm_name = gdk_x11_screen_get_window_manager_name (strut->screen);
+	wm_compatible = g_str_has_prefix (wm_name, "Metacity");
+
 	switch (strut->orientation) {
 	case PANEL_ORIENTATION_TOP:
 		if (!topmost) {
-			strut_orientation = strut_start ? PANEL_ORIENTATION_RIGHT : PANEL_ORIENTATION_LEFT;
-			strut_start = monitor_y;
-			strut_end = (strut_size / scale) + strut_start - 1;
-			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			if (wm_compatible) {
+				strut_orientation = strut_start ? PANEL_ORIENTATION_RIGHT : PANEL_ORIENTATION_LEFT;
+				strut_start = monitor_y;
+				strut_end = (strut_size / scale) + strut_start - 1;
+				strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			} else {
+				strut_size = 0;
+			}
 		} else if (monitor_y > 0) {
 			strut_size += monitor_y;
 		}
 		break;
 	case PANEL_ORIENTATION_BOTTOM:
 		if (!bottommost) {
-			strut_orientation = strut_start ? PANEL_ORIENTATION_RIGHT : PANEL_ORIENTATION_LEFT;
-			strut_end = monitor_height - 1;
-			strut_start = strut_end - (strut_size / scale) + 1;
-			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			if (wm_compatible) {
+				strut_orientation = strut_start ? PANEL_ORIENTATION_RIGHT : PANEL_ORIENTATION_LEFT;
+				strut_end = monitor_height - 1;
+				strut_start = strut_end - (strut_size / scale) + 1;
+				strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			} else {
+				strut_size = 0;
+			}
 		} else if (monitor_y + monitor_height < screen_height) {
 			strut_size += screen_height - (monitor_y + monitor_height);
 		}
 		break;
 	case PANEL_ORIENTATION_LEFT:
 		if (!leftmost) {
-			strut_orientation = strut_start ? PANEL_ORIENTATION_BOTTOM : PANEL_ORIENTATION_TOP;
-			strut_start = monitor_x;
-			strut_end = (strut_size / scale) + strut_start - 1;
-			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			if (wm_compatible) {
+				strut_orientation = strut_start ? PANEL_ORIENTATION_BOTTOM : PANEL_ORIENTATION_TOP;
+				strut_start = monitor_x;
+				strut_end = (strut_size / scale) + strut_start - 1;
+				strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			} else {
+				strut_size = 0;
+			}
 		} else if (monitor_x > 0) {
 			strut_size += monitor_x;
 		}
 		break;
 	case PANEL_ORIENTATION_RIGHT:
 		if (!rightmost) {
-			strut_orientation = strut_start ? PANEL_ORIENTATION_BOTTOM : PANEL_ORIENTATION_TOP;
-			strut_end = monitor_width - 1;
-			strut_start = strut_end - (strut_size / scale) + 1;
-			strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			if (wm_compatible) {
+				strut_orientation = strut_start ? PANEL_ORIENTATION_BOTTOM : PANEL_ORIENTATION_TOP;
+				strut_end = monitor_width - 1;
+				strut_start = strut_end - (strut_size / scale) + 1;
+				strut_size = (1 + strut->allocated_strut_end - strut->allocated_strut_start) * scale;
+			} else {
+				strut_size = 0;
+			}
 		} else if (monitor_x + monitor_width < screen_width) {
 			strut_size += screen_width - (monitor_x + monitor_width);
 		}


lukefromdc (changes requested):
I just tested with a second panel on the right side of a left primary monitor, and a second smaller monitor with ugly results. We have a problem with two unequal sized monitors here:

*On Marco with two screens, struts set aside at what appears to be edge of same width as smaller monitor even though on my setup the left monitor is the larger one. Nasty tiling behavior results unless tiling is disabled. A window that has once been on the left side of this boundary cannot again be moved right of it without triggering an attempt to tile if tiling is enabled in Marco.   The strut size appears to be the entire width of the left monitor in excess of the width of the right monitor, presumably plus the width of the panel.

*No effect with ssd apps in compiz-reloaded (my branch) on maximizing under right side panel on left screen. In general struts do not seem to work there for ssd apps, not sure if this is true on compiz-reloaded from master branch. 

With mate-panel from master and only ONE panel, struts work perfectly for CSD apps on my compiz-reloaded branch, Unfortunately, with this PR on CSD apps the oversize/non-spec struts totally confused my branch of compiz-reloaded, causing unconstrained expansion on attempt to maximixe. On trying to maximize an empty gedit window that turned both screens white. This does not occur with gedit, my branch of compiz-reloaded and mate-panel from master as of 6-8-2018. With that combination Gedit maximizes to the right edge of the left screen (under the right panel) but to the top edge of the bottom panel as one would expect.

My working branch of compiz is at 
https://gitlab.com/lukefromdc/compiz/tree/dev-focus+extents_WIP


*No effect on autohide moving panel to other monitor in any WM, this occurs same as before

lukefromdc:
The Pluma window on the left side of the screen in "maximized" to the space taken by the right panel and strut set aside by this. The terminal window is entirely to the right of that and unaffected barring an attempt to maximize. Desktop icons temporarily turned off to redact desktop contents.

desktop_noicons_exess_struts

vkareh:
Damn that's ugly - I'll force different resolutions and re-test (both my monitors are the same resolution)

vkareh:
@lukefromdc - okay, I realized I was calculating some edge coordinates incorrectly. I updated the commit. This seems to fix the issue for marco, I have no issues with CSD windows (will test more just in case).

There were some really weird behaviors when positioning the primary screen either under or to the right of the secondary (which inverts the order of coordinates), whenever the screens were different sizes AND the position did not snap to either edge of the secondary screen. This behavior is also present on master, so I did not do anything about it (will probably try to address in a separate pull request).

I haven't checked Compiz yet - I'll be testing this there now.

vkareh:
@lukefromdc - with Compiz from the Ubuntu 18.04 archives, my fix doesn't work - seems like Compiz doesn't respect `_NET_WM_STRUT_PARTIAL` or something along those lines...

With your branch, I was able to compile successfully, but it crashes immediately upon running - any hints on how to run this locally? Should I install to `/usr` instead of `/usr/local`?

lukefromdc:
Retest results with latest changes here, my branch of compiz again, and Marco:
One monitor, compiz:

Gedit follows struts, maximizes fine with ONE monitor
Pluma also maximizes properly in compiz-reloaded with ONE monitor
Both bottom and right side panel edges respected


Two monitors, smaller monitor to right, compiz:

Gedit and Pluma both maximize to right screen edge, under panel.
SSD apps: pluma, ff, Caja partially overlap bottom panel but not by much
CSD apps gedit, gnome-disks over edges of ONE screen but no longer effect the
second screen. No struts respected, bottom panel also covered.

One monitor, marco:

Gedit follows struts, maximizes fine with ONE monitor
Pluma also maximizes properly in compiz-reloaded with ONE monitor
Both bottom and right side panel edges respected

Two monitors, smaller monitor to right, marco:

Gedit and Pluma both maximize to right screen edge, under panel.
SSD apps: pluma, ff, Caja partially overlap bottom panel but not by much
CSD apps gedit, gnome-disks maximize to top edge of panel

lukefromdc:
Somewhat redacted gnome-disks, maximized beyond screen edges in my branch of compiz

gnome-disks_over_maxiumized

vkareh:
@lukefromdc - when trying to install to `/usr`, I get the following error:
```
libtool:   error: error: cannot install 'libdecoration.la' to a directory not ending in /usr/local/lib/compiz
```
I started with `./autogen.sh --prefix=/usr`

lukefromdc:
You probably need to start your build entirely over, clean the build directory or re-extract it first. I build with 
./autogen.sh --prefix=/usr --libdir=/usr/lib/x86_64-linux-gnu  --sysconfdir=/etc --enable-shared=yes

raveit65:
Currently i am trying to reproduce https://github.com/mate-desktop/mate-panel/issues/803
with that setup (without PR).
- Vertical setup and external monitor is top monitor, laptop monitor is bottom and primary.
- Using a bottom and top panel
- All is with non-hidipi resolution

1. I don't find any application where i don't see the window decoration if window is maximize on primary monitor.
2. I can't move any window from bottom monitor to top monitor, --> the top panel on primary monitor is a barrier.

I really like to find a way to reproduce the issues before i can test the fix....

raveit65:
Ok, https://github.com/mate-desktop/mate-panel/issues/303#issuecomment-170754699
i can reproduce.

raveit65:
Ok, i found out why i can't move windows from bottom (primary) monitor to top monitor with vertical setup.
With that monitor position it doesn't work:
mate windows-properties-middle
With that position it works:

raveit65:
Fix works well here with marco as WM.
With compiz-reloaded from fedora repos it gets more worse.
Without the PR only the bottom of a window is missing if top monitor is primary.
With the PR bottom and top of a maximize window is under the panels.
Is something missing in compiz-reloaded?
@vkareh 
Compiz-reloaded should be installed in your fedora VM.

vkareh:
@raveit65 - Is there a way to have two monitors in the VM? I use qemu, but not sure how to do that.

Yeah - it seems like compiz may be missing something? I'll poke around in the Compiz source.

Did you have any issues with CSD windows?

raveit65:
@vkareh 
I found this youtube HowTo. It sounds simple but the guy use one command in a terminal and sadly the video is blurred at this point. I can't read it with my 4 eyes ;)
But maybe you are more lucky than me ;)
https://www.youtube.com/watch?v=UySlAfz7e7Q

CSD windows are fine with marco. I specially tested gnome-disks and dconf-editor.
With compiz they have same problems than **normal** windows.

raveit65:
I forgot to say, using compiz in qemu VM isn't really funny if the image isn't located on a ssd or nvme drive.
Using virtio driver with 3D acceleration helps.

raveit65:
@vkareh 
Using quemu VM with 2 monitors is really easy.
The trick is to edit /etc/libvirt/qemu/vm-name.xml
```
    <video>
      <model type='virtio' heads='2' primary='yes'>
        <acceleration accel3d='yes'/>
      </model>
      <address type='pci' domain='0x0000' bus='0x00' slot='0x02' function='0x0'/>
    </video>
```
This is with virtio driver and 'heads' are the monitors.
Restart libvirtd 'systemctl restart libvirtd.service'
Connect virt-manger with libvirtd again.
Start the VM with virt-manager, but only starting not open!
Virt-manager doesn't support 2 connections.
Use remote-viewer 'remote-viewer spice://localhost:5900 &'
After login to mate session in guest enable the second monitor like in the video.
Move the windows to your monitors and run them in fullscreen.
That's it ;)

vkareh:
Okay, so I have compiz running on Ubuntu MATE, and as long as the monitors are aligned (non-aligned monitors have all sorts of issues in `master` anyway) I can get the same issue you're reporting: the bottom struts now seem to be gone as well.

However, when I try to move a small window to either top or bottom panels, I can definitely get resistance at _both_ panels, so the struts _are_ there, they are just not being respected for maximized windows. This is also true on `master`, so Compiz is doing something there regardless of my patch.

On the other hand, if I use the maximumize plugin, it calculates the remaining workspace area correctly with this patch - so that seems like progress.

The funny thing is that the issue with the missing bottom panel is due to the top panel being re-calculated on my patch. If I comment out the hack for _just_ the top panel, the bottom one behaves correctly. It seems as though the top panel "pushes" the window down so that it ends up hidden behind the bottom panel, but then the size itself doesn't take the panels into account... still researching, but I might end up adding an `if (wm == 'marco')` conditional somewhere :confused:

lukefromdc:
If you watch your desktop background while adding and removing panels, you will see the desktop move to make space for the panels. If its pushing the other panel's struts too, that's a problem

vkareh:
@lukefromdc - that doesn't seem to happen in marco.

I've also tested metacity and it works well with this patch. I think I'm just going to blacklist Compiz from this behavior, as I don't really have much bandwidth to submit patches to Compiz, and since Compiz is used in other desktop environments, this has the potential of breaking stuff. There's also the fact that there are _two_ different compiz projects, so I think for this patch I'm just going to skip Compiz for now...

lukefromdc:
Blacklisting compiz so it doesn't break is good enough for me on this.  Both branches of compiz are very old, very complex, and I KNOW compiz-reloaded is a very small team.

lukefromdc:
If you get this to where is works in Marco and is blacklisted in compiz, I will approve it for merge

muktupavels:
@vkareh Any reason why you try to workaround instead of thinking about proper solution?

I doubt that you will find something that will work more or less good for everyone... You can have 2 monitors with different size that already gives you 3 possible placement options for one side. For side by side monitor placement they can be aligned by top edge, by bottom edge or not aligned. Also remember that users are not limited by 2 monitors...

I am really thinking that we should think about improving / extending specification and work for real fix.

For example right now I am thinking about new `_NET_WM_STRUT_XXX` property and I think that new property should be rectangle with area that is / should be reserved for panel / dock. That is easy part, but most importantly we need to think about backward compatibility... Most important (I think) part here is that WM should be able to produce or even MUST create same `_NET_WORKAREA` with new strut. For WM it should not matter if application use `_NET_WM_STRUT_PARTIAL` or new `_NET_WM_STRUT_XXX` it should produce same  `_NET_WORKAREA`.

For example it could be like this:
`_NET_WM_STRUT_XXX, edge, x, y, width, height, CARDINAL[5]/32`

For `edge` it should be 0 for left, 1 for right, 2 for top and 3 for bottom. This is quick idea and I have not verified that it is enough to produce same result as if we would have used `_NET_WM_STRUT_PARTIAL`. Also I think there MUST be requirement for this new strut. One edge should match monitor edge and if that requirement is not met then WM should ignore strut.

This could be first step and alone it would not be full fix, but now WM will be able to know real area that should be reserved. Knowing these areas WM should be able to not maximize windows under panels. Or in other words WM already can calculate per monitor workareas for use in WM. I guess this could be already big improvement. :)

Of course calculating per monitor workareas we still need to use `_NET_WM_STRUT_PARTIAL` if client does not provide new strut property.

Now that WM can create/calculate work-area for each monitor we need a way to provide this info for toolkits and/or any other clients/apps that make use of this info. For example there could be `_NET_WORKAREA_Mx` property where x is monitor number.

vkareh:
@muktupavels - Thanks for chiming in, I think you're spot-on. This should be part of the spec. I just subscribed to the wm-spec mailing list so that I can at least get more informed before starting the conversation (although that list looks pretty silent at the moment).

In the meantime, I think we could try to implement this new property on the mate desktop and use it as testing grounds for it as a first proof of concept.

I can try to put some verbiage together based on your description above, and following the convention from the xdg-specs repository so that we have something to base it off of. I know you're not a MATE user, but would you be interested in implementing this in gnome/metacity as well?

muktupavels:
Yes, I know that it should be part of the spec. Of course before trying to get it in spec we need to carefully think how this will work and more importantly we need solution that is backward compatible with existing applications, toolkits and window managers. This is what I have now for strut property:

> _NET_WM_STRUT_AREA
> 
> _NET_WM_STRUT_AREA, x, y, width, height, CARDINAL[4]/32
> 
> This property MUST be set by the Client if the window is to reserve space at the edge of the monitor. The property contains 4 cardinals specifying the area of the reserved space. All coordinates are root window coordinates. The client MAY change this property at any time, therefore the Window Manager MUST watch for property notify events if the Window Manager uses this property to assign special semantics to the window.
> 
> The Window Manager MUST ignore _NET_WM_STRUT_PARTIAL and _NET_WM_STRUT if this property is set. This will ensure that Clients can safely set all _NET_WM_STRUT properties.
> 
> The purpose of struts is to reserve space at the borders of the monitors. This is very useful for a docking area, a taskbar or a panel, for instance. The Window Manager should take this reserved area into account when constraining window positions - maximized windows, for example, should not cover that area.
> 
> Struts MUST be specified in root window coordinates, that is, they are not relative to the edges of any view port or Xinerama monitor, but there is one requirement - the Client MUST ensure that one of the largest edge of area is positioned at monitor edge. If this requirement is not met then Window Manager MUST ignore this strut!
> 
> The values in second example in _NET_WM_STRUT_PARTIAL description converted to _NET_WM_STRUT_AREA would be 1280 for x, 718 for y, 1024 for width and 50 for height. In this example width is larger then height that means that Window Manager must check if panel is at top or bottom edge of monitor. 718 + 50 == 768, requirement is met.
> 
> The Window Manager MUST ignore struts that are at edges between two monitors when calculatating _NET_WORKAREA.

This actually might be enough for strut property. Now we need something for workareas. I every time forgot that existing _NET_WORKAREA included workarea for each desktop / workspace. Currently I see two options:
- _NET_WORKAREAS that would start with number of monitors followed by all workspaces for first monitor in same form as it is in _NET_WORKAREA, then next monitor...
- _NET_WORKAREA_M0 that would be same _NET_WORKAREA, but for monitor 0, _NET_WORKAREA_M1 for second monitor and so on.

Then should decide if values are in root coordinates or relative to each monitor...

> I know you're not a MATE user, but would you be interested in implementing this in gnome/metacity as well?

Yes, otherwise I would not spend time on this.

I looked at `metacity` and it seems that there are already per monitor workareas - `meta_workspace_get_work_area_for_xinerama`. So we already have info / data to use for _NET_WORKAREA_M0...

For example if two monitors are top edge aligned and has two different heights and you have bottom panels on both monitors you probably have noticed that there is gap between panel and menus. That is because of workarea. If gtk would have same info as metacity then it would just work.

So I think we should start with workarea thing, because we already have needed info and we can work on fix for gtk. I will try to set workarea info for monitors and then will look in gtk to make use of that. If I will succeed then I will create merge proposals for mutter, gtk and metacity. :)

raveit65:
Guys, nice idea, but this sounds that it will take some months before it is ready for production.
So why not using this code from PR as temporary bugfix, which can be reverted when you'r e ready with a new global API?
No need to let users live with a bug.

muktupavels:
Well I have not tried to understand what this hack does, but I think that it most likely will "fix" something and at same time introduce new unexpected bugs, because you are unlikely to test all possible monitor placements + all possible panel placements in all configurations...

Also I already have quick patch for `metacity` and also for `gtk-3-22` branch that makes use of `_NET_WORKAREA_Mn` property. And quick test shows that it does fix gtk menus and/or tooltips for bottom panel on larger monitor. To make it clearer, this would fix #760 bug.

Will try to clean up and create proper commit messages and then I will push test branches to metacity and gtk. :) Then up to you to make similar change to marco and rebuild gtk for testing.

raveit65:
Why not making a PR for marco?

vkareh:
@raveit65 - we can always cherry-pick those changes. I'll continue on this hack for Marco as a temporary measure, but I'll work on these spec changes and try them out on MATE.

muktupavels:
https://gitlab.gnome.org/GNOME/mutter/commits/wip/muktupavels/work-areas
https://gitlab.gnome.org/GNOME/gtk/commits/wip/muktupavels/work-areas

lukefromdc:
Looks like GTK 3.22.31 will have a lot of changes:
https://gitlab.gnome.org/GNOME/gtk/commits/wip/muktupavels/work-areas
will be necessary to use the preferred fix though it is not yet merged to master

 Another one (which has been merged)
 c8b0a0dea5cd129d8e023443dde5e44319d4631d
resolves https://bugzilla.gnome.org/show_bug.cgi?id=765327 "GtkPlug scaled to half the expected size on HiDPI screens" and may require revisiting some of our work on scaling out-of-process applets. My guess is we'd have to version check for 3.22.31 but builds with it would be far less buggy in HiDPI.

muktupavels:
1) https://gitlab.gnome.org/GNOME/gnome-panel/merge_requests/3
This is all I needed in `gnome-panel`.

2) https://gitlab.gnome.org/GNOME/metacity/merge_requests/1
This include not required changes. I thought that I will have to deal with screen vs monitor struts, but it **seems** that `metacity` works fine without more changes.

3) https://gitlab.gnome.org/GNOME/mutter/commits/wip/muktupavels/work-areas
No merge request, still needs `_NET_WM_STRUT_AREA` changes.

4) https://gitlab.gnome.org/GNOME/gtk/tree/wip/muktupavels/work-areas
Also no merge request, but seems that it might not require more changes.

My testing shows that this works, but obviously I have not tested many monitor and/or panel configurations to confirm that this works 100%.

vkareh:
@lukefromdc, @raveit65 - I've updated this to only affect Marco and Metacity.

raveit65 (approved):
Does what it promised for marco and metacity WM.
With compiz-reloaded the behaviour is same like before.
Ready to go.

lukefromdc (approved):

lukefromdc:
This works, I found one nuisance issue remaining: if you change window managers, you must restart mate-panel before the new window manager is recognized. Otherwise, you get the behavior expected for the previous window manager. In other words, this is prevented from working after a compiz-marco WM switch, and if (as is common) someone starts their session with marco and switches to compiz-reloaded, they will get quite a suprise if they try to maximize a CSD app with a panel on an inside edge.

raveit65:
> This works, I found one nuisance issue remaining: if you change window managers, you must restart mate-panel before the new window manager is recognized. Otherwise, you get the behavior expected for the previous window manager.

I think best way to use compiz in MATE is to start it with session.
```
[rave@satellite ~]$ gsettings get org.mate.session.required-components windowmanager
'compiz'
```
This should solve the problem.

lukefromdc:
If we are going to merge this without fixing the restart issue, we need to document the need to restart the panel on changing WM's. End users won't know this and may assume they have to restart their session to get out of the problems. Will probably trigger bug reports against compiz.

raveit65:
Where should we document this?
In the meanwhile i will push that to fedora. As upstream and downstream maintainer of compiz-reloaded i am happy to inform users about this very minimal side-effect.
I don't expect that users switch the WM whole the day.

vkareh:
Hi, sorry guys - it's been a crazy week at work and I have family staying over. I haven't had time to finish this, but will push something as soon as possible.

It would annoy me to have to restart the panel when changing WMs, and users will be quite confused and report bugs about it. I'll try to find and push a solution today.

vkareh:
@lukefromdc - I ran some tests and turns out that whenever I change the window manager, the new window hints are applied correctly. How are you changing window managers that it breaks for you? I'm running `marco --replace &`, `metacity --replace &`, and `compiz --replace &` and maximizing windows work as expected for those window managers (marco/metacity with the new struts, compiz without).

I tried connecting the GdkScreen to a `window_manager_changed` signal, but it behaves in the same way already, as the window hints are redrawn when the window manager is changed (the panel being a window itself, of course)

lukefromdc:
What I got (with an inside edge panel and two monitors) was that if I started with compiz-reloaded and ran "marco --replace" the new panel struts did not work until the panel was restarted. If I started the panel with marco running and then ran "compiz-replace" I got unconstrained expansion on maximize of CSD apps. Possible this may be specific to compiz-reloaded and not the 0.9 branch of compiz, so if using compiz 0.9 you might not see it?  There is also a longstanding issue in compiz-reloaded where multimonitor setups with two different size monitors get overmaximization of CSD windows on the smaller monitor.

Will run another test just in case, but I am strongly opposed to letting anything out the door that will break use of "compiz-replace" so badly that some end users on maximizing a CSD app will not know what to do other than power cycle or (if they can terminate the x server) end their session.

lukefromdc:
Also note that my tests are with two monitors horizontally next to oneanother, not vertically one above the other

lukefromdc:
I would like to test connecting the window_manager_changed signal, how did you do it? Since I have the problem and you do not, it might fix the issue here?

lukefromdc:
A retest without the window_manager_changed (I do not know how to apply it), gave me the same problems as before. Note that I was using "compiz --replace" from the panel run dialog and not "compiz --replace &" in terminal though results should be the same. Monitors are small monitor to the right of a large monitor. Any chance the ugly mess in compiz-reloaded appears only when there are two different monitors? I do not have two identical monitors of any size to test.

lukefromdc:
For some reason, with this version of mate-panel or with Master I get this error when starting the panel under Marco or switching to Marco. I never see it with compiz:
`
(mate-panel:3064): GLib-GObject-CRITICAL **: 22:33:13.491: g_object_set: assertion 'G_IS_OBJECT (object)' failed
`

lukefromdc:
I put a g_message test message into panel_struts_allocate_struts, and found it runs when the panel is first started or when a panel is moved or resized. It does not run when the window-manager is changed, but that is as expected. If we could get it to re-run when the window manager is changed this would be ready to go.

lukefromdc:
panel_struts_set_window_hint does run multiple times on starting the panel-or on changing window managers, so we are getting the new window manager name and it may be enough to recall panel_struts_allocate_struts from it, though some way of not doing that 8 times when there are two panels would be nice. Maybe having it cache the window manager name and re-run only if it has changed?

vkareh:
Okay, that's helpful, I was looking inside the set_window_hint function instead of the allocate_struts one. I can also see the former being called a bunch of times unnecessarily. Maybe we can wrap the call in a queue/idle cycle, effectively debouncing it.

I'll add the notification callback for `panel_struts_allocate_struts` and push again early tomorrow. Thanks again for testing this!

vkareh:
To note, also, I've been working a bit on a proof of concept implementation of the previous conversation with @muktupavels about adding a new _NET_WM_STRUT_* property, and so far it looks promising. It will take longer to get a proper solution, but it will be able to replace this pull request eventually with a much nicer and cleaner version that won't need all the conditionals for different window managers. I'll probably open a feature request issue tomorrow so that we can move the conversation over there.

lukefromdc:
I did some testing with a bunch of g_message lines, and found that set_window_hint function is indeed where this is set-and that the condtionals if(if/else) elseif blocks where wm_compatible is used to set or unset struts are not being run everytime this function runs. We get a long list of runs, on starting the panel those blocks execute, but on later runs and ALL runs after a WM change they do not. This is getting called from panel_toplevel_size_request, which calls panel toplevel update geometry and I am trying to trace the flow to find out where this is breaking down. The window manager name I know is being successfully detected on change, because I have g_message lines printing it out

lukefromdc:
I tried calling panel_struts_allocate_struts from the set_window_hint function and it worked but had absolutely NO effect. On changing window manager back and forth, the conditional blocks that never executed again had somehow set struts on startup that are never changed, because those blocks never get re-run. Even affirmatively setting struts to zero on a no-match didn't help for some reason

lukefromdc:
I finally found something that seemd to work, no idea in the world if it is good or bad code or anywhere near correct but it behaves well, even with compiz:

https://github.com/mate-desktop/mate-panel/commit/43015c60809331879417932984b6dddb633c2a14
This is a very simplified subset of the changes in the PR, applying the 
strut_size += screen_width - (monitor_x + monitor_width);
type of blocks(there are four of them, each slightly different for the different panel positions) unconditionally. I have only tested this with a 1080X1920 monitor as primary and a 1024X768 as a secondary to the right. Trying to reset it as a top monitor corrupted rendering of all the existing windows on my screen, so gave up on that.

None the less, this gave maximization to and not beyond panel edges for both inside and outside panels, in any positions, with both Marco and Compiz, and both SSD and CSD windows. I had exactly none of the bugs I got with earlier tests here.

An autohidden panel does set a strut of zero, not sure if that is intended or shared with this PR's code. When recalled it comes up over a maximized window instead of forcing that window to move. Disabling autohide will push any windows in the panel space back to make room. Autohidden inside panels still show up on the adjacent screen, as with this PR's existing code.

lukefromdc:
I created alternate PR https://github.com/mate-desktop/mate-panel/pull/821 as the branch can merge into master automatically but cannot be auto-merged into this PR due to conflicts.  It can be tested and merged, used as a starting point for more work on this branch, whatever folks want to do with it..

vkareh:
@lukefromdc - that branch shouldn't work, as it doesn't actually take into account the inside edges, so an inside panel will set the struts on the corresponding outside edge instead (if you tested with multiple panels, sometimes the issue is not visible as struts overlap each other). That's why this is such a weird hack, it doesn't actually set inside-edge struts, instead it pivots struts so that they have similar geometry, but orthogonal orientation.

I'm testing a version with this call inside `panel-toplevel.c`:
```c
	g_signal_connect(gtk_widget_get_screen(GTK_WIDGET(toplevel)),
		         "window_manager_changed",
		         G_CALLBACK(window_manager_changed),
		         toplevel);
```
That `G_CALLBACK`, all it does is call `panel_toplevel_update_struts` - that forces a full geometry calculation and allocation, but because it happens _during_ the WM transition, the WM is reported as "unknown", and so the panel is moved out to an outside edge. I will work on a solution to that (maybe put a delay to stabilize it a bit). I don't have compiz-reloaded in Ubuntu (just the compiz from the Ubuntu archives), so not sure how to test that yet.

vkareh:
I really want to get this issue fixed, but it seems like it breaks things across window managers, which makes sense because this is such an ugly hack.

I think instead of trying to tweak this to manually account for each window manager and hoping not to break them, I will probably close this and focus on a solution that involves updating the spec to support per-monitor struts. I've already started work on it, and the changes to the panel are minimal and have no effect on window managers that don't support it.

Sorry guys :disappointed: and thanks for testing this so thoroughly - I've gotten a lot of insight on how to fix this thanks to you :relieved:

lukefromdc:
I agree: breaking an existing specification is likely to create issues all over the place, and we have no way to test them all.

raveit65:
@vkareh 
I don't see any progress with the PR for gtk+ since several months.
https://gitlab.gnome.org/GNOME/gtk/merge_requests/198
And it sounds like that gtk+ devs doesn't like a new specification.
Maybe we should use what you did here for the moment.
We can always revert your solution when gtk+ is ready to merge @muktupavels PR.
Honestly , i don't see any reason to let our users bleeding only because of gtk+ devs.
And i am using your commit without any problems the whole summer ;)

vkareh:
that's reassuring to hear that you've had success with this commit :)

I can try to revive it - Based on some early trials, I suspect that we don't need the GTK+ patches, we can update how we calculate these on Marco directly, then update the panel code accordingly. It would be much cleaner (as compared with the panel patch here) and that way we can make sure it doesn't break flow on Compiz.

And yes, when GTK+ is finally patched we can revisit this

raveit65:
I thought @muktupavels  idea of adding new _NET_WM_STRUT_XXX depends of gtk3 commit?

raveit65:
@vkareh 
3 months later and no reaction from gtk+ developer
https://gitlab.gnome.org/GNOME/gtk/merge_requests/198#note_317450
Seems my personal KungFu from fedora doesn't help here.
Can we have a proper solution without gtk+ changes for 1.22 in January?

alexarnaud:
Le 21/12/2018 à 17:53, raveit65 a écrit :
> 3 months later and no reaction from gtk+ developer
> https://gitlab.gnome.org/GNOME/gtk/merge_requests/198#note_317450
> Seems my personal KungFu from fedora doesn't help here.
> Can we have a proper solution without gtk+ changes for 1.22 in January?

I've not followed this issue but the merge request seems to be a changed 
so I doubt it could be shipped into GTK 3.24 branch anyway. It could be 
part of GTK4.

Best regards,
Alex.

muktupavels:
There is no proper solution without spec changes...

It won't be part of 3.24 or 4 until there is good solution and it is accepted as spec update. If / when that will happen I don't see any reason why it would not be accepted for 3.24. :)

The problem apparently is that I am only one who is interested in proper solution... Please don't waste time searching partial workarounds. Lets focus on real problem and try to find best solution.

For example at this point I am not sure if my new proposal is better... So I would prefer to hear ideas from others and hopefully together we will be able to find good solution.

muktupavels:
Where did he say no?

raveit65:
Did he support your PR?

raveit65:
My experience is that it goes very fast if he agree to a request.

muktupavels:
Have you tested all possible configurations? Have you tested GTK apps, QT apps and any app that could directly read/use _NET_WORKAREA? Are you ready to "fix" bug and then get new bugs that might be even worse then this?

Have you or anyone else checked how this hack affects generated `_NET_WORKAREA`? Specially for you I build `mate-panel` with two commits from this pull request. My monitor configuration for only one test:
screenshot from 2018-12-22 21-26-17

First monitor (800x600) is empty, second monitor (1336x1008) is primary monitor and has both - top and bottom panels. Do you want to guess generated workarea? It was:

> alberts@alberts-VirtualBox:~$ xprop -root | grep "_NET_WORKAREA(CARDINAL)"
> _NET_WORKAREA(CARDINAL) = 0, 624, 1312, 960, 0, 624, 1312, 960, 0, 624, 1312, 960, 0, 624, 1312, 960

Can you spot what is wrong? Workarea effectively excludes first monitor... It might appear as it is working because GTK ignores workarea on non-primary monitors. Set first monitor as primary and now you can play with menus. I used gnome-terminal... Menu from terminal opens over bottom panel. But it might be even worse under some specific configuration.

As you see there can be and will be new problems even with GTK+ apps and I don't know what QT does with workareas. Not to speak about apps that might directly use `_NET_WORKAREA`...

If you want hackish solution you just need to use custom property to communicate better info about structs to wm in way that will produce same `_NET_WORKAREA`. In short you could fix problem with minimizing under panels without spec updates, but it will be custom fix limited to mate-panel and marco. and GTK still would put menus over panels.

raveit65:
....let see what happens
https://src.fedoraproject.org/cgit/rpms/mate-panel.git/commit/?h=f28
If breaks user experience i can simply revert it ;)
In mean while we can wait another half a year or years that gtk+ will take notice from your PR.

raveit65:
> If you want hackish solution you just need to use custom property to communicate better info about structs to wm in way that will produce same _NET_WORKAREA. In short you could fix problem with minimizing under panels without spec updates, but it will be custom fix limited to mate-panel and marco. and GTK still would put menus over panels.

I am sure you know it better, as always.
But how should should if gtk+ don't merge your request? 
And i if it works who do a new PR and when?

muktupavels:
I don't know better! I am just saying what I think based on what I know - it does not mean that I am right. :) Feel free to prove that I am wrong...

That merge request as is won't be merged. We need to find better solution...

https://mail.gnome.org/archives/wm-spec-list/2018-December/msg00000.html
Lets see if I will get some response to this.

strasharo:
Cross-posting from [mate-control-center#412](https://github.com/mate-desktop/mate-control-center/issues/412), since this PR is referenced there as the origin of the bug even though it was never merged here.

For context: this PR closed unmerged in 2018, but its diff was picked up as a downstream patch by at least Fedora (`mate-panel_0001-struts-Create-struts-for-panels-on-inside-edges-of-m.patch`, still shipped in current Fedora `mate-panel` packages as of 2026) — which is presumably true for other distros too. So the "inside edge struts" logic from this PR is what a lot of real-world MATE users are actually running, despite it never landing upstream.

The bug reported in [mate-control-center#412](https://github.com/mate-desktop/mate-control-center/issues/412): `expand=true` panels get placed at `(monitor.x + monitor.width, monitor.y)` instead of `(monitor.x, monitor.y)` whenever the primary monitor's origin isn't `(0,0)`. I bisected this tonight and it's not limited to vertically-stacked monitors — reproduces in a purely horizontal 3-monitor layout too, and identically across two unrelated GPU/driver stacks, which points at the struts placement math introduced here rather than anything driver-specific.

Given this PR's diff is what's actually shipping to users (even if unofficially), it might be worth another look at the non-zero-origin case at some point — no pressure, just flagging it since it's a real (if niche) papercut for anyone with a multi-monitor primary that isn't leftmost/topmost. Happy to help narrow down the exact line if that'd be useful.

PR #821, opened 2018-06-22
struts: Unconditionally create struts for panels on inside edges of monitors

lukefromdc:
Fix https://github.com/mate-desktop/mate-panel/issues/303
Partial fix for https://github.com/mate-desktop/mate-panel/issues/803

Alternate PR to https://github.com/mate-desktop/mate-panel/pull/816
works in all window managers on system tested. Tested with one monitor and with one 1080x1920 monitor plus one 1024x768 monitor positioned right of the primary. Tested with marco and with compix-reloaded, and switching between them.

### DIFF
--- a/mate-panel/panel-struts.c
+++ b/mate-panel/panel-struts.c
@@ -250,13 +250,15 @@ panel_struts_allocate_struts (PanelToplevel *toplevel,
 void
 panel_struts_set_window_hint (PanelToplevel *toplevel)
 {
-	GtkWidget  *widget;
-	PanelStrut *strut;
-	int         strut_size;
-	int         monitor_x, monitor_y, monitor_width, monitor_height;
-	int         screen_width, screen_height;
-	int         leftmost, rightmost, topmost, bottommost;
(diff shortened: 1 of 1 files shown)


vkareh:
@lukefromdc - thanks for giving this a serious try. I think at this point we're more likely to break things for users with different monitor configurations. I think we should probably put our energy into updating the window manager spec to support per-monitor struts. This way we can support this fix properly and natively without regards to window manager or monitor configuration.

lukefromdc:
OK, I assume we are done with this then until window managers can be updated? Closing

lukefromdc:
Leaving the branch for now in case anyone wants to play with it

lukefromdc:
I found windows respecting inside panel space on my particular multimonitor configuration, but as both of us have said that is ONE configuration. Better to have issues with inside panels than problems like the mess I was getting with gedit. This goes double with the fact that an inside panel interferes with one common use of multimonitors, which is to span things like games (and possibly photos or videos) across multiple monitors.

Issue #617, opened 2017-07-15
Panel is without shadow


TomaszGasior:
#### Expected behaviour
Panel should have shadow such as windows.

#### Actual behaviour
MATE panel have not any shadow. 

Composition is enabled in Marco settings.
Even when I try to restart marco (by `marco --replace`) or mate-panel (by `killall mate-panel`), shadow is not visible.
I can have shadow in panel only by using external compositor such as compton with proper configuration.

#### Package version
marco 1.18.1
mate-panel 1.18.3

#### Linux Distribution
Arch Linux

otwesou:
for what it's worth, the panel has had a shadow at various points in the past (back in the Gnome days, I believe) ... it would be there during one version, gone the next, back again, and it seems to have been gone for years now. i'd love to see it return on a permanent basis :-)

elcste:
I like having the panel blend into my desktop, so I prefer the current lack of shadow. Maybe it can be made an option, in the great MATE tradition?

TomaszGasior:
I agree. Option should be created for it. And this option should work **always* regardless of used compositor or window manager and regardless of components staring order (WM was started before panel or after).

marcxjo:
I can confirm that the panel has a shadow when running Compton. For whatever it matters, Fedora 27, x86-64 architecture, panel v1.19.4.

Could this possibly be a Marco issue?

lukefromdc:
I agree, it must be possible to disable panel shadows, and the issue with transparent panels would have to be fixed

AnwarShah:
Any development on this issue. Shadowless panel is ugliest

lukefromdc:
Shadows are not even possible without either compositing or pseudo-compositing, the latter is used in mate-panel when a user set background is selected and a compositor is not present. The code for that is very complex and to add shadows it would be necessary to enlarge the area copied from the desktop image to accomodate the dimensions of the shadow.   This would require modifying code that has been a source of bugs before when GTK versions changed-and it would also make the panel more resource intensive when running the system(default) background. It would be a major job and we are a small team.

With compositing, a shadow requires that a window take up an area beyond its decoration(if decorated) or contents(if undecorated like a menu or panel). The shadow is rendered on the window in this extra window area beyond the contents or decoration, it is not rendered on the contents of the window beneath.

[...]

Issue #344, opened 2015-08-30
panel missing border (for themes that should have border)


AprilVRose:
Prior to the recent update to Mate 1.10 in Fedora 21, TraditionalOk drew a border at the edge of the panels; in Fedora 22 (and Fedora 21 after the Mate 1.10 update), this border is gone. See attachment Fedora22_TraditionalOk.png which illustrates this by using a background which is the same color as the panel.

![fedora22_traditionalok](https://cloud.githubusercontent.com/assets/14043682/9568878/c03da876-4f25-11e5-8061-c738ca860286.png)

This becomes a larger problem when the panel is set to autohide -- when the user hovers the mouse at the top edge of the screen to bring the panel up, it covers the titlebar of a maximized application, and since the panel is the same color as the application, there's no distinction between the panel and the application window. See attachments comparing Fedora21_Pre-Mate_1.10.png and Fedora22_Mate_1.10.png.

![fedora21_pre-mate_1 10](https://cloud.githubusercontent.com/assets/14043682/9568879/c9a08b86-4f25-11e5-9d0a-1fb6ca3a9230.png)
![fedora22_mate_1 10](https://cloud.githubusercontent.com/assets/14043682/9568881/cfde9f10-4f25-11e5-82a7-bb332a1ba7ac.png)

The same issue can apply with the bottom panel in certain applications.

Steps for reproduction:
1) Change theme to TraditionalOk
2) Maximize the "Appearance Preferences" window (or any window)
3) Set the top panel to Autohide
4) Move your mouse to the top of the screen to ensure the top panel is visible ... in Mate 1.8 you will see a border drawing a distinction between the panel and the window that is under it; in Mate 1.10 there is no such border.

By the way, this issue isn't specific to TraditionalOk -- GreenLaguna is also affected. BlackMate is also affected, but the border color was so close to the panel color in that theme that the difference is not really noticeable. The other themes never had panel borders so are not affected.

In Fedora 21, I can confirm that the change which removed the border is in one of the following packages (or between these versions and the unaffected versions noted below):
mate-panel (1.10.1-1)
 mate-panel-libs (1.10.1-1)
mate-desktop (1.10.1-3)
 mate-desktop-libs (1.10.1-3)
pluma (1.10.2-1)
 pluma-data (1.10.2-1)

The following are the package versions prior to update (not affected by the missing border issue):
mate-panel (1.8.1-4)
 mate-panel-libs (1.8.1-4)
[...]

Issue #193, opened 2014-05-10
Mate Panel Shadow


TaylanTatli:
Mate panel shadow not working on startup. But when i run "marco --replace" it's working.
##

mrnhmath:
Is this still relevant with GTK3 release? Or can it be closed in #617 favour and vice-versa?

otwesou:
@mrnhmath, #617 still applies for me with gtk3. as for this bug, running marco --replace didn't result in a shadow being added for me.

Yupy:
Not working anymore with GTK3 thats why I ditched the MATE desktop I hate that flat look.

I've tried many many ways to make it appear even with compiz doesen't appear.

Ikey from Solus did a pull request: https://github.com/mate-desktop/mate-panel/pull/593 this might help?

raveit65:
And there are open questions with this PR.......

Dolsilwa:
Does anybody has any news on panel shadow in MATE?

Yupy:
@Dolsilwa, I doubt they will ever put back the panel shadow, is a bad choice they removed it :( this made me stop using the MATE DE.

That flat look with some darker wallpapers you can barely see it when the themes have a darker color.

raveit65:
Guys, do you really think we removed it on purpose ?
This is a fallout from switch to gtk+3.
If someone have an idea to fix it we'll include the fix shortly.
Fixes are welcome.....
But if this is a reason for to use another desktop, please do.
Sorry, we can't help you at the moment.

Dolsilwa:
@Yupy You can add panel shadow in MATE through Compton, but it's not a perfect solution (when you maximize an app, panel shadow is drawn on window titlebar).

@raveit65 I know that it was not removed on purpose :-) I know that it's gtk3 transparency fault (or so i understand). And i was just curious, because year ago there were some ideas and some commits done to bring back panel shadow. Right now i'm adding artificial shadow directly to wallpaper (in Gimp), so it's not an issue that would make me stop using MATE :-) MATE is most vesatile GTK-based DE, and is my favourite. That being said, MATE lacks a bit in "looks" department. Right now, the best default look has Solus, partly because when panel is only at the bottom of desktop, you don't need panel shadow. But Ubuntu-MATE (which i use) looks a bit ugly with flat, shadow-less panel at the top of desktop. But with panel shadow, default Ubuntu-MATE just looks a lot better (Ubuntu-MATE 16.10 below).

[...]

PR #828, opened 2018-06-29
panel-toplevel: add position css class

raveit65:
This makes it possible to add theme settings for fixing 
https://github.com/mate-desktop/mate-panel/issues/617
https://github.com/mate-desktop/mate-panel/issues/344
https://github.com/mate-desktop/mate-panel/issues/193

An example css for Menta and other light themes which can be used in
~.config/gtk-3.0/gtk.css
```
.mate-panel-menu-bar.horizontal.top > grid {
    border-style: outset;
    border-width: 0px 0px 1px 0px;
    border-color: rgba(0, 0, 0, 0.23);
}

.mate-panel-menu-bar.horizontal.bottom > grid {
    border-style: outset;
    border-width: 1px 0px 0px 0px;
    border-color: rgba(0, 0, 0, 0.23);
}

.mate-panel-menu-bar.vertical.left > grid {
    border-style: outset;
    border-width: 0px 1px 0px 0px;
    border-color: rgba(0, 0, 0, 0.23);
}

.mate-panel-menu-bar.vertical.right > grid {
    border-style: outset;
    border-width: 0px 0px 0px 1px;
    border-color: rgba(0, 0, 0, 0.23);
}
```

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -246,6 +246,51 @@ static void panel_toplevel_calculate_animation_end_geometry(PanelToplevel *tople
 static void panel_toplevel_update_monitor(PanelToplevel* toplevel);
 static void panel_toplevel_set_monitor_internal(PanelToplevel* toplevel, int monitor, gboolean force_resize);
 
+static void
+update_style_classes (PanelToplevel *toplevel)
+{
+	GtkStyleContext *context;
+
+	context = gtk_widget_get_style_context (GTK_WIDGET (toplevel));
+
+	/*ensure the panel BG can always be themed*/
+	/*Without this gtk3.19/20 cannot set the BG color and resetting the bg to system is not immediately applied*/
+	gtk_style_context_add_class(context,"gnome-panel-menu-bar");
+	gtk_style_context_add_class(context,"mate-panel-menu-bar");
+
+	gtk_style_context_remove_class (context, GTK_STYLE_CLASS_HORIZONTAL);
+	gtk_style_context_remove_class (context, GTK_STYLE_CLASS_VERTICAL);
+	gtk_style_context_remove_class (context, GTK_STYLE_CLASS_RIGHT);
+	gtk_style_context_remove_class (context, GTK_STYLE_CLASS_LEFT);
+	gtk_style_context_remove_class (context, GTK_STYLE_CLASS_TOP);
+	gtk_style_context_remove_class (context, GTK_STYLE_CLASS_BOTTOM);
+
+	switch (toplevel->priv->orientation) {
+	case PANEL_ORIENTATION_TOP:
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_HORIZONTAL);
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_TOP);
+		break;
+
+	case PANEL_ORIENTATION_LEFT:
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_VERTICAL);
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_LEFT);
+		break;
+
+	case PANEL_ORIENTATION_BOTTOM:
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_HORIZONTAL);
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_BOTTOM);
+		break;
+
+	case PANEL_ORIENTATION_RIGHT:
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_VERTICAL);
+		gtk_style_context_add_class (context, GTK_STYLE_CLASS_RIGHT);
+		break;
+
+	default:
+		g_assert_not_reached ();
+		break;
+	}
+}
 
 GSList* panel_toplevel_list_toplevels(void)
 {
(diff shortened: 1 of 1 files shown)


muktupavels:
Your CSS is misleading... If I would read this CSS then I would think that this will affect **only** menu bar applet.  Why toplevel has these classes?

raveit65:
Complete css selector i much longer.
`PanelToplevel.horizontal.top.background.mate-panel-menu-bar.gnome-panel-menu-bar.mate-custom-panel-background > grid {}`

mate-panel-menu-bar.gnome-panel-menu-bar was added in the past for some reasons.
I only added the new position css classes.

lukefromdc:
Years ago (MATE 1.9/1.10 era), when I first started working with mate-panel and GTK3, I found that the main menu from the menubar would follow .mate-panel-menu-bar menu while all other menus followed only the GTK theme default. Thus, I added that style class to the toplevel windows of all the other menus, so they would all follow the same theme as the default menubar's main menu. I then made use of that in my UbuntuStudio_Legacy theme.  During the GTK 1.20 transition, the comments in the code now in question show there were issues with getting the  system background reapplied, and the same style class was applied there. I am NOT sure about this, but I think it may have been somewhere on the toplevel even before then, as this was enough to make the panel respond to the existing themes at the time when deselecting a user selected background.

Anyway, that now leaves us with the single style class .mate-panel-menu-bar used too widely. One possible way of refactoring this would be to shorten it just to .mate-panel but still applying the older .mate-panel-menu-bar style class to the panel menu bar itself(the main menu bar applet). Themes could then be backward compatable by listing both style classes with the use of .mate-panel-menu-bar outside the menubar applet deprecated. 

Changes for such a refactor would be needed in mate-panel, mate-applets, mate-settings-daemon, mate-power-manager, mate-polkit, and all themes. Users would need to change the themes at the same time (or before if backward compatable) as all of these, so at least it would have to be for 3.22 and at that time theme names could advance to 3.24 as GTK also is doing so.

raveit65:
I think `gnome-panel-menu-bar` was added to support older and now outdated gnome themes like Adwaita.
Outdated because gnome themes don't style any application explicit any more.
What we can do is to rename css names to be consistent to gnome-panel.
Eg. PanelToplevel --> panel-toplevel.
But this is independent from PR.

raveit65:
But more important is that i wasn't able to use a box-shadow in css.
It works only if i use `inset`
```
.mate-panel-menu-bar.horizontal.top > grid {
    box-shadow: inset  0px  1px shade (@theme_selected_bg_color, 1.3);
}
```
```
inset  0px  1px shade (@theme_selected_bg_color, 1.3),  -->top
inset  1px  0px shade (@theme_selected_bg_color, 1.3), --> left
inset -1px  0px shade (@theme_selected_bg_color, 1.3),  --> right
inset  0px -1px shade (@theme_selected_bg_color, 1.3);  --> bottom
```
@lukefromdc 
Maybe you are more luckier than me.

lukefromdc:
Not sure how big a window the panel toplevel actually uses, but if the window is the same size as the panel than there is no place to render a box shadow.  I am not an expert on box shadows, but maybe inset renders over the contents and the rest outside the contents and there is no window space there to draw them?

lukefromdc:
The whole _GTK_FRAME_EXTENTS issue in window managers is about rendering the extra space around CSD windows to handle box shadows. The panel is an undecorated window, I don't see any .csd style class applied, so I'm wondering if it gets _GTK_FRAME_EXTENTS of zero px and can't render shadows. A GtkMenu's window cannot be accessed from GtkInspector but those I know act like CSD windows, which is why I had to force transparency on them when the _GTK_FRAME_EXTENTS issue in older compiz version caused them to render in RGB rather than RGBA. Had the toplevel also had this issue, shadows would have worked in Marco with and only with compositing, but transparency of the panel in a system theme would have been impossible with any compiz version except recent compiz-reloaded or the gtk-frame-extents branches.

The only place I ever saw shadows with someone's test branch of mate-panel to support them was rendered underneath a transparent panel rather than outside its borders.

lukefromdc:
Merged the PR as the style classes work and will allow borders to work and if we ever fix the window issue to allow box shadows will would need this anyway

raveit65:
In gtk+-2 versions the shadow was outside, that is what users blame us.
I tried to add other shadow settings from themes, eg from CSD windows settings
`box-shadow: 0 0 0 1px @wm_border, 0 2px 8px 3px @wm_shadow;`
But for some reasons it works only with `inset` here.
Maybe this is a sign for a more general problem.
@muktupavels 
Does shadows works with gnome-panel?

raveit65:
I will cherry-pick that to 1.20 branch.

muktupavels:
Grid is same (?) size as window/toplevel, probably clipped away, but no idea... Maybe something else...

To get outset shadow for panel you probably should do that on `panel-toplevel decoration`, but again I don't know requirements that must be met to get it working. Toplevel should be ARGB window, it must have allocated more height so there is extra space for shadow. This space probably should be set in `_GTK_FRAME_EXTENTS` / `gdk_window_set_shadow_width`:
https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#gdk-window-set-shadow-width

Then you must deal with turning on/off compisiting manager, otherwise shadow area will be black when composting manager is off...

No, `gnome-panel` does not have shadow.

raveit65:
Yeah, i noticed `panel-toplevel decoration` in gtk3-inspector, but it still ignore any shadow setting.
Well, for the moment using an `outset` border-style looks OK for me with Menta themes.

raveit65:
> Grid is same (?) size as window/toplevel, probably clipped away

I think so, for some reasons only `grid` accept a border or shadow. PanelToplevel or PanelFrame don't accept that.....

PR #829, opened 2018-06-29
more GDK deprecation fixes

raveit65:
Please test.

### DIFF
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -442,6 +442,7 @@ na_tray_child_new (GdkScreen *screen,
 {
   XWindowAttributes window_attributes;
   Display *xdisplay;
+  GdkDisplay *display;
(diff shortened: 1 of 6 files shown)


raveit65:
Using compiz-reloaded is really messy here since i use a HIDPI monitor, all commits what @vkareh did for marco needs to be done for compiz too, eg. small minimize/maximize/close icons.
And for some reasons i don't have not more than 5 workspaces with compiz today -_-, but unrelated to this PR. Maybe i should updated libwnck3 for fedora.
Not using compiz is really a shame for a maintainer, but i am tired about asking other people for fixes.
Same as in Mate.
And in the end i am only a good rpm packager not a real C developer.

lukefromdc:
Con confirm issue is with compiz-as similar rendering issues occasionally appear on dialogs, comboboxes, and other popup windows and a bit more often with gtk 3.23. This PR seems fine

PR #832, opened 2018-07-01
replace deprecated gtk_menu_popup

raveit65:
There is one occurrence of `gtk_menu_popup` left in menu.c, but replacing it make problems with that context-menu inside the the menu of menubar and main-menu.
Please test.

### DIFF
--- a/mate-panel/applet.c
+++ b/mate-panel/applet.c
@@ -736,13 +736,7 @@ applet_show_menu (AppletInfo     *info,
 	if (!gtk_widget_get_realized (info->menu))
 		gtk_widget_show (info->menu);
 
-	gtk_menu_popup (GTK_MENU (info->menu),
-			NULL,
-			NULL,
-			(GtkMenuPositionFunc) mate_panel_applet_position_menu,
-			info->widget,
-			event->button,
-			event->time);
+	gtk_menu_popup_at_pointer (GTK_MENU (info->menu), NULL);
 }
 
 static gboolean
--- a/mate-panel/panel-action-protocol.c
+++ b/mate-panel/panel-action-protocol.c
@@ -76,8 +76,7 @@ panel_action_protocol_main_menu (GdkScreen *screen,
 	panel_toplevel_push_autohide_disabler (panel_widget->toplevel);
 
 	gtk_menu_set_screen (GTK_MENU (menu), screen);
-	gtk_menu_popup (GTK_MENU (menu), NULL, NULL,
-			NULL, NULL, 0, activate_time);
+	gtk_menu_popup_at_pointer (GTK_MENU (menu), NULL);
 }
 
 static void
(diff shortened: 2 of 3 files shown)


lukefromdc:
Here's the panel menu bar issue on a vertical panel, you can also see why this applet is not suited to a vertical panel anyway. Desktop icons hidden for the usual privacy reasons

panel_menus_side_by_side

raveit65:
Confirmed, that turns out that we have 2 problems with
`applet-context-menu: replace deprecated gtk_menu_popup' and the menu-bar.
1. Another `GdkGravity widget_anchor` is needed for a vertical panel, `GDK_GRAVITY_NORTH_WEST` should be a better choice.
2. The gap between menu-bar (widget) and menu depends on position at panel.
with menubar at top of vert. panel
menubar-context-menu-vertical-at-top
with menu-bar at middle of vert. panel
menubar-context-menu-vertical-at-middle
with unexpand vert. panel
menubar-context-menu-vertical-unexpand

raveit65:
A test with
```
+	gtk_menu_popup_at_widget (GTK_MENU (info->menu),
+	                          widget,
+	                          GDK_GRAVITY_NORTH_WEST,
+	                          GDK_GRAVITY_NORTH_WEST,
+	                          NULL);
```
moves the menu under the menubar of vertical panel, but only if the menubar is at top of a vertical panel.
So, problem still exists if the menubar is at middle of vertical panel.
And i found out that the problem exists with a horizontal panel too, if the menubar position isn't at the left side of a panel.
menubar-context-menu-horizontal-at-middle

I will try to use `gtk_menu_popup_at_pointer` which is a bit weird for selecting the menus with keyboard navigation and the pointer is somewhere at the monitor.

raveit65:
Ok, i found another fallout caused by
`libmate-panel-applet: replace deprecated gtk_menu_popup `
Right-click menu from na-tray and windows-list applets don#t stay open any more if using pointer event.
I will revert that commit.

For `applet-context-menu: replace deprecated gtk_menu_popup` i use `gtk_menu_popup_at_pointer` now, which fixes problems with menubar applet.
I can also revert this commit if you don't agree with new behavior for using keyboard navigation.

raveit65:
I am fine with that because most usage of the panel is with a pointer.
Of course it is a bit weird if a menu opens under the pointer but the pointer isn't near by the panel.
This affect mostly our accessibility support.
So let's call @alexarnaud
Is this OK for you?

lukefromdc:
Fine by me, we need to make sure we don't break anything for Orca users

raveit65:
Done, last 2 commit can be squashed if we merge this.

raveit65:
@lukefromdc Should i will delete the old gtk_menu_popup branches?

lukefromdc:
They can go, their work is done

raveit65:
At 2 places the deprecations are their, but fixing them causes some issues here.
Branches are deleted.

Issue #746, opened 2018-02-25
notification area applets cut in half vertically 


dimdam:
#### Expected behaviour
full size indicators

#### Actual behaviour
chopped up icons

#### Steps to reproduce the behaviour
Its a very random issue ,happens on booting/rebooting but not all the time .
for example as shown in the picture the sound applet is cut in half :
![2018-02-25 14-44-00](https://user-images.githubusercontent.com/7516919/36645955-3a9f6cfa-1a79-11e8-886e-74b9b0b20027.png)

#### MATE general version
1.20.0

#### Linux Distribution
Manjaro MATE

#### Link to downstream report of your Distribution

https://bugs.manjaro.org/#ticket/zoom/17962

Additional info:

Kernel: 4.14.21-1-MANJARO
Shell: bash 4.4.19
Theme: Adapta-Eta-Maia [GTK2/3]
 Icons: Papirus-Maia [GTK2/3]
CPU: Intel i7-5500U (4) @ 3.000GHz
GPU: AMD Radeon R7 M260/M265
 GPU: Intel HD Graphics 5500

info-cppsp:
dup? #518 

is that the notification area applet or the indicator applet?
(right click - about)

what if you move it more to the left?

lukefromdc:
I see the icon is cut vertically. That is an old bug affecting all tray icons, looks like indicator icons get treated the same as all others. That's a clue as to where in the code the trouble (possibly a failure to (re)allocate enough size for the icon when first drawn must be: in code common to everything in the tray, or else in something that was cut and pasted when GNOME first wrote the code for sni support in gnome-panel. 

Confirm you are using an indicator for the volume control applet and not the default mate-media "mate-volume-control-applet" applet? mate-volume-control-applet is a traditional tray applet without an indicator mode

dimdam:
@info-cppsp @lukefromdc it is a notification area applet .

lukefromdc:
OK, it's not an indicator and has nothing to do with sni support then. I did notice that when I built network-manager-applet with SNI support it never got cut in half again, something which used to be a problem from time to time.  That suggests the code is in the tray icon code separate from sni support code

dimdam:
@lukefromdc  Can you please guide me on where i should get help about this?

dimdam:
Another image showing this random issue:
[...]

PR #833, opened 2018-07-01
Fix tray icon cutting

lukefromdc:
Remove explicit width/height setting and GTK2 leftovers that may be responsible for cutting tray icons vertically.
Possible fix for https://github.com/mate-desktop/mate-panel/issues/746
will test in the laptop which often has this problem

### DIFF
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -115,6 +115,7 @@ na_tray_child_style_set (GtkWidget *widget,
    */
 }
 
+#if !GTK_CHECK_VERSION (3, 23, 0)
 static void
 na_tray_child_get_preferred_width (GtkWidget *widget,
                                    gint      *minimal_width,
@@ -156,6 +157,7 @@ na_tray_child_get_preferred_height (GtkWidget *widget,
   *minimal_height = *minimal_height / scale;
   *natural_height = *natural_height / scale;
 }
+#endif
 
 /* The plug window should completely occupy the area of the child, so we won't
  * get an expose event. But in case we do (the plug unmaps itself, say), this
@@ -382,8 +384,10 @@ na_tray_child_class_init (NaTrayChildClass *klass)
 
   widget_class->style_set = na_tray_child_style_set;
   widget_class->realize = na_tray_child_realize;
+#if !GTK_CHECK_VERSION (3, 23, 0)
   widget_class->get_preferred_width = na_tray_child_get_preferred_width;
   widget_class->get_preferred_height = na_tray_child_get_preferred_height;
+#endif
   widget_class->draw = na_tray_child_draw;
 
   /* we don't really care actually */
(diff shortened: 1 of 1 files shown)


lukefromdc:
Laptop test was inconclusive because it wasn't cutting tray icons today (possibly last round of Debian Unstable updates helped)

raveit65:
Hmm, this gives me gaps between the tray applets in a quick test. (ignore small dropbox icon, non-sni problem with HIDPI)
tray-applet
From Left to Right.   dropbox, kaffeine, KeepassXC, Mate-display and nm applets.

lukefromdc:
I did not get those gaps, in fact I got the exact same spacing as before. I have the xrandr applet (tray),  nm-applet(indicator), and mate-volume-control applet (tray) by default, and also got no issues with Audacious or with mate-polkit. Wonder what was different on your end? I suspected HiDPI so tried window-scaling =2 but still did not get the gaps. Panel top or bottom, HiDPI or not I got the same results. I will now try rolling GTK back to 3.22.30 from 3.23.0 to see if that was the difference.

lukefromdc:
OK, I just found your problem: it exists only with HiDPI, and is fixed by using GTK 3.23. I now can duplicate it, meaning I should be able to fix it. Will play with this some more.

lukefromdc:
Removed the second commit and force-pushed, as it turns out removing that code removed the original fix for the HiDPI case. I didn't originally notice that, because GTK 3.23 also fixes it. The preferred width/height code should not be able to trim any icon narrower than 16px, as it sets that as a minimum.

lukefromdc:
I just found that GTK 3.23.0 behaves enough differently from GTK 3.22 in hidpi mode that both in Master and with just my first commit here, tray icons get cut too narrow on both the left and right sides when window-scaling = 2. 

Since removing the code that causes that (set a minimum width or height and divide by the scale factor) causes a gap between icons with GTK 3.22, I had to insert a gtk version check and remove  na_tray_child_get_preferred_width and  na_tray_child_get_preferred_height when and only when building with GTK 3.23 or later

Thanks to GNOME, we are back to gtk version checking again, though in this case it seems like it was a problem in GTK that caused setting window-scaling = 2 to allocate twice the needed space to the icons anyway.

lukefromdc:
Just had the rightmost icon (at edge of applet) cut on a session start, so this is not done yet

lukefromdc:
Since the HiDPI issues with GTK 3.23 occurred on master too, I think the removed code from the first commit is certainly no longer needed, and the second commit is needed as a fix for GTK 3.23 anyway. Still looking for the last part of the problem but cleaning up old GTK2 leftovers was what fixed the problems in the cpufreq applet

raveit65:
Maybe another team member can reproduce origin issue for a better review?

lukefromdc:
I also found that if the rightmost icon gets cut, anything that forces a panel redraw (such as changing the size) will rerender that icon full size.

lukefromdc:
Merging, as this stops some icon cutting even if not all icon cutting yet. I did get a cut icon for the xrandr popup on a restart with a test build with the tray out of process, followed by proper icon rendering on a bunch of subsequent panel restarts.

I will look at refactoring the whole icon drawing code in the same file, as also has a lot of allocation code and is again leftover from GTK2, which has an entirely different drawing model. Beginning with GTK 3.10, GTK made the backgrounds of some widgets transparent by default and regardless of window managers  by drawing on the window beneath, so some of the low level drawing code might now be replaceable. That will be in a second, follow-on PR to this once I get it working.

PR #834, opened 2018-07-03
Tray: simplify force-redraw, use code that actually works

lukefromdc:
Prevent most corruption of tray icons and reduce by half incomplete rendering of them, replace a never-built code block that could not be built, and replace (deprecated) direct expose handling with gtk_widget_hide/gtk_widget_show_all

Note that this is needed for GTK 3.23 but may help remaining icon issues with GTK 3.22 as well and is simpler code in any case.

### DIFF
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -561,48 +561,11 @@ na_tray_child_force_redraw (NaTrayChild *child)
 
   if (gtk_widget_get_mapped (widget))
     {
-#if 1
-      /* Sending an ExposeEvent might cause redraw problems if the
-       * icon is expecting the server to clear-to-background before
-       * the redraw. It should be ok for GtkStatusIcon or EggTrayIcon.
-       */
-      GdkDisplay *display;
-      Display *xdisplay; 
-      XEvent xev;
-      GdkWindow *plug_window;
-      GtkAllocation allocation;
-
-      display = gtk_widget_get_display (widget);
-      xdisplay = GDK_DISPLAY_XDISPLAY (display);
-
-      plug_window = gtk_socket_get_plug_window (GTK_SOCKET (child));
-      gtk_widget_get_allocation (widget, &allocation);
-
-      xev.xexpose.type = Expose;
-      xev.xexpose.window = GDK_WINDOW_XID (plug_window);
-      xev.xexpose.x = 0;
-      xev.xexpose.y = 0;
-      xev.xexpose.width = allocation.width;
-      xev.xexpose.height = allocation.height;
-      xev.xexpose.count = 0;
-
-      gdk_x11_display_error_trap_push (display);
-      XSendEvent (xdisplay,
-                  xev.xexpose.window,
-                  False, ExposureMask,
-                  &xev);
-      /* We have to sync to reliably catch errors from the XSendEvent(),
-       * since that is asynchronous.
-       */
-      XSync (xdisplay, False);
-      gdk_x11_display_error_trap_pop_ignored (display);
-#else
-      /* Hiding and showing is the safe way to do it, but can result in more
-       * flickering.
-       */
-      gdk_window_hide (widget->window);
-      gdk_window_show (widget->window);
-#endif
+    /* Hiding and showing is the safe way to do it, but can result in more
+     * flickering.
+     */
+    gtk_widget_hide(widget);
+    gtk_widget_show_all(widget);
     }
 }
 


lukefromdc:
You probably won't find any tray icon problems with GTK 3.22.30, I could not force the laptop to cut icons with that GTK version, getting perfect rendering session start after session start.

 GTK 3.23.0 on that same laptop was a buggy mess for mate-power-manager's icon. This helped by reducing the frequency of problems with it, and reducing the worst case from icon missing and replaced with partial copy of another to just the offending icon now showing up. Anything that removed the expose code helped with that. Unknown if GTK 3.23 is introducing new problems or is  just a convenient way for me to set up a machine that has the same issues others get with older GTK versions.

Note that na_tray_child_force_redraw runs exactly six times on starting the panel on my desktop, so it won't bog down the CPU with an unending cycle of reshowing a widget like what happened with m-p-m's inhibit applet.

muktupavels:
Did you try to find commit in GTK that introduce this change?

lukefromdc:
Haven't tried that, but presumably it would be something concerning expose handling. As soon as the (not recommended in years) direct expose handling is taken out, the rendering of partial copies of the wrong icon and that sort of corruption stops.  Also, that code is one of two blocks set apart by the #if 1/else conditional. No idea what that sets up (could find it from configure.ac if I really needed to) but as soon as I tried to build the  gdk_window_hide (widget->window)/gdk_window_show (widget->window) block unconditionally it threw build errors concerning window not being part of widget. I don't think that block has been built in many years.

Gnome-panel has most of the same code, so you might want to try it with GTK 3.23 (I normally use a fairly new build) in a slow laptop and see what you get, and if differences elsewhere let you get away with it.

A bisect would be difficult, as each test would require moving a built GTK 3.23 package to the laptop, or else i would have to build GTK itself on a converted Braswell Cromebook (slow). Also the GTK commit history has a lot of branch merges in it, as one might expect on such a large project. The much faster desktop is very "tolerant" and it may take a lot of restarts to get a cut icon on it, and I never, ever get corruption regardless of panel or GTK version. Seeing the bug requires a slow machine busy with something like starting the session, and I only got it with GTK 3.23. That does not mean nobody else with a different configuration isn't getting it with an older (even a much older) GTK version, I'm only testing two machines and both are on nearly identical Debian Unstable setups.

I've bisected GTK before, but not where finding the bug required moving to another machine, I will not go through the commit history in git-gui and see if the word "expose" comes up anywhere though, as that might provide a starting point. As far as I have seen, direct expose handling in a program using GTK is deprecated anyway.

lukefromdc:
In looking at the GTK 3.23 commit history here is what I found:

Nothing new in Expose,plug, gdk_x11_display, GDK_WINDOW_XID, error_trap_push, error_trap_pop,gtk_widget_get_display, GDK_DISPLAY_XDISPLAY

Allocation finds "CssGadget: Round px values up for min-width|height" and this appears to be aimed at fixing problems in font rendering

"gtksocket: Adjust X sizes by scale-factor" probably is what fixed the bug that caused HiDPI try icons to be spaced twice as far apart. Thus in my last round of tray work I had to limit the previous fix for this (in the tray code) to GTK 3.22 builds only or HiDPI icons got badly cut up. Nice bugfix on GTK's end though.  Nothing else new to GTK 3.23 for GtkSocket

The rest of the removed code is Xserver stuff I don't think is changed by installed GTK version at all.

lukefromdc:
I've also tested this with both GTK versions, getting excellent results with GTK 3.22.30 and much better results on 3.23.0 than without it.

raveit65:
I saw that with other applets, eg. nm-applet too if  i remember correctly.

monsta:
Oh and I forgot to mention that I need non-HiDPI examples :slightly_smiling_face:

raveit65:
Hard to say for me with a HIDPI box :)
Just in this moment i see that with all applets in a f28 VM (gtk+-3.22.x) with HIDPI resolution.
Eg. mate-display-applet, volume-applet, nm-applet

monsta:
Ok, picked it to 1.20 as https://github.com/mate-desktop/mate-panel/commit/2c548d163faa0a15deeb8ec4de85f758bda4d4b0... it works fine, it just would be nice to know what's actually fixed.

lukefromdc:
I found I get someone less corruption of icons with it, but I now suspect that might have been chance, though we still get simpler and more understandable code. Same bugs in XFCE say a lot-but it turns out there is also a lot of shared code there

lukefromdc:
On further though, I now think this made the force-redraw code work better, so the case where an application momentarily sends a corrupted icon, then sends it properly rendered gets fixed. If the application continues to send the corrupted icon, no amount of forced redrawing at the panel will fix it. If the redraw code did not always work, than a momentarily corrupted icon would just keep on showing.

Thus in my laptop tests with this considerably  reduced but did not entirely eliminate problems with mate-power-manager's battery icon.

PR #836, opened 2018-07-11
fish: avoid deprecated 'gtk_window_set_wmclass'

sc0w:


raveit65:
And which window class is used now if it isn't **dialog** any more?

raveit65:
Ok , got it, from docs 
It was 
wmclass_name  = fish
wmclass_class = Fish

So isn't that needed for eg. compiz plugin windows-rules to add actions to a specific kind of window?

raveit65:
The compiz windows-rules plugin can use several parameter. wmclass_name, wmclass_class and window-title.
So we don't break it.
But for me this is another example of a very questionable decision of gtk+ to declare this as deprecated.

Issue #843, opened 2018-07-19
applet widths too narrow with gtk 3.23.1


Safari77:
#### Expected behaviour
applets displayed ok

#### Actual behaviour
applet widths cut to half

#### Steps to reproduce the behaviour
add clock applet when using gtk-3.23.1

#### MATE general version
1.21

#### Package version
1.21.1 git 7369482f848486c5e35f2883346b23b1cacc6f37

#### Linux Distribution
Fedora 28

#### Link to downstream report of your Distribution
![mate-panel-1 21 1 half](https://user-images.githubusercontent.com/3876865/42925245-ccf2a3fe-8b2d-11e8-92b7-e44c64c247c1.png)

Safari77:
```
gtk_widget_size_allocate(): attempt to underallocate MatePanelApplet's child GtkToggleButton 0x55f01c74fd80. Allocation is 131x32, but minimum required size is 262x33.
gtk_widget_size_allocate(): attempt to underallocate GtkToggleButton's child ClockBox 0x55f01cc7fa00. Allocation is 129x30, but minimum required size is 260x31.
gtk_widget_size_allocate(): attempt to underallocate ClockBox's child GtkLabel 0x55f01cc7ed50. Allocation is 260x30, but minimum required size is 260x31.
```

raveit65:
I don't see gtk+-3.23.1 in fedora repos.
https://koji.fedoraproject.org/koji/packageinfo?packageID=10393

lukefromdc:
You will get a lot more warnings with any odd-numbered (development) build of GTK3 with default build options. Release versions use(unless changed at configuration) --enable-debug=minimal, which minimizes warnings, while development versions default to --enable-debug=yes, and this is intended to allow finding and fixing these issues, which exist with or without the warnings. Do not use --enable-debug=no, as it removes runtime checks that can stop segfaults in programs with issues.

A build of GTK 3.22.30 with --enable-debug=yes would determine if the reported warnings concern issues limited to GTK 3.23 only or also present in GTK 3.22.30. They might be new, as I recall warning fix commits during the GTK 3.19 and 3.21 cycles. New GTK versions can change behavior or many widgets, thus the debug=yes default during development cycles

Safari77:
There were already several commits with notes about GTK 3.23 in mate-panel git, it led me believe it (3.23) might actually work.

raveit65:
[...]

PR #845, opened 2018-07-20
GTK 3.23: fix applet size allocation

lukefromdc:
Limit workaround for GTK 3.22 double width/height allocation to GTK 3.22 builds only. Fix https://github.com/mate-desktop/mate-panel/issues/843 in which out-of-process applets get only half their needed width with window-scaling=2. This issue is caused by
https://github.com/mate-desktop/mate-panel/commit/ff04f332136272a14c588da3f2d582ba0ff01be1
which is a fix for for out-of-process applets getting double their needed width. A bug in GTK 3.22 caused that and is fixed in GTK 3.23

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1055,21 +1055,23 @@ mate_panel_applet_get_preferred_width (GtkWidget *widget,
 				       int       *minimum_width,
 				       int       *natural_width)
 {
-	MatePanelApplet *applet = MATE_PANEL_APPLET (widget);
-	gint scale;
-
 	GTK_WIDGET_CLASS (mate_panel_applet_parent_class)->get_preferred_width (widget,
 										minimum_width,
 										natural_width);
 
+#if !GTK_CHECK_VERSION (3, 23, 0)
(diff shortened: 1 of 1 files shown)


lukefromdc:
@Safari77 , please test.

lukefromdc:
I tested with both GTK 3.23 and GTK 3.22 and got proper applet widths for out of process applets both ways.

raveit65:
Waiting for 3.23.x in fedora rawhide for better testing.

lukefromdc:
Fixed the unused variables in GTK 3.23 builds, test builds with both GTK versions built fine without these warnings, and ran the same as before

raveit65:
Confirmed, build warnings are gone.

alexarnaud:
@lukefromdc Do you plan to backport it to 1.20? It seems users are impacted in Debian Sid, see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=910345

Best regards,
Alex.

lukefromdc:
I cannot even build 1.20 without rolling back several other packages. Unless its a cherrypick I don't have the time to mess with backports of anything to any package, I have my hands full as it is. Fortunately we have other devs here.

lukefromdc:
An attempt to cherrypick this failed, someone will have to backport it.

Issue #809, opened 2018-05-10
mate-panel high memory usage if build with in-process


prarit:
#### Expected behaviour

mate-panel does not consume 9G of memory

#### Actual behaviour

mate-panel has consumed 9G of memory.  Eventually attempting to start applications from the mate-panel will fail with memory errors.  Starting the application from a previously opened terminal does work.

#### Steps to reproduce the behaviour

Unknown.  I am using the mate desktop on my day-to-day workstation.  I am not doing anything unusual.

#### MATE general version

mate-about shows version 1.20.1

mate-desktop-1.20.1-3.fc27.x86_64

#### Package version

mate-panel-1.20.1-5.fc27.x86_64

#### Linux Distribution

F27

prarit:
##Link to downstream report of your Distribution

https://bugzilla.redhat.com/show_bug.cgi?id=1576930

P.

raveit65:
Do you use windows-applets via wine in na-tray?
See https://bugzilla.redhat.com/show_bug.cgi?id=1575091

prarit:
On 05/10/2018 03:02 PM, raveit65 wrote:
> Do you use windows-applets via wine in na-tray?
> See https://bugzilla.redhat.com/show_bug.cgi?id=1575091

Nope, I'm not using wine or any windows apps.   I can reproduce the problem
right now.  These are the open applications I have: xterm, thunderbird, hexchat,
firefox, mate terminal, and libreoffice.  In previous instances where this has
happened I've had everything but libreoffice open.

P.

prarit:
Current mate-panel memory consumption is over 50%:
2203 prarit    20   0 14.706g 0.013t  26864 S   5.3 50.9  98:26.53 mate-panel 

If I attempt to launch any application through the panel I get a "Failed to fork (cannot allocate memory)" error.

prarit:
OOC (I couldn't find an answer searching online) is there a way to restart the mate-panel?  I know of "--reset" but that kills all my settings.

raveit65:
'killall mate-panel', restart happens automatic via session management.

raveit65:
which is 'kill -9'

raveit65:
> If I attempt to launch any application through the panel I get a "Failed to fork (cannot allocate memory)" error.

With a panel launcher, from menu or with `alt+f2`

prarit:
On 05/10/2018 11:21 PM, lukefromdc wrote:
[...]

PR #846, opened 2018-07-22
mate-panel-applet: setup_x_error_handler for in-process applets too

lukefromdc:
This is for the tray: GtkPlug/GtkSocket widgets are used to embed tray icons (GtkStatusIcons) in the tray even if the tray is in-process and does not use the plug/socket widgets to be itself embedded in the panel.


lukefromdc:
This is one of the few codepath differences for in-process applets.  Offering this for testing to see if it makes any difference in the memory leak issue
https://github.com/mate-desktop/mate-panel/issues/809
Having run this test panel a few minutes I am seeing 39.2MB of RAM in use for the panel, similar to what I used to see back in 2008 with gnome-panel (though that was on 32 bit and about 30MB)

raveit65:
I see no problems with a quick runtime test and na-tray is build `in-process`.

Issue #848, opened 2018-07-22
Scaling notification area icons not work for some apps


amir28:
#### MATE general version
mate-session 1.20.0

#### Package version
mate-panel-1.20.2-1

#### Linux Distribution
ArchLinux

## Description
Hello, I use ArchLinux with mate desktop. I want make small notification area's icon. I wrote this code in the file ~/.config/gtk-3.0/gtk.css:
```
.mate-panel-menu-bar na-tray-applet,
panel-toplevel.background na-tray-applet {
  -NaTrayApplet-icon-padding: 0;
  -NaTrayApplet-icon-size: 15px;
}
```
After reseting the mate panel, I saw this:
![screenshot at 2018-07-22 21-08-00](https://user-images.githubusercontent.com/22644254/43047870-6e1c7c92-8df3-11e8-86bf-51dfb1f81db7.png)
some applications' icons like Uget and Pacmac is very big but some applications' icons like VLC and Telegram Desktop is OK.
I have Ubuntu Mate in dual boot. This problem isn't be seen in My UbuntuMate.

lukefromdc:
I see here you have a large panel height, but are setting much smaller icons in gtk.css. The fact that some but not all icons respond to this means the GTK theme is controlling some of them (looks like that is the status notifier applets) and the panel height is controlling the rest of it. By default the code tries to scale all icons in the panel to fit the panel height, and the case of a smaller icon size being forced in the theme or in gtk.css was never tested and I have no seen any experiments with it.
My guess is that the panel height is the last thing controlling icon scaling for tray icons, and the GTK theme the last thing (as in having the final word) controlling the height of status icons.

The status notifier code contains GTK_ICON_SIZE_MENU, while this code does not appear in the conventional tray (status icon)  code.

amir28:
OK, What i can doing??

lukefromdc:
If you remove that code to set the small icon size, all of them should display at or near full panel size. At the moment I don't know of anything else and am on my way out the door for an event. It might be a major overhaul to support reduced-size icons and at the moment my focus has been on finding a bad memory leak. Will take a look at it at some point but no promises, maybe one of the other devs here will think of something

raveit65:
Seems issue is fixed.

Issue #818, opened 2018-06-13
panel-list- and clock-applet icons seem to be wrong size and strange overlap issue with HIDPI resolution, if build out-of-process


tonyabc:
#### Expected behaviour
I would expect the Active Apps in the panel to show the icon for that App and be size appropriate for the panel.

#### Actual behaviour
Active App Icons are vertically smaller then they should be and they are not taking the correct space on the panel, also there seems to be 'ghost' icons that are the correct size but not clickable to select the app.

https://imgur.com/a/8gYshw9

#### Steps to reproduce the behaviour
I upgraded from F26 Mate to F28 Mate using DNF and 'proper' F26 --> F27 --> F28 flow using dnf upgrade network option.  Since the new setup the panel has become borked.

#### MATE general version
1.20.2-1,FC28.x86_64

#### Linux Distribution
Fedora 28 Mate Compiz

tonyabc:
Can't work out how to embed so just posted the link

raveit65:
Now the link works.
Are you using HIDPI Monitor? 
This looks like https://github.com/mate-desktop/mate-panel/pull/755#issuecomment-374868245
But i get this problem only sometimes with HIDPI resolutions and switching the workspace fixes the redraw issue.
Btw. Is this report from you?
https://bugzilla.redhat.com/show_bug.cgi?id=1590569#c13

tonyabc:
It's a 4K Screen on a Dell 9550 laptop so yes HIDPI.

Switching workspace doesn't fix the problem, although looking at that I notice the 'bunching' seems only to be firefox active app icons.

Yes I did post that redhat bugzilla as that was a separate bork that happened.

raveit65:
Is this screenshot with compiz?
Compiz don't use workspaces (from technical point).
Can you switch to marco and test if switching workspaces helps?

tonyabc:
A very good point.......slipped my mind I've been using Compiz so long.....

Anyhoo, switching to Marco and switching workspaces didn't fix the issue...

Here's two shots of the Panel, one with a white background and then with black so you can see what's visible.

https://imgur.com/a/fzsM646

tonyabc:
Just to update you and I'm not sure if this is linked or a totally separate bug.

I just opened Libreoffice Writer and it kills the Emerald Theme Manager, even doing a reload of Compiz won't work while Writer is still open.

[...]

Issue #793, opened 2018-04-27
Skype icon starts tiling after panel size 29


jarrard:
![skypematepanelissue](https://user-images.githubusercontent.com/17428910/39346303-969f6612-4a2c-11e8-8f4e-761b56d28e5d.jpg)

Expected behaviour
Skype Icon to scale correctly or just a single icon to appear centered at its scale bracket size.

Actual behaviour
The Skype icon starts to tile itself resulting in a 1 1/2 icon at panel size 29 and a double icon at panel size 41.

Steps to reproduce the behaviour
Occurs when panel is set to size 29 or above.

MATE general version
1.20.1

Package version
mate-panel-1.20.1-1

Linux Distribution
Manjaro MATE x86_64

Link to downstream report of your Distribution
unknown

mmishut:
![mate-panel](https://user-images.githubusercontent.com/38864264/39440002-2d2e5462-4cb2-11e8-8fc0-c8e2a4165c5c.png)
Agree about a problem.
In my case it occurs with panel size 22 and more.

Package:
mate-panel (1.20.1-3ubuntu1)

Linux:
Ubuntu MATE 18.04

muktupavels:
My guess it might be skype bug, after skype update I see similar thing with gnome-panel... Is that only app or similar bug happens with other icons too?

jarrard:
The icons are provided by mate-panel (you can find them) and not skype I believe. Also the issue affects any skype spin-off that uses those icons such as ghetto-skype.

valeryvdr:
I am also affected, freshly installed Debian Stretch + Mate 1.18.0. Skype tray icon is tailing at panel size = 24

danieltomasz:
Im on gnome panel  and I have the same issue

vkotovv:
I have a similar problem with Guake icon (and also Skype icon). As you can see, Guake icon on the screenshot (it it the 1st one) starts tiling also. Happened to me after Mate 1.18 -> 1.20.1 update.

![selection_066](https://user-images.githubusercontent.com/1636249/42692534-b38c368c-86b4-11e8-85e1-e2d1b0ba77a4.png)

lukefromdc:
By any chance does this happen with all apps that offer only small png icons and nothing scalable for the tray?

iLGunners:
![image](https://user-images.githubusercontent.com/12131543/42925359-c1498b4a-8b4f-11e8-91ff-0cc38ad872a7.png)

[...]

PR #849, opened 2018-07-25
Fix applet issues

vkareh:
* workspace-switcher had a runtime error when calling `g_object_set` with a `NULL` object.
* na-tray had a runtime error wen calling `gdk_pixbuf_scale_simple` with a `dest_width` or `dest_height` of `<= 0` (we only need to check that `icon_size > 1` since that's its default value, it will never be anything lower than 16px or so when the icon is actually loaded). This error prevented the icon from scaling properly and caused certain icons to render too large or the applet itself to render in an un-scaled state.
* window-list: when WnckTasklist loaded initially, it's not yet populated with whatever application icons are open, so the `size_hints` came back as `{0, 0}`, which confused libmate-panel-applet when setting it from dbus, since when it was time to actually set the size hints, the WnckTasklist was already populated, but the size_hints did not reflect this. The caused the window-list to render in an un-scaled state.

_May_ fix the following issues: #848, #818, #793, #746

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -42,7 +42,7 @@ typedef struct {
 
 	GtkOrientation orientation;
 	int size;
-#ifndef WNCKLET_INPROCESS
+#if !defined(WNCKLET_INPROCESS) && !GTK_CHECK_VERSION (3, 22, 31)
 	gboolean needs_hints;
 #endif
 
@@ -322,7 +322,7 @@ static void applet_size_allocate(GtkWidget *widget, GtkAllocation *allocation, T
 
 	g_assert(len % 2 == 0);
 
-#ifndef WNCKLET_INPROCESS
+#if !defined(WNCKLET_INPROCESS) && !GTK_CHECK_VERSION (3, 22, 31)
 	/* HACK: When loading the WnckTasklist initially, it reports size hints as though there were
 	 * no elements in the Tasklist. This causes a rendering issue when built out-of-process in
 	 * HiDPI displays. We keep a flag to skip size hinting until WnckTasklist has something to
@@ -427,7 +427,7 @@ gboolean window_list_applet_fill(MatePanelApplet* applet)
 
 	tasklist->size = mate_panel_applet_get_size(applet);
 
-#ifndef WNCKLET_INPROCESS
+#if !defined(WNCKLET_INPROCESS) && !GTK_CHECK_VERSION (3, 22, 31)
 	tasklist->needs_hints = FALSE;
 #endif
 
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -115,11 +115,11 @@ na_tray_child_style_set (GtkWidget *widget,
    */
 }
 
-#if !GTK_CHECK_VERSION (3, 23, 0)
+#if !GTK_CHECK_VERSION (3, 22, 31)
 static void
 na_tray_child_get_preferred_width (GtkWidget *widget,
                                    gint      *minimal_width,
-                                  gint      *natural_width)
+                                   gint      *natural_width)
 {
   gint scale;
   scale = gtk_widget_get_scale_factor (widget);
@@ -384,7 +384,7 @@ na_tray_child_class_init (NaTrayChildClass *klass)
 
   widget_class->style_set = na_tray_child_style_set;
   widget_class->realize = na_tray_child_realize;
-#if !GTK_CHECK_VERSION (3, 23, 0)
+#if !GTK_CHECK_VERSION (3, 22, 31)
   widget_class->get_preferred_width = na_tray_child_get_preferred_width;
   widget_class->get_preferred_height = na_tray_child_get_preferred_height;
 #endif
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1059,11 +1059,11 @@ mate_panel_applet_get_preferred_width (GtkWidget *widget,
 										minimum_width,
 										natural_width);
 
-#if !GTK_CHECK_VERSION (3, 23, 0)
+#if !GTK_CHECK_VERSION (3, 22, 31)
 	MatePanelApplet *applet = MATE_PANEL_APPLET (widget);
 	if (applet->priv->out_of_process) {
 		/* Out-of-process applets end up scaled up doubly with GTK 3.22.
-         * For these builds divide by the scale factor to ensure
+		 * For these builds divide by the scale factor to ensure
 		 * they are back at their own intended size.
 		 */
 		gint scale;
@@ -1082,12 +1082,13 @@ mate_panel_applet_get_preferred_height (GtkWidget *widget,
 	GTK_WIDGET_CLASS (mate_panel_applet_parent_class)->get_preferred_height (widget,
 										minimum_height,
 										natural_height);
-#if !GTK_CHECK_VERSION (3, 23, 0)
+
+#if !GTK_CHECK_VERSION (3, 22, 31)
 	MatePanelApplet *applet = MATE_PANEL_APPLET (widget);
 	if (applet->priv->out_of_process) {
 		gint scale;
 		/* Out-of-process applets end up scaled up doubly with GTK 3.22.
-         * For these builds divide by the scale factor to ensure
+		 * For these builds divide by the scale factor to ensure
 		 * they are back at their own intended size.
 		 */
 		scale = gtk_widget_get_scale_factor (widget);


applets/notification_area/status-notifier/sn-item-v0.c
```diff
@@ -256,7 +256,7 @@ update (SnItemV0 *v0)
       if (!pixbuf){
           /*try to find icons specified by path and filename*/
           pixbuf = gdk_pixbuf_new_from_file(v0->icon_name, NULL);
-          if (pixbuf){
+          if (pixbuf && icon_size > 1) {
```
muktupavels: All this code seems weird to me:
- icon_name is supposed to be an icon name. why load it as file?
- what is random `-2`? how that goes together with comment? icon is scaled always...
- looking at code what is effective_icon_size? How can icon_size be <= 0?

@vkareh you say that icon_size default value is 1 !?, but in init it is set to 16... What and why sets icon_size to 1? That warning is because you have 1x1 icon and you try to scale it to -1x-1, because of that random number `2`...

applets/wncklet/workspace-switcher.c
```diff
@@ -120,7 +120,8 @@ static void update_properties_for_wm(PagerData* pager)
 				gtk_widget_show(pager->workspace_names_scroll);
 			if (pager->display_workspaces_toggle)
 				gtk_widget_show(pager->display_workspaces_toggle);
-			g_object_set (pager->cell, "editable", TRUE, NULL);
+			if (pager->cell)
```
muktupavels:
```
cell = g_object_new(GTK_TYPE_CELL_RENDERER_TEXT, "editable", TRUE, NULL);
pager->cell = cell;
```

`g_object_new` does not return `NULL`. I think problem might be `applet_realized` function that seems like only way that could cause your problem...

So instead of adding extra NULL check I would say that code should be updated to make sure that this function is not called before everything is correctly setup. Same is true with other ifs in that switch.

applets/wncklet/window-list.c
```diff
@@ -319,7 +320,23 @@ static void applet_size_allocate(GtkWidget *widget, GtkAllocation *allocation, T
 
 	g_assert(len % 2 == 0);
 
-	mate_panel_applet_set_size_hints(MATE_PANEL_APPLET(tasklist->applet), size_hints, len, 0);
+	/* When loading the WnckTasklist initially, it reports size hints as though there were no
+	 * elements in the Tasklist. We keep a flag to skip size hinting until WnckTasklist has
+	 * something to show. */
```
muktupavels: How this is a problem? Are not size hints updated when there are new elements? I forgot how this works...

vkareh: The size hints are updated. WnckTasklist works just fine, but I think there's a race condition in how we set those hints initially that is only apparent with GDK_SCALE=2. Doing this here defers that initial size hint until the applet is fully loaded.

muktupavels: You think or you have found something that proves that? Why would this work differently with GDK_SCALE=2?

What size hints you get with scale == 1? What numbers you get with scale == 2?

applets/notification_area/status-notifier/sn-item-v0.c
vkareh: Correct, I can see the `icon_size = MAX (1, ...);` above, so that means I could probably just skip this entire chunk of code based on the `icon_size`. I'll look at `effective_icon_size` - it seems to be only set if `icon_size` is `<= 0`, and then it's used to replace `icon_size` itself during size allocation - so this code is not necessarily super clean. I'll work on that.

I don't know about loading `icon_name` as a file, I agree that's strange, but in my tests it seems to be used and working (it actually loads an icon)

applets/wncklet/window-list.c
vkareh: I have not yet found proof of this. WnckTasklist works fine regardless of GDK_SCALE - it loads size hints as `{0, 0}` initially, then updates to `{0, 402}` (in my system with my screen resolution, etc). That's all good. The problem seems to be that the applet is built out-of-process, and in the time between when the first `{0, 0}` hint and the second `{0, 402}` hint are sent, something happens that gets the applet locked into a non-scaled state. This doesn't happen with in-process applet, nor with GDK_SCALE=1. So the issue is clearly somewhere else, but this code here sidetracks it until we can fix whatever causes this in libmate-panel-applet.

Maybe I'll add a `/* HACK: ... */` message to make this clear.

vkareh: I've updated this code to only run for out-of-process wncklet, and made the comment a bit more explicit.

applets/wncklet/window-list.c
```diff
@@ -42,6 +42,9 @@ typedef struct {
 
 	GtkOrientation orientation;
 	int size;
+#ifndef WNCK_INPROCESS
```
muktupavels: Typo our you have both `WNCK_INPROCESS` and `WNCKLET_INPROCESS`?

raveit65:
I can confirm that wnck-list applet with HDPI is fixed.
But it seems that real fixes are located by gtk+
by Ubuntu:
```
From: marmuta <marmvta@gmail.com>
Date: Sun, 12 Jun 2016 15:42:01 +0200
Subject: Unscale plug size before sending XConfigureEvent

https://bugzilla.gnome.org/show_bug.cgi?id=765327

Bug-Ubuntu: https://launchpad.net/bugs/1572331
```
by gtk+
https://gitlab.gnome.org/GNOME/gtk/commit/c8b0a0dea5cd129d8e023443dde5e44319d4631d

Sadly the gtk+ version fixes wnck-list-applet but breaks a lot of other applets,
see eg, workspace-switcher, clock, na-tray and show-desktop have half horizontal size.
bildschirmfoto zu 2018-07-26 17-45-30

Ubuntu patch doesn't break the other applets, i am just testing if it fixes the issue with wnck-list.
But it seems to be good after 1 hour testing.

applets/wncklet/window-list.c
vkareh: duh, typo of course - good catch :)
I just fixed and pushed.

lukefromdc:
https://gitlab.gnome.org/GNOME/gtk/commit/c8b0a0dea5cd129d8e023443dde5e44319d4631d

Is in GTK 3.23, and fixes for the tray and out of process applets to accomodate it in GTK3 builds are already in mate-panel addressing exactly this issue. The combination might fix these issues and be a use case for updating to GTK 3.24 and a GTK 3.24 panel build when it comes out.

lukefromdc:
I am guessing Ubuntu is applying https://gitlab.gnome.org/GNOME/gtk/commit/c8b0a0dea5cd129d8e023443dde5e44319d4631d to their GTK 3.22 builds?

lukefromdc:
If Ubuntu patches GTK to act differently than all other distros we may have an ugly mess on our hands unless GTK accepts that patch upstream.

vkareh:
so the 3.23 patch fixes window-list but breaks all other applets? sounds like
we should wrap my window-list fix with `#if !GTK_CHECK_VERSION (3, 23, 0)`

@raveit65 - when you applied the gtk patch, are you testing from the mate-panel master tree? Do you have this commit in it: https://github.com/mate-desktop/mate-panel/commit/a130fb0379a270f6ed80fa21ba2a69e184b4cc2a ?

raveit65:
Yes, i tested both gtk+ and ubuntu version from latest maste + open PRs.
See my spec file in SRPM from archive i have given you.
Patch 26 is `fix applet size allocation`
3.21 tarball +
```
# https://github.com/mate-desktop/mate-panel/pull/813
Patch1:         mate-panel_0001-workspace-switcher-provide-themed-fallback-color.patch
# https://github.com/mate-desktop/mate-panel/pull/816
Patch2:         mate-panel_0001-struts-Create-struts-for-panels-on-inside-edges-of-m.patch
Patch3:         mate-panel_0002-struts-Use-inside-edge-panels-only-on-Marco-and-Meta.patch
# https://github.com/mate-desktop/mate-panel/pull/820
Patch4:         mate-panel_0001-Panel-gtk.c-Show-menu-icons-only-when-menus-have-ico.patch
# https://github.com/mate-desktop/mate-panel/pull/823
Patch5:         mate-panel_0002-menu.c-Fix-Keep-spaces-in-the-left-if-menus-have-ico.patch
# local test, panel-force-quit
Patch6:         mate-panel_0001-panel-force-quit-replace-deprecated-gdk_flush.patch
Patch7:         mate-panel_0002-panel-force-quit-replace-deprecated-gdk_error_trap-f.patch
Patch8:         mate-panel_0003-panel-force-quit-replace-deprecated-gdk_device_grab.patch
# local test, deprecations-2
Patch9:         mate-panel_0001-panel-widget-replace-deprecated-gdk_flush.patch
Patch10:         mate-panel_0002-main-replace-deprecated-gdk_flush.patch
Patch11:         mate-panel_0003-panel-util-replace-deprecated-gdk_flush.patch
Patch12:         mate-panel_0004-panel-xutils-replace-deprecated-gdk_error_trap-funct.patch
Patch13:         mate-panel_0005-libmate-panel-applet-replace-deprecated-gdk_error_tr.patch
Patch14:         mate-panel_0006-na-tray-child-replace-deprecated-gdk_error_trap_push.patch
# local test, gtk_menu_poup
Patch15:         mate-panel_0001-panel-menu-button-replace-deprecated-gtk_menu_popup.patch
# test style classes
Patch16:         mate-panel_0001-panel-toplevel-add-position-css-class.patch
# local test, gtk_menu_poup
Patch17:         mate-panel_0001-panel-context-menu-replace-deprecated-gtk_menu_popup.patch
Patch18:         mate-panel_0002-applet-context-menu-replace-deprecated-gtk_menu_popu.patch
Patch19:         mate-panel_0003-panel-actions-replace-deprecated-gtk_menu_popup.patch
Patch20:         mate-panel_0004-panel-actions-use-gtk_menu_popup_at_pointer.patch
# https://github.com/mate-desktop/mate-panel/pull/833
Patch22:         mate-panel_0001-na-tray-child-remove-na_tray_child_size_allocate.patch
Patch23:         mate-panel_0002-GTK-3.23-don-t-cut-tray-icons-too-narrow-in-HiDPI-mo.patch
# https://github.com/mate-desktop/mate-panel/commit/9de3a86
Patch24:         mate-panel_0001-Tray-simplify-force-redraw-use-code-that-actually-wo.patch
# https://github.com/mate-desktop/mate-panel/commit/7369482
Patch25:         mate-panel_0001-fish-avoid-deprecated-gtk_window_set_wmclass.patch
# https://github.com/mate-desktop/mate-panel/pull/845
Patch26:         mate-panel_0001-GTK-3.23-fix-applet-size-allocation.patch
# https://github.com/mate-desktop/mate-panel/pull/846
Patch27:         mate-panel_0001-mate-panel-applet-setup_x_error_handler-for-in-proce.patch
# https://github.com/mate-desktop/mate-panel/pull/849
Patch28:         mate-panel_0001-applets-Fix-runtime-error-messages.patch
#Patch29:         mate-panel_0002-window-list-Skip-setting-size-hints-until-WnckTaskli.patch
```
Last one from this PR is commented out.

In any case we need to make the panel code compatible for 3.24 which will come probably with next fedora or ubuntu releases in autumn/winter.
I think it is also possible that gtk+ will release a 3.22.31 version as the fix is in 3.22 branch.
In this case or for maintainers who like to fix stable distro releases like me, a branch for gtk+-3.22 with
`#if !GTK_CHECK_VERSION (3, 22, 0)`
is useful.

lukefromdc:
Using `#if !GTK_CHECK_VERSION (3, 22, 0`) would limit a code block to GTK 3.20 and earlier, which we no longer support. To limit to GTK 3.22 and exclude 3.23/3/24 you would use #if GTK_CHECK_VERSION (3, 23, 0) now or for code written after the release of GTK 3.24 it would be ` #if !GTK_CHECK_VERSION (3, 24, 0)`

raveit65:
You'r right.
I just wanted to say a different thing.
Than we need a branch for 3.22 without any version check if this fix goes into a 3.22.31 release.
It is also possible to ask fedora gtk+ maintainer to add the fix as patch on top of gtk+-3.22.30.

raveit65:
@vkareh 
What should we do?
`applets: Fix runtime error messages` is unrelated to gtk+ fixes, so this can be merged when PR is ready?
'window-list: Skip setting size hints until WnckTasklist` we don't need for >= gtk+-3.23.1, right?
So this needs a version check too?
Of course this is a temporarily solution and should be reverted if gtk+ release the real fix with a new 3.22.31 release.
I would be fine too use this only as patch for fedora without merging it here, but it affects other distros too.

vkareh:
@raveit65 it affects pretty much all distros except Ubuntu. So distro patches might not be a good solution for now. Let me add the version checks today and test with the different gtk versions that you sent me before we come to a decision, but I think this is necessary

muktupavels:
I would suggest to not workaround GTK bugs in app. What you will do when some distros will have patch and some not? More conditional "fixes" like `if UBUNTU or FEDORE ... else ...`?

So there are two patches that gives you different results... How about creating small reproducer that shows your problems? Like it was done in #765327 bug? And providing fix for that?

Then you can tell affected users to open bugs for their distros asking to patch gtk...

raveit65:
> I would suggest to not workaround GTK bugs in app.

We have open report from user which needs to be fixed in current distro releases.
And it isn't clear that gnome will release a new 3.22.31 release.
Maybe you like that users have to live with bugs more than necessary?
I don't like that....
Any way, i will use this fix for fedora.

> More conditional "fixes" like `if UBUNTU or FEDORE ... else ...`?

Why fedora? Only ubuntu used a patched gtk+-3 version.

muktupavels:
@raveit65 are you saying that fedora never patches gtk? do they always wait new release or live with bugs?

I am not saying users should live with bugs, but if bug is in GTK it should be fixed there, no? Make sure that GTK has full fix (if that is not the case now) and then it is up to distro maintainers to update GTK if they care about their users.

raveit65:
Not sure if we talk about the same thing.
We have 2 general situations and 1 special situation which isn't sure that it will come.
1. current distro releases with gtk+-3.22.30. For this situation we need this PR. Only gtk+ in ubuntu is patched and doesn't need the fix from PR.
2. For future distro releases with gtk+-3.24 we have to add `if !GTK_CHECK_VERSION (3, 23, 0)` for `wnck-list part of this PR` and `previous optimisations for HIDPI in code`. All of this are not needed any more with fix from gtk+ devs.
3. When distros update to a new gtk+-3.22.31 release we need a special branch without `previous optimisations for HIDIPi in code` and `wnck-list part of this PR`

I don't see where we need a special check for fedora or any other distro except ubuntu ;)

vkareh:
@raveit65 - even so, the wnck fix doesn't really pose a problem if it gets run with the new version of GTK - all it does is defer the size hinting for another render cycle. I'm currently testing to make sure, but it seems fine in both Ubuntu and Fedora (with both patched and unpatched gtk)

So, really, the only thing we would need is `#if !GTK_CHECK_VERSION (3, 23, 0)` for the instances in which we divide the size by the scale factor. I'm currently working on this and will add a new commit to this pull request.

@muktupavels - We don't need to check for distros, just for gtk version and whether it's in-process or not. This particular GTK patch is on Ubuntu (and it's not even a complete one, as they don't scale the size hints), except for the issue is not present in Ubuntu with or without our mate-panel fixes. Guarding against GTK version is sufficient here.

vkareh:
@raveit65 - hmmm... not sure whether they will have 3.22.31 or not, but their GtkSocket fix was applied to the dev 3-22 branch. This means that if ever they release 3.22.31, it will include the scale fix. In that case, I can guard against `#if !GTK_CHECK_VERSION (3, 22, 31)` to be sure.

lukefromdc:
I just did another build of this with GTK 3.23 and didn't have any problems with it at window-scaling=1 (my normal setup) in-process or out-of-process. Keeping the in-process build (again my usual setup) installed for now.

raveit65:
> I can guard against #if !GTK_CHECK_VERSION (3, 22, 31) to be sure.

A++++++

applets/wncklet/workspace-switcher.c
vkareh: @muktupavels - good point, it seems `window_manager_changed` is called during `applet_realized`, except for the window manager hasn't changed - there's already listener for that. However, calling this avoids some duplicate code used in setting the window manager and initializing a few things, so those other `ifs` were most likely added for that same reason.

I think the code is not perfect like this, but my patch just follows the same methodology in place since this function was added (quite a while ago too: https://github.com/GNOME/gnome-panel/commit/250492f57753d9c3197ff27fee97038f957fb4e3).

So for now I'll leave it like this to fix the runtime error, but I'll add to my to-do list to refactor this file.

raveit65:
I can confirm that latest changes are working with
- fedora 28 and gtk+-3.22.30
- fedora 28 and gtk+-3.22.30 + ubuntu patch
- fedora 29 and gtk+-3.23.1

Ready to go.

raveit65:
@vkareh @lukefromdc @monsta 
I cherry-picked all related commits to 1.20 branch and i like to do a new release the next days.
Can you take a look at it, please?
Did i add all related commits?
Any other commit for a bugfix missing?

raveit65:
For a short build test it is save to downgrade mate-menus to latest 1.20.x version.
Don't reload the panel in this time :)

lukefromdc:
Not sure whether you intended to include 
https://github.com/mate-desktop/mate-panel/commit/9de3a86d736f70c4d45550649ba62a935369b146
"Tray: simplify force-redraw, use code that actually works" 
but this one is not present in the 1.20 commit history I saw

raveit65:
We can add it as it sounds like a bug fix, but cherry-picking failed.

lukefromdc:
Test builds with GTK 3.23.1 as this stands, with mate-menus rolled back worked fine both in and out of process, and I didn't get rendering problems with window-scaling=2/

PR #325, opened 2015-07-13
na_tray.c replace GtkBox by GtkTable

braikar:
Update from the old patch, https://bug531371.bugzilla-attachments.gnome.org/attachment.cgi?id=140510  to make the notification area suitable for wide vertical panels.

Issue #157, opened 2014-01-17
Usable vertical panel?


Amet13:
That's my beautiful horizontal MATE panel:
![screenshot](https://f.cloud.github.com/assets/5184847/1941635/68431048-7f8d-11e3-901a-cdf47b0730f2.png)
But I have a widescreen monitor and I want use a free place for example on right.
That's my vertical panel, how to use this I don't know.
![screenshot-1](https://f.cloud.github.com/assets/5184847/1941669/fe20bef8-7f8d-11e3-8dc2-72bef7ec38a5.png)

Problems for me:
1. Gradient
2. Vertical mintmenu & clock applets
3. Fixed tray icons
4. Large windows list's icons

Any ideas? Thanks.
##

damarev:
Bug has been around since 2002: https://bugzilla.gnome.org/show_bug.cgi?id=86382

This is a major UX issue:
- The "window list" on a vertical panel is just unusable.
- Nowadays widescreens are the standard. 
- From many points of view, a vertical panel is the most logical option.

slacka:
@fraterlinux 
I ran into the exact same issue with Mint 17.  In issue #58 stefano-k said:

> Because 1.8 uses libwnck, so you need an updated libwnck. The issue is not is mate-panel

I couldn't find any information on what version of libwnck was required or how to fix Mint, so I ended up downgrading to Mint 16.

@stefano-k 

Could you provide any more info to help Mint and now PCLinuxOS users? Are you sure it's not a mate-panel issue as symptoms are identical to how the old mate-panel used to behave?

Taknamay:
Have you considered implementing #144 for this issue? I think it would fix the window list situation.

kafran:
state_of_the_art = True if mate-panel == xfce-panel + lxpanel else False

zsoltpeterbasak:
The bugs related to this on the Gnome tracker had patches in comments, you should check those out. Long time ago I had my Gnome patched and it gave me a very nice vertical window list. (Tint2 is a good example how it could be done. Having both horizontal AND vertical list of windows is the best IMO.)

GotDemBallsOfSteel:
[...]

PR #856, opened 2018-08-22
Fix on wide panels (vert/horz)

braikar:
This takes care of some issues from https://github.com/mate-desktop/mate-panel/issues/157
And it's a final remake of the own issue I had created: https://github.com/mate-desktop/mate-panel/pull/325 (at the time mate-panel was being ported to Gtk3 so it was messy, now that it's finally only Gtk3, I've made the proper changes)

This change limits the size taken by all BUTTON_WIDGETs when the panel is wide, because for the moment when for example the panel is horizontal 200 pixels wide/high, the widgets take up 200x200 for an icon of max size 48.
I've set it to take now (for this example) 50x200 or 200x50 (if it's a vertical panel 200px wide)
If users want to leave the same space as before around the icons they just have to move them on the panel anyway and leave empty space around and it can be as before.. (but it might affect other applets I don't know of)
The GTK_ARROW is also resized properly on BUTTON_WIDGETs with property "has_arrow"

I've setup 50 as a limit, because I found by my own surprise 48 was hardcoded as the limit for the icon size.
For further discussion:
Couldn't it be a panel wide user set option, so that the user sets the maximum icon size somewhere and we can get the value in any applet needing it?
I though about changing the panel properties dialog and add a new option like 'max_icon_size' or something like that, so that the user decides on the maximum icon size him/herself. Then that would be saved in the gsettings of the panel.
But getting that value as a gsettings from "org.mate.panel.." in any applet that needs it doesn't seem to be the right way to do it? I haven't seen a single applet accessing the main panel gsettings it is attached to..?

changes

Then it also finally sets the way for a notification area that should make everyone happy.
I replaced the GtkBox that has been there since forever by a GtkGrid.

screenshot at 2018-08-21 14-05-58
As you can see I added a Preferences dialog, with two options, a tick box to keep them aligned on one line as it was before, they automatically resize to the width of the panel, up to 48px (icon size) centered (the change I made in BUTTON_WIDGETs applies now on the notification area items too, they don't take up 200x200 on a wide panel anymore).
And if the "Display icons only on one line" is unticked, one can choose the size of the icons directly and they'll re-arrange nicely in a gird.
As you can see the two examples where I set 28 size for one and 43 for the other. They don't adapt to all sizes though, once the size is big enough or small enough to fill the width/height, it resizes..
And as you can see the settings are set in a GSchema. I'll just need help to figure out how to change the Makefile to automatically compile it, I've spent hours on it without figuring it out. I installed the GSchema manually..
screenshot at 2018-08-21 14-14-01

PROBLEMS left:
The makefile is buggy, I couldn't figure out how to get the gschema file to compile and be included in the mape-panel package, somehow it is left out, in the build process dpkg doesn't validate it?
So to test it, you need to install the new notification_area gschema manually for the moment
If someone could help, I'm not a pro of makefiles ;)
Else the main problems left are in finishing the event handlers, the grid doesn't redraw when the panel loads for the first time. I don't know which event to target? realize? size-allocate?
The grid doesn't redraw properly when the panel size is modified, same, I don't know which event to target for that change.
And the grid sorting code of the icons is a bit messy `reorder_items_with_data, refresh_grid` in `na-grid.c`. I'm sure it can be made much more neatly, but since I'm leaving for 3 weeks to a place with no internet in 3 days, I'm already publishing it so that anyone else who wants can work on it. I'll get back to it when I come back

One small thing is that the GtkDialog for preferences is not attached to a parent hence it produces:
"GtkDialog mapped without a transient parent. This is discouraged." in the syslog
I took parts of the code from the clock applet, hence the clock applet has the same issue.

I also had to coment out the na-resource.c/na-resource.h stuff in the actual Makefile, because these files don't exist as of yet?

### DIFF
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -25,9 +25,11 @@
 #include <string.h>
 
 #include <mate-panel-applet.h>
+#include <mate-panel-applet-gsettings.h>
 
 #include <glib/gi18n.h>
 #include <gtk/gtk.h>
+#include <gio/gio.h>
 
 #include "main.h"
 #include "na-grid.h"
@@ -40,7 +42,10 @@
 
 struct _NaTrayAppletPrivate
 {
-  GtkWidget *grid;
+  GtkWidget               *grid;
+
+  GSettings               *settings;
+  gint                     min_icon_size;
 
 #ifdef PROVIDE_WATCHER_SERVICE
   GfStatusNotifierWatcher *sn_watcher;
@@ -102,6 +107,23 @@ get_gtk_orientation_from_applet_orient (MatePanelAppletOrient orient)
   g_assert_not_reached ();
 
   return GTK_ORIENTATION_HORIZONTAL;
+}
+
+static void
+gsettings_changed_min_icon_size (GSettings    *settings,
+                                 gchar        *key,
+                                 NaTrayApplet *applet)
+{
+  applet->priv->min_icon_size = g_settings_get_int (settings, key);
+
+  na_grid_set_min_icon_size (NA_GRID (applet->priv->grid), applet->priv->min_icon_size);
+}
+
+static void
+setup_gsettings (NaTrayApplet *applet)
+{
+  applet->priv->settings = mate_panel_applet_settings_new (MATE_PANEL_APPLET (applet), NA_TRAY_SCHEMA);
+  g_signal_connect (applet->priv->settings, "changed::" KEY_MIN_ICON_SIZE, G_CALLBACK (gsettings_changed_min_icon_size), applet);
 }
 
 static void help_cb(GtkAction* action, NaTrayApplet* applet)
@@ -203,6 +225,11 @@ na_tray_applet_realize (GtkWidget *widget)
                                               NA_RESOURCE_PATH "notification-area-menu.xml",
                                               action_group);
   g_object_unref(action_group);
+
+  setup_gsettings (applet);
+
+  // load min icon size
+  gsettings_changed_min_icon_size (applet->priv->settings, KEY_MIN_ICON_SIZE, applet);
 }
 
 static void
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -30,7 +30,7 @@
 #include <gtk/gtk.h>
 
 #include "main.h"
-#include "na-box.h"
+#include "na-grid.h"
 
 #ifdef PROVIDE_WATCHER_SERVICE
 # include "libstatus-notifier-watcher/gf-status-notifier-watcher.h"
@@ -40,7 +40,7 @@
 
 struct _NaTrayAppletPrivate
 {
-  GtkWidget *box;
+  GtkWidget *grid;
 
 #ifdef PROVIDE_WATCHER_SERVICE
   GfStatusNotifierWatcher *sn_watcher;
@@ -225,14 +225,14 @@ na_tray_applet_style_updated (GtkWidget *widget)
   if (parent_class_style_updated)
     parent_class_style_updated (widget);
 
-  if (!applet->priv->box)
+  if (!applet->priv->grid)
     return;
 
   gtk_widget_style_get (widget,
                         "icon-padding", &padding,
                         "icon-size", &icon_size,
                         NULL);
-  g_object_set (applet->priv->box,
+  g_object_set (applet->priv->grid,
                 "icon-padding", padding,
                 "icon-size", icon_size,
                 NULL);
@@ -247,8 +247,8 @@ na_tray_applet_change_background(MatePanelApplet* panel_applet, MatePanelAppletB
     parent_class_change_background (panel_applet, type, color, pattern);
   }
 
-  if (applet->priv->box)
-    na_box_force_redraw (NA_BOX (applet->priv->box));
+  if (applet->priv->grid)
+    na_grid_force_redraw (NA_GRID (applet->priv->grid));
 }
 
 static void
@@ -260,10 +260,10 @@ na_tray_applet_change_orient (MatePanelApplet       *panel_applet,
   if (parent_class_change_orient)
     parent_class_change_orient (panel_applet, orient);
 
-  if (!applet->priv->box)
+  if (!applet->priv->grid)
     return;
 
-  gtk_orientable_set_orientation (GTK_ORIENTABLE (applet->priv->box),
+  gtk_orientable_set_orientation (GTK_ORIENTABLE (applet->priv->grid),
                                   get_gtk_orientation_from_applet_orient (orient));
 }
 
@@ -286,10 +286,10 @@ na_tray_applet_focus (GtkWidget        *widget,
 {
   NaTrayApplet *applet = NA_TRAY_APPLET (widget);
 
-  /* We let the box handle the focus movement because we behave more like a
+  /* We let the grid handle the focus movement because we behave more like a
    * container than a single applet.  But if focus didn't move, we let the
    * applet do its thing. */
-  if (gtk_widget_child_focus (applet->priv->box, direction))
+  if (gtk_widget_child_focus (applet->priv->grid, direction))
     return TRUE;
 
   return GTK_WIDGET_CLASS (na_tray_applet_parent_class)->focus (widget, direction);
@@ -353,10 +353,10 @@ na_tray_applet_init (NaTrayApplet *applet)
 #endif
 
   orient = mate_panel_applet_get_orient (MATE_PANEL_APPLET (applet));
-  applet->priv->box = na_box_new (get_gtk_orientation_from_applet_orient (orient));
+  applet->priv->grid = na_grid_new (get_gtk_orientation_from_applet_orient (orient));
 
-  gtk_container_add (GTK_CONTAINER (applet), GTK_WIDGET (applet->priv->box));
-  gtk_widget_show (GTK_WIDGET (applet->priv->box));
+  gtk_container_add (GTK_CONTAINER (applet), GTK_WIDGET (applet->priv->grid));
+  gtk_widget_show (GTK_WIDGET (applet->priv->grid));
 
   atko = gtk_widget_get_accessible (GTK_WIDGET (applet));
   atk_object_set_name (atko, _("Panel Notification Area"));
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -40,9 +40,18 @@
 
 #define NOTIFICATION_AREA_ICON "mate-panel-notification-area"
 
+typedef struct
+{
+  GtkWidget *preferences_dialog;
+  GtkWidget *min_icon_size_spin;
+} NAPreferencesDialog;
+
 struct _NaTrayAppletPrivate
 {
   GtkWidget               *grid;
+
+  NAPreferencesDialog     *dialog;
+  GtkBuilder              *builder;
 
   GSettings               *settings;
   gint                     min_icon_size;
@@ -116,6 +125,10 @@ gsettings_changed_min_icon_size (GSettings    *settings,
 {
   applet->priv->min_icon_size = g_settings_get_int (settings, key);
 
+  if (applet->priv->dialog)
+    gtk_spin_button_set_value (GTK_SPIN_BUTTON (applet->priv->dialog->min_icon_size_spin),
+                               applet->priv->min_icon_size);
+
   na_grid_set_min_icon_size (NA_GRID (applet->priv->grid), applet->priv->min_icon_size);
 }
 
@@ -124,6 +137,78 @@ setup_gsettings (NaTrayApplet *applet)
 {
   applet->priv->settings = mate_panel_applet_settings_new (MATE_PANEL_APPLET (applet), NA_TRAY_SCHEMA);
   g_signal_connect (applet->priv->settings, "changed::" KEY_MIN_ICON_SIZE, G_CALLBACK (gsettings_changed_min_icon_size), applet);
+}
+
+static void
+na_preferences_dialog_min_icon_size_changed (NaTrayApplet  *applet,
+                                             GtkSpinButton *spin_button)
+{
+  applet->priv->min_icon_size = gtk_spin_button_get_value_as_int (spin_button);
+  g_settings_set_int (applet->priv->settings, KEY_MIN_ICON_SIZE, applet->priv->min_icon_size);
+}
+
+static gboolean
+na_preferences_dialog_hide_event (GtkWidget    *widget,
+                                  GdkEvent     *event,
+                                  NaTrayApplet *applet)
+{
+  gtk_widget_hide (applet->priv->dialog->preferences_dialog);
+  return TRUE;
+}
+
+static void
+na_preferences_dialog_response (NaTrayApplet *applet,
+                                int           response,
+                                GtkWidget    *preferences_dialog)
+{
+  switch (response)
+    {
+    case GTK_RESPONSE_CLOSE:
+      gtk_widget_hide (preferences_dialog);
+      break;
+    default:
+      break;
+    }
+}
+
+static void
+ensure_prefs_window_is_created (NaTrayApplet *applet)
+{  
+  if (applet->priv->dialog)
+    return;
+
+  applet->priv->dialog = g_new0 (NAPreferencesDialog, 1);
+  
+  applet->priv->dialog->preferences_dialog = GTK_WIDGET (gtk_builder_get_object (applet->priv->builder, "notification_area_preferences_dialog"));
+
+  gtk_window_set_icon_name (GTK_WINDOW (applet->priv->dialog->preferences_dialog), NOTIFICATION_AREA_ICON);
+  
+  applet->priv->dialog->min_icon_size_spin = GTK_WIDGET (gtk_builder_get_object (applet->priv->builder, "min_icon_size_spin"));
+  g_return_if_fail (applet->priv->dialog->min_icon_size_spin != NULL);
+  
+  gtk_spin_button_set_range (GTK_SPIN_BUTTON (applet->priv->dialog->min_icon_size_spin), 7, 100);
+  gtk_spin_button_set_value (GTK_SPIN_BUTTON (applet->priv->dialog->min_icon_size_spin), applet->priv->min_icon_size);
+
+  g_signal_connect_swapped (applet->priv->dialog->min_icon_size_spin, "value_changed",
+                            G_CALLBACK (na_preferences_dialog_min_icon_size_changed),
+                            applet);
+
+  g_signal_connect_swapped (applet->priv->dialog->preferences_dialog, "response",
+                            G_CALLBACK (na_preferences_dialog_response), applet);
+
+  g_signal_connect (G_OBJECT (applet->priv->dialog->preferences_dialog), "delete_event",
+                    G_CALLBACK (na_preferences_dialog_hide_event), applet);
+}
+
+static void
+properties_dialog (GtkAction    *action,
+                   NaTrayApplet *applet)
+{
+  ensure_prefs_window_is_created (applet);
+
+  gtk_window_set_screen (GTK_WINDOW (applet->priv->dialog->preferences_dialog),
+                         gtk_widget_get_screen (GTK_WIDGET (applet)));
+  gtk_window_present (GTK_WINDOW (applet->priv->dialog->preferences_dialog));
 }
 
 static void help_cb(GtkAction* action, NaTrayApplet* applet)
@@ -173,6 +258,7 @@ static void about_cb(GtkAction* action, NaTrayApplet* applet)
 		"Vincent Untz <vuntz@gnome.org>",
 		"Alberts Muktupāvels",
 		"Colomban Wendling <cwendling@hypra.fr>",
+		"Fabien Broquard <braikar@gmail.com>",
 		NULL
 	};
 
@@ -200,6 +286,9 @@ static void about_cb(GtkAction* action, NaTrayApplet* applet)
 }
 
 static const GtkActionEntry menu_actions [] = {
+	{ "SystemTrayPreferences", "document-properties", N_("_Preferences"),
+	  NULL, NULL,
+	  G_CALLBACK (properties_dialog) },
 	{ "SystemTrayHelp", "help-browser", N_("_Help"),
 	  NULL, NULL,
 	  G_CALLBACK (help_cb) },
@@ -230,6 +319,10 @@ na_tray_applet_realize (GtkWidget *widget)
 
   // load min icon size
   gsettings_changed_min_icon_size (applet->priv->settings, KEY_MIN_ICON_SIZE, applet);
+
+  applet->priv->builder = gtk_builder_new ();
+  gtk_builder_set_translation_domain (applet->priv->builder, GETTEXT_PACKAGE);
+  gtk_builder_add_from_resource (applet->priv->builder, NA_RESOURCE_PATH "notification-area-preferences-dialog.ui", NULL);
 }
 
 static void
@@ -238,6 +331,8 @@ na_tray_applet_dispose (GObject *object)
 #ifdef PROVIDE_WATCHER_SERVICE
   g_clear_object (&NA_TRAY_APPLET (object)->priv->sn_watcher);
 #endif
+
+  g_clear_object (&NA_TRAY_APPLET (object)->priv->builder);
 
   G_OBJECT_CLASS (na_tray_applet_parent_class)->dispose (object);
 }
--- a/applets/notification_area/na-grid.c
+++ b/applets/notification_area/na-grid.c
@@ -25,7 +25,6 @@
  * don't make a big deal out of it. */
 
 #include "config.h"
-
 #include <gtk/gtk.h>
 
 #include "na-grid.h"
@@ -33,18 +32,29 @@
 #include "system-tray/na-tray.h"
 #include "status-notifier/sn-host-v0.h"
 
-#define ICON_SPACING 1
-#define MIN_GRID_SIZE 3
+#define MIN_ICON_SIZE_DEFAULT 24
+
+typedef struct
+{
+  GtkOrientation  orientation;
+  gint            index;
+  NaGrid         *grid;
+} SortData;
 
 struct _NaGrid
 {
-  GtkBox parent;
+  GtkGrid    parent;
 
-  gint icon_padding;
-  gint icon_size;
+  gint       icon_padding;
+  gint       icon_size;
 
-  GSList *hosts;
-  GSList *items;
+  gint       min_icon_size;
+  gint       cols;
+  gint       rows;
+  gint       length;
+
+  GSList    *hosts;
+  GSList    *items;
 };
 
 enum
@@ -54,7 +64,7 @@ enum
   PROP_ICON_SIZE
 };
 
-G_DEFINE_TYPE (NaGrid, na_grid, GTK_TYPE_BOX)
+G_DEFINE_TYPE (NaGrid, na_grid, GTK_TYPE_GRID)
 
 static gint
 compare_items (gconstpointer a,
@@ -85,16 +95,95 @@ compare_items (gconstpointer a,
 }
 
 static void
-reorder_items (GtkWidget *widget,
-               gpointer   user_data)
+sort_items (GtkWidget *item,
+            SortData  *data)
 {
-  NaGrid *nb;
-  gint position;
+  gint col, row, left_attach, top_attach;
 
-  nb = NA_GRID (user_data);
+  /* row / col number depends on whether we are horizontal or vertical */
+  if (data->orientation == GTK_ORIENTATION_HORIZONTAL)
+    {
+      col = data->index / data->grid->rows;
+      row = data->index % data->grid->rows;
+    }
+  else
+    {
+      row = data->index / data->grid->cols;
+      col = data->index % data->grid->cols;
+    }
 
-  position = g_slist_index (nb->items, widget);
-  gtk_box_reorder_child (GTK_BOX (nb), widget, position);
+  /* only update item position if it has changed from current */
+  gtk_container_child_get (GTK_CONTAINER (data->grid),
+                           item,
+                           "left-attach", &left_attach,
+                           "top-attach", &top_attach,
+                           NULL);
+
+  if (left_attach != col || top_attach != row)
+    {
+      gtk_container_child_set (GTK_CONTAINER (data->grid),
+                               item,
+                               "left-attach", col,
+                               "top-attach", row,
+                               NULL);
+    }
+
+  /* increment to index of next item */
+  data->index++;
+}
+
+static void
+refresh_grid (NaGrid *self)
+{
+  GtkOrientation orientation;
+  GtkAllocation allocation;
+  gint rows, cols, length;
+  
+  orientation = gtk_orientable_get_orientation (GTK_ORIENTABLE (self));
+  gtk_widget_get_allocation (GTK_WIDGET (self), &allocation);
+  length = g_slist_length (self->items);
+
+  if (orientation == GTK_ORIENTATION_HORIZONTAL)
+    {
+      rows = MAX (1, allocation.height / self->min_icon_size);
+      cols = MAX (1, length / rows);
+      if (length % rows)
+        cols++;
+    }
+  else
+    {
+      cols = MAX (1, allocation.width / self->min_icon_size);
+      rows = MAX (1, length / cols);
+      if (length % cols)
+        rows++;
+    }
+  
+  if (self->cols != cols || self->rows != rows || self->length != length)
+    {
+      self->cols = cols;
+      self->rows = rows;
+      self->length = length;
+      
+      SortData data;
+      data.orientation = gtk_orientable_get_orientation (GTK_ORIENTABLE (self));
+      data.index = 0;
+      data.grid = self;
+
+      g_slist_foreach (self->items,
+                       (GFunc) sort_items,
+                       &data);
+    }
+}
+
+void
+na_grid_set_min_icon_size (NaGrid *grid,
+                           gint    min_icon_size)
+{
+  grid->min_icon_size = min_icon_size;
+  
+  g_return_if_fail (NA_IS_GRID (grid));
+  
+  refresh_grid (grid);
 }
 
 static void
@@ -111,10 +200,17 @@ item_added_cb (NaHost *host,
                           G_BINDING_SYNC_CREATE);
 
   self->items = g_slist_prepend (self->items, item);
-  gtk_box_pack_start (GTK_BOX (self), GTK_WIDGET (item), FALSE, FALSE, 0);
+
+  gtk_widget_set_hexpand (GTK_WIDGET (item), TRUE);
+  gtk_widget_set_vexpand (GTK_WIDGET (item), TRUE);
+  gtk_grid_attach (GTK_GRID (self),
+                   GTK_WIDGET (item),
+                   self->cols - 1,
+                   self->rows - 1,
+                   1, 1);
 
   self->items = g_slist_sort (self->items, compare_items);
-  gtk_container_foreach (GTK_CONTAINER (self), reorder_items, self);
+  refresh_grid (self);
 }
 
 static void
@@ -128,52 +224,25 @@ item_removed_cb (NaHost *host,
 
   gtk_container_remove (GTK_CONTAINER (self), GTK_WIDGET (item));
   self->items = g_slist_remove (self->items, item);
-}
-
-static void
-update_size_and_orientation (NaGrid         *self,
-                             GtkOrientation  orientation)
-{
-  /* FIXME: do we really need that?  comes from NaTray */
-  /* FIXME: if we do, do that in overridden preferred size handlers */
-
-  /* note, you want this larger if the frame has non-NONE relief by default. */
-  switch (orientation)
-    {
-    case GTK_ORIENTATION_VERTICAL:
-      /* Give grid a min size so the frame doesn't look dumb */
-      gtk_widget_set_size_request (GTK_WIDGET (self), MIN_GRID_SIZE, -1);
-      break;
-    case GTK_ORIENTATION_HORIZONTAL:
-      gtk_widget_set_size_request (GTK_WIDGET (self), -1, MIN_GRID_SIZE);
-      break;
-    }
-}
-
-static void
-orientation_notify (GObject    *object,
-                    GParamSpec *pspec,
-                    gpointer    data)
-{
-  update_size_and_orientation (NA_GRID (object),
-                               gtk_orientable_get_orientation (GTK_ORIENTABLE (object)));
+  refresh_grid (self);
 }
 
 static void
 na_grid_init (NaGrid *self)
 {
-  GtkOrientation orientation;
-
   self->icon_padding = 0;
   self->icon_size = 0;
 
+  self->min_icon_size = MIN_ICON_SIZE_DEFAULT;
+  self->cols = 1;
+  self->rows = 1;
+  self->length = 0;
+
   self->hosts = NULL;
   self->items = NULL;
-
-  orientation = gtk_orientable_get_orientation (GTK_ORIENTABLE (self));
-  update_size_and_orientation (self, orientation);
-
-  g_signal_connect (self, "notify::orientation", G_CALLBACK (orientation_notify), NULL);
+  
+  gtk_grid_set_row_homogeneous (GTK_GRID (self), TRUE);
+  gtk_grid_set_column_homogeneous (GTK_GRID (self), TRUE);
 }
 
 static void
@@ -276,6 +345,14 @@ na_grid_unrealize (GtkWidget *widget)
 }
 
 static void
+na_grid_size_allocate (GtkWidget     *widget,
+                       GtkAllocation *allocation)
+{
+  GTK_WIDGET_CLASS (na_grid_parent_class)->size_allocate (widget, allocation);
+  refresh_grid (NA_GRID (widget));
+}
+
+static void
 na_grid_get_property (GObject    *object,
                       guint       property_id,
                       GValue     *value,
@@ -336,6 +413,7 @@ na_grid_class_init (NaGridClass *klass)
   widget_class->realize = na_grid_realize;
   widget_class->unrealize = na_grid_unrealize;
   widget_class->style_updated = na_grid_style_updated;
+  widget_class->size_allocate = na_grid_size_allocate;
 
   g_object_class_install_property (gobject_class, PROP_ICON_PADDING,
     g_param_spec_int ("icon-padding",
@@ -357,7 +435,6 @@ na_grid_new (GtkOrientation  orientation)
 {
   return g_object_new (NA_TYPE_GRID,
                        "orientation", orientation,
-                       "spacing", ICON_SPACING,
                        NULL);
 }
 
--- a/applets/notification_area/na-grid.c
+++ b/applets/notification_area/na-grid.c
@@ -21,22 +21,22 @@
  */
 
 /* Well, actuall'y is the Tray itself, the container for the items.  But
- * NaTray is already taken for the XEMBED part, so for now it's called NaBox,
+ * NaTray is already taken for the XEMBED part, so for now it's called NaGrid,
  * don't make a big deal out of it. */
 
 #include "config.h"
 
 #include <gtk/gtk.h>
 
-#include "na-box.h"
+#include "na-grid.h"
 
 #include "system-tray/na-tray.h"
 #include "status-notifier/sn-host-v0.h"
 
 #define ICON_SPACING 1
-#define MIN_BOX_SIZE 3
+#define MIN_GRID_SIZE 3
 
-struct _NaBox
+struct _NaGrid
 {
   GtkBox parent;
 
@@ -54,7 +54,7 @@ enum
   PROP_ICON_SIZE
 };
 
-G_DEFINE_TYPE (NaBox, na_box, GTK_TYPE_BOX)
+G_DEFINE_TYPE (NaGrid, na_grid, GTK_TYPE_BOX)
 
 static gint
 compare_items (gconstpointer a,
@@ -88,10 +88,10 @@ static void
 reorder_items (GtkWidget *widget,
                gpointer   user_data)
 {
-  NaBox *nb;
+  NaGrid *nb;
   gint position;
 
-  nb = NA_BOX (user_data);
+  nb = NA_GRID (user_data);
 
   position = g_slist_index (nb->items, widget);
   gtk_box_reorder_child (GTK_BOX (nb), widget, position);
@@ -100,11 +100,11 @@ reorder_items (GtkWidget *widget,
 static void
 item_added_cb (NaHost *host,
                NaItem *item,
-               NaBox  *self)
+               NaGrid *self)
 {
   g_return_if_fail (NA_IS_HOST (host));
   g_return_if_fail (NA_IS_ITEM (item));
-  g_return_if_fail (NA_IS_BOX (self));
+  g_return_if_fail (NA_IS_GRID (self));
 
   g_object_bind_property (self, "orientation",
                           item, "orientation",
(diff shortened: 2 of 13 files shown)


braikar:
Sorry, I manually modified all the files on github, I forgot that change (I just added the commit), it's done. na-box has been renamed to na-grid everywhere..

braikar:
I couldn't prevent myself from fixing almost everything.
So the makefile is updated, now the notification-area gschema is into the package.
And I fixed the panel loading and panel resize. It's almost perfect.
Frankly, the only display bug I can find is if I keep on changing repeatedly the icon sizes (no one does that..), once in a while they end up clipped.

Remains the minor issue:
> "GtkDialog mapped without a transient parent. This is discouraged." in the syslog

lukefromdc:
A wide panel showing both the menubutton and main menu bar:
full_wide_panel

Partial wide panel showing very nice window list, and tray with icons in one vertical line. 
partial_wide_panel

Setting "display icons only in one line" to false will show them across the width of a wide vertical panel and also across the height of a a (rarely used) wide horizontal panel

Note here the applet handle icon issue (which may not be easy to fix), which cannot be seen except if the system panel theme background rather than the usual partially transparent custom background is used on a wide left panel:
bad_applet_handle_on_panel

braikar:
Hi, great thanks.
For the applet handle, it was always there for me, even in vertical position (whichever way the notification area is setup, one-line or grid), but it is only 2-3 pixels wide..
I hadn't encountered the issue, I will look at it later.

For the panel menu bar, I had started working on it, but the changes are not properly working yet.
screenshot at 2018-08-23 15-22-40

I can get it to display properly on wide panel (vertical or horizontal), but once it is aligned like that, items on top of each others, it doesn't resize back to, inline items, when the panel size is set to be slim.
It's a bit harder to code because I have to detect the natural size of the applet that will trigger the "wide_panel_position", text doesn't resize like icons..

I will work on that slowly to get it to work properly too. For the moment I don't have time anymore.

lukefromdc:
I have my normal horizontal panel at the bottom, don't want to have to reconfigure that. The left side panel in my images was made just to test this, as my normal setup is a single slim panel at the bottom. 

That small white square above the window-list(with just one window on that workspace) is the narrow applet handle icon (just one of the three dashes is rendered), which renders on a vertical panel of any dimensions with the system background. The wider the panel the easier it is to miss that. Not that important as the system background is not going to be used that often with a wide left panel. I tried to fix that when we originally redid the code for that icon, but there was something that made it extremly difficult. I don't remember for sure but it may have been not having orientation passed to the necessary function.

braikar:
Ok.
I just finished cleaning up the code and re-arranging some functions/code that was written a bit hastily.
Apart from the problems you mentioned that I can't fix (I never even encountered that problem..), it's all good.

lukefromdc:
Nice cleanup, another test build built and ran fine. Here is an image of the part of a test panel with some applets added and the latest changes:

raveit65:
Are you familiar with `git rebase - i last-commit-id-before-you-want-change-git-log` command?
With that command you can edit/squash/drop/re-order commits local.
It opens a self-explanatory vi editor in terminal. Don't worry it's simple :)
After all is edited do a `git push -f` to update your repo.
This will update the PR automatic.
I am using always single commits to build a RPM for testing.
But 30 single commits are to much.
Thank you.

braikar:
@raveit65 nope, I'm not really familiar with anything ;) I'm learning to use git as I'm coding these changes.. I knew for sure my way of doing is not right yet! Thanks for the help.
I'm trying to get better at it.
Thanks I'll now try to use rebase -i commit_id
Could you answer these questions for me please?

A. I read about it and thought it wasn't important that I publish so many commits, because they're commits in the master branch of my repo/user? (but I agree it would still make my branch more organized! i'll work on avoiding so many commits and using rebase :) )

B. Once the pull request is accepted into the master branch of mate-desktop, can't it be squashed as a single commit anyway? (Obviously if these were commits on the master branch of mate-desktop directly I'd be much more careful!)

C. Should I still rebase/squash everything on my branch? If it's just for me, I don't mind these first commits looking a bit bad ;)

D. If I rebase all the changes from my master branch (I'll try it later), will they be grouped in this pull request as well?

Thanks!

raveit65:
Please don't rebase everything in one commit. This is too much code change for one commit.
But currently i see 30 commits in this PR.
I am sure some of theme can be squashed together.

> C. Should I still rebase/squash everything on my branch? If it's just for me, I don't mind these first commits looking a bit bad ;)

Every commit from this PR will be merge to master with a merge.

> D. If I rebase all the changes from my master branch (I'll try it later), will they be grouped in this pull request as well?

Yes, a `git push -f` will update the PR automatic and The PR is grouped in the same way like you repo.

braikar:
Thanks, but I don't get it..
I need specific info else I'm going to create always more commits and wrong pull requests..

Exact sequence i did last :
-- git clone https://github.com/braikar/mate-panel
-- cd mate-panel
-- git status (it's all good)
-- git rebase -i 038138df469caf5d110c86695a7a55c0aa5932d9 (to merge all the last code clean up commits)
-- git commit
--> responded: On branch master
Your branch and 'origin/master' have diverged,
and have 1 and 6 different commits each, respectively.
  (use "git pull" to merge the remote branch into yours)
so I did
-- git pull
-- git commit
-- git push

and it created a new commit instead of what I expected to be a merge of all commits?

raveit65:
Ok, start with
-- git clone https://github.com/braikar/mate-panel
-- cd mate-panel
-- git rebase -i 0e32e29fcc1dfda6b8764c7a7fda11aa9830065e

This will open an editor in terminal and you should see all commits which you did.
The editor is self-explanatory.
I hope you know minimal vi commands
After you did changes in vi editor close him with `wg!`
Rest depends of what you did in vi editor.

0e32e29fcc1dfda6b8764c7a7fda11aa9830065e is a commit of mine, the last one before you start to commit.
I suggest to do only one change for learning how it works.
If the change is done, do
`git push -f`
to push local change to github repo. (-f = --force)
PR will update automatic.

> Your branch and 'origin/master' have diverged,
and have 1 and 6 different commits each, respectively.
(use "git pull" to merge the remote branch into yours)

This will gone with the `git push -f` command which overrides github repo with your new local changes.
Of course this can't be undo, be careful.
You can also create a new branch from master for training and learning.
`git checkout -b new-branch-name` (after the cd mate-panel command)

I hope this makes it more clear.

braikar:
@raveit65 great, many thanks for your help!
It took me a bit of try and test to see how it works, but I thing I got it, I re-arranged/grouped all the changes by each file and grouped the finals clean ups.

raveit65:
Any news?

braikar:
Just a quick hello, I wrote earlier, I'd be on holiday in a place with very bad internet (it's hard to even access here..) ;) I have this week left, I'll continue when I get back.. Talk to you soon.

braikar:
Hi, I won't open an issue for the GResource issue, it's not important, as I can test without it, these changes won't affect the end result even if I can't resolve the dependencies.. It just changes the makefile and one line in main.c anyway. I will revert those changes to keep the GResource.
I've looked at all the small inconsistencies left, compile warning for incorrect casts etc. also a small configure.ac change (need to include the gschema in there). I will commit/rebase all the changes by the end of this weekend and then it should be good to go.

braikar:
I will comment the changes correctly with the last rebase. basically some more clean up of incorrect casts, and I reverted back to using Gresource (for xml menu and preferences dialog glade ui) in the code and the makefile.
The change in configure.ac AC_OUTPUT is something I had missed at the beginning, if it's not there "@GETTEXT_PACKAGE@" in the gschema xml file does not get replaced.

Can you @lukefromdc or @raveit65 test to see if it all compiles and runs fine?
Thank you.

lukefromdc:
OK, I just did another test build, install, and run. Everything worked fine. This was a build with all possible applets in-process, over GTK 3.24 and Glib 2.58.0 (both locally compiled) on Debian Unstable.

Same test results as before, works really nicely and fixes some old and aggravating issues for wide panel users. We really need to get this merged prior to the 3.22 release, this is important.

raveit65:
Can you squash patch 18 into 10, please?
This will make it more clean to see what is really change in make file.
Meanwhile i will start testing.

applets/notification_area/main.c
```diff
@@ -416,44 +406,27 @@ static void
...
 	action_group = gtk_action_group_new ("NA Applet Menu Actions");
 	gtk_action_group_set_translation_domain (action_group, GETTEXT_PACKAGE);
 	gtk_action_group_add_actions (action_group, menu_actions, G_N_ELEMENTS(menu_actions), applet);
-	ui_path = g_build_filename (NOTIFICATION_AREA_MENU_UI_DIR, "notification-area-menu.xml", NULL);
-	//mate_panel_applet_setup_menu_from_resource (MATE_PANEL_APPLET (applet),
-	//					      NA_RESOURCE_PATH "notification-area-menu.xml",
-	//					      action_group);
-	mate_panel_applet_setup_menu_from_file (MATE_PANEL_APPLET(applet), ui_path, action_group);
-	g_free (ui_path);
+	mate_panel_applet_setup_menu_from_resource (MATE_PANEL_APPLET (applet),
```
raveit65: Why not reverting this where you comment out the code?

raveit65:
I can't say this is good, because for me this is too confusing :)

applets/notification_area/main.c
braikar: I have removed the original changes where I undid the GResource changes, so this commit does not exist anymore.

braikar:
I've been at it all night to make it as clean as possible. All the "in-between" changes are gone, so now the diffs are quite easy to read. Except that one maybe: https://github.com/mate-desktop/mate-panel/pull/856/commits/e3fb3ff101e09620279d9acaff9c7946ed0dae28 the differences are too big to be shown as a diff..

applets/notification_area/na-grid.c
```diff
@@ -0,0 +1,523 @@
...
+
+	if (c1 < c2)
+		return -1;
+	else if (c1 > c2)
+		return 1;
+
+	id1 = na_item_get_id (item1);
+	id2 = na_item_get_id (item2);
+
+	return g_strcmp0 (id1, id2);
```
sc0w: If you need to do a comparison here, I think you must to do:

`return (g_strcmp0 (id1, id2) == 0);`

Without `"== 0"` , it doesn't work as expected.

raveit65:
Thanks a lot, this is much better and more clear what every commit does :)
I think it is save to squash commit 12 into 10, or not?
Looks like they belongs together...
It's just a one liner in configure.ac.

raveit65:
I tested this in a VM.
- na-preferences doesn't come up
- na-tray has vertical direction on vertical panel
- menu-applet seems to be ok on vertical panel , just smaller

braikar:
@raveit65 I've merged the 2 commits you mentioned it made sense :)
But I've removed some code based on your reaction to testing, you expected it to work right out of the box, 
> na-tray has vertical direction on vertical panel
yes, because that's the expect default behavior, on first load it is set to display as it was before, icons aligned in the direction of the panel. That's why I had set an option "display icons in one line" with a tick box.
I actually removed it now, as you can see on the screenshot.
na_tray
The preferences are simply the icon size, nothing else. It simply makes more sense.
That's the reasoning that came to my mind.
When a new panel is loaded, its width is 24 (at least on my system?) I don't know if that's a global constant everywhere, but the default panel is for sure quite slim, even if it's 26-28, it's fine.
The default minimum icon size of the notification area applet is set to 26, which means, if given space the icons will expand to a maximum of 26x26 (if the panel is below that size they will expand to the maximum allowed width/height anyway), having  26 as a default size, means that to display a grid of 2 icons wide, it will require at least 52pixels wide, so if the panel is expanded to 30-40-50 the notification area will keep on displaying on a single line, until 52px is reached. That should fit most users who like to keep a narrow panel (vertical and horizontal), the ones who want a wide panel will have to figure out there is an option to set the icon size. ANd if they want to keep a single line of very wide icons (which seems totally unrealistic anyway), by setting the icon size to 100 for example, the "display icons on one line" behavior can still work (even if I think no one will ever use that).
So having the icons display in one line will be a default anyway, unless someone expands beyond 52pixels wide, or if they have a 30px wide panel and still want 2 rows/columns they can set the icon size to 15 anyway, etc..

Sorry for the changes again, but it made more sense based on your reaction @raveit65

**Remains one issue if it really is the case**
@lukefromdc can you test and see if you can access the preferences dialog in this last version? I don't understand why you, @raveit65, can't find it, I'm almost sure it's due to the handle of the applet which is 2-3 pixels wide, so getting the popup menu is actually hard when someone's not used to it. I'm 99.99% sure the preferences dialog works.. That's why your reaction made me make that change, even if someone doesn't know there's a preference dialog to change the icon size, by default it won't always display as one line aligned in the direction of the panel, the icon sizes will scale by default between 26-51px wide in a grid. I wonder even if I shouldn't set the default to 20 or close to that value?

braikar:
> i think the UI-file isn't installed or loaded.

Thanks, a detail I didn't know, I have to add the resource in the gresource xml file..
It should work now. (I'll merge will another commit later, I don't have time now)
If 24px is the default size in icon-themes I'll set it to 24?

If someone wants to use 20 trays, They'll change the code themselves. ;) And it doesn't work FYI, I noticed when testing that the notification area can only be added once to any of all the panels on the desktop. The second time it's added, it's always empty... No idea if it's meant to be like that, I guess it makes sense, no one needs it twice.

raveit65:
> And it doesn't work FYI, I noticed when testing that the notification area can only be added once to any of all the panels on the desktop. The second time it's added, it's always empty...

I meant a lot of tray-applets inside the na-applet, like nm-applet, volume-applet, dropbox-applet, etc...

raveit65:
Preferences UI is there with latest commit.
And i think it's a good idea to control the number of rows with the icon size.
extreme setting :)
vertical-panel-2

applets/notification_area/main.c
```diff
@@ -353,16 +501,17 @@ na_tray_applet_init (NaTrayApplet *applet)
...
-  gtk_widget_show (GTK_WIDGET (applet->priv->box));
+  gtk_container_add (GTK_CONTAINER (applet), applet->priv->grid);
+  gtk_widget_show (applet->priv->grid);
 
   atko = gtk_widget_get_accessible (GTK_WIDGET (applet));
   atk_object_set_name (atko, _("Panel Notification Area"));
 
   mate_panel_applet_set_flags (MATE_PANEL_APPLET (applet),
                           MATE_PANEL_APPLET_HAS_HANDLE|MATE_PANEL_APPLET_EXPAND_MINOR);
+
```
raveit65: Do we need this extra blank line ?   :)

applets/notification_area/main.c
```diff
@@ -317,6 +464,7 @@ na_tray_applet_class_init (NaTrayAppletClass *class)
 
   parent_class_change_orient = applet_class->change_orient;
   applet_class->change_orient = na_tray_applet_change_orient;
+
```
raveit65: Do we need this extra blank line ? :)

lukefromdc:
I think holding up a bugfix over tabs in source code (that final users won't see) is not a good idea. We can fix the tabs later if need be.

lukefromdc:
I think holding up a bugfix over tabs in source code (that final users won't see) is not a good idea. We can fix the tabs later if need be.

applets/notification_area/na-grid.h
```diff
@@ -32,25 +32,26 @@
...
-#define NA_TYPE_BOX			(na_box_get_type ())
-#define NA_BOX(obj)			(G_TYPE_CHECK_INSTANCE_CAST ((obj), NA_TYPE_BOX, NaBox))
-#define NA_BOX_CLASS(klass)		(G_TYPE_CHECK_CLASS_CAST ((klass), NA_TYPE_BOX, NaBoxClass))
-#define NA_IS_BOX(obj)			(G_TYPE_CHECK_INSTANCE_TYPE ((obj), NA_TYPE_BOX))
-#define NA_IS_BOX_CLASS(klass)		(G_TYPE_CHECK_CLASS_TYPE ((klass), NA_TYPE_BOX))
-#define NA_BOX_GET_CLASS(obj)		(G_TYPE_INSTANCE_GET_CLASS ((obj), NA_TYPE_BOX, NaBoxClass))
+#define NA_TYPE_GRID			(na_grid_get_type ())
+#define NA_GRID(obj)			(G_TYPE_CHECK_INSTANCE_CAST ((obj), NA_TYPE_GRID, NaGrid))
+#define NA_GRID_CLASS(klass)		(G_TYPE_CHECK_CLASS_CAST ((klass), NA_TYPE_GRID, NaGridClass))
+#define NA_IS_GRID(obj)			(G_TYPE_CHECK_INSTANCE_TYPE ((obj), NA_TYPE_GRID))
```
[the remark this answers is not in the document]
braikar: This actually looks right in a file editor, but the tab indents in github don't all look right, I changed it to spaces..

braikar:
All done.
I just removed the complete code cleanup part and stuck to the coding style of na-tray..
It's really weird, all the files in mate-panel/libmate-panel-applet are indented with tabs, some files with 4s paces in dents.
The 3 other applets have almost only tabs also.
And then notification area, with indents of 2spaces, and a weird style of indenting the if-else/switch brackets with 2 more spaces ;)
main.c in the notification area has some parts with tab indents also.. the about/help and menu actions parts, I left it as is.

That's why I felt like changing to tabs at first (follow the global style used almost everywhere else, the indent style of notification area is really an outcast ^^), but as you requested and it actually make sense I didn't change the style at all, so the changes are just the changes for the vertical/wide panels issues and nothing else.
If the style is to be changed that's better in another commit  :)

applets/notification_area/na-grid.h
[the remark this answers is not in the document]
raveit65: This depends on tab sizes in editors, which can be different by everyone.
For this reason i don't use tabs personal.

raveit65:
Some commits have same commit header.
```
[rave@satellite mate-panel]$ git format-patch 2bccd4e4ac308918e2cbdf7f53f230bd41d7dedb
0001-Wide-panels-updated-resize-limits-on-Widgets-and-arr.patch
0002-Wide-panels-GtkBox-to-GtkGrid-display.patch
0003-Wide-panels-GtkBox-to-GtkGrid-display.patch
0004-Wide-panels-add-GSchema-and-key-for-new-preference-o.patch
<cut>
```
See how it looks if someone is browsing you own repo.
https://github.com/braikar/mate-panel/commits/master

raveit65:
Hmm, building failed with lates changes here.
```
na-grid.c:286:18: error: unknown type name 'NaBox'; did you mean 'NaHost'?
                  NaBox  *self)
                  ^~~~~
                  NaHost
In file included from /usr/include/glib-2.0/gobject/gobject.h:27,
                 from /usr/include/glib-2.0/gobject/gbinding.h:29,
                 from /usr/include/glib-2.0/glib-object.h:23,
                 from /usr/include/glib-2.0/gio/gioenums.h:28,
                 from /usr/include/glib-2.0/gio/giotypes.h:28,
                 from /usr/include/glib-2.0/gio/gio.h:26,
                 from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:28,
                 from /usr/include/gtk-3.0/gtk/gtk.h:30,
                 from na-grid.c:28:
na-grid.c: In function 'add_host':
na-grid.c:337:40: error: 'item_removed_cb' undeclared (first use in this function); did you mean 'item_added_cb'?
                            G_CALLBACK (item_removed_cb), self, 0);
                                        ^~~~~~~~~~~~~~~
/usr/include/glib-2.0/gobject/gclosure.h:70:40: note: in definition of macro 'G_CALLBACK'
 #define G_CALLBACK(f)    ((GCallback) (f))
                                        ^
na-grid.c:337:40: note: each undeclared identifier is reported only once for each function it appears in
                            G_CALLBACK (item_removed_cb), self, 0);
                                        ^~~~~~~~~~~~~~~
/usr/include/glib-2.0/gobject/gclosure.h:70:40: note: in definition of macro 'G_CALLBACK'
 #define G_CALLBACK(f)    ((GCallback) (f))
                                        ^
make[5]: *** [Makefile:792: na-grid.lo] Error 1
make[5]: Leaving directory '/media/Programme/Linux/Mate-Desktop/git-version/github-matedesktop/mate-panel-compile/applets/notification_area'
make[4]: *** [Makefile:895: all-recursive] Error 1
make[4]: Leaving directory '/media/Programme/Linux/Mate-Desktop/git-version/github-matedesktop/mate-panel-compile/applets/notification_area'
make[3]: *** [Makefile:608: all] Error 2
make[3]: Leaving directory '/media/Programme/Linux/Mate-Desktop/git-version/github-matedesktop/mate-panel-compile/applets/notification_area'
make[2]: *** [Makefile:451: all-recursive] Error 1
```

braikar:
Sorry for the compile error, i did it so quickly this morning I copied stuff from the old code and forgot one renamign from NaBox to NaGrid.
I can see how the commits look, but I don't know what's the best convention to use, since many the changes are for the same purpose, I don't know what other descriptions to use, it's the same changes in many files..?
I grouped main.c and main.h together and na-box.c/h->na-grid.c/h together
I don't know what else to do now, if you think other descriptions/groupings are needed just tell me what I should change :) I frankly don't know if it's good or not.

muktupavels:
https://github.com/mate-desktop/mate-panel/pull/856/commits/8528cbfba7262c6263dd0bb0592f908d3158135f

I would say this is huge no... Can you build that commit? If not then it should not be separate commit.

Also that commit is not good for reviewing. According to commit message it includes other changes, but diff just shows that it is completely new file.

To make it easier to review you could do one commit that renames only files - `na-box.c/h` to `na-grid.c/h` and related changes that are needed to make sure that build does not fail. So you need update `Makefile.am` and also `na-grid.c` to include `na-grid.h` not `na-box.h`.

Then you could do commit that changes it from `GtkBox` to `GtkGrid` and changes to make sure that it builds and works. After that - commit that adds preferences.

Of course it is just example, its your work and you probably should know better how to logically separate changes to achieve end result avoiding commits that breaks build.

raveit65:
I complete agree with @muktupavels hints.

braikar:
@muktupavels @raveit65 thanks for the help and guidance.
Then if I were to make commits that don't break build, it would end up as 3 separate commits (as far as i can think of it at the moment).

1. commit (convert box to grid)
change na-box.c/h to na-grid.c/h and changing all the code in there
this requires 2 lines changes in test tray (rename box to grid..)
and a couple of changes in main.c (again rename box to grid mostly)
and the make file rename box to grid.

2. commit (add the gsettings)
 creating the gschema.xml file
adding the gschema in POTFILES, in AC_OUTPUT and in the makefile
and adding some lines of code in main.c/main.h

3. commit (adding the preferences menu)
creating the galde.ui file
adding the ui file to potfiles
changing the makefile again
and main.c/main.h

That's the only way it would work on each commit.
With the current state of how I made the commits, the build will break at every commit, they need to be all applied all at once..

muktupavels:
@braikar sounds good, but... really consider first commit as two commits. Do first commit that **only** renames files and "fixes" includes. Then do real conversion from box to grid as second commit. That way it would be easier to review changes that are done to change from box to grid.

braikar:
@muktupavels about the renaming from na-box.c to na-grid.c I was also wondering that myself.. I don't know, I have been learning about using git as I was making these changes, so it's sure, I made the mistake of deleting and not renaming!
I will rework how the commits are made tonight with what you just explained :)
1. rename box/Box/BOX to grid/Grid/GRID
2. set the GtkBox to GtkGrid
3. set GSettings
4. add preferences
:)

applets/notification_area/na-grid.c
```diff
@@ -54,7 +54,7 @@ enum
   PROP_ICON_SIZE
 };
 
-G_DEFINE_TYPE (NaBox, na_box, GTK_TYPE_BOX)
+G_DEFINE_TYPE (NaGrid, na_grid, GTK_TYPE_BOX)
```
muktupavels: Should not this be `GTK_TYPE_GRID` here? I see that you changed that in next commit, but it should be in this commit...

applets/notification_area/na-grid.c
```diff
@@ -353,19 +353,19 @@ na_box_class_init (NaBoxClass *klass)
 }
 
 GtkWidget *
-na_box_new (GtkOrientation  orientation)
+na_grid_new (GtkOrientation  orientation)
```
muktupavels: Remove that extra space in this (first commit) not next one.

applets/notification_area/na-grid.h
```diff
@@ -32,25 +32,25 @@
...
+struct _NaGridClass
 {
-  GtkBoxClass parent_class;
+  GtkGridClass parent_class;
 };
 
-GType           na_box_get_type         (void);
-GtkWidget      *na_box_new              (GtkOrientation orientation);
-void            na_box_force_redraw     (NaBox *box);
+GType           na_grid_get_type        (void);
```
muktupavels: 35 - 50 lines could be simply two lines:
```
#define NA_TYPE_GRID (na_grid_get_type ())
G_DECLARE_FINAL_TYPE (NaGrid, na_grid, NA, GRID, GtkGrid)
```

applets/notification_area/main.c
```diff
@@ -186,6 +186,16 @@ static const GtkActionEntry menu_actions [] = {
...
 
+static void
+na_tray_applet_size_allocate (GtkWidget     *grid,
+                              GtkAllocation *allocation,
+                              gpointer       data)
+{
+  NaTrayApplet *applet = data;
+  g_return_if_fail (NA_IS_GRID (grid));
+
+  set_grid_display_mode (NA_GRID (grid), MIN_ICON_SIZE_DEFAULT);
```
muktupavels: Hmm, why this in size-allocate if min size does not depend on allocation? I think you can probably call this after creating new grid or even do this in `na_grid_init`, seems that you already do that.

I guess you want this here to call `refresh_grid`, right? But then you get size-allocate signal on grid itself so you should be able to handle that in `na-grid.c`.

applets/notification_area/na-grid.c
```diff
@@ -54,7 +66,7 @@ enum
   PROP_ICON_SIZE
 };
 
-G_DEFINE_TYPE (NaGrid, na_grid, GTK_TYPE_BOX)
+G_DEFINE_TYPE (NaGrid, na_grid, GTK_TYPE_GRID)
```
muktupavels: This change should be in previous commit.

applets/notification_area/na-grid.h
```diff
@@ -48,6 +50,7 @@ struct _NaGridClass
 };
 
 GType           na_grid_get_type        (void);
+void            set_grid_display_mode   (NaGrid *grid, gint min_icon_size);
```
muktupavels: You should use `na_` prefix here... But I think you can handle what you want in `na-grid.c` and this is not even needed.

applets/notification_area/main.c
```diff
@@ -104,6 +109,32 @@ get_gtk_orientation_from_applet_orient (MatePanelAppletOrient orient)
...
 }
 
+static void
+gsettings_changed_min_icon_size (GSettings    *settings,
+                                 gchar        *key,
+                                 NaTrayApplet *applet)
+{
+  applet->priv->min_icon_size = g_settings_get_int (settings, key);
+
+  set_grid_display_mode (NA_GRID (applet->priv->grid), applet->priv->min_icon_size);
```
muktupavels: Ah, now I see the use for this function, but then you probably should add it in this commit, because in previous commits this function call looks unneeded. Also function name seems wrong and does not match what it does... To me looks like this should be called `na_grid_set_min_icon_size`

applets/notification_area/main.c
```diff
@@ -214,6 +245,9 @@ na_tray_applet_realize (GtkWidget *widget)
                                               action_group);
   g_object_unref(action_group);
 
+  setup_gsettings (applet);
+  load_gsettings (applet);
```
muktupavels: `load_gsettings` seems unneeded / duplicated code. You can call `gsettings_changed_min_icon_size (applet->priv->settings, KEY_MIN_ICON_SIZE, applet);` to load initial settings.

applets/notification_area/main.c
```diff
@@ -259,6 +352,9 @@ na_tray_applet_dispose (GObject *object)
 #endif
 
   G_OBJECT_CLASS (na_tray_applet_parent_class)->dispose (object);
+
+  if (&NA_TRAY_APPLET (object)->priv->builder)
+    g_clear_object (&NA_TRAY_APPLET (object)->priv->builder);
```
muktupavels: `g_clear_object` is `NULL` safe so you don't need to check if builder is NULL before calling g_clear_object.

muktupavels:
> Since I cannot test if it all compiles fine (as I can only test without the gresource..)

Why?

applets/notification_area/na-grid.c
braikar: same as above, all the names are grid, but after this first commit the gtkobject is still a box..? SHoudl I still change it even if the applet might probably crash on running?

applets/notification_area/na-grid.h
braikar: I tried to set a size-allocate function on the grid, that was the first thing I did (i added it to the GtkGridClass, it didn't work to update the grid. I'm not an expert in Gtk.. Is there a difference between setting it as a g_signal_callback and on the Class virtual functions? If I knew how to make it work I would not have created that function, I agree, it's not its place..

applets/notification_area/main.c
braikar: as above, if I knew how to get rid of size-allocate in main.c, that call would simply be to provide the min_icon_size to the grid and hence wouldn't appear as a dud "et_grid_display_mode (NA_GRID (applet->priv->grid), MIN_ICON_SIZE_DEFAULT);" in the previous as at that point there's no size to change.. That call there is to update the grid, without it there the grid didn't update..

applets/notification_area/main.c
braikar: ok, I will change that

applets/notification_area/na-grid.c
muktupavels: Well this looks incomplete to me. If it was supposed to be still `GtkBox` then why did you change `GtkBox parent;` to `GtkGrid parent;`? Sound that in this commit it should be `GtkBox parent;` still.

applets/notification_area/main.c
braikar: :+1:

applets/notification_area/na-grid.h
muktupavels: Well you can do `g_signal_connect (self, "size-allocate", callback, data);` in na-grid.c, no? Other option is to use `widget_class->size_allocate = na_grid_size_allocate;`

braikar:
@muktupavels I didn't manage to install the dependencies for `mate_panel_applet_setup_menu_from_resource`
since that's not in 1.20, but starting 1.21. I tried to compile/make mate-panel 1.21, it needed a dependency of libmate-menu 1.21, and when trying to compile that one it never worked. I'm not an expert, it didn't come with the debian folder of a standard package and I couldn't find the info on how to set it up quickly.

applets/notification_area/na-grid.h
braikar: I did set it up as widget_class->size_allocate as the first option, it didn't work.. but I haven't tried g_signal_connect on the gird directly, that's why I'm asking if there's a specific difference between setting it as a virtual_method on the class, or as a signal callback? In any case I'll try again to figure out how to get rid of the size-allocate in main.c.

muktupavels: Well it should work in both cases... What exactly did not work?

raveit65:
With latest changes or from this morning (the version with 9 commits) na-tray-applet runs with high cpu load. (System in idle mode)
I tested it the first time on bare metal.
na-applet --> 12 % cpu load
of course this increase the load of Xorg 
Xorg --> 16 % cpu load

@lukefromdc 
Can you test again please

raveit65:
@braikar 
Yes, you need mate-menus-1.21.0 installed.
Than you need to rebuild mozo, mate-panel and mate-screensaver against mate-menus-1.21.0.

braikar:
@raveit, strange, the order of the commits is changed, but I made sure the end state was exactly the same (except some spaces/indents)
I need to figure out how to install mate-menus 1.21! I just tried, I can't get further than running aclocal/autocont etc. until configure runs. Then running "dpkg-buildpackage -rfakeroot -b" gives me the error "tail: cannot open 'debian/changelog' for reading: No such file or directory"
Obvious because there's no debian folder. What I had done, was to get the debian folder from the mate-menus 1.20 package to try and compile, but that didn't work..

applets/notification_area/na-grid.h
braikar: Strange, I just tried as a g_signal.. it works perfectly! Yet I remember clearly I had tried as a class fucntion before and it didn't work.. Anyway, now that I know it works, I'll get on it tomorrow and all the inconsistencies you pointed out. Tonight I can't continue anymore.

raveit65:
I am not really an expert how to rebuild debian installer packages, maybe some other can help here.

lukefromdc:
I only use Checkinstall to build by Debian packages, I don't try to use Debian's normal files due to their checking for changes (presumed corruption or mitm) in the source packages. I can refer to the documentation to build a stock Debian package from its source and debian packages, but that's it. That's why all my debs are built one binary package to one source package. They install, but they are not direct replacments for Debian's provided packages that break one source package up into multiple binary packages.

Checkinstall can build a Deb from any arbitrary source, but watch out for it picking up all files in /usr/share/mine if anything is installed there, plus the icon cache and compiled gschemas. I interrupt Checkinstall,  copy the installer directory, remove those duplicate files, manually edit changelogs and DEBIAN/control, and then run dpkg -b to build the installer by hand.

braikar:
Thanks, before continuing (If there's time, When do you plan to release mate 1.22?), I opened an issue, I need help to learn how to properly work with git and the repos here, else I'm always going to have tons of troubles testing on non-yet fully released versions.. https://github.com/mate-desktop/mate-panel/issues/878

raveit65:
We don't have a fixed release date, but for sure not before next year.
A soft target is beginning of next year.

raveit65:
For dependencies of packages you can take a look here at
https://github.com/mate-desktop/debian-packages
They are written in control file of debian folder
https://github.com/mate-desktop/debian-packages/blob/master/mate-panel/debian/control
You need to have libmate-menu-dev-1.21.0 installed.
As there is no repo for linuxmint with mate-1.21.x you need to compile/install mate-menus for by hand or updating dep package.
As a newsbe i would work in a virtual machine to avoid crashing your main box.

I am still wondering that nobody has a repo with mate-1.21.x and debian packages online?

raveit65:
> Then there's another way of doing things I don't understand, since you can test against 1.21, how do you install all the dependencies?

I have a local repo for fedora where i have all 1.21.x RPMs and i test all PRs with using dnf/yum packagemanager.
An example:
`dnf install mate-menus-devel`

This will install the build requires.

braikar:
Sorry, again a problem with git. I'll have to refix the commits orders... Still learning

braikar:
@muktupavels I corrected the code for all your requests. And my mistakes.

applets/notification_area/na-grid.c
```diff
@@ -84,17 +98,157 @@ compare_items (gconstpointer a,
...
+  * then sorting them again in reorder_items_with_data()
+  *
+  * I couldn't find how to force the update with another mechanism?
+  * Just sorting them doesn't work.
+  * Removing them/re-adding them alone doesn't cause an update either.
+  * All these steps have to happen for the grid to update when the panel loading occurs, when a new item is added or when the min icon size is changed in the preferences...
+  */
+  for (item = self->items; item; item = item->next)
+    {
+      if (GTK_IS_WIDGET (item))
```
muktupavels: Can `items` list contain something that is not `GtkWidget`? If not then this check should not be needed...

applets/notification_area/na-grid.c
```diff
@@ -84,17 +98,157 @@ compare_items (gconstpointer a,
...
+        gtk_container_remove (GTK_CONTAINER (self), GTK_WIDGET (item));
+    }
 
-  nb = NA_GRID (user_data);
+  for (item = self->items; item; item = item->next)
+    {  
+      if (GTK_IS_WIDGET (item))
```
muktupavels: Same here...

muktupavels:
Just question / idea, but maybe it should not be `NaBox` or `NaGrid`, but for example `NaItems`, `NaIcons` or `NaTrayItems`?

First commit touches `testtray.c`, but I don't see changes in second commit. You should update test.

braikar:
For the name, there's a comment in `na-grid.h` (that was in `na-box.h` before)..

> /* Well, actuall'y is the Tray itself, the container for the items.  But
    * NaTray is already taken for the XEMBED part, so for now it's called NaGrid,
    * don't make a big deal out of it. */

;)

`NaTray` is a name used in the folder system-tray.. and there's already a `na-item` in main and `na-tray-child` in system-tray.
I find the naming convention not bad for so many names redundancy.
As for `testtray.c` there are no changes needed, all it does is create a window in which the tray/grid is added.. Since it's treated as a GtkWidget by all the rest of the code, a change from GtkBox to Grid has no effect there.

applets/notification_area/na-grid.c
muktupavels: My guess is that when you remove widget from container it is destroyed. Widgets have floating reference - https://developer.gnome.org/gobject/stable/gobject-The-Base-Object-Type.html#GInitiallyUnowned.

In short you need to temporarily increase reference to make sure it is not destroyed. Once it is added back you remove your temporarily reference.

braikar: I've been testing, actually it seems all that code becomes unnecessary. since I have the size-allocate set on the grid as it should, the reordering works almost perfectly simply with reorder_items_with_data(). I guess that having the size-allocate on the panel widget, plus the notify::size-changed on the grid, just triggered many resizes that conflicted with each other?
Btw. since I'm not expert in gtk, what does issue "notify::size-changed" signals? is it necessary to have both, size-allocate and notify::size-changed?

muktupavels: What is `notify::size-changed`? Quickly looking I don't see such property on `GtkGrid`, `GtkContainer` and/or `GtkWidget`...

braikar: Ok I don't know where I picked that up.. I've been testing even orientation notify is now useless with size_allocate directly on the GtkGrid. I just removed all that now unnecessary code.

applets/notification_area/na-grid.c
```diff
@@ -128,52 +236,32 @@ item_removed_cb (NaHost *host,
...
 na_grid_init (NaGrid *self)
 {
-  GtkOrientation orientation;
-
   self->icon_padding = 0;
   self->icon_size = 0;
 
   self->hosts = NULL;
   self->items = NULL;
+  self->grid_props = g_new0 (GridProps, 1);
```
muktupavels: You should free this in finalize.

applets/notification_area/na-grid.c
```diff
@@ -128,52 +236,32 @@ item_removed_cb (NaHost *host,
...
+  self->min_icon_size = MIN_ICON_SIZE_DEFAULT;
+  
+  gtk_grid_set_row_homogeneous (GTK_GRID (self), TRUE);
+  gtk_grid_set_column_homogeneous (GTK_GRID (self), TRUE);
 
-  orientation = gtk_orientable_get_orientation (GTK_ORIENTABLE (self));
-  update_size_and_orientation (self, orientation);
-
-  g_signal_connect (self, "notify::orientation", G_CALLBACK (orientation_notify), NULL);
+  g_signal_connect (self, "size-allocate", G_CALLBACK (na_grid_size_allocate), NULL);
```
muktupavels: I would prefer if this was `widget_class->size_allocate = ...;`, but that should not change functionality - so whatever MATE developers (I am not) prefer.

applets/notification_area/na-grid.h
```diff
@@ -33,8 +33,9 @@
 G_BEGIN_DECLS
 
 #define NA_TYPE_GRID (na_grid_get_type ())
-G_DECLARE_FINAL_TYPE (NaGrid, na_grid, NA, GRID, GtkBox)
+G_DECLARE_FINAL_TYPE (NaGrid, na_grid, NA, GRID, GtkGrid)
 
+void            na_grid_set_min_icon_size       (NaGrid *grid, gint min_icon_size);
```
muktupavels: Typically each parameter is in new line.

applets/notification_area/na-grid.c
```diff
@@ -25,26 +25,40 @@
...
 
-#define ICON_SPACING 1
-#define MIN_GRID_SIZE 3
+#define MIN_ICON_SIZE_DEFAULT 24
+
+typedef struct
+{
+  gint cols;
+  gint rows;
+  gint length;
```
muktupavels: `length` seems unused

applets/notification_area/na-grid.c
braikar: I just set the cols/rows properties on the `_NaGrid struct` directly, and removed the unused length property.

braikar:
I made all the changes again, except the size_allocate one, as explained, it doesn't work when set on the widget_class. I'll try to figure out why.
Just a small git question, Every time I make changes I see that I completely change the commit by a forced push, I don't know how to do it any other way. Is there another way to do it to just update, not change the commit?

lukefromdc:
I would suggest installing git-gui, which makes working with git a lot easier, especially for a newcomer to it. To avoid a force-push, you would simply make your changes to the files that are going to change, then "stage" them, then "commit" them, and finally push the result as a new commit that does not have to be forced. A forced-update is used when you have to "rebase" to catch up with master, when you have "amended previous commit" to change a commit you have already pushed, or when you have squashed or reordered commits. That's its job, and one objective is to ensure that a forced update is never necessary on the master branch because it would force everyone using it to reset to a point before the update and rerun git pull. This in turn could break their local branches if rebase does not run clean (automatically).

applets/notification_area/na-grid.c
muktupavels: Did you chain-up? `GTK_WIDGET_CLASS (na_grid_parent_class)->size_allocate (grid, allocation);`?

muktupavels:
Since you are fixing existing commits it is OK to force push change, there is no really other way. Making new commits with small changes does not sound good to me...

applets/notification_area/na-grid.c
```diff
@@ -84,17 +93,102 @@ compare_items (gconstpointer a,
...
+                           "height", &bottom,
+                           NULL);
+  if (left != col || right != 1 || top != row || bottom != 1)
+    {
+      gtk_container_child_set (GTK_CONTAINER (data->grid),
+                               item,
+                               "left-attach", col,
+                               "width", 1,
+                               "top-attach", row,
+                               "height", 1,
```
muktupavels: Do you  really need to "update"  width and height here?

applets/notification_area/na-grid.c
braikar: No, now I know what's the difference between the g_signal_connect and setting it on the class. Thanks

braikar:
@muktupavels before I push the new changes, I was wondering there are many calls to
`gtk_orientable_get_orientation (GTK_ORIENTABLE (self));`
but `orientation` is attached to the `GObject`
`  return g_object_new (NA_TYPE_GRID,
                       "orientation", orientation,
                       NULL);`

is it necessary to always call `gtk_orientable_get_orientation`, isn't it already updated on the `GObject` property whenever the orientation changes? if yes, how do I fetch it?

braikar:
I've answered my own question.. `gtk_orientable_get_orientation` does exactly the `g_object_get` I repalced it with, but it adds a g_return_val_if_fail.. So I don't know, since here `orientation` property is known to be defined, the security check is not necessary?

muktupavels:
`GtkGrid` implements `GtkOrientable` interface:
https://developer.gnome.org/gtk3/stable/GtkGrid.html

I think `gtk_orientable_get_orientation (GTK_ORIENTABLE (grid));` should be same as `g_object_get (grid, "orientation", &orientation, NULL);`. In both cases you get orientation from object and would return same value.

I think interface might be used here to avoid need to create gtk_grid_[get/set]_orientation, gtk_box_[get/set]_orientation... You can see how many widgets implements that interface here:
https://developer.gnome.org/gtk3/stable/gtk3-Orientable.html#GtkOrientable

Orientation is updated when code does that. Applet emits "change-orient" signal when it gets "Orient" D-Bus signal from panel. To be on safe side I would probably refresh grid also on orientation change.

braikar:
> To be on safe side I would probably refresh grid also on orientation change.

I removed the `notify::orientation` callback because I noticed while testing that any orientation change always triggers a `size-allocate`. Adding the callback just makes the applet reorder the grid twice (at the current state of how the gtk events happen). I can add a comment to explain why I didn't set the orientation callback, in case it changes..

muktupavels:
I would do differently. Call `refresh_grid` when there are something that could affect it, in grid calculate new cols, rows and then check if cols, rows and items length has changed. If things are same then you don't need to do anything. :)

I don't see that setting "orientation" property would trigger size-allocate... For example when you do `gtk_grid_set_row_spacing` it calls `gtk_widget_queue_resize`, but there is no such thing for orientation property.

https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkgrid.c#L231
https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkorientable.c#L78

For example - `GtkBox` does that:
https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkbox.c#L383

Maybe "bug" in gtk, don't know...

braikar:
Good points :) I actually had used a lot of older code I never bothered to cleanup (carried froward since the time of unverified gnome2 patches!..)

I changed some function names too, makes more sense, PackData, was there because before it used to be sorting in multiple boxes..

> I don't see that setting "orientation" property would trigger size-allocate...

It's not something that happens in the code, but it's specific to the panel. The only way an orientation change happens, is when the user decides to set the panel somewhere else, when that happens, for the notification area, no matter what size it had, the width becomes the height and the height becomes the width, so the size-allocate is triggered anyway.

muktupavels:
I was thinking that you might get D-Bus signal after size-allocation has already run...

https://github.com/mate-desktop/mate-panel/blob/master/mate-panel/libmate-panel-applet-private/panel-applet-frame-dbus.c#L187

But seems that panel will queue resize after D-Bus call. :)

applets/notification_area/na-grid.c
```diff
@@ -275,6 +344,14 @@ na_grid_unrealize (GtkWidget *widget)
   GTK_WIDGET_CLASS (na_grid_parent_class)->unrealize (widget);
 }
 
+static void
+na_grid_size_allocate (GtkWidget     *widget,
+                       GtkAllocation *allocation)
+{
+  refresh_grid (NA_GRID (widget));
```
muktupavels: Hmm, I would call `refresh_grid` after chaining-up. Otherwise `gtk_widget_get_allocation` might return previous allocation (too lazy to check if that is really true).

braikar: I tried to queue it first, hello disco panel. ;) It works once in a while, but in many cases it causes heavy flicker on loading..

muktupavels: queue?

braikar: Bad choice of words. I meant queue as "put after".. so I tried to put `refresh_grid` after `GTK_WIDGET_CLASS..->size_allocate (..)` it did flicker a lot (not always) on panel load. I just corrected that, the check to be sure size_allocate is present on parent class, seems to solve it. I guess there's some order of execution on panel load when sometimes the grid size_allocate happens too early. I think the code works really well now. Thank you for your reviews. :)
I'm done for today, unfortunately I can't spend every free time on that everyday.

muktupavels: Problem might be that changing / updating grid in size allocate queues new resize...

`if (GTK_WIDGET_CLASS (na_grid_parent_class)->size_allocate)` should not be needed, I think it should never be `NULL`!

raveit65:
Cool, with current state the high cpu load from na-applet in idle mode is gone :)

lukefromdc:
This works very nicely as of Oct 18, 2018. Is this ready to be merged? Looks ready to go from here

braikar:
Thanks to all the reviews from @raveit65 and @muktupavels I cleaned it up as much as I could.
All that can be done is removing line 351
`if (GTK_WIDGET_CLASS (na_grid_parent_class)->size_allocate)`
in commit https://github.com/mate-desktop/mate-panel/pull/856/commits/b8e80482539b387089bfdf06a1f6b0af7a7ea639

I had left that "if" check until I had enough tests to see if it works fine without (in theory it should, but I had doubts..), I've had many reboots since then with that check removed and it works all fine.
It's good to go as is, but to make it even neater I should just rebase to remove that line and I see nothing else I can change anymore.. Should I do it, and then we call it done?

raveit65:
yes.

braikar:
Ok, done.

raveit65:
I just tested it again, all seems to be fine.
Ready to go.

lukefromdc:
A final touch would be an update from the documentation team for mate-user-guide for the new tray preferences dialog

PR #863, opened 2018-08-30
menu-bar: transfer focus correctly on alt-F1

### LINKED ISSUE #851: alt-F1 shortcut is not transferring focus correctly
Hello,

The background is in https://gitlab.gnome.org/GNOME/gtk/issues/85 . One of the conclusions, in https://gitlab.gnome.org/GNOME/gtk/issues/85#note_264804 , is that mate-panel is not properly transferring focus on alt-F1 keyboard shortcut. (on alt-F2 it is working properly).

#### How to reproduce

- install gtk3, which contains the fix for https://gitlab.gnome.org/GNOME/gtk/issues/85 (can be seen by gdk/x11/gdkdevicemanager-core-x11.c's APPEARS_FOCUSED testing for both has_focus and has_focus_window).
- run the mate desktop
- run accerciser, in its Event Monitor, open window element and enable window:activate and window:deactivate
- run e.g. mate-terminal, accerciser shows window:activate for it
- press alt-f1

#### Actual result

accerciser shows only window:activate for the mate-panel

#### Expected result

accerciser should show a window:deactivate for mate-terminal, and window:activate for mate-panel

It used to work only by luck before, only because gtk used to deactivate itself during a keyboard grab. But as discussed in https://gitlab.gnome.org/GNOME/gtk/issues/85 that behavior poses accessibility feedback issues, is not coherent, and keyboard grab feedback will not be available in wayland anyway. Thus @ebassi saying in https://gitlab.gnome.org/GNOME/gtk/issues/85#note_264804 that not transferring focus properly is the actual bug.

#### MATE general version

1.20

#### Package version

1.20

#### Linux Distribution

Debian

sthibaul:
The background is in https://gitlab.gnome.org/GNOME/gtk/issues/85 . One of
the conclusions, in https://gitlab.gnome.org/GNOME/gtk/issues/85#note_264804
, is that mate-panel needs to properly transfer focus on alt-F1 keyboard
shortcut.

It used to work only by luck before, only because gtk used to
deactivate itself during a keyboard grab. But as discussed in
https://gitlab.gnome.org/GNOME/gtk/issues/85 that behavior poses
accessibility feedback issues, is not coherent, and keyboard grab
feedback will not be available in wayland anyway. Thus @ebassi saying
in https://gitlab.gnome.org/GNOME/gtk/issues/85#note_264804 that not
transferring focus properly is the actual bug.

This change uses gdk_window_focus to explictly switch to the menu bar after
saving which X Window had the focus, and on menu bar deactivation,
restore focus to that X Window.

Fixes #851

### DIFF
--- a/mate-panel/panel-menu-bar.c
+++ b/mate-panel/panel-menu-bar.c
@@ -28,6 +28,9 @@
 
 #include "panel-menu-bar.h"
 
+#include <X11/Xlib.h>
+#include <gdk/gdkx.h>
+
 #include <string.h>
 #include <glib/gi18n.h>
 
@@ -65,6 +68,8 @@ struct _PanelMenuBarPrivate {
 	GSettings* settings;
 
 	PanelOrientation orientation;
+
+	Window interrupted_window;
 };
 
 enum {
@@ -79,6 +84,15 @@ static gboolean panel_menu_bar_reinit_tooltip(GtkWidget* widget, PanelMenuBar* m
 	g_object_set(menubar->priv->applications_item, "has-tooltip", TRUE, NULL);
 	g_object_set(menubar->priv->places_item, "has-tooltip", TRUE, NULL);
 	g_object_set(menubar->priv->desktop_item, "has-tooltip", TRUE, NULL);
+
+	return FALSE;
+}
+
+static gboolean panel_menu_bar_deactivate (GtkWidget* widget, PanelMenuBar* menubar)
+{
+	GtkWidget *toplevel = gtk_widget_get_toplevel (widget);
+	panel_util_set_current_active_window (toplevel, menubar->priv->interrupted_window);
+	menubar->priv->interrupted_window = None;
 
 	return FALSE;
 }
@@ -108,6 +122,7 @@ static void panel_menu_bar_setup_tooltip(PanelMenuBar* menubar)
 
 	/* Reset tooltip when the menu bar is not used */
 	g_signal_connect(GTK_MENU_SHELL (menubar), "deactivate", G_CALLBACK (panel_menu_bar_reinit_tooltip), menubar);
+	g_signal_connect(GTK_MENU_SHELL (menubar), "deactivate", G_CALLBACK (panel_menu_bar_deactivate), menubar);
 }
 
 static void panel_menu_bar_update_visibility (GSettings* settings, gchar* key, PanelMenuBar* menubar)
@@ -411,10 +426,17 @@ void panel_menu_bar_popup_menu(PanelMenuBar* menubar, guint32 activate_time)
 {
 	GtkMenu* menu;
 	GtkMenuShell* menu_shell;
+	GtkWidget* toplevel;
+	GdkWindow* window;
 
 	g_return_if_fail(PANEL_IS_MENU_BAR(menubar));
 
 	menu = GTK_MENU(menubar->priv->applications_menu);
+
+	toplevel = gtk_widget_get_toplevel (GTK_WIDGET (menubar));
+	menubar->priv->interrupted_window = panel_util_get_current_active_window (toplevel);
+	window = gtk_widget_get_window (toplevel);
+	panel_util_set_current_active_window (toplevel, GDK_WINDOW_XID(window));
 
 	/*
 	 * We need to call _gtk_menu_shell_activate() here as is done in
--- a/mate-panel/panel-menu-button.c
+++ b/mate-panel/panel-menu-button.c
@@ -26,9 +26,12 @@
 
 #include "panel-menu-button.h"
 
+#include <X11/Xlib.h>
+
 #include <string.h>
 #include <glib/gi18n.h>
 #include <gio/gio.h>
+#include <gdk/gdkx.h>
 
 #include <matemenu-tree.h>
 
@@ -91,6 +94,8 @@ struct _PanelMenuButtonPrivate {
 	char                  *menu_path;
 	char                  *custom_icon;
 	char                  *tooltip;
+
+	Window                 interrupted_window;
 
 	MenuPathRoot           path_root;
 	guint                  use_menu_path : 1;
@@ -437,12 +442,23 @@ panel_menu_button_recreate_menu (PanelMenuButton *button)
 	button->priv->menu = NULL;
 }
 
+static gboolean panel_menu_button_menu_deactivate (GtkWidget* widget, PanelMenuButton* button)
+{
+	GtkWidget *toplevel = gtk_widget_get_toplevel(widget);
+	panel_util_set_current_active_window(toplevel, button->priv->interrupted_window);
+	button->priv->interrupted_window = None;
+
+	return FALSE;
+}
+
 void
 panel_menu_button_popup_menu (PanelMenuButton *button,
 			      guint            n_button,
 			      guint32          activate_time)
 {
 	GdkScreen *screen;
+	GtkWidget *toplevel;
+	GdkWindow *window;
 
 	g_return_if_fail (PANEL_IS_MENU_BUTTON (button));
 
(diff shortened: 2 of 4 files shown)


sthibaul:
lukefromdc, le jeu. 30 août 2018 15:44:56 -0700, a ecrit:
> In the (unchanged) menu button, we get NO window:activate or window:deactivate
> events at all,

Are you opening it with alt-F1, or by clicking with the mouse on it?  I
indeed didn't implement it in the mouse click case (but that should be
easy).

lukefromdc:
I was explicitly testing this with alt-F1 while keeping the mouse elsewhere

sthibaul:
lukefromdc, le jeu. 30 août 2018 16:37:08 -0700, a ecrit:
> I was explicitly testing this with alt-F1 while keeping the mouse elsewhere

Ok. I that with the marco WM?

lukefromdc:
I always use compiz unless I am testing Marco. Switching requires moving all windows onto the same workspace and breaks my workflow, so Marco is only tested when I have known reason to do so.

sthibaul:
lukefromdc, le jeu. 30 août 2018 16:58:21 -0700, a ecrit:
> I always use compiz unless I am testing Marco.

Ok. Indeed with compiz it doesn't work, I wonder why while it does with
Marco, I'll have a look.

lukefromdc:
For some reason I ran into a high CPU usage issue (about 15% of an 8-"core" AMD Bulldozer CPU) while letting this run a while, not sure if it related to having tested it with Accerciser though.

sthibaul:
Ok, it seems compiz does tricky timing computation which makes the switch not happen with the gdk operation. I have pushed making it a stronger window activation request, which works fine in my tests.

Re-reading what you wrote, just to be sure: I usually just let accerciser print events for all widgets, and filter only on the type of event. window:activate/deactivate are very seldom anyway so it's readable.

lukefromdc:
Also confirmed to still work with Marco, just in case

lukefromdc:
I looked over the memory management. Again, I am no expert at this, but it looks like all the variables called as dynamically allocated are then set to point to existing objects (windows etc) that should not be freed.

Example: 
`GtkWidget *toplevel = gtk_widget_get_toplevel (GTK_WIDGET (menubar));`
sets toplevel as a pointer to the existing toplevel widget for the menubar (its window?) so that gets freed (and can be freed only) when the menubar itself is removed or the panel restarted.

muktupavels:
@sthibaul Should not this be done by window manager?

Anyway I see that you edited menu bar applet... What about main menu applet? It also can be activated with `Alt + F1`:
https://github.com/mate-desktop/mate-panel/blob/master/mate-panel/panel-action-protocol.c#L52

lukefromdc:
What I was saying is that memory management looks good, with only new pointers to existing data added if I am correct. What we need to avoid is calling a new local variable on the heap that does NOT point to existing (already managed) contents in RAM and then failing to free it. When such a local function closes, the pointers are removed but the contents in RAM remain and cannot be reached, "leaking" that RAM by putting it off-limits to everything until the entire program is closed. I don't see any such problem here. I am trying to look over as many PR's for that sort of issue as I can, and am learning in the process. We have way too many memory leaks in the existing code so at least we need to be careful not to add new ones. This PR looks good in that regard. 

Also, I agree that the menu button (which does not seem to interact with the activate/deactivate code at all) needs to be dealt with

sthibaul:
lukefromdc, le ven. 31 août 2018 17:31:15 -0700, a ecrit:
> I don't see any such problem here. I am trying to look over
> as many PR's for that sort of issue as I can, and am learning in the process.

Ah, OK :)

Thanks for that work, it's tedious but so useful :)

sthibaul:
@muktupavels : the change here is precisely to tell the window manager what it is supposed to do.

Indeed, the main menu applet probably needs the same.

muktupavels:
@sthibaul Yes, but not sure if application should save active window, then re-focus it. It sounds more like job for window manager. Of course panel is not same random regular application, but still...

I am worried that this could introduce other problems? For example you have focused app1, you move focus to panel and save app1 as active/previously focused app. You launch app2 and it opens faster then wm gets message to re-focus app1. Launching app2 you would expect that it has focus, but panel/wm will move it to app1.

Not sure if that can happen, I have not looked what code does. Simply mentioning in case you have not thought about this already.

raveit65:
Does it make sense to add another commit for menu-applet here?
Instead of reviewing the same thing in another PR.

lukefromdc:
Makes sense to me

raveit65:
@sthibaul 
Can you do the same for menu-applet?
Than we can finish your PR.

sthibaul:
Yes, I'll work on it, I have just been busy with the new school year.

sthibaul:
I have now added the support for the menu button too.

I have checked for the race, the deactivate signal gets called before the execution of the launcher, so the server timestamp for the switch message will be before the creation of the window by the newly started application.

lukefromdc:
OK, I have just confirmed that this now works for both the menubar and the menu button

lukefromdc:
Rebased and merged with https://github.com/mate-desktop/mate-panel/commit/f0f4c5e1217eefd46edf9f98633fb32967c67461

lukefromdc:
An attempt to cherrypick failed, it will require a manual backport

alexarnaud:
@lukefromdc Do you want @sthibaul  do the manual backport or do you want to do it yourself?

Best regards,
Alex.

alexarnaud:
@raveit65 @lukefromdc OK @cwendling will do a pull request for that and I'll test it to confirm that it works with Orca. I'm also impacted by the issue as a Orca user.

Best regards,
Alex.

lukefromdc:
I'm rather maxed out right now, better for me if @sthibaul or another dev can deal with the backport right now.

monsta:
I have a problem after this change. I use mintMenu in Mint 19. There's no menu-button or menubar on my panel. So, Alt-F1 doesn't show the menu window anymore, and there are some warnings in `~/.xsession-errors`:

```
(mate-panel:1698): Gtk-WARNING **: 16:28:15.118: no trigger event for menu popup

(mate-panel:1698): Gtk-CRITICAL **: 16:28:15.118: gtk_menu_popup_at_rect: assertion 'GDK_IS_WINDOW (rect_window)' failed
```

sthibaul:
This looks very probable to me, yes: 17ac8aa passes NULL as trigger_event, and thus rect_window remains NULL, and then GDK_IS_WINDOW(rect_window) can't work.

raveit65:
but https://github.com/mate-desktop/mate-panel/commit/17ac8aab4d139cb917619a3a981300f51b750c17 isn't gtk_menu_popup_at_rect function from xsession-errors.
gtk_menu_popup_at_pointer and gtk_menu_popup_at_rect are different functions, see docs

sthibaul:
I prefer to read the source :)
gtk_menu_popup_at_pointer just calls gtk_menu_popup_at_rect with appropriate parameters.

lukefromdc:
Is there any chance this came from 
https://github.com/mate-desktop/mate-panel/commit/17ac8aab4d139cb917619a3a981300f51b750c17
which replaced deprecated  replace deprecated gtk_menu_popup  or is this confirmed to come from this PR? 

Also, I looked at the change files, and didn't see any code removed from this PR that would trigger the menu. In fact I did not know that it was ever possible to show the menu with alt-F1 if no menu applet is installed. Is there any chance that the recent port of mintmenu to be compatable with the latest changes in mate-panel is the culprit here?

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/885 fixes the regression caused by https://github.com/mate-desktop/mate-panel/commit/17ac8aab4d139cb917619a3a981300f51b750c17 , please test

raveit65:
Did any one with mintmenu installed really test that 17ac8aa causes the issue?
This can be done with reverting the commit in a local build.

monsta:
Yeah, reverting that commit fixes the problem for me, thanks.
I'll check #885 later.

monsta:
> Is there any chance that the recent port of mintmenu to be compatable with the latest changes in mate-panel is the culprit here?

No, I don't think it would intercept Alt-F1 or something. It would be the same if I used any other 3rd party applet, e.g. Brisk Menu (from Solus project).

PR #864, opened 2018-09-04
Migrate dbus-glib to gdbus.

yetist:
Please test and review it, thank you.


raveit65:
Indents can be improved in a lot of places or not?
Looks like we have a problem with using different 'tab sizes` in editors.
Personal, i don't add new tabs in code and i replace tabs with spaces in every part where i do code changes, to avoid the tab size problem.
And why last commit?
In firefox browser old  indents are looking better than new ones.
https://github.com/mate-desktop/mate-panel/pull/864/commits/98dff560a8f2f503834dfaccd29297619b05deb6
In pluma it looks complete different.
pluma-example
Please do not replace a mixed tab+spaces line with tabs.
Spaces is always the better choice.

Ok, i start testing now.

rcaridade145:
@raveit65 what i've noticed in pluma is that sometimes i have to press tab twice for it to visually display a tab. However if  i save and open with nano or vi i'll have 2 tabs.

raveit65:
I see von different how often i have to press the tab key with using pluma or vi.
Problem is that the tab size can be different in editors.
4 spaces =1 tab,  8 spaces =1 tab, 16 spaces =1 tab, 1024 spaces =1 tab........
For this reason i try to avoid using tabs.

lukefromdc:
Any progress on this? it is being held up over the tab/space issue. Should I fix that myself and open a new PR?

raveit65:
Beside from tab/space issue it looks ready to go for me.

raveit65:
yeah, go ahead.

lukefromdc:
OK, I will get on that tomorrow or the next day

monsta:
First and second commits can be squashed...

The code in `panel-session-manager.c` seems to work, as I still can invoke logout and shutdown dialogs from the menu. Not sure how to test the changes in `panel-dbus-service.c`, in particular, the removal of reaction to `"NameOwnerChanged"`.

mate-panel/libpanel-util/panel-dbus-service.c
```diff
@@ -54,6 +51,12 @@ static void panel_dbus_service_get_property (GObject    *object,
 					     guint       prop_id,
 					     GValue     *value,
 					     GParamSpec *pspec);
+void
+panel_dbus_service_on_signal (GDBusProxy *proxy,
```
monsta: This is a declaration of a function which doesn't exist. It does nothing as is. 

@yetist: did you plan to implement that function as a replacement for the removed `panel_dbus_service_on_name_owner_changed`?

muktupavels:
Seems that `PanelDBusService` is used only by `PanelSessionManager`... You won't get `NameOwnerChanged` for it unless you know how to restart gnome-session without terminating your session. Also `GDBusProxy` handles that signal for you - https://gitlab.gnome.org/GNOME/glib/blob/master/gio/gdbusproxy.c#L1772.

raveit65:
@yetist 
Can we finish the PR?
I want to avoid that your PR doesn't clean apply to master after some huge other open PRs, eg. wayland port.

lukefromdc:
We really need to get this done soon, with all the other PRs coming through anything like this that is being held gets farther behind.

yetist:
@raveit65 
Now add "NameOwnerChanged" signal process, and some indent, please review again, thank you.

muktupavels:
Why "NameOwnerChanged" is needed? In case you get that signal won't you subscribe again to that signal?

yetist:
> Why "NameOwnerChanged" is needed? In case you get that signal won't you subscribe again to that signal?

I'm not sure if there's a possibility of using mate-panel with other sessions, or switching / restarting / repalacing that session during use, and if that's the case, it's necessary to monitor the NemeOwnerChanged signal.

Mate-panel monitors the NameOwnerChanged signal on the session bus of org.freedesktop.DBUS, so when the change occurs, the connection does not change, so we don't need to subscribe the signal again.

muktupavels:
On Thu, Nov 1, 2018 at 2:55 AM Wu Xiaotian <notifications@github.com> wrote:

> Why "NameOwnerChanged" is needed? In case you get that signal won't you
> subscribe again to that signal?
>
> I'm not sure if there's a possibility of using mate-panel with other
> sessions, or switching / restarting / repalacing that session during use,
> and if that's the case, it's necessary to monitor the NemeOwnerChanged
> signal.
>
Did you miss that I told you that `GDBusProxy` proxy handles that signal
for you?

> Mate-panel monitors the NameOwnerChanged signal on the session bus of
> org.freedesktop.DBUS, so when the change occurs, the connection does not
> change, so we don't need to subscribe the signal again.
>
Don't need, but that will happen, no? When you get that signal you will
call `panel_dbus_service_ensure_connection`, right? It will re-create proxy
and will subscribe, again, no? Or am I reading code wrong?

yetist:
> On Thu, Nov 1, 2018 at 2:55 AM Wu Xiaotian ***@***.***> wrote: Why "NameOwnerChanged" is needed? In case you get that signal won't you subscribe again to that signal? I'm not sure if there's a possibility of using mate-panel with other sessions, or switching / restarting / repalacing that session during use, and if that's the case, it's necessary to monitor the NemeOwnerChanged signal.
> Did you miss that I told you that `GDBusProxy` proxy handles that signal for you?
> Mate-panel monitors the NameOwnerChanged signal on the session bus of org.freedesktop.DBUS, so when the change occurs, the connection does not change, so we don't need to subscribe the signal again.
> Don't need, but that will happen, no? When you get that signal you will call `panel_dbus_service_ensure_connection`, right? It will re-create proxy and will subscribe, again, no? Or am I reading code wrong?

Yes, subscribe signal code inside re-create proxy block is wrong, so I need move that code to other place or just drop that code about "NameOwenerchanged" signal?

muktupavels:
From what I see I would say that handling `NameOwnerChanged` is not needed because `GDBusProxy` does that already for you, at least it looks like it does. Maybe best thing is to write small test case to verify that...

raveit65:
I think there are more places where intends are not correct or a new code line doesn#t start under a bracket of the line before.
https://patch-diff.githubusercontent.com/raw/mate-desktop/mate-panel/pull/864.diff
But i am really tired to say people who knows much more of coding like me, they should work more cleaner.

raveit65:
Btw. can you squash some commits and rebase the PR.
No need that intends fixes are a new commit.
You can do that with `git rebase -i <last-commit-id-before-PR>`

muktupavels:
Well it looks like `dbus_connection` is also not needed!? You can use `g_dbus_proxy_new_for_bus_sync`, like you did in set-timezone.c.

yetist:
> Well it looks like `dbus_connection` is also not needed!? You can use `g_dbus_proxy_new_for_bus_sync`, like you did in set-timezone.c.

Yes, I think so, just get proxy and then call dbus method in panel-session-manager.c. If do that, panel-dbus-service.{h,c} should removed.

yetist:
@muktupavels @raveit65 
I droped panel-dbus-service.{h,c} files, get proxy and call dbus method in panel-session-manager.c now, is it OK?

muktupavels:
Speaking about `PanelSessionManager`... Put `GDBusProxy` in `PanelSessionManager` struct! Don't use static here, not needed.

Also it looks like proxy could be `NULL`, but you use it in `g_dbus_proxy_call_sync` assuming it will never be `NULL`.

yetist:
What about to write code in panel-action-button.c, and remove panel-session-manager.c file?

muktupavels:
Is it only place where it is used? Anyway check that with MATE devs... I would probably keep it.

yetist:
> Speaking about `PanelSessionManager`... Put `GDBusProxy` in `PanelSessionManager` struct! Don't use static here, not needed.
> 
> Also it looks like proxy could be `NULL`, but you use it in `g_dbus_proxy_call_sync` assuming it will never be `NULL`.

All done, waiting for review.

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -42,136 +49,102 @@ panel_session_manager_class_init (PanelSessionManagerClass *klass)
...
+				       NULL,
+				       &error);
+	if (_ret == NULL) {
 		g_warning ("Could not ask session manager if shut down is available: %s",
 			   error->message);
 		g_error_free (error);
-
 		return FALSE;
+	} else {
+		g_variant_get(_ret, "(b)", &is_shutdown_available);
```
muktupavels: missing space before `(`...

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -42,136 +49,102 @@ panel_session_manager_class_init (PanelSessionManagerClass *klass)
...
 }
 
 void
 panel_session_manager_request_logout (PanelSessionManager           *manager,
 				      PanelSessionManagerLogoutType  mode)
 {
-	GError *error;
-	DBusGProxy *proxy;
+	GError *error = NULL;
+	GVariant *_ret;
```
muktupavels: Why underscore is needed?

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -42,136 +49,102 @@ panel_session_manager_class_init (PanelSessionManagerClass *klass)
...
-		return;
-	}
-
-	proxy = panel_dbus_service_get_proxy (PANEL_DBUS_SERVICE (manager));
-
-	if (!dbus_g_proxy_call (proxy, "Logout", &error,
-				G_TYPE_UINT, mode, G_TYPE_INVALID,
-				G_TYPE_INVALID) &&
-	    error != NULL) {
+	g_return_if_fail (manager->proxy != NULL);
```
muktupavels: `g_return_if_fail` is meant for programmer errors:
https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-if-fail

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -42,136 +49,102 @@ panel_session_manager_class_init (PanelSessionManagerClass *klass)
...
-					   "org.gnome.SessionManager");
-
-	error = NULL;
-	if (!panel_dbus_service_ensure_connection (PANEL_DBUS_SERVICE (obj),
-						   &error)) {
-		g_message ("Could not connect to session manager: %s",
-			   error->message);
+	GError *error = NULL;
+
+	manager->proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
```
muktupavels: Would not be better to move this to `ensure_proxy` or something like that? And use it in all functions that use proxy? Not sure if that can happen, but if proxy creating fails then you will never try again to create proxy. Old code would try to create proxy if it is `NULL`...

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -22,18 +22,25 @@
...
-						   GObjectConstructParam *construct_properties);
+struct _PanelSessionManager {
+	GObject     parent;
+	GDBusProxy *proxy;
+};
+
+G_DEFINE_TYPE (PanelSessionManager, panel_session_manager, G_TYPE_OBJECT);
+
+static void
+panel_session_manager_finalize (GObject *object)
```
muktupavels: You need to chain-up in finalize... `G_OBJECT_CLASS (...)->finalize (...);`

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -22,18 +22,25 @@
...
 
-static GObject *panel_session_manager_constructor (GType                  type,
-						   guint                  n_construct_properties,
-						   GObjectConstructParam *construct_properties);
+struct _PanelSessionManager {
+	GObject     parent;
+	GDBusProxy *proxy;
+};
+
+G_DEFINE_TYPE (PanelSessionManager, panel_session_manager, G_TYPE_OBJECT);
```
muktupavels: Trailing semicolon is not needed.

applets/clock/set-timezone.c
```diff
@@ -171,75 +178,63 @@ free_data (gpointer d)
...
+			data->callback (data->data, NULL);
 	}
 }
 
 static void
 set_time_async (SetTimeCallbackData *data)
 {
-        DBusGConnection *bus;
-        DBusGProxy      *proxy;
+        GDBusProxy      *proxy;
```
muktupavels: Unneeded space.

applets/clock/set-timezone.c
```diff
@@ -171,75 +178,63 @@ free_data (gpointer d)
...
-						      set_time_notify,
-						      data, free_data,
-						      INT_MAX,
-						      /* parameters: */
-						      G_TYPE_INT64, data->time,
-						      G_TYPE_INVALID,
-						      /* return values: */
-						      G_TYPE_INVALID);
+		g_dbus_proxy_call (proxy,
+				"SetTime",
```
muktupavels: Does not align correctly...

applets/clock/set-timezone.c
```diff
@@ -171,75 +178,63 @@ free_data (gpointer d)
...
 				data->callback (data->data, error);
 			else
 				g_error_free (error);
-		}		
+		} else {
+			if (data->callback)
+				data->callback (data->data, NULL);
+		}
+	} else {
+		g_variant_get (variant, "()");
```
muktupavels: What this is supposed to do?

applets/clock/set-timezone.c
```diff
@@ -171,75 +178,63 @@ free_data (gpointer d)
 }
 
 static void
-set_time_notify (DBusGProxy     *proxy,
-		 DBusGProxyCall *call,
-		 void           *user_data)
+set_time_notify (GObject *source_object,
+		GAsyncResult *res,
```
muktupavels: Does not align correctly.

applets/clock/set-timezone.c
```diff
@@ -171,75 +178,63 @@ free_data (gpointer d)
...
-			/* these errors happen because dbus doesn't
-			 * use monotonic clocks
-			 */	
-			g_warning ("ignoring no-reply error when setting time");
-			g_error_free (error);
-			if (data->callback)
-				data->callback (data->data, NULL);
-		}
-		else {
+	GError     *error = NULL;
```
muktupavels: Align all variables or don't... Do not mix.

applets/clock/set-timezone.c
```diff
@@ -171,75 +178,63 @@ free_data (gpointer d)
...
+	GDBusProxy *proxy;
+	GVariant   *variant;
+
+	proxy = get_bus_proxy ();
+	variant = g_dbus_proxy_call_finish (proxy, res, &error);
+	if (variant == NULL) {
+		if (error != NULL) {
 			if (data->callback)
 				data->callback (data->data, error);
```
muktupavels: Are callback supposed to free error? Or we have memory leak here?

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,81 @@
...
-					   "/",
-					   "org.mate.SettingsDaemon.DateTimeMechanism");
-
-	dbus_g_proxy_begin_call_with_timeout (proxy,
-					      action,
-					      notify_can_do,
-					      callback, NULL,
-					      INT_MAX,
-					      G_TYPE_INVALID);
+	GDBusProxy      *proxy;
```
muktupavels: Unneeded space.

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,81 @@
...
 #define CACHE_VALIDITY_SEC 2
 
 typedef  void (*CanDoFunc) (gint value);
 
 static void
-notify_can_do (DBusGProxy     *proxy,
-	       DBusGProxyCall *call,
-	       void           *user_data)
+notify_can_do (GObject *source_object,
+		GAsyncResult *res,
```
muktupavels: Does not align correctly.

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,81 @@
...
-					      INT_MAX,
-					      G_TYPE_INVALID);
+	GDBusProxy      *proxy;
+
+	proxy = get_bus_proxy ();
+	if (proxy == NULL)
+		return;
+
+	g_dbus_proxy_call (proxy,
+			action,
```
muktupavels: Not aligned.

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,81 @@
...
+		proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
+				G_DBUS_PROXY_FLAGS_NONE,
+				NULL,
+				DATETIME_DBUS_NAME,
+				DATETIME_DBUS_PATH,
+				DATETIME_DBUS_NAME,
+				NULL,
+				&error);
+		if (proxy == NULL)
+		{
```
muktupavels: `{` should not be in new line.

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,81 @@
...
-	if (bus == NULL) {
-        	error = NULL;
-        	bus = dbus_g_bus_get (DBUS_BUS_SYSTEM, &error);
-        	if (bus == NULL) {
-                	g_warning ("Couldn't connect to system bus: %s", 
-				   error->message);
-                	g_error_free (error);
+	GError            *error = NULL;
+	static GDBusProxy *proxy = NULL;
+	if ( proxy == NULL) {
```
muktupavels: extra unneeded space before `proxy`.

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,81 @@
...
+		proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SYSTEM,
+				G_DBUS_PROXY_FLAGS_NONE,
```
muktupavels: Alignment.

applets/clock/set-timezone.c
muktupavels: My personal opinion is that tabs should not be used, only spaces. To keep existing style use combination of tab/tabs and spaces. For example:
```
[tab]my_function (param1,
[tab][spaaaaaaces]param2);
```

mate-panel/libpanel-util/panel-session-manager.c
yetist: ok, yes, I forgot it.

applets/clock/set-timezone.c
yetist: Good. It looks like a better way.
Is there any possibility of using tools?

applets/clock/set-timezone.c
```diff
@@ -28,74 +28,80 @@
...
 
 typedef  void (*CanDoFunc) (gint value);
 
 static void
-notify_can_do (DBusGProxy     *proxy,
-	       DBusGProxyCall *call,
-	       void           *user_data)
+notify_can_do (GObject *source_object,
+	       GAsyncResult *res,
+	       gpointer user_data)
```
muktupavels: Still not aligned.

```
notify_can_do (GObject      *source_object,
               GAsyncResult *res,
               gpointer      user_data)
```

applets/clock/set-timezone.c
```diff
@@ -171,75 +177,61 @@ free_data (gpointer d)
 }
 
 static void
-set_time_notify (DBusGProxy     *proxy,
-		 DBusGProxyCall *call,
-		 void           *user_data)
+set_time_notify (GObject *source_object,
+		 GAsyncResult *res,
+		 gpointer user_data)
```
muktupavels: Same thing...

applets/clock/set-timezone.c
muktupavels: Any comment on this?

applets/clock/set-timezone.c
```diff
@@ -171,75 +177,61 @@ free_data (gpointer d)
...
+			data->callback (data->data, NULL);
 	}
 }
 
 static void
 set_time_async (SetTimeCallbackData *data)
 {
-        DBusGConnection *bus;
-        DBusGProxy      *proxy;
+        GDBusProxy *proxy;
```
muktupavels: Indentation...

mate-panel/libpanel-util/panel-session-manager.c
```diff
@@ -42,136 +51,102 @@ panel_session_manager_class_init (PanelSessionManagerClass *klass)
...
+	manager->proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
+							G_DBUS_PROXY_FLAGS_NONE,
+							NULL,
+							"org.gnome.SessionManager",
+							"/org/gnome/SessionManager",
+							"org.gnome.SessionManager",
+							NULL,
+							&error);
+	if (manager->proxy == NULL)
+	{
```
muktupavels: Should not be in new line.

mate-panel/libpanel-util/panel-session-manager.c
muktupavels: ?

yetist:
@raveit65 
I'm sorry, I didn't handle it in time to keep you waiting.

Mixed space and tab can cause too many alignment problems. I use tab by default, but it looks like Mate is using space?

raveit65:
You're using tab size=8, right?
I am using tab size=4 in pluma, which is default.
The result of this mismatch is that the resulting patch of your PR looks weird in my pluma.
Try it out for yourself and change tab size, maybe than you understand the problem with mixing spaces and tabs in one line.

raveit65:
I squashed all in one commit, which is better for a meaningful git history.
merged
https://github.com/mate-desktop/mate-panel/commit/a34f531da4f03eecfc38414295ebe70742dfad71
But next time i won't merge PRs any more with tab vs. spaces problems.

muktupavels:
@raveit65 did you saw more indentation problems? Last changes looks more or less good to me. There was two places where function parameters could be aligned. But there are one change that I don't like - he removed `g_return_val_if_fail`, but should have replaced it with early return probably with warning that proxy is not available or something like that.

I probably was not clear enough, my fault. Right now if proxy is NULL it will crash, it could crash also with `g_return_val_if_fail`:

> If G_DISABLE_CHECKS is defined then the check is not performed. You should therefore not depend on any side effects of expr .

raveit65:
wtf.....
@muktupavels 
Indentations was OK if you use tab size = 8 in editor.

@yetist 
Can you take a look again to fix last concern?

muktupavels:
I was look at diff here in github... Well if file is already using tabs you typically want to stick to same style. Looking again at diff I see he wrongly aligned arguments - for arguments you want use tabs (same time you used for first line) and everything else should be spaces.

raveit65:
Resulting commit looks OK for me if open in browser
https://github.com/mate-desktop/mate-panel/commit/a34f531da4f03eecfc38414295ebe70742dfad71
Or if you use tab size =8 in editor.
Any way, a bigger problem is the issue which you mentioned.
How can i fix that?

muktupavels:
In simplest form:
`if (!manager->proxy) return;`

Or if you want to print debug or warning message:
```
if (!manager->proxy) {
    g_debug / g_warning ("...");
    return;
}
```

lukefromdc:
Just how bad are the tab/indent issues here? If we are holding up this whole PR over it, we are also holding up the Wayland port and risking further diff growth as other things get worked on. I propose we get this merged and open a new PR to work on the indent issues

lukefromdc:
OK, I see this is merged and the remaining indent issues can be dealtt with separately. That needs to be fixed, but the way is now open for the wayland work

yetist:
> In simplest form:
> `if (!manager->proxy) return;`
> 
> Or if you want to print debug or warning message:
> 
> ```
> if (!manager->proxy) {
>     g_debug / g_warning ("...");
>     return;
> }
> ```

I tested on my system with follow code:
```
#include <gio/gio.h>
int main(int argc, char** argv)
{
    GDBusProxy *proxy;
    GError *error = NULL;
    proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
            G_DBUS_PROXY_FLAGS_NONE,
            NULL,
            "org.gnome.SessionManager1",    // Here use wrong name, wrong path
            "/org/gnome/SessionManager2",   // and wrong interface for test,
            "org.gnome.SessionManager3",    // like mate-session not running.
            NULL,
            &error);
    if (proxy == NULL) {
        g_warning ("Unable to connect session manager daemon: %s\n", error->message);
        g_error_free (error);
    } else {
        g_print ("proxy is not null\n");
    }
    return 0;
}
```
But I get output `proxy is not null` always,  so I think no need to check `manager->proxy == NULL` when call dbus method,  maybe I was wrong.

yetist:
More test:
```
$ cat test.c 
#include <gio/gio.h>
int main(int argc, char** argv)
{
    GDBusProxy *proxy;
    GVariant *ret;
    GError *error = NULL;
    gboolean is_shutdown_available;

    proxy = g_dbus_proxy_new_for_bus_sync (G_BUS_TYPE_SESSION,
            G_DBUS_PROXY_FLAGS_NONE,
            NULL,
            "org.gnome.SessionManager1",    // Here use wrong name, wrong path
            "/org/gnome/SessionManager2",   // and wrong interface, to test
            "org.gnome.SessionManager3",    // mate-session not running.
            NULL,
            &error);
    if (proxy == NULL) {
        g_warning ("Unable to connect session manager daemon: %s\n", error->message);
        g_error_free (error);
    } else {
        g_print ("proxy is not null\n");
    }

    g_assert(error == NULL);

    ret = g_dbus_proxy_call_sync (proxy, "CanShutdown",
            g_variant_new ("()"),
            G_DBUS_CALL_FLAGS_NONE,
            -1,
            NULL,
            &error);
    if (ret == NULL) {
        g_print ("ERROR: %s\n", error->message);
        g_error_free (error);
    } else {
        g_variant_get (ret, "(b)", &is_shutdown_available);
        g_variant_unref (ret);
    }

    g_print("shutdown available is %d\n", is_shutdown_available);

    return 0;
}
$ gcc -o test test.c `pkg-config --cflags --libs gio-2.0`
$ ./test 
proxy is not null
ERROR: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.SessionManager1 was not provided by any .service files
shutdown available is 0
```
If `proxy` is not null,  atfer call `g_dbus_proxy_call_sync` , will print warning message, if proxy is NULL,  call `g_dbus_proxy_call_sync` will crash.
But the question is that after call `g_dbus_proxy_new_for_bus_sync`, it always return no null, is that ok?  At lease on my system it looks like this, maybe my system has error, or I was wrong.
Does anyone have time to test the above code?

muktupavels:
`g_dbus_proxy_new_for_bus_sync` is documented that it can return `NULL`... So I would say that you should handle that! Anyway, best thing is to read source code not guess when it can be `NULL`.

raveit65:
i got same results with test.c
```
[rave@mother Schreibtisch]$ ./test 
proxy is not null
ERROR: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.SessionManager1 was not provided by any .service files
shutdown available is 0
```

raveit65:
Same result in multiuser-target without running X (runlevel 3) after a `killall dbus-daemon`
```
[rave@mother Schreibtisch]$ ./test 
proxy is not null
ERROR: GDBus.Error:org.freedesktop.DBus.Error.ServiceUnknown: The name org.gnome.SessionManager1 was not provided by any .service files
shutdown available is 0
```

PR #885, opened 2018-10-29
Ensure alt-F1 can show the menu even if no menu applet on panel

lukefromdc:
Fix regression introduced by https://github.com/mate-desktop/mate-panel/commit/17ac8aab4d139cb917619a3a981300f51b750c17 and ensure that menu shown follows same theme as menu shown from panel menu button or panel menu bar.

### DIFF
--- a/mate-panel/panel-action-protocol.c
+++ b/mate-panel/panel-action-protocol.c
@@ -58,6 +58,8 @@ panel_action_protocol_main_menu (GdkScreen *screen,
 	GdkVisual *visual;
 	GtkWidget *toplevel;
 	GtkStyleContext *context;
+	GdkSeat *seat;
+	GdkDevice *device;
 
 	info = mate_panel_applet_get_by_type (PANEL_OBJECT_MENU_BAR, screen);
 	if (info) {
@@ -88,6 +90,11 @@ panel_action_protocol_main_menu (GdkScreen *screen,
 	context = gtk_widget_get_style_context (GTK_WIDGET(toplevel));
 	gtk_style_context_add_class(context,"gnome-panel-menu-bar");
 	gtk_style_context_add_class(context,"mate-panel-menu-bar");
+
+	seat = gdk_display_get_default_seat (gdk_display_get_default());
+	device = gdk_seat_get_pointer (seat);
+	gdk_event_set_device (event, device);
+
 	gtk_menu_popup_at_pointer (GTK_MENU (menu),event);
 }
 
--- a/mate-panel/panel-action-protocol.c
+++ b/mate-panel/panel-action-protocol.c
@@ -50,11 +50,14 @@ static Atom atom_mate_panel_action_kill_dialog = None;
 
 static void
 panel_action_protocol_main_menu (GdkScreen *screen,
-				 guint32    activate_time)
+				 guint32    activate_time, GdkEvent  *event)
 {
 	PanelWidget *panel_widget;
 	GtkWidget   *menu;
 	AppletInfo  *info;
+	GdkVisual *visual;
+	GtkWidget *toplevel;
+	GtkStyleContext *context;
 
 	info = mate_panel_applet_get_by_type (PANEL_OBJECT_MENU_BAR, screen);
 	if (info) {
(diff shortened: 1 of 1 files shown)


lukefromdc:
Behavior in test here is as follows: same as before if panel menu button or panel menu bar is on panel. If neither is on the panel, show the main menu at the pointer and with the same theme it would have from the applet when Alt-F1 (or user's alternate keybinding) is pressed.

Memory management: The three new variables are from my theme support boilerplate, and they are pointers to existing data that should not be freed (unless I am wrong on one or more of them). Also the event is passed in from the calling function, I am assuming that should not be freed from the called function.

Style: I separated declaring them from initializing them to avoid adding more code incompatable with very old C standards that don't like mixed declarations and code. The rest of the file at least in that area uses Tabs so I followed that pattern.

monsta:
Also I never noticed any theme issues without this PR...

raveit65:
I think we need something like that when we change `NULL` to a GdkEvent.
https://github.com/mate-desktop/mate-terminal/commit/9081ba937a78666df689f9bf6cdf81fa5225fbab
I recall that this fixed the same warning about GdkSeat.

Ihmo, panel menus should simply follow normal css selector menu or menubar.
I never saw that this was a problem with new css language from gtk+-3.22.

lukefromdc:
The code to support a custom theme in panel menus was merged all the way back in the 1.11 dev cycle. If it is removed from the existing panel and applet code I will have to patch it back into every build I install (every time), and my theme will depend on those patches to work.

That is because in my GTK theme all panel menus have a completely different styling than the main GTK theme, and I never found a way to do that without changing the menu code itself to support that.  I was hoping others would use that, but if others insist on removing it I can do the extra work to patch it back in locally on my machines (which I won't remove it from). I already have to do this with the network-manager applet. 

I am not particularily worried about this particular use of it (as I never use the panel that way) but wanted on seeing that this feature exists to keep it consistant with other panel menus, which I have put a lot of effort into keeping in sync. BTW,  menus from the panel menubar applet follow this even in gnome-panel (assuming .gnome-panel-menu-bar menu is set) as those menus are children of the menubar it seems.

I would however be completely open to refactoring it for style, to clean up the selectors used. Also there is the option of stripping out hardcoded transparency support for compiz as compiz reloaded has merged the GtkFrameExtents work, though the compiz 0.9 branch has refused to do so.

muktupavels:
Well, that sounds like GTK bug or you are abusing API...

https://developer.gnome.org/gdk3/stable/gdk3-Event-Structures.html
There are many event types... I would say/guess that API is supposed to be used with GdkEventKey, GdkEventButton and maybe few other. For example I doubt that it supposed to be used with GdkEventProperty or GdkEventWindowState events...

I would suggest to open bug for GTK about this. If API should work with all events then GTK should set GdkSeat for all events, otherwise it should have proper warning/message that invalid event is used with API and that should be mentioned in documentation.

I would say that linked commit is workaround not fix...

lukefromdc:
OK, just tested the workaround and like in the terminal it works. If GTK ever fixes this in both 3.22 and 3.24 we can revisit this

muktupavels:
If you will not report bug it will never be fixed... Ideally you would prepare fix and create merge request for GTK. Or maybe just ask on irc if that API should work with all events.

lukefromdc:
I do not have a functional account on GNOME, and thus can submit neither bug reports nor PR's there. 

I have a LOT of issues with account formation on some sites, as I use extremely locked-down browsers disabling trackers, as much fingerprinting as I can, referrers, and many other things. A previous GNOME account no longer works and I do not have time to debug it. Someone else here will have to post that report.

lukefromdc:
@monsta , the only time you would notice a theme issue without this is if you used a theme that sets a different style for panel menus then the rest of the GTK menus, as my UbuntuStudio_Legacy theme does. One part of that overall series of theme support commits (calendar window transparency) is also used to allow the transparent calendar windows in the Submarine themes. 

 I'm a little disappointed that themers have never used the rest of my 2015 theme support code (merged in 1.11). I now know enough coding though that I may be able to revisit my old idea of making panel menus themable from the panel-prefs dialog like the panel background is, rather like in the mintmenu applet or cairo-dock. To keep the code light, that whole block could exit from the top in each instance if the system theme is selected.

lukefromdc:
Had to fix commit message (gtkseat>gdkseat)

lukefromdc:
Keeping the branch for now due to the branch force-push prior to merge in case anyone finds a further need for it.

PR #890, opened 2018-11-15
Check if manager->proxy is NULL

yetist:
Please see #864 code review.


raveit65:
@lukefromdc 
I will make a new release to have a stable version before wayland port.

PR #891, opened 2018-12-02
panel-gtk: Fix memory leaks

sc0w:
'g_strconcat' needs to be freed


muktupavels:
https://github.com/mate-desktop/mate-panel/commit/86701517e7d7cb3d2c08a40d76af97308f18902c#r29476695

lukefromdc:
As I am not quite as familiar with some of the GTK functions here, asking for another review, hoping for another set of eyes before merging

PR #894, opened 2018-12-03
Remove legacy X11 package detection

wmww:
This was a suggested change to my Wayland PR, but since I just got it from prexisting code, I thought it would be best to propose it separately.

Original comment by @muktupavels:
> Well it is 2018... I would just require pkg-config file and would remove old-style detection...
https://gitlab.freedesktop.org/xorg/lib/libx11
https://gitlab.freedesktop.org/xorg/lib/libxau
This tells that x11.pc and xau.pc is available as minimum for 9 years.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -169,26 +169,7 @@ gl_CHECK_TYPE_STRUCT_DIRENT_D_TYPE
 
 dnl X development libraries check
 
-PKG_CHECK_MODULES(X, x11 xau, :, [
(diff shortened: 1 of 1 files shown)


muktupavels:
@lukefromdc I suggest to learn what `PKG_CHECK_MODULES` does:
https://autotools.io/pkgconfig/pkg_check_modules.html

lukefromdc:
I really do NOT know much about the build system and am out of time at the moment, so I suggest others look at this

PR #895, opened 2018-12-03
Wayland protocol files

wmww:
This is part of my Wayland port, that can be easily merged without effecting existing functionality (build files aren't touched, so new files are ignored).


yetist:
It seems that these files are generated by the `wayland-scanner` command. Is it possible to provide only XML files and automatically generate these files at compile time?

wmww:
Yes it is, but why?

muktupavels:
XDG Shell is from `wayland-protocols`, right? If you are generating files at build time you would automatically get any updates to protocol file. For example fixing typos and/or other documentation changes.

yetist:
The new files in this PR can be generated from these xml files:
https://github.com/wayland-project/wayland-protocols/blob/master/stable/xdg-shell/xdg-shell.xml
https://github.com/swaywm/wlr-protocols/blob/master/unstable/wlr-layer-shell-unstable-v1.xml

muktupavels:
I know, but that is manual process... Where is problem to depend on `wayland-protocols`? Something like this:
```
PKG_CHECK_MODULES([PROTOCOLS], [wayland-protocols], [
  protocolsdir=`$PKG_CONFIG --variable=pkgdatadir wayland-protocols`
  AC_SUBST([PROTOCOLSDIR], [$protocolsdir])
])
```

Then in makefile you create list of files / protocols you want to generate and add rule to do that. And if in future you will want to add/use other protocol you just add it to list.

Seems that `wlr-protocols` also have pkg-config file...

wmww:
@muktupavels There's no advantage of depending on `wayland-protcols` and `wayland-scanner`. Protocols aren't like libraries. They don't change, until we choose to upgrade them, that is definitely a manual process. Furthermore, AFAIK neither of the protocols used are in the major distros' `wayland-protcols` package yet. I could vender in the .xml instead of the .c/.h and depend on `wayland-scanner`, but again, why? There's nothing to get out of generating fresh code each build, except the possibility that something goes wrong.

muktupavels:
I still think that generated files should not be in git... D-Bus interface files does not change too, but still everyone use `gdbus-codegen` to generate files at build time, at least I have not seen anyone adding generated files to git.

raveit65:
My investigation shows me that we can't create the files during building because they aren't in all major distros, or am i wrong?

yetist:
@raveit65 
`wayland-scanner` is part of wayland, so you can found it in the `wayland-devel` package: https://fedora.pkgs.org/29/fedora-i386/wayland-devel-1.16.0-1.fc29.i686.rpm.html.

If `wayland` is provided in a distribution, `wayland-scanner` must exist in a devel/dev-like development package, otherwise it is a bug in the distribution.

@wmww 
I think you should only add XML files into git, and use `wayland-scanner` to generate the header/code files at build time.

In addition to the reasons given by @muktupavels , I think there are the following points:
1. Keep in sync with `wayland`.
`wayland-scanner` will be upgraded with the upgrade of `wayland`. Future versions of `wayland-scanner` generated code may change. This change needs to be known in time so that other code can be known and updated in time.
2. More automation.
Changes in the XML file protocol are automatically processed.
3. Keep the git repo clean.
Changes in the content of files such as `-gen.h/c` should not be recorded in the GIT repo.

yetist:
Please run the follow scripts for test:
```bash
#!/bin/bash

gen_files=(
https://github.com/mate-desktop/mate-panel/raw/99ad2e03104288ef8c6215a83c37212662aa3749/mate-panel/wayland-protocols/wlr-layer-shell-unstable-v1-client-protocol.h
https://github.com/mate-desktop/mate-panel/raw/99ad2e03104288ef8c6215a83c37212662aa3749/mate-panel/wayland-protocols/wlr-layer-shell-unstable-v1-protocol.c
https://github.com/mate-desktop/mate-panel/raw/99ad2e03104288ef8c6215a83c37212662aa3749/mate-panel/wayland-protocols/xdg-shell-client-protocol.h
https://github.com/mate-desktop/mate-panel/raw/99ad2e03104288ef8c6215a83c37212662aa3749/mate-panel/wayland-protocols/xdg-shell-protocol.c
)

xml_files=(
https://github.com/wayland-project/wayland-protocols/raw/master/stable/xdg-shell/xdg-shell.xml
https://github.com/swaywm/wlr-protocols/raw/master/unstable/wlr-layer-shell-unstable-v1.xml
)

echo download the generated files from pr ...
mkdir the_pr
pushd the_pr
for i in ${gen_files[@]};do
        wget $i
done
popd

echo download the xml files ...
for i in ${xml_files[@]};do
        wget $i
done

mkdir new
pushd new
wayland-scanner client-header ../xdg-shell.xml xdg-shell-client-protocol.h
wayland-scanner public-code ../xdg-shell.xml xdg-shell-protocol.c

wayland-scanner client-header ../wlr-layer-shell-unstable-v1.xml wlr-layer-shell-unstable-v1-client-protocol.h
wayland-scanner public-code ../wlr-layer-shell-unstable-v1.xml wlr-layer-shell-unstable-v1-protocol.c
popd

echo now, you can see the diff ...
diff -uNr the_pr new
```

raveit65:
Didn't know that we need `wayland-scanner`
```
[root@mother rave]# dnf provides */wayland-scanner
wayland-devel-1.15.0-1.fc28.x86_64 : Development files for wayland
Repo        : @System
Matched from:
Filename    : /usr/bin/wayland-scanner
```

muktupavels:
You need only if files will be generated at build time and/or you will want to regenerate them at some point... I also think that at least `xdg-shell.xml` should not be copied, but used from `wayland-protocols`.

I guess problem is with `wlr-protocols` because it is not available... But then how related compositors are built in distros if all dependencies are not available? Or they are not available too?

wmww:
You're right that future changes to libwayland could potentially require generated code to change, and I didn't realize `wayland-scanner` was shipped with `wayland` (that makes this quite convenient). I'll generate them in the build process.

I would still strongly advise against using anything from a `wayland-protocols` package. If there is a version of a protocol newer then what I used, we won't benefit (or maybe it will break the build, I'm not sure). If there is a version older it will break the build. If some distro doesn't have a protocol we expect it to it will break the build.

> But then how related compositors are built in distros if all dependencies are not available?

I don't know about others. I know Mir keeps all its .xml protocol files in the source tree.

wmww:
Replaced by #898

Issue #778, opened 2018-04-15
"Time" Applet Lacks Padding on the Left-and-Right Sides


MingcongBai:
#### Expected behaviour

The "time" applet on the MATE Panel should have a couple pixels of padding for its left and right sides, especially when placed on the right corners.

#### Actual behaviour

Currently the applet has no padding whatsoever, the time displayed appears to be "touching" the edge of the display, which makes the applet seem incomplete ("is my time displayed completely?") and could result in cut-offs on certain displays. Please refer to the screenshot below.

![2018-04-15 12-25-57](https://user-images.githubusercontent.com/5006263/38781328-a9699d7c-40a8-11e8-9355-1bad15885abe.png)

#### Steps to reproduce the behaviour

Create a panel and put the "time" applet to the right-most edge of the panel (which should be 100% in width).

#### MATE general version

MATE Panel 1.20.1.

#### Package version

MATE Panel 1.20.1.

#### Linux Distribution

AOSC OS.

#### Link to downstream report of your Distribution

N/A.

DrReD:
In my machine this also leads the the following problem:
[Fedora Bugzilla#1405795](https://bugzilla.redhat.com/show_bug.cgi?id=1405795)

(I can supply the Arial font for testing, but it is the usual Microsoft Arial Font distributed with office 2016)

MingcongBai:
Well mine simply doesn't have any padding to the right...

rkw-inspiron:
I found that adding a few pixels in mate-panel/applets/clock/clock.c line 1283 "" padding: 0px;\n" solves this

lukefromdc:
I do NOT want forced padding on the clock or any other applet on my setup that cannot be overridden by themes. The real question is why was this set up with hardcoded 0px in the first place: was the applet jumping on focus or something?

lukefromdc:
One more point: since that padding value is loaded with GTK_STYLE_PROVIDER_PRIORITY_APPLICATION , you can override it with something like

#clock-applet-button {
	padding: 3px;
}

Added to ~/.config/gtk-3.0/gtk.css and you can create this file with just those lines if it does not exist. This will give you the padding you (or any distro) want without having to modify and recompile the source code. Theme values from that file load with GTK_STYLE_PROVIDER_PRIORITY_USER and thus override everything else (USER should never be used in application code for this reason).

rkw-inspiron:
[...]

PR #896, opened 2018-12-03
clock.c: Force negative margin, not padding on clockbutton 

lukefromdc:
This function is no longer needed to make the clock fit in any panel, and no longer seems to affect warnings given in terminal with a debug build of GTK 3.24. Not forcing zero padding allows themers more control, and retaining the clock-applet-button widget name allows themers to keep any unwanted padding off the clockbox. Tested with all themes

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1273,15 +1273,16 @@ clock_update_text_gravity (GtkWidget *label)
 }
 
 static inline void
-force_no_button_padding (GtkWidget *widget)
+force_button_negative_margin (GtkWidget *widget)
 {
         GtkCssProvider  *provider;
 
         provider = gtk_css_provider_new ();
         gtk_css_provider_load_from_data (provider,
-                                         "#clock-applet-button {\n"
-                                         " padding: 0px;\n"
-                                         " margin: 0px;\n }",
+                                         "#clock-applet-button.flat.toggle > box.horizontal > box.horizontal > image {\n"
+                                         "margin-top: -4px;\n"
+                                         "margin-bottom: -4px;\n"
+                                         "}",
                                          -1, NULL);
         gtk_style_context_add_provider (gtk_widget_get_style_context (widget),
                                         GTK_STYLE_PROVIDER (provider),
@@ -1300,7 +1301,7 @@ create_main_clock_button (void)
         button = gtk_toggle_button_new ();
         gtk_button_set_relief (GTK_BUTTON (button), GTK_RELIEF_NONE);
 
-        force_no_button_padding (button);
+        force_button_negative_margin (button);
 
         return button;
 }


lukefromdc:
To allow themers to put padding on the clock applet if they want to-but default should stay at zero so users don't have to waste space. See
https://github.com/mate-desktop/mate-panel/issues/778

raveit65:
This looks much better and i don't get any mentioned warnings about clock-applet in x-session-errors log.
But the weather icons are cropped.
Before:
clock-applet-wether-icon-good
with PR:
clock-applet-cuted-weater-icon

This is with Blue-Submarine theme and HIDPI resolution.

muktupavels:
Should be fixed in theme by removing top & bottom padding!?

lukefromdc:
Anything in the theme would be overriden by APPLICATION prioriry as the code stands in master

muktupavels:
What else would cause that? I think `#clock-applet-button` is `GtkButton` so padding could be inherited from `button`?

raveit65:
Please have a look in theme before you post.
Again, there is no setting for `image` widget in theme.

muktupavels:
```
#PanelApplet button,
#PanelApplet button.flat,
#PanelApplet button.toggle
#PanelApplet button.flat.toggle {
    ...
    padding: 4px;
}
```

raveit65:
Please install MATE desktop and test it with adwaita theme before posting.

lukefromdc:
About padding on this widget: since the widget is constrained to the panel size, it is applied to the inside and can cut off content. The tasklist buttons are like this as well, it was an issue in writing my panel theme. 

Thus, a case can be made for keeping the function and modifying it to force zero top and bottom padding, or we can remove it (one less jump, one less code block) and expect themes not to set broken top/bottom padding for the clock applet.

raveit65:
For me it looks like that the `horizontal.box` for the weather icon/label is at the wrong places in widget order in code.
see gtk3-inspector.
clock-applet-weather-icon
The clockbox for the weather widgets should be below the `clock-applet-button` (GtkToggleButton),
and not below the first `horizontal.box` for the clock itself.
Or not?

raveit65:
> Thus, a case can be made for keeping the function and modifying it to force zero top and bottom padding, or we can remove it (one less jump, one less code block) and expect themes not to set broken top/bottom padding for the clock applet

I don't think that it is wrong to set a padding for GtkButton in themes.
Again, it affects also themes like adwaita, which don't set anything for applications. 
Or do you think Adwaita theme did something wrong?

I think the problem is more that the clockbox for weather is at the wrong place in code.

muktupavels:
There is nothing wrong with `Adwaita`. It will not ship application specific css, it does not have GNOME app specific css either. If you want support `Adwaita` then probably your only option is to detect when `Adwaita` is in used and load custom css like GNOME apps do.

P.S. Hardcoding only top & button also is not that simple because when panel is vertical you would want to change left & right...

P.S.S. Your screenshot looks ok to me. Weather icon and label is in one box. Then "weather" box and clock/date label is in other box. It is probably that way so you can easily hide weather when it is disabled.

raveit65:
> There is nothing wrong with Adwaita. It will not ship application specific css, it does not have GNOME app specific css either. If you want support Adwaita then probably your only option is to detect when Adwaita is in used and load custom css like GNOME apps do.

I didn't say that something is broken with Adwaita theme.
Sadly , you don't test anything here before you post something,
otherwise you would see that adwaita has the same issue.
In other words, something is wrong in code.

And if we want to hardcode something again to fix it for Adwaita support, please set top/bottom margin for the weather icon.
No need to force zero top/bottom padding on the button.
```
#clock-applet-button.flat.toggle > box.horizontal > box.horizontal > image {
    margin-top: -4px;
    margin-bottom: -4px;
}
```

raveit65:
I tried to re-order the widgets like this https://gist.github.com/raveit65/4a0afb5c2e2ebc2dd1f53d726e0d8dac
I packed both boxes in one main box betow the button in widget order,
but it doesn't help with adwaita and all other themes.
clock-weather-adwaita



Still wondering why the `image` widget doesn't display smaller and fits in his horz. box.

muktupavels:
For testing try to increase panel height so there is enough space for padding.top + icon.height + padding.bottom...

raveit65:
https://github.com/mate-desktop/mate-panel/blob/master/applets/clock/clock.c#L1993
Looks like that `icon_size` for the weather icon is read by `mate_panel_applet_get_size` from normal panel icon-size.
So, `icon_size` will always be full panel icon-size.
That means we need to hardcore `vertical top/bottom padding=0` for  `clock-applet-button` or
set an negative top/bottom margin for the `image`,
because Adwaita and probably all copies from adwaita set a `padding= 4px 8px` for `button` widget.
https://gitlab.gnome.org/GNOME/gtk/blob/master/gtk/theme/Adwaita/gtk-contained.css#L203

But maybe your guys have a better idea....

raveit65:
> For testing try to increase panel height so there is enough space for padding.top + icon.height + padding.bottom...

It depends on panel size and icon-theme size. Eg. 
panel-size = 40px --> icon-theme size =48px --> weather-icon is cut.
panel-size = 52px --> icon-theme size =48px --> weather-icon displays well.

lukefromdc:
I had not heard before of setting a negative margin but I see now what it's purpose would be here.

raveit65:
I tested it again with adwaita theme. Adding this setting to ~/.config/gtk-3.0/gtk.css fixes the problem for all panel sizes.
```
#clock-applet-button.flat.toggle > box.horizontal > box.horizontal > image {
    margin-top: -4px;
    margin-bottom: -4px;
}
```
Of course this helps for other themes too.

lukefromdc:
OK, what do we want to do with this? One option is to keep the origjnal code, but replace the hardcoded padding with 

#clock-applet-button.flat.toggle > box.horizontal > box.horizontal > image {
    margin-top: -4px;
    margin-bottom: -4px;
}

I can do that and force-push to this branch if that is the best move

raveit65:
I think this is OK.

lukefromdc:
You mean as it sits or after the proposed change?

raveit65:
I would prefer to replace replace the hardcoded padding with
```
#clock-applet-button.flat.toggle > box.horizontal > box.horizontal > image {
margin-top: -4px;
margin-bottom: -4px;
}
```

lukefromdc:
Force-pushed and renamed to reflect the change. Worked fine in a local test and I was unable to force the theme to cut the weather icons by setting any amount of padding on the button as a whole

monsta:
So I get the icon cut after this change...

With this change (https://github.com/mate-desktop/mate-panel/commit/3e73d66a4e9f41958ca5148bd76d49ab28904793):

bad

With previous commit in master (https://github.com/mate-desktop/mate-panel/commit/a0d306870068a9cd2d637018275679671d0ec190):

good

This is in Debian Testing, GTK+ 3.24.2, Trad.Green theme + mate-faenza icons.

I get the same result in Debian Stable with GTK+ 3.22.11, Trad.Green theme + Mint-X icons.
Actually, I also tested our other themes and Adwaita, it happens with them too.
Panel size is 24 in both cases.

monsta:
Hmm, Greybird theme doesn't cut the icon...

lukefromdc:
I don't have the mint-X or faenza icon theme installed, but I could not get the weather icon to cut with any icon theme or gtk theme I have installed: not GNOME/Adwaita's icons, not mate or menta, nor my UbuntuStudio_Legacy icon theme. 

Do we want to revert this and just declare "can't pad the clock" a WONTFIX?

raveit65:
An alternative is to use only vertical padding for the clock-button (padding-top, padding-bottom).
This will give us space right and left from the button and avoid the cut icon without any padding or margin.
But honestly, i am still wondering that only @monsta had a problem with the margin setting.

monsta:
I can reproduce it with other icon themes too.

HighContrast (also ContrastHigh - it has the same icon):

highcontrast

MATE (also Adwaita, Menta, Gnome - the same icon):

mate

raveit65:
@monsta 
Ok, we know already that it doesn't work for you for some reasons.
More important is to know if using only vertical padding would work?
```
                                         "#clock-applet-button {\n"
                                         "padding-top: 0px;\n"
                                         "padding-bottom: 0px;\n"
                                         "}",
                                         -1, NULL);
```





Btw. your screenshot isn't very helpfull for me , to much grey, I do not recognize where something should cut. Better use the sun icon.
Please add a second location where the sun is shining for a screenshot. :)

raveit65:
Please try https://github.com/mate-desktop/mate-panel/pull/901
Here i use only vertical padding for the clock-button.

PR #873, opened 2018-09-25
Initial Wayland support

wmww:
This gives initial support for MATE panel to work under Wayland. It uses the Layer shell protocol, which is currently implemented by Sway and other wlroots based compositors, and will soon be implemented in Mir. The Wayland backend is incomplete, but its enough to get started with.

By default, the panel builds with X and Wayland support (if it finds both development libraries on your system). To control which backend(s) are built, pass in `--enable-x11` and/or `--enable-wayland`. If either backend is specified, the other is disabled unless it is also explicitly specified.

If its built with support for both platforms, it detects which is being used by GTK at runtime. This PR should not change the panel's behavior when run under X11, regardless if Wayland support is built or not.

This is the first step in a larger effort to port MATE to Wayland.

PR #898, opened 2018-12-07
Add XML Wayland protocols

wmww:
This adds the xml Wayland protocols from which we can use `wayland-scanner` to generate libwayland interface code at build time. A future PR will add the make files necessary to generate the code. This will be used by the upcoming Wayland backend you can find in #873. This is an alternative to committing the generated code, as I was doing in #895.

### DIFF
--- a/mate-panel/wayland-protocols/wlr-layer-shell-unstable-v1.xml
+++ b/mate-panel/wayland-protocols/wlr-layer-shell-unstable-v1.xml
@@ -1,0 +1,285 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<protocol name="wlr_layer_shell_unstable_v1">
+  <copyright>
+    Copyright © 2017 Drew DeVault
+
+    Permission to use, copy, modify, distribute, and sell this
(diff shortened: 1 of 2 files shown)


ammen99:
@wmww Not really sure about it, but shouldn't the protocol files installed to `/usr/share` be used (speaking about `xdg-shell`)?

muktupavels:
@ammen99 It should be, check https://github.com/mate-desktop/mate-panel/pull/895 for his opinion. :)

ammen99:
> @ammen99 It should be, check #895 for his opinion. :)

Thanks, read some of the comments there.

Well, xdg-shell is a stable protocol, meaning it won't change. Plus, this thing here is about a wayland port of mate-panel - and the compositors that support it either require wlroots (which depends on the system wayland-protocols) or on Mir, which seems to be also quite new. My thought is that if the user wants wayland support, then they will already have a relatively up-to-date system (compared to upstream protocols). Plus xdg-shell stable isn't that new, IIRC it was merged into upstream some half-year ago.

Then again, this is just my 5 cents, just something that you may want to consider.

PR #899, opened 2018-12-10
Wayland additions to build system

wmww:
This adds most of the changes needed to our build system for #873 (Wayland Support).

It autodetects which backends (x11 or wayland) are available, or can be forced to build one or both with the `--enable-x11` and `--enable-wayland` options. As in the other PR, if a backend is specified specifically only that backend is built, and an error will occur if it is not available. Both backends may be specified, in which case both must be available for the build to succeed.

I also added the `mate-panel/wayland-protocols` subdirectory to the build system. The makefile in it generates the protocol files from the xml at build time.

PR #900, opened 2018-12-17
Replace wayland-scanner private-code option with code

wmww:
Should fix the problem noted here: https://github.com/mate-desktop/mate-panel/pull/899#issuecomment-447844869

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -167,30 +167,31 @@ AM_CONDITIONAL(WNCKLET_INPROCESS, test -n "$WNCKLET_COMPILE_INPROCESS")
 # For the run dialog
 gl_CHECK_TYPE_STRUCT_DIRENT_D_TYPE
 
-must_enable_x11=no
-must_enable_wayland=no
-try_all_backends=yes
-
+# $enable_x11 and $enable_wayland will be set to "yes", "no" or "auto"
 AC_ARG_ENABLE(x11,
               [AC_HELP_STRING([--enable-x11],
-                              [Enable X11 support and disable Wayland (unless it's also enabled explicitly)
-                              @<:@default=yes@:>@])],
-              [must_enable_x11=yes
-               try_all_backends=no],)
+                              [Explicitly enable or disable X11 support
+                              (default is to enable only if X development libraries are detected)
+                              @<:@default=auto:>@])],
+              [enable_x11=$enableval],
+              [enable_x11=auto])
 
 AC_ARG_ENABLE(wayland,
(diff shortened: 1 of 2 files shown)


wmww:
The problem is that apparently on Debian stable, the `wayland-scanner` is too old to support the `private-code` option. Replacing it with just `code` is deprecated, but works fine for now. This PR has nothing to do with the `--enable-*` flags.

raveit65:
Do we really should support old debian stable for wayland support?
Debian stable will never be updated to next mate-1.22.
For the record, debian stable use mate-1.18 which is EOL.
I am in doubt that someone will test experimental wayland support for mate-panel on deprecated debian-stable.
So why introducing deprecated code?

raveit65:
And i am in doubt that Ubuntu 18.04 will ever been update to next mate-1.22 or a future version with mate-panel for wayland. Maybe ubuntu maintainer have a plan to update LTS version.
@flexiondotorg  ^^^^^^

Your decision, if you think it is useful to support older distro releases without mate-panel for wayland, i will follow you.

To be realistic, i don't think that a working wayland session for Mate is ready before mate-1.24/26

wmww:
It doesn't matter if someone wants to get a Wayland session running on Ubuntu 18.04 or not. This is just to fix the build system so it doesn't crap out on those distros when not given an explicit `--enable-x11`. Even if it won't be used much, there's no reason I see to go to extra trouble to disallow building with Wayland on the latest Ubuntu LTS.

raveit65:
Should be tested with debian-stable or ubuntu-18.04 installed.
Sorry, i don't use those distros.

wmww:
I've tested on Ubuntu 18.04, and @monsta [has tested](https://github.com/mate-desktop/mate-panel/pull/899#issuecomment-448579890) on Debian stable.

wmww:
I fixed the issue of `make clean` not removing generated code, and I made some adjustments to the build options.

As before, both backends are built by default, but now explicitly enabling one does not disable the other. Rather, they're both set to `auto` by default, which will cause them to build if available, but not cause an error otherwise. If the `--enable-*` flag is passed for either, it causes an error if the libraries are not found, and if `--disable-*` it causes that backend to not be built.

raveit65:
> I've tested on Ubuntu 18.04, and @monsta has tested on Debian stable.

and @monsta  is part of core-team and is privilege to merge PR

lukefromdc:
As expected, passing --disable-x11 yields

`configure: error: No support yet for non-X11 builds`

PR #901, opened 2018-12-24
clock: force_no_button_vertical_padding

raveit65:
fixes weather-icon problems with some themes

with TOK theme:
clock-applet-tok-theme
with adwaita:
clock-applet-adwaita

Sorry, i can't test linuxmint themes, because it isn't allowed to spam my mate-appearance-properties with tons of similar themes.

@monsta @lukefromdc 
Does this work for you?

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1273,15 +1273,17 @@ clock_update_text_gravity (GtkWidget *label)
 }
 
 static inline void
-force_button_negative_margin (GtkWidget *widget)
+force_no_button_vertical_padding (GtkWidget *widget)
 {
         GtkCssProvider  *provider;
 
         provider = gtk_css_provider_new ();
         gtk_css_provider_load_from_data (provider,
-                                         "#clock-applet-button.flat.toggle > box.horizontal > box.horizontal > image {\n"
-                                         "margin-top: -4px;\n"
-                                         "margin-bottom: -4px;\n"
+                                         "#clock-applet-button {\n"
+                                         "padding-top: 0px;\n"
+                                         "padding-bottom: 0px;\n"
+                                         "margin-top: 0px;\n"
+                                         "margin-bottom: 0px;\n"
                                          "}",
                                          -1, NULL);
         gtk_style_context_add_provider (gtk_widget_get_style_context (widget),
@@ -1301,7 +1303,7 @@ create_main_clock_button (void)
         button = gtk_toggle_button_new ();
         gtk_button_set_relief (GTK_BUTTON (button), GTK_RELIEF_NONE);
 
-        force_button_negative_margin (button);
+        force_no_button_vertical_padding (button);
 
         return button;
 }


raveit65:
This is weird.
This is exactly the state before the commit with using negative margin, only that the horizontal padding isn't set to zero.

raveit65:
It is possible that you have to restart the applet after size change.
But this is a old problem and you will get this before the commit with negative margin was added.

raveit65:
PR is updated!
I set the vertical margin also back to zero, now code looks like this.
```
                                         "#clock-applet-button {\n"
                                         "padding-top: 0px;\n"
                                         "padding-bottom: 0px;\n"
                                         "margin-top: 0px;\n"
                                         "margin-bottom: 0px;\n"
                                          "}",
                                          -1, NULL);
```
So, only the horizontal hard coding of margin/and padding is removed, comparing to master before https://github.com/mate-desktop/mate-panel/pull/896

lukefromdc:
At the moment, with access to partially cloudy/sun partially cloudy/moon and round moon weather icons showing I could not get a visual difference in either the TOK theme (gtk+icons) or my own theme, and in all cases it was in fact necessary to restart the panel to avoid clipping the icon if the panel had been resized smaller. I tested 26px, 24px, and 20px and all got close to the correct icon side-but (I think all) all had a slight tendecy to clip the top rays when for a moment I got a full sun icon to show. I even tested old master from before the previous PR relating to this was merged and saw exactly the same icon sizes rendered

lukefromdc:
Here are results I got with adwaita-dark , partial sun, and three different panel versions:

Before the previous commit relating to this issue:
adwaita_dark_pr_26px

With git master as of 12-25-2018:
adwaita_dark_26px_master

with this PR:
adwaita_dark_pr_26px

Again, I don't have Linuxmint themes installed so can't test them but we can see icon clipping in all three panel/clock applet versions here, and of course the panel was restarted each time.

Note that clipping rays on the sun is far more visible on dark themes, though BlackMATE was only minimally affected if at all:
26px_blackmate_oldmaster

lukefromdc:
Thus, I still think we have a general problem relating to rendering the weather icon too large for the available space in the panel, if that is a custom widget we may need to deal directly with the allocations for that image, probably be subtracting a few px from the height allocation to force the icon to render smaller. Resizing the panel just 2px larger and NOT restarting it seemd to stop the icon cutting, so maybe that's all we have to reduce it by but a bit more space (maybe 4px) might look better on all but the slimmest panels

lukefromdc:
Here is what I got by subtracting 2px from the results of` mate_panel_applet_get_size to set the icon size:
`
adwaita_dark_26px_new_commit

usl_dark_26px_new_commit

muktupavels:
What is that random `2`? I would understand if goal would be to create empty space around icon, but even then I would say that it should be done by using css. Also does not that mean that weather icon now will be smaller icon then other icons on panel?

The real problem here (or at least partially) is that panel is violating gtk sizing rules - container (Panel) must be big enough to hold all child widgets. In short if weather icon with padding, margin and any other things requires minimum height 60px (as example) then panel must have at least 60px height. Failing that you are just asking for problems and with debug builds also for `... attempt to underallocate ...` warnings. Unfortunately it did not look like easy fix last time I looked at it.

monsta:
We'll need to scale the icon anyway if we want it to fit and not to be cut...

monsta:
With panel size 27-29 the icon is still sometimes cut from the bottom. In addition, with Trad.Green theme, there's a border around the applet which appears on mouseover. Well, for some reason the icon is drawn *over* that border.

weather

muktupavels:
That is good example why random numbers don't fix problems...

raveit65:
I don't see that with traditional  themes and our default icon theme not with our default theme

lukefromdc:
It What I am seeing on the bottom of your screenshot of tradtional-green is that the entire applet is getting a border which is offset from the bottom of the panel when moused over and is NOT centered on the panel. That may be an issue to be fixed in the theme, unless we want to scale the icon at returned panel height minus 4px.

The 2px figure is not random, it's the minimum value I found to be effective in testing,though 17-20px panels  might also need 4px as I did see some cutting below 20px but I didn't want to make the icon too small or make it blurry.

lukefromdc:
I did get an offset border if the panel was on the top, but that might have been 1 or 2px being drawn offscreen with my setup:

traditional-green-24px-top

muktupavels:
To me that looks like margin on button...

lukefromdc:
My advice is to use this now to mostly fix the problem for users, then look at refactoring the code as Albert suggested. The icon rendering might need to be redone entirely, and should be done ideally in the same way that gave the sharpest results on HiDPI screens for icons elsewhere in the panel.

yetist:
It works for clock applet, but does not keep the icons on the panel to display the same size. Perhaps different people will choose to use `mate_panel_applet_get_size` minus a value different from `2` as the icon size on their own applet.
So, my suggestion is, would it be better to add a new function to mate-panel, such as `mate_panel_applet_get_icon_size`?

raveit65:
@monsta 
You concealed a important detail in you last post with Traditional theme.
Which icon-theme are you using?

And are you sure that the problem doesn't exists with hardcoding vertical/horizontal padding/margin before https://github.com/mate-desktop/mate-panel/commit  /3e73d66a4e9f41958ca5148bd76d49ab28904793 ?

I think we are talking about a problem which doesn't have anything to do with this PR.
This is my result with Traditional theme, mate-icone-theme, 29px and HIDPI resolution.
weather-appet-trad-29px
No cutting edges and the icon image is centered.

yetist:
When the panel size changes, it may be necessary to recalculate the icon size. Now the weather icon does not change with the panel size, so we can get another issue:
Please test as follow:
1. set panel size to 44+.
2. restart panel/clock-applet, now you will see weathor icon bigger, maybe icon size value is `panel_size -2` .
3. now set panel size from 44 down to 20, the icon will be cut off.

Initial icon:

after change panel size:

raveit65:
@yetist 
Did you restart applet after reducing size from 44 down to 20?

This is another issue which will happen with master before https://github.com/mate-desktop/mate-panel/commit/3e73d66a4e9f41958ca5148bd76d49ab28904793
...and already posted here https://github.com/mate-desktop/mate-panel/pull/901#issuecomment-449830030

yetist:
> @yetist
> Did you restart applet after reducing size from 44 down to 20?
Don't restart applet after that. If restart panel/clock-applet, it will be ok, because it is the first run of clock-applet.

raveit65:
And sorry guys, my initial PR was to set only vertical padding/marging in gtk_css_provider
```
                                         "#clock-applet-button {\n"
                                         "padding-top: 0px;\n"
                                         "padding-bottom: 0px;\n"
                                         "margin-top: 0px;\n"
                                         "margin-bottom: 0px;\n"
                                          "}",
                                          -1, NULL);
```
Compare to before https://github.com/mate-desktop/mate-panel/commit/3e73d66a4e9f41958ca5148bd76d49ab28904793
```
                                        "#clock-applet-button {\n"
                                         " padding: 0px;\n"
                                         " margin: 0px;\n }",
                                          -1, NULL);
```
So all problems you posted here doesn't have anything to do with PR.
They have to be in master before.
It is impossible that only removing horizontal padding/margin can trigger your problems.

raveit65:
> Don't restart applet after that. If restart panel/clock-applet, it will be ok, because it is the first run of clock-applet.

Why? This is another problem and isn't caused by using `2`.

raveit65:
Again, you need to restart applet after you change panel size.
This isn't trelated to PR:

raveit65:
Closing this as people posting unrelated problems to PR.
I will use this code as patch for fedora.
```
                                         "#clock-applet-button {\n"
                                         "padding-top: 0px;\n"
                                         "padding-bottom: 0px;\n"
                                         "margin-top: 0px;\n"
                                         "margin-bottom: 0px;\n"
                                          "}",
                                          -1, NULL);
```
So issue with missing space left and right from clock-button is solved for me.
Feel free to open another PRs to fix your problems which are unrelated to this PR.

yetist:
> Again, you need to restart applet after you change panel size.
> This isn't related to PR.

Sorry, the above reply is wrong, originally intended to write `did't`, but written as `don't`.
I didn't restart applet after adjusting it to a smaller size, so I saw this problem. 

The first commit is ok, the second commit works, but use `2` is not a good choice,  we can calculate the standard size of the icon based on the panel size.
```
  if (panel_size <= panel_max_icon_size + spacing)
    {
      if (panel_size < 22 + spacing)
        panel_icon_size = 16;
      else if (panel_size < 24 + spacing)
        panel_icon_size = 22;
      else if (panel_size < 32 + spacing)
        panel_icon_size = 24;
      else if (panel_size < 48 + spacing)
        panel_icon_size = 32;
      else if (panel_size < 64 + spacing)
        panel_icon_size = 48;
      else
        panel_icon_size = 64;
    }
```
Like gnome-panel https://gitlab.gnome.org/GNOME/gnome-panel/blob/master/libgnome-panel/gp-applet.c#L183, gnome-panel provide `gp_applet_get_panel_icon_size` function, may be mate can provide  similar functions for applets to use, so that icons displayed on the panel can be kept the same size.

In fact, there is another unrelated issue which is not suitable for discussion here, the applet need to connect the `change-size` signal of panel. Maybe we can improve it later.

raveit65:
@lukefromdc 
Personal i like the idea to reduce the icon size 2px because it it is more equal with the font size of clock-applet, but OK.
I will do.

lukefromdc:
Once we get a correct fix for the icon sizing, we can revisit this as it won't be possible for a themer wanting to pad the top and bottom of a clock applet on a vertical panel to do so. That is a low priority, but making sure the icon never cuts on any panel size on any system is a higher one and also requires connection to the `change-size `signal as said before. All this is for another PR though

lukefromdc:
https://github.com/mate-desktop/mate-panel/pull/903 enables icon resizing with changes in panel size. Code is still a bit kludgy, but it connects to the the change-size signal.

raveit65:
Ok, maybe we need to set a horizontal padding with the GtkCssProvider.
I think it looks good if the applet have some space left and right, also if a theme don't set that.
I will check that the next days.

raveit65:
Forcing a horizontal padding works fine for all themes if applet is build-in-process.
I will do a PR  where we can discuss if this is needed.

PR #902, opened 2018-12-30
Tray: Fix too-wide icon spacing after mate-polkit has been used and icon dismissed (privileges dropped) with  sni

lukefromdc:
Note that vertical and horizontal panels require different options in the grid, as "homogenous" is needed in rows for horizontal panels and columns in vertical panels to prevent icon rendering problems including icons growing without limit when the tray is moved from a horizontal to a vertical panel

### DIFF
--- a/applets/notification_area/na-grid.c
+++ b/applets/notification_area/na-grid.c
@@ -145,6 +145,8 @@ refresh_grid (NaGrid *self)
 
   if (orientation == GTK_ORIENTATION_HORIZONTAL)
     {
+      gtk_grid_set_row_homogeneous (GTK_GRID (self), TRUE);
+      gtk_grid_set_column_homogeneous (GTK_GRID (self), FALSE);
       rows = MAX (1, allocation.height / self->min_icon_size);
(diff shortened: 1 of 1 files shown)


lukefromdc:
You need mate-polkit compiled with SNI support enabled and SNI enabled in your panel configuration to test this.   You also need mate-panel from git master after https://github.com/mate-desktop/mate-panel/commit/c7d431d36f8bf6e9b94bab37c19dd51041e117c9
which ported the tray applet layout to GtkGrid but without this PR applied as your test case for the issue. 

Then, to duplicate the doublespaced tray icons, mount any internal disk or do anything else that brings up the mate-polkit passphrase dialog, then click the key icon in the tray and click the menu item to drop privileges. The tray stays wide and the icons respace wider.

Now apply this PR, and the icons stay at their intended spacing and the tray shrinks back down by one item. Test this on both a horizontal and a vertical panel, and test dragging the tray from the horizontal to the vertical panel. It should work every time and you should not see any rendering issues on moving the applet.

PR #903, opened 2018-12-31
Clock: ensure weather icon size is correct for panel height

lukefromdc:
Connect to the "change_size" signal and reload the icon at the correct size

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -179,6 +179,8 @@ struct _ClockData {
         gboolean   can_handle_format_12;
 
         GSettings *settings;
+
+        const gchar *weather_icon_name;
 };
 
 /* Used to count the number of clock instances. It's there to know when we
@@ -1970,12 +1972,37 @@ show_temperature_changed (GSettings    *settings,
 }
 
 static void
+weather_icon_updated_cb (MatePanelApplet *applet,
+                             gint icon_size,
+                             gpointer       data)
+{
+        ClockData *cd = data;
+        GtkIconTheme *theme;
+        cairo_surface_t *surface;
+        gint icon_scale;
+
+        if (cd->weather_icon_name == NULL)
+                return;
+
+        theme = gtk_icon_theme_get_for_screen (gtk_widget_get_screen (GTK_WIDGET (cd->applet)));
+
+        icon_size = mate_panel_applet_get_size (MATE_PANEL_APPLET (cd->applet));
+        icon_scale = gtk_widget_get_scale_factor (GTK_WIDGET (cd->applet));
+
+        surface = gtk_icon_theme_load_surface (theme, cd->weather_icon_name, icon_size, icon_scale,
+                                               NULL, GTK_ICON_LOOKUP_GENERIC_FALLBACK, NULL);
+
+        gtk_image_set_from_surface (GTK_IMAGE (cd->panel_weather_icon), surface);
+
+        cairo_surface_destroy (surface);
+}
+
+static void
 location_weather_updated_cb (ClockLocation *location,
                              WeatherInfo   *info,
                              gpointer       data)
 {
         ClockData *cd = data;
-        const gchar *icon_name;
         const gchar *temp;
         GtkIconTheme *theme;
         cairo_surface_t *surface;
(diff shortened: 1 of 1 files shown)


lukefromdc:
If anyone knows a better way to connect that callback I am open to it. The callback is based on a simplified version of `location_weather_updated_cb` and like it has to accept three arguments to work for some reason, not just the widget sending the signal (the applet) and the data pointer. If I remove that middle line its an instant segfault when it is run, so I used

```
static void
weather_icon_updated_cb (MatePanelApplet *applet,
                             gint icon_size,
                             gpointer       data)

```
and it works fine. Icon size is reset in the function, it is used in the function, and the function returns void yet with the callback format I have to put SOMETHING in that middle line. I really don't know enough about gsignals to know what is going on with that. It works but I suspect it could be done better and then force-pushed.

Also note that watching the icon change size with the panel clearly reveals the icon cutting issue: it looks like we've got only a 16px icon and a 24 px icon, and the 24 px full moon especially and to a lesser extent the full sun (and icons based on them) are getting rendered just slightly too large to fit.  Delaying the switch by (again) 2px as the panel is made larger would decisively stop the cutting, which is worst at exactly one panel size: 25px. The default 24px fits almost perfectly with my theme and this PR in place as it stands now.

muktupavels:
I don't know if that is case, but I would expect that `change-size` gives same size as `size-allocate` callback... In other words - there is already code that updates clock when panel size changes! I would say that problem is elsewhere...

lukefromdc:
The word "allocate" does not turn up anywhere in the entire clock.c file. Also, I found that although we have the full range of .png icons in libmateweather plus SVG icons, only the 16 and 24 px icons seem to be actually used. I have more digging to do...

lukefromdc:
I found that with this, (and possibly previous work) the MATE icons are never cutting. It looks like when my own icon theme is in use, the 32px icons are getting selected starting at a 25px applet height, something that should not happen in any icon theme. All of those icons fit properly in their own image space, and none of these files exceed the expected px size limits.

It looks like with the MATE icons the weather icons are scaling smoothly with the panel with this, implying the SVG icons might be the ones in use.

muktupavels:
1) `panel_button_change_pixel_size`...

2) I did not do any research, but I have feeling that this is regression from HiDPI fixes or something like that. Carefully read `gtk_icon_theme_load_surface` documentation...

lukefromdc:
I went through the` gtk_icon_theme_load_surface` docs, and found a way to make the applet size itself to properly fit the panel (and stay at its original size so as not to get blurry) without having to refactor the whole applet. Note that some icons for weather are identical except for empty space around them between 22 and 24px versions, but this makes the panel switch to the next larger icon at the 22, 24, 32, and 48px panel heights to match available icons but never cut them. It stops the use of the 32px icon in 25-31px panels, we almost but not quite got away with that due to the exact size of those icons.

The allocation warnings (as found earlier by a debug build of GTK) will still need to be dealt with but that's a matter for another PR as it may be a global refactor of the panel applet handing code. There is allocation handling code in mate-panel-applet.c but it might not be working correctly anymore.

lukefromdc:
@raveit65 , @monsta , @yetist , can you test this as it now stands?

lukefromdc:
I just tested running the panel in terminal, with my usual non-debug version of GTK3.24. I didn't get any warnings at all, including on showing/hiding weather or changing weather stations (and icons).

Are you running a full debug build of GTK3 by any chance? I tested that a while back and got torrents of warnings from panel applets.  That one wasn't included, but the clock applet just got a lot of padding/margin changes in master-not all of them limited to the weather icon.

`gtk_widget_get_allocation` 

is not in any function I added or changed, and it looks like the warning might be for the button as a whole. Can you recheck with the weather not shown? If you still get the warning, it is not related to this PR.

yetist:
Well, I retested it and found that the warning still existed before using the PR, so it is not related to this PR.

lukefromdc:
OK, thanks. I figured it has to relate to the clockbox as a whole. I ran into some kind of problem with a debug build of GTK3 (configure log said full debug) as the bucket of warnings I get last time didn't come up in terminal, syslog, xsession-errors etc. I didn't get any panel warnings at all with GTK 3.24.2 from a Dec 13 pull of the git branch, rebuilt with --enable-debug=full . Must have done something wrong with the debug build of GTK but don't know what. Even tried a full new session after installation.

PR #904, opened 2019-01-03
Remove no longer needed forced vertical padding

lukefromdc:
force_no_button_vertical_padding should no longer be needed after https://github.com/mate-desktop/mate-panel/commit/9935eeb51c6fef98573246421ff7e834ff385638 "Clock: Fix weather icon size selection for panel height"

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1274,29 +1274,6 @@ clock_update_text_gravity (GtkWidget *label)
         pango_context_set_base_gravity (context, PANGO_GRAVITY_AUTO);
 }
 
-static inline void
-force_no_button_vertical_padding (GtkWidget *widget)
-{
-        GtkCssProvider  *provider;
-
-        provider = gtk_css_provider_new ();
-        gtk_css_provider_load_from_data (provider,
-                                         "#clock-applet-button {\n"
-                                         "padding-top: 0px;\n"
-                                         "padding-bottom: 0px;\n"
-                                         "margin-top: 0px;\n"
-                                         "margin-bottom: 0px;\n"
-                                         "}",
-                                         -1, NULL);
-        gtk_style_context_add_provider (gtk_widget_get_style_context (widget),
-                                        GTK_STYLE_PROVIDER (provider),
-                                        GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
-        g_object_unref (provider);
-
-
-        gtk_widget_set_name (widget, "clock-applet-button");
-}
-
 static GtkWidget *
 create_main_clock_button (void)
 {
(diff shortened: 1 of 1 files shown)


lukefromdc:
The force_no_button_vertical_padding was an effort to control the weather icon being cut. The icon cutting was caused by selection of oversized icons, which now has been stopped. Also, on a vertical function this would limit the flexability of themers by overriding any padding set at the ends of the applet.

I found no difference at all in my own theme, TOK, or Adwaita-dark in the appearance of the weather icon with this. Note that in TOK the border shown on mouseover will render over the edges of the tallest icons but they are still drawn when the icon height is the same as the panel height e.g a 22px panel (24px icons are mostly the same) or a 32 px panel. Stopping that would compromise all other themes on those panel heights by forcing use of the next smaller size icon. Same in master and with this PR, no difference there.

lukefromdc:
Absolutely, sorry I missed the widget name in all of this. Feel free to push this to the branch, I am on the road at the moment but will be back later. Very much for keeping the widget name . Thanks for catching that!

raveit65:
I think the cropped icon is back with every theme which use a padding.
The problem is the the transition area to the next smaller icon. Eg. Blue-submarine.
53px --> good.
clock-applet-53px
52px --> starts cropping
clock-applet-52px
50px --> cropped
clock-applet-50px
48px --> cropped
clock-applet-48px
47px --> good
clock-applet-47px

lukefromdc:
OK, I will close this PR then, because to delay the transition to each larger icon size just to support a padding around the icon (which a themer could also set by accident from buttons in general) would in my judgement be a bad tradeoff. 

I will look to.see if we can get the panel orientation to set rhis properly for vertical panels though. I assume that should be a new PR?

raveit65:
> I will look to.see if we can get the panel orientation to set rhis properly for vertical panels though. I assume that should be a new PR?

Yes, but  i am not really sure if we need this for vertical panels.
Clock-applet looks really weird on a vertical panel. 
Up to you if you like to do this extra work.

raveit65:
A typical user for the clock-applet on a vertical panel :)
https://i.kym-cdn.com/photos/images/original/001/072/172/6b0.jpg

lukefromdc:
I just found it take a very wide panel to show the clock applet other than drawn vertically for that cat-and that neither the weather or the temperature was even rendered.
Miniumum panel width for a horizontal clock on a vertical panel with 12 hour time is 64px, and 70px to render it without cutting the text of the time without seconds. With 24 hour time 41 px to render horizontally, and about 50px not to cut the text. The standalone weather applet will work at any of these widths, but the clock's weather is NEVER rendered on a vertical panel no matter what locations, panel widths, or options are set. Space was allocated above the rest of the clock for them but even on a narrow panel with the clock drawn vertically they did not render.

Thus, this can't be "fixed" on a vertical panel unless the applet as a whole is also fixed for that use case. A vertical panel can show the very basic clock on a wide enough panel but that's it. Closing as not needed

PR #905, opened 2019-01-07
Remove X specific functions from fish applet

wmww:
Allows the applet to run on Wayland without backend-specific code. Split off of #873

### DIFF
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -32,12 +32,10 @@
 #include <time.h>
 
 #include <cairo.h>
-#include <cairo-xlib.h>
 
 #include <glib/gi18n.h>
 #include <glib-object.h>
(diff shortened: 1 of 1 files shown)


wmww:
Argggggg... sorry, fixed now.

Cairo refuses to hand out the  size of a surface without backend-specific code, so I just store it now.

wmww:
Argggggg... sorry, fixed now.

raveit65:
Thanks, this works fine under X11, as a side effect you fixed a weird flickering with HDPI resolutions of this applet.  :smile: 
Can you please squash second commit into the first one?

lukefromdc:
Didn't catch the request to squash the two commits (which was above the last approval), sorry about that. Can't do that now without an ugly force-push to master that would break any local repo updated in between

PR #908, opened 2019-01-14
clock: set  horizontal padding for clock-button

raveit65:
For some reasons padding settings for `button` widget in gtk+  themes are still ignored if clock-applet is build-in-process.
Forcing a horizontal padding is to make sure we have space left and right from the button.
I think it looks better to have a space there.
`padding-left/right: 4 px;` is used by our and gtk+ default themes. So this is a good value.


raveit65:
> Do we care about putting the clock on a wide vertical panel or is this something for a later fix in the clock? If no or the latter we can merge this now.

For me it ishard to use the applet with a vertical panel, so i don't care.

PR #911, opened 2019-01-16
Libmate-panel make X dependecies optional

wmww:
Required for future Wayland support

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -33,20 +33,23 @@
 
 #include <glib/gi18n-lib.h>
 #include <cairo.h>
-#include <cairo-xlib.h>
 #include <gdk/gdk.h>
-#include <gdk/gdkx.h>
 #include <gdk/gdkkeysyms.h>
 #include <gtk/gtk.h>
+
+#ifdef HAVE_X11
+#include <cairo-xlib.h>
+#include <gdk/gdkx.h>
 #include <gtk/gtkx.h>
 #include <X11/Xatom.h>
+#include "panel-plug-private.h"
+#endif
 
 #include "mate-panel-applet.h"
 #include "panel-applet-private.h"
 #include "mate-panel-applet-factory.h"
 #include "mate-panel-applet-marshal.h"
 #include "mate-panel-applet-enums.h"
-#include "panel-plug-private.h"
 
 #define MATE_PANEL_APPLET_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), PANEL_TYPE_APPLET, MatePanelAppletPrivate))
 
@@ -449,6 +452,8 @@ mate_panel_applet_set_locked_down (MatePanelApplet *applet,
 	g_object_notify (G_OBJECT (applet), "locked-down");
 }
 
+#ifdef HAVE_X11
+
 static Atom _net_wm_window_type = None;
 static Atom _net_wm_window_type_dock = None;
 static Atom _net_active_window = None;
@@ -533,6 +538,8 @@ mate_panel_applet_find_toplevel_dock_window (MatePanelApplet *applet,
 	return None;
 }
 
+#endif // HAVE_X11
+
 /* This function
  *   1) Gets the window id of the panel that contains the applet
  *	using XQueryTree and XGetWindowProperty to find an ancestor
@@ -543,6 +550,7 @@ void
 mate_panel_applet_request_focus (MatePanelApplet	 *applet,
 			    guint32	  timestamp)
 {
+#ifdef HAVE_X11
 	GdkScreen  *screen;
 	GdkWindow  *root;
 	GdkDisplay *display;
@@ -550,6 +558,9 @@ mate_panel_applet_request_focus (MatePanelApplet	 *applet,
 	Window	    dock_xwindow;
 	Window	    xroot;
 	XEvent	    xev;
+
+	if (!GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
+		return;
 
 	g_return_if_fail (PANEL_IS_APPLET (applet));
 
@@ -582,6 +593,7 @@ mate_panel_applet_request_focus (MatePanelApplet	 *applet,
 		    xroot, False,
 		    SubstructureRedirectMask | SubstructureNotifyMask,
 		    &xev);
+#endif
 }
 
 static GtkAction *
@@ -834,15 +846,20 @@ mate_panel_applet_position_menu (GtkMenu   *menu,
 	GtkAllocation   allocation;
 	GtkRequisition  requisition;
 	GdkDevice      *device;
+	GdkMonitor     *monitor;
 	GdkScreen      *screen;
 	int             menu_x = 0;
 	int             menu_y = 0;
 	int             pointer_x;
 	int             pointer_y;
+	GdkRectangle    screen_geom;
 
 	g_return_if_fail (PANEL_IS_APPLET (widget));
 
 	applet = MATE_PANEL_APPLET (widget);
+
+	monitor = gdk_display_get_monitor_at_window (gtk_widget_get_display (widget), gtk_widget_get_window (widget));
+	gdk_monitor_get_geometry (monitor, &screen_geom);
 
 	screen = gtk_widget_get_screen (widget);
 	gtk_menu_set_screen (menu, screen);
@@ -874,9 +891,9 @@ mate_panel_applet_position_menu (GtkMenu   *menu,
 					       allocation.width - requisition.width);
 			}
 		}
-		menu_x = MIN (menu_x, WidthOfScreen (gdk_x11_screen_get_xscreen (screen)) - requisition.width);
+		menu_x = MIN (menu_x, screen_geom.width - requisition.width);
 
-		if (menu_y > HeightOfScreen (gdk_x11_screen_get_xscreen (screen)) / 2)
+		if (menu_y > screen_geom.height / 2)
 			menu_y -= requisition.height;
 		else
 			menu_y += allocation.height;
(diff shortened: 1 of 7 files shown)


raveit65:
Most applets under x11 are gone :/
I just build my rpm with those options for testing PR.
```
%configure                                        \
           --disable-static                       \
           --disable-schemas-compile              \
           --enable-x11                           \
           --disable-wayland                      \
           --libexecdir=%{_libexecdir}/mate-panel \
           --enable-introspection                 \
           --disable-gtk-doc                      \
           --with-in-process-applets=none
```
`--with-in-process-applets=none` is the current default for fedora.
Config summary:
```
              mate-panel 1.21.3
              =================

        prefix:                        /usr
        exec_prefix:                   /usr
        libdir:                        /usr/lib64
        bindir:                        /usr/bin
        sbindir:                       /usr/sbin
        sysconfdir:                    /etc
        localstatedir:                 /var
        datadir:                       /usr/share
        source code location:          .
        compiler:                      gcc
        cflags:                        -O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection
        Maintainer mode:               no
        Use *_DISABLE_DEPRECATED:      no
        Applets to build in-process:   (none)
        Wayland support:               no
        X11 support:                   yes
        XRandr support:                yes
        Build introspection support:   yes
        Build gtk-doc documentation:   no
```
clock- applet, wnck-appets, na-tray, and multiload, cpu-freq from mate-applets and sensor-applet won't start any more. Probably other applets too.
I see only menu-bar-applet, panel-launchers and panel-action-buttons :/
Is building under X11 tested ?

Beside from that we shouldn't introduce deprecated code again which we had already fixed, if building for X11.
```
mate-panel-applet.c: In function 'mate_panel_applet_get_screen_geometry':
mate-panel-applet.c:2436:3: warning: 'gdk_screen_get_width' is deprecated [-Wdeprecated-declarations]
   *width  = gdk_screen_get_width(screen);
   ^
In file included from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:31,
                 from mate-panel-applet.c:36:
/usr/include/gtk-3.0/gdk/gdkscreen.h:58:14: note: declared here
 gint         gdk_screen_get_width             (GdkScreen   *screen);
              ^~~~~~~~~~~~~~~~~~~~
mate-panel-applet.c:2437:3: warning: 'gdk_screen_get_height' is deprecated [-Wdeprecated-declarations]
   *height = gdk_screen_get_height(screen);
   ^
In file included from /usr/include/gtk-3.0/gdk/gdkapplaunchcontext.h:31,
                 from mate-panel-applet.c:36:
/usr/include/gtk-3.0/gdk/gdkscreen.h:60:14: note: declared here
 gint         gdk_screen_get_height            (GdkScreen   *screen);
              ^~~~~~~~~~~~~~~~~~~~~
```

lukefromdc:
If building for wayland, ALL applets will have to be in-process as out of process uses Xembed. Gnome-panel converted all applets (including the entire gnome-applets package) to in process only, probably for this reason

wmww:
I'm now trouble getting out-of-process applets working at all (even at the previous release tag). Will try harder tomorrow

wmww:
I've now tried on multiple distros (Arch and Pop!_OS) and multiple versions of MATE Panel (this branch, master, 1.21.3 tag and 1.21.2 tag), and I still can't get any out-of-process applets working. I've done the full clean -> uninstall -> autogen -> make -> install process each time. Any advice?

libmate-panel-applet/mate-panel-applet.c
```diff
@@ -2370,6 +2416,27 @@ mate_panel_applet_factory_setup_in_process (const gchar               *factory_i
...
+#ifdef HAVE_X11
+	if (GDK_IS_X11_DISPLAY (gdk_screen_get_display (screen))) {
+		*width  = WidthOfScreen (gdk_x11_screen_get_xscreen (screen));
+		*height = HeightOfScreen (gdk_x11_screen_get_xscreen (screen));
+	} else
+#endif
+	{ // Not using X11
+		// TODO: get a monitor somehow, and use gdk_monitor_get_geometry ()
+		*width  = gdk_screen_get_width(screen);
+		*height = gdk_screen_get_height(screen);
```
sc0w: Please, don't add deprecated functions in new code

raveit65:
>  I've done the full clean -> uninstall -> autogen -> make -> install process each time. Any advice?

Which configure options are you using?
I would use the --prefix=/usr option
I noticed that after i tried this PR, the applets didn't start any more with normal build for some reasons.
Try out logout from session and reset the panel with `mate-panel --reset --replace`

wmww:
After switching to a third install on a separate machine (this time with a real MATE session, instead of just bits and pieces), I was able to actually reproduce. Seems to be a bug in the new code that determines applet size.

wmww:
So much trouble for such a little mistake 🤦. Should be fixed now, please give it another spin.

libmate-panel-applet/mate-panel-applet.h
```diff
@@ -123,6 +129,8 @@ int  mate_panel_applet_factory_setup_in_process (const gchar               *fact
 							  gpointer                   data);
 
 
+void mate_panel_applet_get_screen_geometry (GdkScreen *screen, int* width, int* height);
```
muktupavels: Why this is added as public API? And why applet should care about screen size?

I would use:
```
void mate_panel_applet_get_monitor_geometry (MatePanelApplet *applet,
                                             int             *width,
                                             int             *height);
```

This way you can get correct monitor from applet.

libmate-panel-applet/mate-panel-applet.c
```diff
@@ -2370,6 +2416,29 @@ mate_panel_applet_factory_setup_in_process (const gchar               *factory_i
...
+#ifdef HAVE_X11
+	if (GDK_IS_X11_DISPLAY (gdk_screen_get_display (screen))) {
+		*width  = WidthOfScreen (gdk_x11_screen_get_xscreen (screen));
+		*height = HeightOfScreen (gdk_x11_screen_get_xscreen (screen));
+	} else
+#endif
+	{ // Not using X11
+		// TODO: get a monitor somehow, and use gdk_monitor_get_geometry ()
+		GdkRectangle geom;
+		gdk_monitor_get_geometry (gdk_display_get_primary_monitor (gdk_screen_get_display (screen)), &geom);
```
muktupavels: Applet / panel can be on non-primary monitor too, no? Also it is documented that it can return null...

libmate-panel-applet/panel-plug-private.h
```diff
@@ -18,6 +18,12 @@
 #ifndef PANEL_PLUG_PRIVATE_H
 #define PANEL_PLUG_PRIVATE_H
 
+#include <config.h>
```
muktupavels: Is this include here really needed?

wmww: It pulls in the `HAS_X11` define, which is checked to insure this file is not being built under incorrect circumstances.

muktupavels: Well I have not seen that someone would include `config.h` in header files... Also when `panel-plug.c` is compiled `config.h` is included first so by the time this file is included that define should be already available, no?

Should not it be more or less same as it would have been done this way:
```
#define HAS_X11
#include "panel-plug-private.h"
```

libmate-panel-applet/mate-panel-applet.h
```diff
@@ -107,7 +111,9 @@ void mate_panel_applet_set_size_hints(MatePanelApplet* applet, const int* size_h
 
 gboolean mate_panel_applet_get_locked_down(MatePanelApplet* applet);
 
+#ifdef HAVE_X11
```
muktupavels: This is public file that will be used by 3rd party applets. Do you expect all applet developers to define this in their own `config.h` files?

raveit65:
Thanks,
out-of-process applets are running fine now under X11.
I just tested 2 different builds:
1. with --enable-x11 and --disable-wayland --> works fine under X11
2. without any specifc option but with all buildl dependencies met. --> works fine under X11

I think building the panel for X11 and wayland should be the default option and use case.
This will give use the opportunity to write session files for login-managers for wayland and X11 like gnome did.
Does it makes sense to test PR under wayland?

Distcheck runs fine and i saw no new build warnings.
Some time i will start to disable deprecation warnings for GtkAction to make build logs better readable.......

Can you please squash commit 4 and 5 into 3, please?
No need to add a new commit for a fix of a commit.

libmate-panel-applet/mate-panel-applet.h
wmww: Removed from public API

libmate-panel-applet/mate-panel-applet.c
wmww: Function removed

libmate-panel-applet/mate-panel-applet.h
wmww: Ifdef removed

wmww:
I believe I resolved all the issues mentioned so far, and I recommitted to be more organized (if you still prefer I squash some/all of them, just ask).

> Does it makes sense to test PR under wayland?

No, it will just segfault, because Wayland support in the rest of the panel is still to be merged.

raveit65:
> ....and I recommitted to be more organized (if you still prefer I squash some/all of them, just ask).

It's Ok when PR isn't ready and you fix issues. 
Would be nice to do that before merging when all is done.
Sure we can do that during merging but than we lost the internal github link from commits to PR.
So, it's better when author of a PR did that when all is done.

PR #913, opened 2019-01-20
eggsmclient: avoid deprecated 'g_type_class_add_private'

sc0w:


mate-panel/libegg/eggsmclient.c
```diff
@@ -46,7 +46,8 @@ struct _EggSMClientPrivate {
 
 #define EGG_SM_CLIENT_GET_PRIVATE(o) (G_TYPE_INSTANCE_GET_PRIVATE ((o), EGG_TYPE_SM_CLIENT, EggSMClientPrivate))
```
muktupavels: Is not this unused now?

mate-panel/libegg/eggsmclient.c
```diff
@@ -46,7 +46,8 @@ struct _EggSMClientPrivate {
...
 
-G_DEFINE_TYPE (EggSMClient, egg_sm_client, G_TYPE_OBJECT)
+G_DEFINE_TYPE_WITH_CODE (EggSMClient, egg_sm_client, G_TYPE_OBJECT,
```
muktupavels: Why not `G_DEFINE_TYPE_WITH_PRIVATE`?

sc0w:
done

lukefromdc:
How do I fully test this? Wondering what in the panel would be involved in saving a session

raveit65:
Not sure, but i think SMclient is used for recent files menu.

lukefromdc:
I normally have gtk set not to store recent files 
`gtk-recent-files-max-age=0`
so testing this will require logging out and back in at least twice. All of these will take some time because of that. Been a very busy weekend with other things, will see if I can at least get to this one tonight.

sc0w:
It works in the function `egg_sm_client_get_state_file` and it seems to be unused in `mate-panel`

yetist:
@sc0w 

Did you find that `eggsmclient.c` is used in several codes, such as:
https://github.com/mate-desktop/engrampa/tree/master/copy-n-paste
https://github.com/mate-desktop/mate-session-manager/tree/master/egg
https://github.com/mate-desktop/mate-panel/tree/master/mate-panel/libegg
https://github.com/mate-desktop/caja/tree/master/libegg
https://github.com/mate-desktop/atril/tree/master/cut-n-paste/smclient
https://github.com/mate-desktop/mate-terminal/tree/master/src
https://github.com/mate-desktop/mate-utils/tree/master/gsearchtool/libeggsmclient

These directories all look the same.
So I think we can maintain this directory independently and use `git submodule` to include the code, which can make the code get the same updates and maintenance.
What do you think?

lukefromdc:
OK, the questionn of what this code does in mate-panel is on the table and I think I have the answer: it controls restarting a crashed or killed panel by the session manager.

on doing a string search of all of mate-panel for the string "smclient" I find it is used in panel-session.c and main.c and does in fact have a function: to tell the session manager to restart the panel if it dies  I have never heard of launching the panel with "MATE_PANEL_DEBUG" as an evironmental variable until now but that is used to tell the session manager NOT to restart it.

Of course, to use this for working on mate-panel, you would need to run
`export env MATE_PANEL_DEBUG`
and in that terminal run "mate-panel --replace" 
only thing is, at that point the replaced panel is not controlled by the session manager anyway with or without this environmental variable and will stay stopped if killed. That is how I get it stopped for such things as testing in GDB. Thus, I do not know if that portion of the code still has a user, but removing all session management code from mate-panel would quite likely kill the ability to auto-respawn it when it crashes, at least without also refactoring the session management code itself.

I  recall in all the work with Caja and the GtkApplication transition that the removal of the session management code from GNOME 3 forced a redo of how Nautilus was started,  with GNOME relying on the desktop file to start it rather than the session management code. In my early work on that I did the same thing, but when I got the session management code ported over and working again we got back normal auto-respawning of Caja if is crashes as well as keeping the ability to save application state across sessions.

Thus, I conclude that so long as this code is maintained individually in each package this PR is ready to do, and the same if this is done in lockstep across all affected packages.

I rather like the submodule idea as though it slightly complicates building it greatly simplifies maintainance (one PR next time around instead of PR's in seven different packages) and at build time it is easy enough to drop the module into the source directory as I do in building GNOME packages such as gedit or nautilus.

PR #915, opened 2019-01-26
Wayland support for applets

wmww:
Various fixes for running applets on non-X11 (read: Wayland) platforms.


wmww:
Just review it alone, all the current PRs are independent.

PR #916, opened 2019-01-26
Remove unused mate_panel_applet_position_menu ()

wmww:
As far as I can tell, this was zombie code.


mate-panel/applet.h
```diff
@@ -111,11 +111,6 @@ void        mate_panel_applet_menu_set_recurse (GtkMenu     *menu,
 
 gboolean    mate_panel_applet_toggle_locked  (AppletInfo *info);
 
-void mate_panel_applet_position_menu (GtkMenu   *menu,
```
muktupavels: Is not this API break? Should not *.so version of library be updated to reflect this?

wmww: What library is this part of? I thought it was just part of the `mate-panel` application.

muktupavels: Oh, sorry... I was responding based on function name / prefix that is used by `libmate-panel-applet`.

wmww: Ah, yes, that is confusing. Clearing that up is another benefit of this PR.

raveit65:
Indeed, looks like that mate_panel_applet_position_menu is a left over from
https://github.com/mate-desktop/mate-panel/commit/002244141bd0205aff0aa3e7028529d1ba2ab22c
and
https://github.com/mate-desktop/mate-panel/commit/8001d65ccfbb4fccd84407680e4e4da5dc16cde4
Thanks, i forgot to remove it.

For external applets the code is here https://github.com/mate-desktop/mate-panel/blob/master/libmate-panel-applet/mate-panel-applet.c#L839
IHMO.

PR #918, opened 2019-01-26
Check X11 support before using X11 specific features

wmww:
Add various compile time and runtime checks to keep X11 specific files running on Wayland.

### DIFF
--- a/mate-panel/panel-action-protocol.c
+++ b/mate-panel/panel-action-protocol.c
@@ -24,6 +24,10 @@
 
 #include <config.h>
 
+#ifndef HAVE_X11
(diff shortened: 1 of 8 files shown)


wmww:
@lukefromdc That is expected behavior. This PR doesn't add Wayland support, it simply adds checks that will fail relatively gracefully when the wrong things are done on Wayland. Further patches are needed to add real support.

raveit65:
@lukefromdc 
Do you see a problem or can we merge it?

lukefromdc:
Now that I know this isn't supposed to work yet in Wayland I am OK with merging

PR #919, opened 2019-01-28
fix crash on removing tray

lukefromdc:
*clear applet gsettings on dispose to stop crash and memory leak introduced in github.com/mate-desktop/mate-panel/commit/10b9c30fc79de5f8d9516ba153df233c4968da65
    
*tray: fix misplaced g_return_if_fail check

### DIFF
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -333,6 +333,7 @@ na_tray_applet_realize (GtkWidget *widget)
 static void
 na_tray_applet_dispose (GObject *object)
 {
+  g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);
 #ifdef PROVIDE_WATCHER_SERVICE
   g_clear_object (&NA_TRAY_APPLET (object)->priv->sn_watcher);
 #endif
--- a/applets/notification_area/na-grid.c
+++ b/applets/notification_area/na-grid.c
@@ -183,9 +183,9 @@ void
 na_grid_set_min_icon_size (NaGrid *grid,
                            gint    min_icon_size)
 {
-  grid->min_icon_size = min_icon_size;
-  
   g_return_if_fail (NA_IS_GRID (grid));
+
+  grid->min_icon_size = min_icon_size;
   
   refresh_grid (grid);
 }


muktupavels:
Can you explain what is problem and why this fixes it? I would expect that by this time you have learnt at least something about `GObject` and how signals works...

lukefromdc:
The problem is that with the tray in-process, removing the tray crashes the panel due to the signal invoking a function that then returns a bus error. This disconnects the signal and avoids the crash. With the tray out of process you won't see the crash.

lukefromdc:
Updated after
https://github.com/mate-desktop/mate-panel/commit/3630b204548036d02abc663509b33ebe5797b3b3

lukefromdc:
I did not say it wasn't CRASHING in out-of-process mode, rather that end users would never see that crash as it does not crash the whole panel when an out of process applet crashes, and the applet is already expected to disappear so nothing for the user to see unless it invokes the dialog for reloading or deleting a crashed applet. I build in-process so not sure if that happens or not.

muktupavels:
Still system could detect that it crashed and ask to submit crash report...

lukefromdc:
I don't have Apport installed, and I don't know if any distro is shipping mate-panel 1.21 so its possible this has never been run with automated crash reporting

braikar:
Hmm that's an important detail I forgot when adding gsettings.
Shouldn't `g_clear_object (applet->priv->settings);` be added after the disconnect to clear the object and free the memory properly?

muktupavels:
Signals are automatically disconnected when object is destroyed. In this case explicit `g_signal_handlers_disconnect_by_data` is not needed.

raveit65:
Crash function is
```
0x00007fffdc888c18 in na_grid_set_min_icon_size (grid=0x0, min_icon_size=24)
    at na-grid.c:186
186	  grid->min_icon_size = min_icon_size;

<cut>

Thread 1 (Thread 0x7ffff7fb9a80 (LWP 7092)):
#0  0x00007fffdc888c18 in na_grid_set_min_icon_size (grid=0x0, min_icon_size=24) at na-grid.c:186
        __func__ = "na_grid_set_min_icon_size"
#1  0x00007ffff4f42c8d in g_closure_invoke () at /lib64/libgobject-2.0.so.0
#2  0x00007ffff4f561e3 in signal_emit_unlocked_R ()
    at /lib64/libgobject-2.0.so.0
#3  0x00007ffff4f5f30a in g_signal_emit_valist () at /lib64/libgobject-2.0.so.0
#4  0x00007ffff4f5f903 in g_signal_emit () at /lib64/libgobject-2.0.so.0
#5  0x00007ffff529b910 in g_settings_real_change_event ()
    at /lib64/libgio-2.0.so.0
#6  0x00007fffeefa503e in ffi_call_unix64 () at /lib64/libffi.so.6
#7  0x00007fffeefa49ff in ffi_call () at /lib64/libffi.so.6
#8  0x00007ffff4f43c56 in g_cclosure_marshal_generic_va ()
    at /lib64/libgobject-2.0.so.0
#9  0x00007ffff4f42ee6 in _g_closure_invoke_va () at /lib64/libgobject-2.0.so.0
#10 0x00007ffff4f5ed84 in g_signal_emit_valist () at /lib64/libgobject-2.0.so.0
#11 0x00007ffff4f5f903 in g_signal_emit () at /lib64/libgobject-2.0.so.0
---Type <return> to continue, or q <return> to quit---
#12 0x00007ffff529c0cc in settings_backend_path_changed ()
    at /lib64/libgio-2.0.so.0
#13 0x00007ffff5296a43 in g_settings_backend_invoke_closure ()
    at /lib64/libgio-2.0.so.0
#14 0x00007ffff4c6425b in g_idle_dispatch () at /lib64/libglib-2.0.so.0
#15 0x00007ffff4c6791d in g_main_context_dispatch () at /lib64/libglib-2.0.so.0
#16 0x00007ffff4c67ce8 in g_main_context_iterate.isra ()
    at /lib64/libglib-2.0.so.0
#17 0x00007ffff4c68012 in g_main_loop_run () at /lib64/libglib-2.0.so.0
#18 0x00007ffff6e62f6d in gtk_main () at /lib64/libgtk-3.so.0
#19 0x0000555555579a0d in main (argc=<optimized out>, argv=<optimized out>)
    at main.c:199
        desktopfile = <optimized out>
        context = <optimized out>
        error = 0x0
        display = <optimized out>
        screen = 0x555555826020
        css = 0x555555c9a740
        provider = 0x555555c9a740
        resource = 0x5555555c7648 "/org/mate/panel/theme/mate-panel.css"
        priority = 1
(gdb) 
```
full stacktrace at https://www.dropbox.com/s/md2j2728v1eeply/panel-stack-trace?dl=0

braikar:
I misplaced a line.
https://github.com/mate-desktop/mate-panel/blob/c7d431d36f8bf6e9b94bab37c19dd51041e117c9/applets/notification_area/na-grid.c#L179-L187

I think
`g_return_if_fail (NA_IS_GRID (grid));`
should be first and that's it.
These checks are there to prevent these calls when the Objects are not properly initialized.

lukefromdc:
Please submit an alternate PR. I do NOT know where the memory leak you decribe is and you have not really pointed me to it. I would throw up my hands and close this if this was gnome-panel.

lukefromdc:
If g_return_if_fail can be optimized out, I can try a direct conditional check for the grid. A check for the applet though did NOT work, as it is not all the way gone yet when the crashing function runs
If the signal is normally disconnected anyway than this is getting run after something it needs is fone but before the signal handler has been automatically disconnected.

lukefromdc:
Using
` g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);`
was enough to stop the crash by itself

lukefromdc:
Had to squash both commits in order to fix the commit message to reflect the changed approach to this. Thanks @braikar for catching the source of the problems

braikar:
You're welcome, these were things I had to deal with too when changing the applet. I still haven't figured out how the code works exactly... the GtkGrid is created with `g_object_new (NA_TYPE_GRID...)` and that's something I really don't understand, because it is added to the applet while still not being realized, so the first calls when gsettings are being loaded plain crash the applet if the checks `g_return_if_fail (NA_IS_GRID (grid));` are not there in na_gric.c.. because the NA_GRID object is not ready yet. Then the code in system-tray/ populates the grid and after that the grid is properly initialized.

On applet dispose your guess is what I thought as well, that unloading triggers a size_allocate signal forwarded to the grid that doesn't exist anymore. But also `g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);` is needed, because after that comes `G_OBJECT_CLASS (na_tray_applet_parent_class)->dispose (object);` which will clear the whole object, but if g_clear_object on gsettings wasn't done, then it would also mean that it would try to clear a pointer to a memory space not allocated anymore, thus a memory leak.

Anyway, thanks for fixing it.

raveit65:
Are you sure?
Using this commit on top of 1.20 branch gives me this build error.
```
main.c: In function 'na_tray_applet_dispose':
main.c:211:49: error: 'NaTrayAppletPrivate' {aka 'struct _NaTrayAppletPrivate'} has no member named 'settings'
   g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);
                                                 ^~
/usr/include/glib-2.0/glib/gmacros.h:241:104: note: in definition of macro 'G_STATIC_ASSERT'
 #define G_STATIC_ASSERT(expr) typedef char G_PASTE (_GStaticAssertCompileTimeAssertion_, __COUNTER__)[(expr) ? 1 : -1] G_GNUC_UNUSED
                                                                                                        ^~~~
/usr/include/glib-2.0/gobject/gobject.h:672:36: note: in expansion of macro 'g_clear_pointer'
 #define g_clear_object(object_ptr) g_clear_pointer ((object_ptr), g_object_unref)
                                    ^~~~~~~~~~~~~~~
main.c:211:3: note: in expansion of macro 'g_clear_object'
   g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);
   ^~~~~~~~~~~~~~
In file included from /usr/include/glib-2.0/glib/glist.h:32,
                 from /usr/include/glib-2.0/glib/ghash.h:33,
                 from /usr/include/glib-2.0/glib.h:50,
                 from ./../../libmate-panel-applet/mate-panel-applet.h:28,
                 from main.c:27:
main.c:211:49: error: 'NaTrayAppletPrivate' {aka 'struct _NaTrayAppletPrivate'} has no member named 'settings'
   g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);
                                                 ^~
/usr/include/glib-2.0/glib/gmem.h:122:24: note: in definition of macro 'g_clear_pointer'
     _pp.in = (char *) (pp);                                                    \
                        ^~
main.c:211:3: note: in expansion of macro 'g_clear_object'
   g_clear_object (&NA_TRAY_APPLET (object)->priv->settings);
   ^~~~~~~~~~~~~~
make[5]: *** [Makefile:803: notification_area_applet-main.o] Error 1
```
We don't use all commits from https://github.com/mate-desktop/mate-panel/pull/856
for 1.20 branch.

lukefromdc:
NOT sure, but someone will need to test 1.20 and see if the tray crashes when it is removed. The offending code is the code adding the gsettings preference to allow changing the minimum icon size. If that is not in the commits used, than the crash should not occur and this should not be possible to backport nor necessary

braikar:
agreed, if there's not gsettings the patch does not need to be applied. But `g_return_if_fail (NA_IS_GRID (grid));` in:
https://github.com/mate-desktop/mate-panel/blob/c7d431d36f8bf6e9b94bab37c19dd51041e117c9/applets/notification_area/na-grid.c#L179-L187
is still relevant to change, that's still a potential way for the applet to crash, I knew it had to be put first, I somehow forgot it.
It's a bit redundant, but you'd need to cut this actual patch in two separate patches to apply the one line change on gsettings separately from the change of position of `g_return_if_fail (NA_IS_GRID (grid));`..

muktupavels:
Are anyone of you reading documentation? `g_return_if_fail` is not some magic thing that should be used to prevent crashes...

https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-if-fail

> If expr evaluates to FALSE, the current function should be considered to have **undefined** behaviour (**a programmer error**). The only **correct** solution to such an error is to change the module that is calling the current function, so that it **avoids** this incorrect call.

raveit65:
> Are anyone of you reading documentation? g_return_if_fail is not some magic thing that should be used to prevent crashes...

Before i merge it yesterday i ask you yesterday at irc channel if this fix is correct. And you said `yes`
And now you blame us again......

And yes, i read docs but sadly i am not so smart as you , so i can't fix it in a way that you're satisfied.
So please provide an alternate fix/PR.

But blaming us all the time that we should read docs doesn't make any sense.
Keep in mind that we are all adults and nobody wants to be treated as a student all the time.

Thank you

braikar:
I didn't start with that!

It was already there before, I understand it's not the proper way to use it and that such checks should not be there in the first place, but the way the applet is structured, apart from changing a lot of things, for the moment that's the simplest way..

https://github.com/mate-desktop/mate-panel/blob/f5cf97810b4ffd0aec43aafc60c20bb00a2dc96a/applets/notification_area/na-box.c#L100-L131
These were there before version 1.20 and as I explained above it is because of how the applet loads, the gtkgrid (or box before) is created using g_object_new(..), so when:

https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/main.c#L306-L331

runs, and as I understands it, normally the applet gtk widget and all its child (the gtkgrid..) should exist and be working and usable.. Because of that weird way of starting up the applet, it isn't, so these checks are necessary to follow how the applet is working at the moment without changing a whole lot about how it is structured.
Same as luke: "i am not so smart as you , so i can't fix it in a way that you're satisfied." You're welcome to do it so we can see how to do it best next time :)
I don't have time to understand how it works in details,  my main domain of coding is not gtk or operating systems UIs in general, as far as I understand it, changing that (I fully understand it's not 100% best practice) would required a lot of changes in the whole way of many many functions and the order in which they all happen.

raveit65:
> ....you're right, there's no need to backport anything, since there are no gsettings, there's no misplaced g_return_if_fail (NA_IS_GRID (grid)); either, that function does not exist without gsettings..

I just tested current 1.20 branch building na-tray-applet in-process.
I doesn't crash the panel when removing \o/
Happily, i can release 1.20.5 now.

muktupavels:
@raveit65 it was response to `is still relevant to change, that's still a potential way for the applet to crash`...

It is used to catch programmer errors. For example when programmer calls xxx function with GtkBox as parameter, but functions expects GtkGrid. For end user that part maybe be optimized out so it would not even matter where it is placed... So if function is called in wrong way it would crash anyway!

@braikar what?
> Because of that weird way of starting up the applet, it isn't, so these checks are necessary to follow how the applet is working at the moment without changing a whole lot about how it is structured.

Can you explain that more detailed? What is not working and why does `g_return_*` "fixes" it? Or did I misunderstood something? If that "fixes" something for you then I guess you get also critical warnings, no?

P.S. Those `g_return_if_fail` lines in linked signal callbacks was not in original code... No idea why it was added.

braikar:
@muktupavels as I'm no expert I cannot say exactly.
What I seem to understand is:
https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/na-grid.c#L331-L334
here the hosts are initialized, then these will populate the grid with the icons whenever they signal back anything to display their icon in the tray.

And when they reply, this function is called
https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/na-grid.c#L447-L453
when `na_tray_applet_change_background(...)` is triggered in the main applet by some cairo redraws for the background of the icons.

Before any of these forced redraws occur, the gtkgrid does not seem to be properly realized yet and thus ,calls to do something with the grid or the `_NaGrid` struct, will fail.
Since the hosts replies are asynchronous, the grid becomes usable whenever one icon is finally added to the grid. Some of the initial calls to update the grid can happen before a host has replied to display the first icon.

https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/main.c#L471
happens, then
https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/main.c#L307
happens.
so
https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/main.c#L323-L326
shouldn't create an error when already changing some of the grid properties, but it does because (how I understand it) no event/signal might have yet forced the grid to be ready to be used and it seems its these asynchronous forced_redraws that trigger the grid to be initialized once and for all.

At least that's how I understand it, it's the way of how the hosts populate the grid asynchronously that cause me to not understand how it works precisely.. I never encountered an applet that worked this way before.

muktupavels:
NaGrid object is ready once `na_grid_new` returns. Realization happens when you call `gtk_widget_show (GTK_WIDGET (applet->priv->grid));`...

Could you provide backtrace with crash? It might be easier to understand what is going on...
gdb mate-panel
run --replace
bt full

braikar:
I thought a bit about it, these errors I had were all at the beginning. (I'm unable to provoke a crash with the actual code)
I remember it was the `refresh_grid(...)` that often crashed the applet, but you helped me figure out all the errors in there (and clean up/simplify the code a whole lot). Around the end, when I was just trying to rework the commits to make them right, I wasn't able to have any error/crash anymore.

I just tried removing all the `g_return_if_fail` that were there even before I made some changes, it seems to be working fine. I'm running it with those changes applied (no `g_return_if_fail` calls at all), I'll let you know if something occurs.
So you're right, I guess https://github.com/mate-desktop/mate-panel/blob/f5cf97810b4ffd0aec43aafc60c20bb00a2dc96a/applets/notification_area/na-box.c#L100-L131
was there because something wasn't working right before too.
Apparently they might be leftovers that can all be removed.

Btw.
This makes me think of: https://github.com/mate-desktop/mate-panel/blob/371e3d7cb45ed908ce2f04d9972129c69ff156eb/applets/notification_area/main.c#L192

I actually put that there because I copied code from another applet, then that code is also useless?

muktupavels:
@raveit65 I did not ask for one you could get from this bug... The problem here was missing unref and this particular bug is fixed. I was asking for different thing and it seems that @braikar did understand what I wanted. :)

@braikar `g_return_if_fail` are not useless and I would say that better to not remove - they maybe be helpful in future. All I am trying to say is that these things should not be used as instrument to prevent crashes and/or workaround real problems. :)

Your last one linked line seems useless to me. If you would return there you would get dialog that is partially setup and I doubt that it is what you want. Assert might be more appropriate in that case as only case when it would return NULL is when ui file does not have widget/object with requested name.

PR #924, opened 2019-02-05
Enable Travis CI support

yetist:
It's ready to go.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,155 @@
+# vim: set ts=2 sts=2 sw=2 expandtab :
+dist: xenial
+sudo: required
+language: bash
+services:
+  - docker
+
+# Use travis branch for test.
+#branches:
+#  only:
+#  - travis
+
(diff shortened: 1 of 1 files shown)


raveit65:
Ok, runs fine with latest changes from https://github.com/mate-desktop/mate-dev-scripts/pull/10
But why does distcheck not run here?
https://travis-ci.org/mate-desktop/mate-panel/jobs/488999917

yetist:
run with `make distcheck >/dev/null`, start at line 4232, and return 0 at 4266 line.

raveit65:
That means it was running but it isn't printed?

yetist:
`make distcheck` only failed on debian system and succeeded on other distro.
logs on debian: 
https://travis-ci.org/mate-desktop/mate-panel/jobs/488999916

raveit65:
Ok, than the `return 0` means it was running successful.
I missed the full output. But OK.

PR #927, opened 2019-02-07
use archlinux/base image

yetist:
Use archlinux/base image should be ok.


raveit65:
I squashed both commits.

PR #930, opened 2019-02-11
Don't build X11 files unless X11 is enabled

wmww:
Actual Wayland support still isn't here, this is just to get things organized in preparation. I also dropped a few seemingly unneeded X11 includes.


wmww:
Bump. This is just build system changes, so it should be easy to verify it doesn't do anything nasty.

PR #931, opened 2019-02-11
Wayland support for the easy files

wmww:
This has a lot of the files that just need minor changes to support Wayland. It doesn't remove all X-specific code from the project and there is still no Wayland backend in place, so this should only be tested under X. If reviewers would like me to break this up into smaller pieces, just ask. Most of the commits are independent.

### DIFF
--- a/libmate-panel-applet/panel-plug-private.h
+++ b/libmate-panel-applet/panel-plug-private.h
@@ -20,6 +20,12 @@
 #ifndef PANEL_PLUG_PRIVATE_H
 #define PANEL_PLUG_PRIVATE_H
 
+#ifdef PACKAGE_NAME // only check HAVE_X11 if config.h has been included
+#ifndef HAVE_X11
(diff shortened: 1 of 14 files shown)


wmww:
Nice, thanks.

This PR doesn't touch `mate-panel-applet-gsettings.h`, or any of the 6 files that include it. My previous PRs have touched some of the including files. FWIW I'm not getting those warnings on Ubuntu 18.10.

raveit65:
Looks good in quick overview, but we're are going in freeze for 1.22 release today.
And i don't want to merge such a big change on the last day to avoid breaking the 1.22 release.
I hope 1.22 is released end of February and we can work again on it.
Thank you for your comprehension.

wmww:
Bump. From what I can tell, the release has happened.

raveit65:
> FWIW I'm not getting those warnings on Ubuntu 18.10.

You don't see that warning if gobject-introspection isn#t enabled.

PR #936, opened 2019-02-16
na-tray-child: avoid deprecated 'gtk_widget_set_double_buffered'

sc0w:
please, don't push to the 1.22 release

### DIFF
--- a/applets/notification_area/system-tray/na-tray-child.c
+++ b/applets/notification_area/system-tray/na-tray-child.c
@@ -101,13 +101,6 @@ na_tray_child_realize (GtkWidget *widget)
 
   gtk_widget_set_app_paintable (GTK_WIDGET (child),
                                 child->parent_relative_bg || child->has_alpha);
-
-  /* Double-buffering will interfere with the parent-relative-background fake
-   * transparency, since the double-buffer code doesn't know how to fill in the
-   * background of the double-buffer correctly.
-   */
-  gtk_widget_set_double_buffered (GTK_WIDGET (child),
-                                  child->parent_relative_bg);
 }
 
 static void


muktupavels:
That function was needed last time I looked at this... If you have done some research and found that this is no longer needed then, please, share this info.

raveit65:
status?

raveit65:
No idea, if this can be droped or not. I will start testing.

raveit65:
I like to see travis working :)

sc0w:
I am not sure how to test

It needs to test with `child->parent_relative_bg` `TRUE` and `FALSE`

or is always `TRUE`?

lukefromdc:
This is conjecture on my part, but one check might be to make sure there are no surprises with or without compositing enabled, as that is the biggest difference in operating conditions the tray will get.

lukefromdc:
If there is a corner-case this creates an issue with, we should leave gtk_widget_set_double_buffered' in place, as it will exist in GTK as long as GtkStatusIcon does (that is all GTK3 versions but no GTK4 versions). That said, either or both could get broken by future commits to Gtk3 that are not checked for behavior of deprecated widgets and functions

sc0w:
If this PR have collateral effects, please show screenshots or videos and I will try to fix it.

vkareh:
I ran with this patch most of my day today and could not notice any difference... I honestly think this is fine to merge, and if anyone ever detects an issue, we can revisit when that happens.

raveit65:
For me the (fake) transparency looks OK in non compositor case and with short a test.
Normally, i don't use this.
Bildschirmfoto zu 2019-03-22 10-32-11
It seems that the tray-applet and new volume-icon have a little different color than the main panel background.
I think the applet sand the panel wrote here a double background.
But this i see this effect without this PR too, so it isn't related.

PR #939, opened 2019-02-22
Change panel_toplevel_get_monitor_geometry signature

wmww:
Instead of filling in 4 `int`s, it now fills in a `GdkRectangle`. This is more readable and less error prone. (I'm working with this code from Wayland support, which is why it matters to me)


wmww:
@vkareh I agree, done

wmww:
This has had two approvals and no objections for 9 days. Time for a bump.

PR #942, opened 2019-02-28
[Security] panel-keyfile: Use 'g_strlcpy' instead of 'strcpy'

sc0w:
Fixes Clang static analyzer warning:

warning: Call to function 'strcpy' is insecure as it does not provide bounding of the memory buffer. Replace unbounded copy functions with analogous functions that support length arguments such as 'strlcpy'. CWE-119

### DIFF
--- a/mate-panel/libpanel-util/panel-keyfile.c
+++ b/mate-panel/libpanel-util/panel-keyfile.c
@@ -126,11 +126,10 @@ panel_key_file_to_file (GKeyFile     *keyfile,
 		gsize  new_length;
 
 		new_length = length + strlen (KEYFILE_TRUSTED_SHEBANG);
-		new_data = g_malloc (new_length);
+		new_data = g_malloc (new_length + 1);
 
-		strcpy (new_data, KEYFILE_TRUSTED_SHEBANG);
-		memcpy (new_data + strlen (KEYFILE_TRUSTED_SHEBANG),
-			data, length);
+		g_strlcpy (new_data, KEYFILE_TRUSTED_SHEBANG, new_length + 1);
+		g_strlcat (new_data, data, new_length + 1);
 
 		g_free (data);
 		data = new_data;


lukefromdc:
I notice in the clang tools report, the same issue gets reported twice, but is the same line of code in both reported bugs and is fixed by this PR. In my opinion, this is important enough to get into 1.22.0 before final release or if necesary an immediate 1.22.1 point release as this sort of thing could be exploitable and we've just published it.

mate-panel/libpanel-util/panel-keyfile.c
```diff
@@ -128,7 +128,7 @@ panel_key_file_to_file (GKeyFile     *keyfile,
 		new_length = length + strlen (KEYFILE_TRUSTED_SHEBANG);
 		new_data = g_malloc (new_length);
 
-		strcpy (new_data, KEYFILE_TRUSTED_SHEBANG);
+		g_strlcpy (new_data, KEYFILE_TRUSTED_SHEBANG, (new_length + 1));
```
muktupavels: You allocate `new_length` bytes and pass `new_length + 1` as `dest_size`...

rbuj:
```patch
diff --git a/mate-panel/libpanel-util/panel-keyfile.c b/mate-panel/libpanel-util/panel-keyfile.c
index ebd3249..5cf4996 100644
--- a/mate-panel/libpanel-util/panel-keyfile.c
+++ b/mate-panel/libpanel-util/panel-keyfile.c
@@ -126,11 +126,10 @@ panel_key_file_to_file (GKeyFile     *keyfile,
                gsize  new_length;
 
                new_length = length + strlen (KEYFILE_TRUSTED_SHEBANG);
-               new_data = g_malloc (new_length);
+               new_data = g_malloc (new_length + 1);
 
-               strcpy (new_data, KEYFILE_TRUSTED_SHEBANG);
-               memcpy (new_data + strlen (KEYFILE_TRUSTED_SHEBANG),
-                       data, length);
+               g_strlcpy (new_data, KEYFILE_TRUSTED_SHEBANG, new_length + 1);
+               g_strlcat (new_data, data, new_length + 1);
 
                g_free (data);
                data = new_data;
```

muktupavels: `g_strdup_printf ("%s%s", KEYFILE_TRUSTED_SHEBANG, data);`

Or even better it has "//FIXME: kill this when bug #309224 is fixed" and that bug has been fixed:
https://bugzilla.gnome.org/show_bug.cgi?id=309224

lukefromdc: That report shows the bug was fixed back in 2013, meaning it is very unlikely any distro would ship a glib version predating the fix and also ship MATE 1.20 or 1.22. Thus we should be able to to (as the FIXME states) remove the entire function.

sc0w: @rbuj ok, done!

@lukefromdc please test again

lukefromdc:
I will do that tomorrow, on the road at the moment

PR #946, opened 2019-03-23
Create empty panel on first launch if no default layout is found

wmww:
This allows the user to add panels and applets if no `default.layout` is found

### DIFF
--- a/mate-panel/panel-layout.c
+++ b/mate-panel/panel-layout.c
@@ -161,9 +161,13 @@ panel_layout_append_group_helper (GKeyFile                  *keyfile,
         dir = "toplevels";
         type = PANEL_GSETTINGS_TOPLEVELS;
     }
-    if (g_strcmp0 (id_list_key, PANEL_OBJECT_ID_LIST_KEY) == 0) {
+    else if (g_strcmp0 (id_list_key, PANEL_OBJECT_ID_LIST_KEY) == 0) {
         dir = "objects";
         type = PANEL_GSETTINGS_OBJECTS;
+    }
+    else {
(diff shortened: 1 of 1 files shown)


raveit65:
This looks good, but when is this the case?
I mean if no default or any other layout is found than installation of mate-panel is incomplete and it is possible that other files are missing too.
Or do i miss something i case of running under wayland?

raveit65:
There is a build warning about a variable which you're use in this new code block.
https://travis-ci.org/mate-desktop/mate-panel/jobs/510363118#L2395
```
panel-layout.c:186:21: warning: 1st function call argument is an uninitialized value
        unique_id = panel_profile_find_new_id (type);
1 warning generated.
```
But i don't get where this is introduce.
I see it the first time in travis logs from this https://github.com/mate-desktop/mate-panel/commit/308d0d31734c59650af5cd4f5542df4b409c672b
but not here
https://github.com/mate-desktop/mate-panel/commit/59a633d981b39cecd530ef00ed9cada69bf809e5
And i don't see in local build log of my fedora box.
:confused:

wmww:
> I mean if no default or any other layout is found than installation of mate-panel is incomplete and it is possible that other files are missing too.

Correct, but it can't hurt to make the panel usable despite missing this file. The `mate-wayland` snap does not install anything yet, but with this fix the panel still appears and works somewhat.

I can't replicate that build warning on my machine either, but the potential problem is obvious from reading the code. I fixed it, and hopefully the build server's compiler is smart enough to stop the warning.

raveit65:
Thanks for fixing the build warning, i don't see them any more in travis logs.

PR #960, opened 2019-05-07
Set compressed="true" for gresource text files

rbuj:
shrink the size of binary files:
```shell
$ ./autogen.sh --prefix=/usr && make && LANG=C ls -la applets/clock/.libs/clock-applet
-rwxrwxr-x. 1 robert robert 748592 May  8 11:47 applets/clock/.libs/clock-applet
$ make distclean
$ git checkout dd161527f822ab152ee97c4ff8f4e111da10d27f
$ ./autogen.sh --prefix=/usr && make && LANG=C ls -la applets/clock/.libs/clock-applet
-rwxrwxr-x. 1 robert robert 818232 May  8 12:25 applets/clock/.libs/clock-applet
```

### DIFF
--- a/applets/clock/clock.gresource.xml
+++ b/applets/clock/clock.gresource.xml
@@ -3,12 +3,12 @@
   <gresource prefix="/org/mate/panel/applet/clock">
     <file compressed="true">clock.ui</file>
     <file compressed="true">clock-menu.xml</file>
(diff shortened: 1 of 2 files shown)


raveit65:
@rbuj 
Again, Doesn't compress the icons not slow down to process them?

lukefromdc:
My guess is it would speed up reading them from disk by a greater amount than slowing down processing in CPU, though I have no way to timer-test the pair of events.

Issue #436, opened 2016-05-19
Panels disappearing after "mate-panel --reset"


seamus-45:
| Package | Version |
| --- | --- |
| OS | Ubuntu Xenial 16.04 |
| MATE | 1.12 |
| GTK | 2.24.30 |
| GLIB2 | 2.48.0 |
### Steps to reproduce

1) launch `mate-panel --reset`
### Result

Layout is resetted, but panels is disappear while still in processes, until `mate-panel --replace` do not runned.
### Expected result

Panels should be reloaded with new layout.
##

flexiondotorg:
You should run `mate-panel --reset --replace`

raveit65:
nice trick, i never known that :)

seamus-45:
I have tried --replace command line switch, but still get same behavior :(

pellcorp:
I actually habe same issue.  I can switch layouts which use same menu bar but switching to a layout which uses the new style menu results in all panels disappearing.  I can switch back to an old style to recover

Tweak works fine.  So what else is tweak calling other than mate-panel --reset --replace --layout somelayout

pellcorp:
It seems to work randomly.  A restart does however apply the change

pellcorp:
Run the same command twice works and it works successfully second time.  Seems consistent.  But it only seems to work when you have --reset --layout and --replace on both commands

pellcorp:
OK so I narrowed if down.

You can call --reset --layout somelayout

And theb call --reset by itself second time and the panel changes apply perfectly every time

monsta:
I can't reproduce it with 1.18.

monsta:
So I was able to reproduce it in a few Ubuntu MATE systems which have GTK+ 3.22.

We have two problems here.
First, some callbacks connected to GSettings might still be invoked while the panel is being deleted. This causes random crashes on `mate-panel --reset`.
Second, if that crash doesn't happen, and all the panels are deleted, we have the next problem. Reloading the default layout after the panel reset might start too early, while the previous layout isn't completely deleted yet. In this case the reloading function doesn't do its work, so the new panels don't appear.

monsta:
[...]

Issue #65, opened 2012-12-18
mate-panel 1.5.x regression: deleting a panel crashes mate-panel


dmashal:
jasmineaura was able to reproduce. 

Right click -> new panel -> delete panel -> mate -panel restarts.

PR #968, opened 2019-05-13
fix panel reset problems

monsta:
Fixes https://github.com/mate-desktop/mate-panel/issues/436

How to test:

1. Resetting the panel layout

- use `mate-panel --reset` to reset the panel(s) to the current default layout
- use `mate-panel --reset --layout=<layout_name>` to set the new default layout and reset the panel(s) to it
- the panel should not crash
- the panel should reappear with the default layout

2. Deleting one panel in a multi-panel layout

- use "Delete this panel" menu item from the right-click menu to delete that panel
- the panel should not crash (https://github.com/mate-desktop/mate-panel/issues/65 should not happen again)

Both problems I've described in https://github.com/mate-desktop/mate-panel/issues/436#issuecomment-491845638 can be more often reproduced in Ubuntu than in Debian Testing, for example. That's the problem of race condition issues...

### DIFF
--- a/mate-panel/panel-profile.c
+++ b/mate-panel/panel-profile.c
@@ -1435,6 +1435,18 @@ panel_profile_delete_removed_ids (PanelGSettingsKeyType    type,
 	g_slist_free (removed_ids);
 }
 
+static gboolean
+load_default_layout_idle (gpointer unused) {
+	if (g_slist_length (panel_toplevel_list_toplevels ()) != 0) {
+		/* some toplevels are not destroyed yet, waiting */
+		return TRUE;
+	}
+
+	/* load the default layout and stop this handler */
+	panel_profile_ensure_toplevel_per_screen ();
+	return FALSE;
+}
+
 static void
 panel_profile_toplevel_id_list_notify (GSettings *settings,
 									   gchar *key,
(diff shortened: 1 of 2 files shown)


lukefromdc:
This is something I cannot test except when I am doing nothing else due to my bare-metal only setup and heavily customized panel layout that can take 1/2 hour to redo

flexiondotorg:
@lukefromdc You can save/restore your panel layouts using MATE Tweak, it's in the Debian archive as `mate-tweak` or available here https://github.com/ubuntu-mate/mate-tweak

@monsta I'm testing this and will report back later.

lukefromdc:
Mate-tweak mostly works, but I got a ton of python errors, lost all custom launchers except GIMP, and the configuration of both the clock and system monitor applets. Will test this now before restoring

lukefromdc:
Just duplicated the crash so I have a baseline

lukefromdc:
The mate-tweak stuff has one benefit despite having to rebuild my launcher layout:greaty cleaned up my dconf database for mate-panel. This came from the same issue that trashed my launchers: not saving numbered objects that lack a unique name (launchers do have names but the panel does not use them it seems)..

monsta:
The test for deleting the panel is just to make sure there's no regression, and #65 won't appear again. That's because I've removed the code that fixed #65 and replaced it with similar code in toplevel's dispose function.

lukefromdc:
I can't reproduce a crash on panel removal now with or without this applied, so good on this end

lukefromdc:
Is this ready to go or are we still waiting on @vkareh ?

monsta:
Thanks, merged. Will pick this to 1.22 and 1.20 too.

PR #973, opened 2019-05-20
travis: use f28 with old libmatemenu

raveit65:
with a docker-image with MATE 1.20

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,7 @@ script:
 env:
   - DISTRO="archlinux/base"
   - DISTRO="debian:sid"
-  - DISTRO="fedora:29"
+  - DISTRO="fedora:28"
   - DISTRO="ubuntu:18.10"
 
 ##########################################################
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ script:
   - ./docker-build --name ${DISTRO} --verbose --config .travis.yml --build autotools
 
 env:
-  - DISTRO="archlinux/base"
+#  - DISTRO="archlinux/base"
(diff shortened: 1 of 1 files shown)


raveit65:
@yetist 
Any idea what we should do with build for archlinux after arch use 1.22 docker-image?
Or should we simply set building disable for 1.20 branch?
I mean archlinux don't need any 1.20 package any more as rolling release, or not?

yetist:
> I mean archlinux don't need any 1.20 package any more as rolling release, or not?

Yes, archlinux don't need any 1.20 package any more, now is 1.22.

yetist:
There is a way to use archlinux on travis: in the `before_scripts` of archlinux, download the mate-menus-1.20 tarball and install it, But this is not worth it, archlinux does not need 1.20.

A similar abi change problem, there will be a better solution after moving to meson in the future, meson supports subproject, as used here: https://gitlab.gnome.org/GNOME/gtk/tree/master/subprojects

raveit65:
So, do you agree with disable archlinux for 1.20 branches with ABI changes?

yetist:
Yes, I agree.

monsta:
We can also switch Debian from Sid to Buster for 1.20. Debian Sid (and Testing) will get 1.22 once Buster is released as the next Debian Stable this year. This might happen even this summer, given that Buster is in freeze for almost a half of year already, and Debian releases usually happen mid-year.

raveit65:
Feel free to test and commit this.

monsta:
Created https://github.com/mate-desktop/mate-panel/pull/975 for that, let's see how it builds.

PR #975, opened 2019-05-24
travis: use Debian Buster which has 1.20

monsta:


raveit65:
I hope you don't ask for a review for every repo :)
Only build requires might be adjusted when build failed.

sc0w:
The PR is ok for me, and I think it can be done the same in all the repositories for 1.20 to avoid problems in the future

PR #978, opened 2019-06-09
Update FSF address

tamplan:
Update [FSF address](https://www.fsf.org/about/contact/)


tamplan:
Learning git needed...

PR #982, opened 2019-06-17
Fix Makefile.am to install in a given prefix path

jumapico:
Currently installing mate-panel fails with the error

     /bin/mkdir -p '/usr/share/gir-1.0'
     /usr/bin/install -c -m 644 MatePanelApplet-4.0.gir '/usr/share/gir-1.0'
    /usr/bin/install: cannot create regular file '/usr/share/gir-1.0/MatePanelApplet-4.0.gir': Permission denied

when configure is called with the --prefix option.

This fix is the same used in:

* caja (libcaja-extension/Makefile.am)
* eom (src/Makefile.am)
* libmatekbd (libmatekbd/Makefile.am)
* mate-desktop (libmate-desktop/Makefile.am)
* pluma (pluma/Makefile.am)

### DIFF
--- a/libmate-panel-applet/Makefile.am
+++ b/libmate-panel-applet/Makefile.am
@@ -133,10 +133,10 @@ MatePanelApplet_4_0_gir_FILES = $(addprefix $(srcdir)/,$(introspection_sources))
 MatePanelApplet_4_0_gir_SCANNERFLAGS = --strip-prefix=MatePanel --pkg-export=libmatepanelapplet-4.0
 INTROSPECTION_GIRS += MatePanelApplet-4.0.gir
 
-girdir = $(INTROSPECTION_GIRDIR)
+girdir = $(datadir)/gir-1.0
 gir_DATA = $(INTROSPECTION_GIRS)
 
-typelibdir = $(INTROSPECTION_TYPELIBDIR)
+typelibdir = $(libdir)/girepository-1.0
 typelib_DATA = $(INTROSPECTION_GIRS:.gir=.typelib)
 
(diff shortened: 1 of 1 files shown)


raveit65:
`sudo make install` should work.

jumapico:
Yes, but if the --prefix option works fine sudo should not be necessary.
Also, is a plus if we can avoid pollute the system directories during development.

lukefromdc:
I have always run make install as sudo, also make uninstall to sweep back out afterwards unless all the files will be overwritten by my locally made .deb files which use the same options I test with. I would not expect make install to work without root permissions, barring of course doing something like prefix=/home/luke/<something>

muktupavels:
He is **not** building with `--prefix=/usr` and/or installing in `/usr` prefix... `sudo` is not answer here!

jumapico:
My case, and the motive behind the pull request.
I want to work in mate-power-manager.  For that I have created a [script for build a project and his dependencies](https://github.com/jumapico/mate-dev-scripts/blob/add-build-script/build/build.mk) (i want to make a pull request later after add all the projects) that build all the projects with the line: `./autogen.sh --prefix="/home/jmpc/workspace-mate/PREFIX" && make && make install`.
Using the previous line I get the error described in the commit message.
Again, maybe I'm wrong, but I want to install all of the project files in the given prefix inside my home directory without using root permission (through sudo).

lukefromdc:
Point is I did not TEST an install into a user-writeable directory with my usual build options, rather just tested to ensure it did not break a conventional build installing into /usr

raveit65:
@monsta @sc0w 
Does this works for you in debian/ubuntu?

PR #984, opened 2019-06-17
Switch out wayland protocols for gtk-layer-shell dependency

wmww:
I originally added a Wayland protocols directory as a step toward Wayland support. I have since built [GTK Layer Shell](https://github.com/wmww/gtk-layer-shell) to abstract the complex Wayland functionality out of any one app, and so this PR removes the protocols and adds an optional dependency on that. GTK Layer Shell currently must be built from source and statically linked. If MATE panel doesn't detect it, it will simply disable Wayland support (unless of course `--enable-wayland` has been explicitly specified, in which case it will error).

Wayland support is still in no sense "working" here. I have a branch with it working (that's what's running in the `mate-wayland` snap), and I'm slowly PRing bits of that.


raveit65:
@wmww 
I just added gtk-layer-shell to travic-CI builds
https://github.com/mate-desktop/mate-panel/commit/4703065becdb10a99dda91db6c8f68cc4486db35
Now all builds use wayland and x11, which should make it easier for you and us to test your amazing wayland work.
Of course they can fail now because of wayland and your gtk-layer-shell :)
You should rebase all PRs to get this commit in your branches.

PR #985, opened 2019-06-18
Allow applets to specify the platform(s) they support (X11/Wayland)

wmww:
* Add a `Platforms` list to the applet metadata files
* Fail to load an applet when on a platform it doesn't support (logs a warning and shows the same error dialog to the user as with other applet errors)
* In the addto window, grey out and make unselectable applets that don't support the current platform

Wayland support still isn't here. To test, I recommend changing the platform value of an applet to be Wayland only, and noting how it is then not loaded or addable on X11.


mate-panel/panel-addto.c
```diff
@@ -681,7 +695,12 @@ static void panel_addto_make_application_model(PanelAddtoDialog* dialog)
 	if (dialog->filter_application_model != NULL)
 		return;
 
-	store = gtk_tree_store_new(NUMBER_COLUMNS, G_TYPE_STRING, G_TYPE_STRING, G_TYPE_POINTER, G_TYPE_STRING);
+	store = gtk_tree_store_new(NUMBER_COLUMNS,
```
raveit65: Any chance of using a space before a bracket, same like in the rest of the file?
I see that at several places, i don't want to mark every occurrence in your PR, please fix.

raveit65:
> I recommend changing the platform value of an applet to be Wayland only,

How can i do that?

raveit65:
Ok, i guess in need to remove X11 from this line to see that the fish won't load under X11?
```
[root@mother rave]# cat /usr/share/mate-panel/applets/org.mate.panel.FishApplet.mate-panel-applet | grep Wayland
Platforms=X11;Wayland;
```

raveit65:
Nice, this works as expected with fish and clock applet under X11 :)
`Platforms=Wayland;`
Both applets are grey-out in add-to-panel dialog.

wmww:
Formatting fixed. This PR adds two string (`Not compatible with Wayland` and `Not compatible with X11`) that should be translated. I wrapped them in `_(...)`, but I don't know what else if anything I ought to do.

raveit65:
After merging we can update transifex resource file (mate-panel.pot) to catch new translation strings, np.

mate-panel/panel-addto.c
raveit65: Looks much better, thanks.

PR #986, opened 2019-06-18
Get panel background ready for Wayland

wmww:
`PanelBackgroundMonitor` pulls the desktop background out of the root window for in-process compositing. This is both impossible and unneeded on Wayland, so I made the files X11-only. Unfortunately the desktop pixmap (`PanelBackground->desktop`) was previously retrieved even when it was not needed, so I had to shuffle some logic around a bit to make that happen only on X11.

### DIFF
--- a/mate-panel/Makefile.am
+++ b/mate-panel/Makefile.am
@@ -43,7 +43,6 @@ panel_sources = \
 	panel-applets-manager.c \
 	panel-shell.c \
 	panel-background.c \
-	panel-background-monitor.c \
 	panel-stock-icons.c \
 	panel-action-button.c \
 	panel-menu-bar.c \
@@ -77,7 +76,8 @@ panel_sources += \
 	xstuff.c \
 	panel-xutils.c \
 	panel-force-quit.c \
-	panel-action-protocol.c
+	panel-action-protocol.c \
+	panel-background-monitor.c
 endif
 
 panel_headers = \
@@ -101,7 +101,6 @@ panel_headers = \
 	panel-applets-manager.h \
 	panel-shell.h \
 	panel-background.h \
-	panel-background-monitor.h \
 	panel-stock-icons.h \
 	panel-action-button.h \
 	panel-menu-bar.h \
@@ -139,7 +138,8 @@ panel_sources += \
 	xstuff.h \
 	panel-xutils.h \
 	panel-force-quit.h \
-	panel-action-protocol.h
+	panel-action-protocol.h \
+	panel-background-monitor.h
 endif
 
 mate_panel_SOURCES = \
--- a/mate-panel/panel-background-monitor.c
+++ b/mate-panel/panel-background-monitor.c
@@ -24,6 +24,12 @@
  *      Mark McLoughlin <mark@skynet.ie>
  */
 
+#include <config.h>
+
+#ifndef HAVE_X11
+#error file should only be compiled when HAVE_X11 is enabled
+#endif
+
 #include <glib.h>
 #include <glib-object.h>
 #include <gdk/gdk.h>
@@ -82,6 +88,7 @@ static guint signals [LAST_SIGNAL] = { 0 };
 
 gboolean gdk_window_check_composited_wm(GdkWindow* window)
 {
+	g_return_val_if_fail (GDK_IS_X11_WINDOW (window), TRUE);
 	return gdk_screen_is_composited(gdk_window_get_screen(window));
 }
 
@@ -182,6 +189,8 @@ panel_background_monitor_new (GdkScreen *screen)
 PanelBackgroundMonitor *
 panel_background_monitor_get_for_screen (GdkScreen *screen)
 {
+	g_return_val_if_fail (GDK_IS_X11_SCREEN (screen), NULL);
+
 	if (!global_background_monitor) {
 		global_background_monitor = panel_background_monitor_new (screen);
 
@@ -360,6 +369,9 @@ panel_background_monitor_get_region (PanelBackgroundMonitor *monitor,
 	GdkPixbuf *pixbuf, *tmpbuf;
 	int        subwidth, subheight;
 	int        subx, suby;
+
+	g_return_val_if_fail (monitor, NULL);
+	g_return_val_if_fail (GDK_IS_X11_WINDOW (monitor->gdkwindow), NULL);
 
 	if (!monitor->gdkpixbuf)
 		panel_background_monitor_setup_pixbuf (monitor);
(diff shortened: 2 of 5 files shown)


wmww:
Either I don't fully understand, or it's already doing that. Compile-time and runtime checks in panel-background.c insure X11-only GDK calls are not made on Wayland, and panel-background-monitor.c is not used on Wayland. My Wayland branch has these commits and their working, so if you've tested on compositing and non-compositing X11, we have everything covered.

Going forward, I would delete panel-background-monitor.c entirely and not attempt to support transparency emulation on non-compositing window managers. It seems to me that there is way too much code and complication introduced than this edge case is worth. Anyway, that's for someone else to decide and PR if they wish, my work here is done.

lukefromdc:
ANY compile-time check means the code is being excluded from some
builds, meaning a build supporting Wayland cannot also support x11. It
would be fine to not build x11 code on a wayland-ONLY build, but this
is not the best way to go for a wayland plus x11 build, which is what
most distros would want to ship

On 6/20/2019 at 4:23 AM, "William Wold"  wrote:

	Either I don't fully understand, or it's already doing that.
Compile-time and runtime checks in panel-background.c insure X11-only
GDK calls are not made on Wayland, and panel-background-monitor.c is
not used on Wayland. My Wayland branch has these commits and their
working, so if you've tested on compositing and non-compositing X11,
we have everything covered. 

	Going forward, I would delete panel-background-monitor.c entirely and
not attempt to support transparency emulation on non-compositing
window managers. It seems to me that there is way too much code and
complication introduced than this edge case is worth. Anyway, that's
for someone else to decide and PR if they wish, my work here is done.

wmww:
> ANY compile-time check means the code is being excluded from some builds

Correct

> meaning a build supporting Wayland cannot also support x11

Incorrect

The system we're using is slightly more sophisticated than you think. We have `HAVE_X11` and `HAVE_WAYLAND` macros which can be turned on and off independently with configure flags:

```bash
./autogen.sh --enable-x11 --disable-wayland  # X11 only
./autogen.sh --disable-x11 --enable-wayland  # Wayland only (currently errors, but will work once I land full Wayland support)
./autogen.sh --enable-x11 --enable-wayland   # X11 and Wayland
./autogen.sh --disable-x11 --disable-wayland # Neither (doesn't make sense, so gives an error)
./autogen.sh                                 # Support for omitted platforms is automatic. Checks installed libraries
```

If a platform is disabled at compile-time, it's libraries aren't linked and its headers aren't included anywhere. There are also files which are only compiled/included on some platforms (for example, `panel-background-monitor.c/.h` will give errors if compiled with X11 turned off, and the build system is set up to keep this from happening.

If both X11 and Wayland are turned on at compile-time, then both libraries are linked, all files are compiled, both sets of ifdefs are enabled and the resulting binary can run on either platform. In that case, runtime checks determine the code path to follow (often these look something like `if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))`). I make sure to put a runtime check directly inside every compile-time check, and frequently compile in all three configurations to make sure everything's working.

lukefromdc:
Thanks, that is exactly what is needed: runtime.checks on builds
supportng both platformswhich I suspect is how most.distros will build
this in the future. Nice bonus to exclude code thatcannot be used
(becomes.dead code) from x11-only or wayland-only builds.

On 6/20/2019 at 2:15 AM, "William Wold"  wrote: 

	ANY compile-time check means the code is being excluded from some
builds  

	Correct  

	meaning a build supporting Wayland cannot also support x11  

	Incorrect 

	The system we're using is slightly more sophisticated than you think.
We have HAVE_X11 and HAVE_WAYLAND macros which can be turned on and
off independently with configure flags: 

	./autogen.sh --enable-x11 --disable-wayland  # X11 only ./autogen.sh
--disable-x11 --enable-wayland  # Wayland only (currently errors, but
will work once I land full Wayland support) ./autogen.sh --enable-x11
--enable-wayland   # X11 and Wayland ./autogen.sh --disable-x11
--disable-wayland # Neither (doesn't make sense, so gives an error)
./autogen.sh                                 # Support for omitted
platforms is automatic. Checks installed libraries 

	If a platform is disabled at compile-time, it's libraries aren't
linked and its headers aren't included anywhere. There are also files
which are only compiled/included on some platforms (for example,
panel-background-monitor.c/.h will give errors if compiled with X11
turned off, and the build system is set up to keep this from
happening. 

	If both X11 and Wayland are turned on at compile-time, then both
libraries are linked, all files are compiled, both sets of ifdefs are
enabled and the resulting binary can run on either platform. In that
case, runtime checks determine the code path to follow (often these
look something like if (GDK_IS_X11_DISPLAY (gdk_display_get_default
()))). I make sure to put a runtime check directly inside every
compile-time check, and frequently compile in all three configurations
to make sure everything's working.

raveit65:
With wayland we lose rgba color transparency of the panel background?
Is it possible to use an image as background with wayland?
Not that i use second point.........:)

raveit65:
> Going forward, I would delete panel-background-monitor.c entirely and not attempt to support transparency emulation on non-compositing window managers. It seems to me that there is way too much code and complication introduced than this edge case is worth. Anyway, that's for someone else to decide and PR if they wish, my work here is done.

I think this is OK for wayland. I don't think that there are much use cases for a non-compositor WM on modern boxes nowadays,  but as always removing that for x11 will give us angry users who use that feature.

raveit65:
Btw. is gtk-layer-shell ready to production?
I think it is time to bring it in fedora.

raveit65:
This works fine under X11.
After testing fake tranparency when compositor is disable i think we should start a discussion to remove that feature, because the background of some applets looks really weird, eg. new volume-applet.
@lukefromdc 
Any concerns left or can this be merged?

lukefromdc:
I think we can proceed with the merge anyway. I thnk
pseudotransparency isonly used in user-ser backgrounds. We will have
to document any removal ofthis, but it is heavy (memory copy jobs
always are), a lot of code, and has beenthe source of multple hassles
in fixing deprecations. In other words becominga CPU burden and a
maintainace burden.

On 6/21/2019 at 5:41 AM, "raveit65"  wrote:

	This works fine under X11.
 After testing fake tranparency when compositor is disable i think we
should start a discussion to remove that feature, because the
background of some applets looks really weird, eg. new volume-applet.
 @lukefromdc
 Any concerns left or can this be merged?

wmww:
> With wayland we lose rgba color transparency of the panel background?

No.

> Is it possible to use an image as background with wayland?

Yes.

The implementation of panel-background was quite bad. If my reading of it was correct, it was doing to CPU background pixbuffer copy even on compositing window managers where it wasn't needed. For this reason previous iterations of my Wayland support dropped a lot of panel-background features, because the non-wayland concept of copying the desktop was tangled through them. This PR attempts to untangle things enough to give Wayland full support for transparency and images.

sc0w:
we have since this PR this cppcheck warning:

`[mate-panel/panel-background-monitor.c:60]: (style) struct member '_PanelBackgroundMonitorClass::changed' is never used.`

maybe it needs `#ifdef HAVE_X11` there?

PR #988, opened 2019-06-19
Panel multiscreen -> panel multimonitor

wmww:
Since GTK 3.10 (older than our current required version of 3.22) GTK has not supported multiple screens. In panel-multiscreen.c, `screens = 1` was hard coded, but the logic to handle multiple screens was still laying around. This renames panel-multiscreen to panel-multimonitor, and removes all logic dealing with screens from it. I needed this for Wayland, but It is also a step toward GTK4 (where `GdkScreen` doesn't exist).

There is still some logic around RANDR in panel-multimonitor that I think is left over from times long past, but that can be dealt with another time by someone else.

### DIFF
--- a/mate-panel/Makefile.am
+++ b/mate-panel/Makefile.am
@@ -50,7 +50,7 @@ panel_sources = \
 	panel-toplevel.c \
 	panel-struts.c \
 	panel-frame.c \
-	panel-multiscreen.c \
+	panel-multimonitor.c \
 	panel-a11y.c \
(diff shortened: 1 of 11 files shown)


wmww:
I think I fixed it. Can you test again?

raveit65:
> The warnings disappeared from log once I restarted the computer.

Those are well known old warnings of GtkStatusIcon applets. I saw them from time to time before this PR.

PR #990, opened 2019-06-20
Panel toplevel use monitors instead of GdkScreen size

wmww:
On top of #988

Use size and position of monitors instead of dimensions of the `GdkScreen`. `GdkScreen` Doesn't make sense in Wayland, and is going away in GTK4.

### DIFF
--- a/mate-panel/panel-multimonitor.c
+++ b/mate-panel/panel-multimonitor.c
@@ -670,3 +670,21 @@ panel_multimonitor_is_at_visible_extreme (int        n_monitor,
 	}
 }
 
+void
+panel_multimonitor_get_bounds (GdkPoint *min,
+			       GdkPoint *max)
+{
+	int i;
+
+	g_return_if_fail (monitor_count > 0);
+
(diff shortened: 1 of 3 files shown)


lukefromdc:
I just confirmed these panel position problems are identical in 
https://github.com/mate-desktop/mate-panel/pull/988 and do not exist in master

wmww:
> NOT sure what you meant by "on top of #988" but as I see commits with the same names I am assuming all those commits are included here and are the same, and that this is thus a superset of the other PR.

Exactly. I just mean that the changes I'm PRing here are built "on top" of the changes proposed over there (so we should land that one before landing this one).

I think the issue was that I changed the signals it uses to trigger a monitor refresh. I think I have a fix in the other PR, and so once that is tested and lands we can take another look at this.

lukefromdc:
If the other PR was merged, this one would then conflict and would have to be rebased

lukefromdc:
I can confirm the panel position fix works here too, looks good to go

PR #998, opened 2019-07-07
Fix popup placement on Wayland

wmww:
This PR effects right-click menus on applets and opening the application menu.

On X, it appears the X struts we create automatically push the menus off the panel. The same does not happen on Wayland, so popups cover up the panel. This fixes the two cases where this is definitely bad, by adjusting the anchors so the popups appear in the right place. This should not change behavior on X (the X server will just no longer need to shove popups out of the way).

There are still cases where on X popups appear off the panel and on Wayland over top of it (ex right clicking on the empty space on the panel). However, the remaining cases use `gtk_menu_popup_at_pointer ()`. IMO this looks fine and is arguably preferable behavior in those cases to what X does.

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -852,10 +852,26 @@ mate_panel_applet_menu_popup (MatePanelApplet *applet,
 	context = gtk_widget_get_style_context (GTK_WIDGET(toplevel));
 	gtk_style_context_add_class(context,"gnome-panel-menu-bar");
 	gtk_style_context_add_class(context,"mate-panel-menu-bar");
+	GdkGravity widget_anchor = GDK_GRAVITY_NORTH_WEST;
+	GdkGravity menu_anchor = GDK_GRAVITY_NORTH_WEST;
+	switch (applet->priv->orient) {
+	case MATE_PANEL_APPLET_ORIENT_UP:
+		menu_anchor = GDK_GRAVITY_SOUTH_WEST;
+		break;
+	case MATE_PANEL_APPLET_ORIENT_DOWN:
+		widget_anchor = GDK_GRAVITY_SOUTH_WEST;
+		break;
+	case MATE_PANEL_APPLET_ORIENT_LEFT:
+		menu_anchor = GDK_GRAVITY_NORTH_EAST;
+		break;
+	case MATE_PANEL_APPLET_ORIENT_RIGHT:
+		widget_anchor = GDK_GRAVITY_NORTH_EAST;
+		break;
+	}
 	gtk_menu_popup_at_widget (GTK_MENU (menu),
 	                          GTK_WIDGET (applet),
-	                          GDK_GRAVITY_NORTH_WEST,
-	                          GDK_GRAVITY_NORTH_WEST,
+	                          widget_anchor,
+	                          menu_anchor,
 	                          event);
 }
 
--- a/mate-panel/panel-menu-button.c
+++ b/mate-panel/panel-menu-button.c
@@ -456,15 +456,31 @@ panel_menu_button_popup_menu (PanelMenuButton *button,
 	gtk_window_set_attached_to (GTK_WINDOW (gtk_widget_get_toplevel (button->priv->menu)),
 				    GTK_WIDGET (button));
 
-	/*using these same anchor points lets default "anchor-hints" properly position the menu
-	 *so that on a vertical panel the menu aligns with the outside edge of a menu button
-	 *placed at the top or bottom of a left or a right panel
-	 */
+	GdkGravity widget_anchor = GDK_GRAVITY_NORTH_WEST;
+	GdkGravity menu_anchor = GDK_GRAVITY_NORTH_WEST;
+	switch (panel_toplevel_get_orientation (button->priv->toplevel)) {
+	case PANEL_ORIENTATION_TOP:
+		widget_anchor = GDK_GRAVITY_SOUTH_WEST;
+		g_message ("PANEL_ORIENTATION_TOP");
+		break;
+	case PANEL_ORIENTATION_BOTTOM:
+		menu_anchor = GDK_GRAVITY_SOUTH_WEST;
+		g_message ("PANEL_ORIENTATION_BOTTOM");
+		break;
+	case PANEL_ORIENTATION_LEFT:
+		widget_anchor = GDK_GRAVITY_NORTH_EAST;
+		g_message ("PANEL_ORIENTATION_LEFT");
+		break;
+	case PANEL_ORIENTATION_RIGHT:
+		menu_anchor = GDK_GRAVITY_NORTH_EAST;
+		g_message ("PANEL_ORIENTATION_RIGHT");
+		break;
+	}
 
 	gtk_menu_popup_at_widget (GTK_MENU (button->priv->menu),
 	                          GTK_WIDGET (button),
-	                          GDK_GRAVITY_NORTH_WEST,
-	                          GDK_GRAVITY_NORTH_WEST,
+	                          widget_anchor,
+	                          menu_anchor,
 	                          NULL);
 }
 


wmww:
* You are correct that the changes to panel-menu-button.c are what fix the main application menu
* Clock and fish should both work, you can test the changes to mate-panel-applet.c with them. If they don't work, uninstall, ./configure with `--with-in-process-applets=all`, rebuild and install. (you probably know that, but no harm in repeating it as it's hard for anyone to figure out that's the problem)
* Is the right click menus overlapping the panel actually a problem? My personal opinion is that it wouln't look good for the main panel menu, but it's preferable for right click menus as that's the normal behavior in other (non-panel) applications. Perhaps even the applet menus should be switched to popup at the cursor. If it's generally agreed that popup up outside the panel is always preferable I can implement that in another PR.
* I'm not seeing the panel deletion bug. I had some trouble with that a while back. Perhaps you need to update and install GTK-layer-shell? If after that you still see the bug, can you explain it in more detail please

lukefromdc:
I don't see any of the issues mentioned as reasons not to merge this, in fact I packaged it up in one of my local .deb packages on the grounds that it is much better than before.

raveit65:
Maybe we can switch more menus to use `gtk_menu_popup_at_widget` ?
```
[rave@mother mate-panel]$ grep -nr gtk_menu_popup *
libmate-panel-applet/mate-panel-applet.c:871:	gtk_menu_popup_at_widget (GTK_MENU (menu),
mate-panel/applet.c:664:	gtk_menu_popup_at_pointer (GTK_MENU (info->menu), NULL);
mate-panel/menu.c:576:	gtk_menu_popup_at_pointer (GTK_MENU (menu), event);
mate-panel/panel.c:371:	gtk_menu_popup_at_pointer (GTK_MENU (menu), NULL);
mate-panel/panel-menu-button.c:480:	gtk_menu_popup_at_widget (GTK_MENU (button->priv->menu),
mate-panel/panel-action-protocol.c:102:	gtk_menu_popup_at_pointer (GTK_MENU (menu),event);
```
I am still wondering that we can't add any `anchor-hints` to `gtk_menu_popup_at_pointer` ?
https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-popup-at-pointer
https://developer.gnome.org/gtk3/stable/GtkMenu.html#GtkMenu--anchor-hints
`These hints determine how the menu should be positioned in the case that the menu would fall off-screen if placed in its ideal position.`

And i am thinking that `gtk_menu_popup_at_pointer` is the right choice for right-click menus. They should open under the pointer.
Opening under `widget` can open the menu at another place if `widget` is the whole panel, which can be confusing.
This is a problem with window-list-applet under X11 , ihmo.

wmww:
> I am still wondering that we can't add any anchor-hints to `gtk_menu_popup_at_pointer`?

You can't add anchor hints because they don't make sense. Anything but lower-right gravity would look weird and it's anchored to a single point (the pointer), so anchor hints don't make sense there either.

raveit65:
Ok.

It seems that changes are working well under X11. menubar open the menus fine on all panel orientations.
But with rootston window i see no difference than before. Eg. The menus opens off-screen if the menubar is at bottom panel.
Is it possible that there is an issue with rootston window?

wmww:
Yes, it's Rootston's fault. From my comment on the previous PR:

> The Wayland compositor is supposed to keep popups from going off screen, which Rootston does not last I checked. I think ddvault was working on that, but if any popups go partly or completely off screen it's a compositor issue, and probably wont happen in Mir.

raveit65:
Ok, than i can only confirm that both commits are working under X11.

But, does `GDK_ANCHOR_FLIP_Y` not help in case of bottom panel.
Or do i misunderstood this example?
https://developer.gnome.org/gtk3/stable/GtkMenu.html#GtkMenu--anchor-hints
` GDK_ANCHOR_FLIP_Y will replace GDK_GRAVITY_NORTH_WEST with GDK_GRAVITY_SOUTH_WEST and vice versa if the menu extends beyond the bottom edge of the monitor.`

raveit65:
> If you really wanted to mimic the X11 behavior, that's doable. You just use gtk_menu_popup_at_rect() and anchor hints to put it where you want it.

Well, i need a working wayland compositor . Add the moment it is very hard to see what is working or not, if most menus opens off-screen on bottom and left panel.

So, @lukefromdc approval should be enough for wayland.

raveit65:
@wmww 
Does mate-wayland work with newer nvidia-card and current nvidia-driver for linux on baremetal?
Than i can use my fedora rawhide installation for testing.
I don#t want to break my current panel config in my normal f30 installation.

raveit65:
wtf, on fedora rawhide baremetal.
```
[rave@mother ~]$ sudo snap install --edge --classic mate-wayland
Warning: /var/lib/snapd/snap/bin was not found in your $PATH. If you've not
         restarted your session since you installed snapd, try doing that.
         Please see https://forum.snapcraft.io/t/9469 for more details.

mate-wayland (edge) 0+git.7ab25c8 from William Wold (wmww) installed
[rave@mother ~]$ mate-wayland
2019/07/10 10:59:19.328952 cmd_run.go:529: WARNING: XAUTHORITY environment value is not a clean path: "/run/lightdm/rave/xauthority"
Waiting for Wayland display to appear at /run/user/1000/wayland-mate...
[2019-07-10 10:59:24.891690] <information> mirserver: Starting
[2019-07-10 10:59:24.905818] < - debug - > mirserver: Not trying logind: "DISPLAY" is set and X need not have claimed the VT
[2019-07-10 10:59:24.932540] < - debug - > mirserver: Not using Linux VT subsystem for session management: Failed to find the current VT
[2019-07-10 10:59:24.932567] < - debug - > mirserver: No session management supported
[2019-07-10 10:59:24.932585] <information> VT switch key handler: No VT switching support available: MinimalConsoleServices does not support VT switching
[2019-07-10 10:59:24.932797] <information> mircommon: Loading modules from: /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/mir/server-platform
[2019-07-10 10:59:24.932843] <information> mircommon: Loading module: /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/mir/server-platform/graphics-mesa-kms.so.16
[2019-07-10 10:59:24.932866] <information> mircommon: Loading module: /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/mir/server-platform/server-mesa-x11.so.16
[2019-07-10 10:59:24.932876] <information> mircommon: Loading module: /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/mir/server-platform/input-evdev.so.7
[2019-07-10 10:59:24.950326] <information> mesa-kms: EGL platform does not support EGL_KHR_platform_gbm extension
[2019-07-10 10:59:24.953163] <information> mesa-kms: Failed to open device node: Invalid argument
[2019-07-10 10:59:24.953249] <information> mirplatform: Found graphics driver: mir:mesa-kms (version 1.3.0) Support priority: 0
[2019-07-10 10:59:24.985944] <information> mirplatform: Found graphics driver: mir:mesa-x11 (version 1.3.0) Support priority: 128
[2019-07-10 10:59:24.986102] <information> mirserver: Selected driver: mir:mesa-x11 (version 1.3.0)
MESA-LOADER: failed to open nouveau (search paths /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: nouveau
MESA-LOADER: failed to open kms_swrast (search paths /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/dri)
failed to load driver: kms_swrast
MESA-LOADER: failed to open swrast (search paths /snap/mate-wayland/144/usr/lib/x86_64-linux-gnu/dri)
failed to load swrast driver
ERROR: /build/mate-wayland/parts/mir-git/src/src/server/graphics/default_configuration.cpp(172): Throw in function mir::DefaultServerConfiguration::the_graphics_platform()::<lambda()>
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
std::exception::what: Exception while creating graphics platform
ERROR: /build/mate-wayland/parts/mir-git/src/src/platforms/mesa/server/display_helpers.cpp(330): Throw in function mir::graphics::mesa::helpers::GBMHelper::GBMHelper(const mir::Fd&)
Dynamic exception type: boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<std::runtime_error> >
std::exception::what: Failed to create GBM device
```
Sorry, using nouveau driver isn't an option, because session is slow and have graphical glitches. with GeForce GT 1030/PCIe/SSE2
In other words nouveau driver is broken for newer cards.
It has a reason that gnome/fedora suggest to use proprietary nvidia driver for fedora wayland session.

In result i have to wait for fix to run mate-wayland in VM.

I am sorry to say, my motivation to spend my time for testing goes to zero without nvidia driver support.

Issue #745, opened 2018-02-25
Widgets on vertical panel don't stick to their position after restarting the panel


o-alquimista:
During my sessions, all widgets are locked to a position, and they remain properly positioned during the current session. However, restarting the panel (or rebooting the system) will cause it to forget the position of my widgets, only when a vertical panel is used. The screenshots show how it looks _before and after._
![before](https://user-images.githubusercontent.com/25424263/36642177-5bedcf8c-1a1a-11e8-957a-fd884b55eb49.png)
![after](https://user-images.githubusercontent.com/25424263/36642181-62dba666-1a1a-11e8-8e40-535913ef470c.png)

### Additional info
- I have never tried this with an older version.
- I did not succeed in getting the stack trace for mate-panel. GDB says "No stack" when running "gdb mate-panel" followed by "backtrace".

#### MATE general version
1.20.0

#### Package version
community/mate-panel 1.20.0-1 (mate)

#### Linux Distribution
Arch Linux

#### Link to downstream report of your Distribution
https://bugs.archlinux.org/task/57631?project=5&opened=27326

info-cppsp:
doesn't seem to be a problem with 1.18.6
I have noticed app launcher moving when I enable / disable the hide buttons on left panel with 1.20.
now app launchers move when enable / disable the hide buttons, but after `mate-panel --replace` they are where they are supposed to be.

I would say try using `mate-panel --replace` and add launchers one after another to a new panel.
maybe one launcher causes this, not the panel

o-alquimista:
I experimented with mate-panel --replace and was able to reproduce the problem only when any of the widgets were placed to the very extremity of the panel. If I position all widgets so they don't touch the ends of the panel, after running --replace they remain in their positions.
![captura de tela em 2018-02-25 12-29-49](https://user-images.githubusercontent.com/25424263/36643293-2740bffc-1a28-11e8-8c8e-6512a1e52717.png)

info-cppsp:
on ub 18.04 I can see that at the bottom of the left panel (near the bottom left corner of the screen)
the position looked up in dconf-editor changed from panel-right-stick off and a position around 900 (I have a fullhd monitor) to panel-right-stick on and position 37. on the panel object I have placed on the panel.

for me this means that for vertical panels the calculation of the panel widget is not fully correct...
not a problem with 1.18.6
tested with left panel size 40

info-cppsp:
[...]

PR #1004, opened 2019-07-22
Fix applets placement on expanded vertical panel

monsta:
This fixes https://github.com/mate-desktop/mate-panel/issues/745.

A regression from https://github.com/mate-desktop/mate-panel/pull/683 caused all applets on expanded vertical panel stick to the top. In addition it made window list (if it's present on that panel) completely disappear...

For testing I've used a horizontal panel with some applet(s) set to stay on the right (tray area or workspace switcher). Then I've moved that panel to the left, making it vertical. Without this fix, all applets were stuck to the top. With this fix, tray area or workspace switcher stay at panel's bottom.

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -2931,10 +2931,8 @@ panel_toplevel_move_resize_window (PanelToplevel *toplevel,
 				   toplevel->priv->geometry.width,
 				   toplevel->priv->geometry.height);
 
-	if(resize || move)
-	{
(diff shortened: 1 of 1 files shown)


monsta:
Not exactly applet-specific, but position-specific: it affects applets with `panel-right-stick` key set to `true`. They also usually have `position` key set to some small value, like 0 or 10. That's their position relative to the right edge (on horizontal panel) or to the bottom (on vertical panel).

lukefromdc:
Does anyone else want to test this or should I go ahead and merge it?

raveit65:
https://github.com/orgs/mate-desktop/teams/core-team/discussions/17

PR #1015, opened 2019-09-11
When determining whether the pos is in the applet, use ad->cells to calc.

### LINKED ISSUE #1012: Adding the launcher to the panel causes the opened web taskbar tab to be obscured by the icon
#### Expected behaviour
After adding the launcher to the panel, the tabs that have been opened in the mate-panel should remain visible.

#### Actual behaviour
After adding the launcher to the panel, the opened tab in the mate-panel is overwritten by the new icon. You cannot switch the window by clicking the tab with the mouse. You need to drag the window list bar to adjust it.

Open the application menu in mate-panel, select the system tool or other group, right click and select Add this launcher to panel

#### MATE general version
1.8.0

#### Package version
1.8.1-2

#### Linux Distribution
fedora21

yetist:
This PR can fix #1012.

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -1917,7 +1917,7 @@ is_in_applet (int pos, AppletData *ad)
 	g_return_val_if_fail (ad != NULL, NULL);
 
 	if (ad->constrained <= pos &&
-	    ad->constrained + ad->min_cells > pos)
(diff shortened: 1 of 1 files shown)


raveit65:
cherry-picked to 1.22 branch https://github.com/mate-desktop/mate-panel/commit/6343e1442382b551be09e664724cbb1efb77daed

Issue #1017, opened 2019-09-11
Segfault on adding new location to calendar


lukefromdc:
#### Expected behaviour
Adding location to calendar works normally

#### Actual behaviour
Panel segfaults on clicking "OK" button in dialog to add location

#### Steps to reproduce the behaviour
Open the calendar popup , it will come up with "locations" selected
click "edit" and then "add" buttons
Select any location
click "OK" and that's when the panel segfaults. When it restarts the location will have been sucessfully added. 
Got this backtrace in gdb:
```
Thread 1 "mate-panel" received signal SIGSEGV, Segmentation fault.
0x00007ffff714fadd in g_type_check_instance_is_fundamentally_a ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
(gdb) thread apply all bt full

Thread 6 (Thread 0x7fffdefff700 (LWP 14569)):
#0  0x00007ffff7fe80c5 in  () at /lib64/ld-linux-x86-64.so.2
#1  0x00007ffff6a8db2f in _dl_catch_exception () at /lib/x86_64-linux-gnu/libc.so.6
#2  0x00007ffff7fe7bba in  () at /lib64/ld-linux-x86-64.so.2
#3  0x00007ffff6a8d06d in  () at /lib/x86_64-linux-gnu/libc.so.6
#4  0x00007ffff6a8db2f in _dl_catch_exception () at /lib/x86_64-linux-gnu/libc.so.6
#5  0x00007ffff6a8dbbf in _dl_catch_error () at /lib/x86_64-linux-gnu/libc.so.6
#6  0x00007ffff6a8d147 in  () at /lib/x86_64-linux-gnu/libc.so.6
#7  0x00007ffff6a8d1d6 in __libc_dlopen_mode () at /lib/x86_64-linux-gnu/libc.so.6
#8  0x00007ffff6a74b9e in  () at /lib/x86_64-linux-gnu/libc.so.6
#9  0x00007ffff6a75498 in __nss_lookup_function () at /lib/x86_64-linux-gnu/libc.so.6
#10 0x00007ffff6a755b2 in __nss_lookup () at /lib/x86_64-linux-gnu/libc.so.6
#11 0x00007ffff6a6758b in getservbyname_r () at /lib/x86_64-linux-gnu/libc.so.6
#12 0x00007ffff6a67173 in getservbyname () at /lib/x86_64-linux-gnu/libc.so.6
[...]

PR #1019, opened 2019-09-17
avoid deprecated 'g_type_class_add_private'

sc0w:


lukefromdc:
I have nm-applet built with indicator support and started with` nm-applet --indicator` so I'm thinking something in how that uses the status notifier (I see /org/ayatana/NotificationItem/nm_applet in the backtrace) would be why I got the segfault and the previous tester did not.

sc0w:
@lukefromdc I am working now with this PR installed and it seems to work fine. How to reproduce the crash step by step?

lukefromdc:
Open the calendar, expand it to include the map/locations/clock faces, click
"edit," then try to add a location. When you have one selected, try to add it, that's
when the crash occurs on my box(I am not at it right now). If this does not occue on
your system you probably cannot duplicate it.

Note that when the panel restarts the location has been sucessfully added, so the
crash must be in the final portion of that code(os as I suspect in something called
by it).

sc0w:
@lukefromdc seems you are mentioned the issue https://github.com/mate-desktop/mate-panel/issues/1017 not related to this PR (I can't reproduce the problem)

lukefromdc:
Oh yeah, that's it-and I just confirmed it again with master, so it is not related to this PR. When I previously tested it, I must have gotten a successful close of the dialog without a crash by chance, as this happens most but not quite all of the time. Sorry for the noise on that, it means this PR should be ready to go as we never had a problem with this code change before, I didn't get any other issues, and the issue I did get is now confirmed (again) to exist on master too. Approving

lukefromdc:
Shit: I got totally confused this morning and asked you to try to duplicate the WRONG issue, which is common to master. Was away from my desktop and there had been much conversation about that other issue.

lukefromdc:
I also just tested reverting the changes to applets/notification_area/status-notifier/sn-item.c and keeping the rest of this, and the resulting build worked and didn't crash on bringing back nm-applet in indicator mode.

lukefromdc:
Playing with the file I found that the first change adding 
`G_ADD_PRIVATE (SnItem)` does not crash the panel on running `nm-applet --indicator` if I am right (I don't really know a lot about this) this should support both the old and the new way of handling this.

Then replacing `item->priv = G_TYPE_INSTANCE_GET_PRIVATE (item, SN_TYPE_ITEM, SnItemPrivate);` with item->priv = sn_item_get_instance_private (item) brings about the crash.
`
item->priv = sn_item_get_instance_private (item)` is used elsewhere in the code before this PR but seems not to work for sn_item_init, maybe it's not ready in time?

sc0w:
not here, I did:

```
$ killall nm-applet
$ nm-applet --indicator
```

and mate-panel doesn't crash, I need to do additional steps?

lukefromdc:
OK,that means you will might be able to duplicate the crash. Maybe you have a build
of nm-applet that does not support indicator mide? Its a build-time option and not all
distros use it. In that case, try any known status notifier, I will try our own mate-polkit 
with it enabled (my normal build) tonight. If you still cannot duplicate it I will be out of
ideas except for.trying to fix it myself.

lukefromdc:
The issue seems to be on any indicator. I just tried mate-polkit built with sni support, and on mounting an internal partition from caja, the panel crashed and went into another restart/crash loop(from the session manager respawning it as usual) as soon as mate-polkit tried to display the keys icon as an indicator.

lukefromdc:
I tested modifying sn_item_init to put a message in terminal if item->priv comes up null
```
static void
sn_item_init (SnItem *item)
{
  item->priv = sn_item_get_instance_private (item);
  if (item->priv == NULL)
      g_message("invalid item->priv");

  gtk_widget_add_events (GTK_WIDGET (item), GDK_SCROLL_MASK);
}
```
and with this got `notification-area-applet-Message: 00:28:30.881: invalid item->priv` just before the segfault, indicating that ` item->priv = sn_item_get_instance_private (item);` in `sn_item_init` is not working here.

sc0w:
I think the problem is this line in `sn-item.c`:

`#define sn_item_get_instance_private(i) (SN_ITEM (i)->priv)`

@lukefromdc I did new changes, please test again

PR #1021, opened 2019-09-20
Load layouts from absolute paths

acperkins:
If the layout name begins with a '/', load from that filename instead of
from PANEL_LAYOUTS_DIR.  This allows loading of layouts from home
directories and other write-permitted locations.


acperkins:
This seemed to work in my testing, but I would appreciate a second look to test this.  I wanted to be able to load layouts from my home directory, instead of having to `sudo cp` them into /usr/share/mate-panel/layouts.

raveit65:
Can you please squash both commits together?
No need to have a second commit which fixes an errror in your first commit.

acperkins:
@raveit65 I've done that and pushed the modified commit now.

PR #1029, opened 2019-10-15
Add date display format for “Clock preferences”

LiuXingxiang-kylinsec:
- In China, we hope that like the Windows operating system, we can set the display format of time. Therefore, I added a long/short date display format for the Chinese environment.
- Long Date Display Format:YYYY年MM月DD日
- Short Date Display Format:YYYY-MM-DD

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -74,6 +74,8 @@
 
 #define KEY_FORMAT		"format"
 #define KEY_SHOW_SECONDS	"show-seconds"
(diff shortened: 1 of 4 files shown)


lukefromdc:
We don't develop against released versions and especially not old versions that are no longer supported. The current release is 1.22, which gets bugfixes but not new features. The current development release is 1.23, which is where master is at right now and will eventually become 1.24 for stable release. Please open a new PR against master, closing this one as it cannot be built against master

PR #1035, opened 2019-10-28
Add date display format for 'Clock preferences'

LiuXingxiang-kylinsec:
Expected behaviour

    In China, we hope that like the Windows operating system, we can set the display format of time. Therefore, I added a long/short date display format for the Chinese environment.
    Long Date Display Format:YYYY年MM月DD日
    Short Date Display Format:YYYY-MM-DD

Actual behaviour
Steps to reproduce the behaviour
MATE general version

    git master master

Package version

    git master master

Linux Distribution

    Centos7.6

Link to downstream report of your Distribution

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -76,6 +76,8 @@
 
 #define KEY_FORMAT                "format"
 #define KEY_SHOW_SECONDS        "show-seconds"
+#define KEY_SHOW_LONGDATE	"show-longdate"
+#define KEY_SHOW_SHORTDATE	"show-shortdate"
 #define KEY_SHOW_DATE                "show-date"
 #define KEY_SHOW_WEATHER        "show-weather"
 #define KEY_SHOW_TEMPERATURE        "show-temperature"
@@ -146,6 +148,8 @@ struct _ClockData {
         ClockFormat  format;
         char        *custom_format;
         gboolean     showseconds;
+        gboolean     showlongdate;
+	gboolean     showshortdate;
         gboolean     showdate;
         gboolean     showweek;
         gboolean     show_weather;
@@ -172,6 +176,8 @@ struct _ClockData {
         int fixed_height;
 
         GtkWidget *showseconds_check;
+        GtkWidget *showlongdate_check;
+        GtkWidget *showshortdate_check;
         GtkWidget *showdate_check;
         GtkWidget *showweeks_check;
         GtkWidget *custom_hbox;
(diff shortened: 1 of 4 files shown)


LiuXingxiang-kylinsec:
I fix display error for longdate,and add a function to set default date format

yetist:
I noticed that the clock applet supports custom format:

https://github.com/mate-desktop/mate-panel/blob/bbdaf1fd77e851bd0b63e4793e1f9309dbe12ebc/applets/clock/org.mate.panel.applet.clock.gschema.xml.in#L30

https://github.com/mate-desktop/mate-panel/blob/bbdaf1fd77e851bd0b63e4793e1f9309dbe12ebc/applets/clock/org.mate.panel.applet.clock.gschema.xml.in#L32

But there are no configuration widget on the UI, is it possible to add a third choice: 12-hours, 24-hours and custom format?

lukefromdc:
I got this build error on rebuilding with the latest changes:
`
org.mate.panel.applet.clock.gschema.xml:57:1  Error on line 57 char 1: Element <key> not allowed inside <key>.  --strict was specified; exiting.`
build options were
`
./autogen.sh  --with-in-process-applets=wncklet,clock,fish,notification-area --prefix=/usr libdir=/usr/lib/x86_64-linux-gnu --sysconfdir=/etc --enable-shared=yes --enable-static=no --disable-maintainer-mode
`
Also I have no idea if file in Po are supposed to be changed as part of other PR's or if translaters do that separately, I know little of that part of the process but have never seen a change there as part of another PR that I can recall.

applets/clock/clock.c
```diff
@@ -682,6 +773,64 @@ update_tooltip (ClockData * cd)
                 else
                         tip = _("Click to view month calendar");
         }
+        if (!cd->showdate) {
```
rbuj: You don't need to show date and time on the tooltip, furthermore tip was set on the previous line.

applets/clock/clock.c
```diff
@@ -146,6 +149,9 @@ struct _ClockData {
         ClockFormat  format;
         char        *custom_format;
         gboolean     showseconds;
+        gboolean     showlongdate;
+	gboolean     showshortdate;
+        gboolean     showdefaultdate;
```
rbuj: You really only need a boolean variable, for instance showlongdate and set its default value.

rbuj:
clock_format for short and long dates should be set in the else block, i.e. if cd->showdate is true.
https://github.com/mate-desktop/mate-panel/blob/8a4ae0116b1a75f6b77e60ef364288c38a60b35f/applets/clock/clock.c#L465-L489

lukefromdc:
What is the status of this?

raveit65:
@LiuXingxiang-kylinsec 
There are open remarks and a file conflict.

raveit65:
@LiuXingxiang-kylinsec
Any news?

po/zh_CN.po
```diff
@@ -313,6 +313,10 @@ msgstr "长日期格式：YYYY年MM月DD日"
 msgid "Show shortdate"
 msgstr "短日期格式：YYYY-MM-DD"
 
+#: ../applets/clock/clock.ui.h:37
```
raveit65, line 316: We have a file conflict with master branch at this point.
Better do not edit this file. Every pull from our translation sever (transifex) will override this file.
We will update mate-panel.pot file with your changes after this is merged.
After that the translators can update this translation string.

lukefromdc:
Is this going to be fixed or should we close it? Lots of outstanbding PR's right now

Issue #835, opened 2018-07-05
Panel widgets should expose an Accessible object to at-spi


sthibaul:
#### Steps to reproduce the behaviour

- Run a mate desktop
- Run orca, enable mouse review ("Speak object under mouse" option in its general configuration panel)
- Move mouse to mate panel wigets, for instance here the clock

#### Expected behaviour

- Orca should speak the panel widget content, in the instance here the clock date & time

#### Actual behaviour

- Orca just says "panel"

#### MATE general version

1.20

#### Package version

1.20.1

#### Linux Distribution

Debian buster

#### Additional information

The issue is that mate-panel widgets do not expose Accessible objects through at-spi. When looking in accerciser, we see mate-panel expose the panel structure (thus why Orca says "panel"), but the final widgets are not exposed.

Of course, for proper complete support, probably each and every panel widget needs to be fixed to expose its content through at-spi. Perhaps mate-panel should at least put the name of the widget in the panel description (just like it does with `panel_a11y_set_atk_name_desc` for "Top Panel"), so at least the user knows which widget it is. We can then work on adding support to widgets, at least the most commonly used.

sthibaul:
Launchers notably do have an accessible description that Orca can read.

sthibaul:
Here is a very crude attempt [patch.txt](https://github.com/mate-desktop/mate-panel/files/2197495/patch.txt) to do this which does work:

- what needs to wear the atk name is the socket since that's the innermost object that Orca sees
- what knows the name of the applet is apparently the MatePanelAppletFrame. For now I just took the id, perhaps there is a better name to be taken, ideally it would be translated.

that's why I had to make `mate_panel_applet_frame_dbus_set_id_descr` get the name from a call to `mate_panel_applet_frame_get_id`, and then call `mate_panel_applet_container_set_id` which calls `panel_a11y_set_atk_name_desc` on the socket.

Also, I had to call mate_panel_applet_frame_dbus_set_id_descr from `mate_panel_applet_frame_dbus_init_properties`, which is late enough for applet_info to be available.

Eventually, I copied over panel_a11y_* just to get it linking, otherwise linking mate-panel-test-applets was failing. I guess we could put these functions to a better place.

How do you feel this could be improved to be integrated?

alexarnaud:
@lukefromdc @raveit65  What are your opinions on this?

@muktupavels It could maybe benefit also to GNOME Panel.

Best regards,
Alex.

[...]

PR #1040, opened 2019-11-26
Provide a fallback accessibility string for applets

johndescs:
Corresponds to https://github.com/mate-desktop/mate-panel/issues/835 and the thread beneath.

### DIFF
--- a/mate-panel/Makefile.am
+++ b/mate-panel/Makefile.am
@@ -196,6 +196,7 @@ mate_panel_test_applets_SOURCES = \
 	panel-modules.c \
 	panel-applet-info.c \
 	panel-applets-manager.c \
+	panel-a11y.c \
 	panel-test-applets.c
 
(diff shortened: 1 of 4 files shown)


alexarnaud:
Confirmed that it does what it promise to do.

johndescs:
Pushed cosmetics enhancements after @cwendling's review.

sthibaul:
Mmm.

Orca says so because it has panel role.
If the role is set to filler, does Orca still speak it when the applet is accessible? Does it speak it when the applet is not accessible?

If that other approach doesn't work, I'd tend to lean on your first option: having applets to overwrite the default description, or clear it if that's not the proper place to expose something.

lukefromdc:
Is this going to get further work (to limit the fallback case to those applets don't provide a useful accessable string) or should we close this?

alexarnaud:
@lukefromdc we don't plan to work on it again so I assume we can close this.

PR #1041, opened 2019-11-27
status-notifier: Render icons as cairo surface

vkareh:
Instead of GdkPixbuf, so that SNI icons can render correctly on HiDPI
displays.

### DIFF
--- a/applets/notification_area/status-notifier/sn-dbus-menu-item.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu-item.c
@@ -197,7 +197,10 @@ sn_dbus_menu_item_new (GVariant *props)
             }
           else if (item->icon_data)
             {
-              image = gtk_image_new_from_pixbuf (item->icon_data);
+              cairo_surface_t *surface;
+              surface = gdk_cairo_surface_create_from_pixbuf (item->icon_data, 0, NULL);
+              image = gtk_image_new_from_surface (surface);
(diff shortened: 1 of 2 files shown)


sc0w:
@lukefromdc the warnings you mentioned aren't related to this PR

I reviewed the logs, and there is no new warnings with this PR

vkareh:
@lukefromdc - just added a new commit that renders menu items inside the popup-menu as cairo surfaces. This, along with a new update to https://gitlab.gnome.org/GNOME/network-manager-applet/merge_requests/72 should render all variations of nm-applet (with and without `--indicator`) correctly.

lukefromdc:
I just tested this with https://gitlab.gnome.org/vkareh/network-manager-applet/commit/d56c26da6d4ddddfcf3b8d54db2fc61b818b5af1#note_658383
which with mate-panel from master renders the menu item icons double-size with window-scaling = 2, and with this PR used together with it gives clean, sharp menuitem icons the correct size

lukefromdc:
Latest commit https://gitlab.gnome.org/GNOME/network-manager-applet/merge_requests/72/diffs?commit_id=b8a47d40681b00141b4d5fffcb70e517d4eda10c to the network-manager-applet PR solves the doublesize icon problem with git master on their end, meaning we can now fix the half-size icon on this end and then users can mix and match mate-panel and network-manager-applet versions without ever getting wrong size icons

PR #1046, opened 2019-12-11
Add clock ATK name

johndescs:
This adds the current displayed time from clock applet to its widget's ATK attributes. When using Orca's mouse review this makes the applet tell something usefull.

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -598,21 +598,26 @@ static void
 update_clock (ClockData * cd)
 {
         gboolean use_markup;
-        char *utf8;
+        char *utf8, *text;
 
         time (&cd->current_time);
         utf8 = format_time (cd);
 
         use_markup = FALSE;
-        if (pango_parse_markup (utf8, -1, 0, NULL, NULL, NULL, NULL))
+        if (pango_parse_markup (utf8, -1, 0, NULL, &text, NULL, NULL))
                 use_markup = TRUE;
+        else
+                text = g_strdup (utf8);
 
         if (use_markup)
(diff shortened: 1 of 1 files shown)


applets/clock/clock.c
```diff
@@ -612,6 +612,8 @@ update_clock (ClockData * cd)
         else
                 gtk_label_set_text (GTK_LABEL (cd->clockw), utf8);
 
+        set_atk_name_description (cd->applet, utf8, NULL);
```
cwendling: If markup is used (see https://github.com/mate-desktop/mate-panel/pull/1046/files#diff-5f61b58fed2b9cfc19d74c7e85f4b891L607), you should not include the markup in the ATK name.  You should use the 5th argument of the [`pango_parse_markup()`](https://developer.gnome.org/pango/stable/pango-Markup.html#pango-parse-markup) call in case it returns `TRUE`.

```c
char *text;

if (pango_parse_markup (utf8, -1, 0, NULL, &text, NULL, NULL))
        use_markup = TRUE;
else
        text = g_strdup (utf8);

…

g_free (text);
```

Agreed that markup is rather uncommon, but it's supported.
BTW, to test:
```console
$ gsettings set org.mate.panel.applet.clock:/org/mate/panel/objects/clock/prefs/ custom-format '<span color="red">%x</span> <b>%T</b>'
$ gsettings set org.mate.panel.applet.clock:/org/mate/panel/objects/clock/prefs/ format 'custom'
```

cwendling: You still need to update the `set_atk_name_description()` to use `text` ;)

johndescs: I don't know what to say appart from "no comment" and "fixed".

PR #1047, opened 2019-12-11
Clock applet: Fix string comparison

clefebvre:
The clock format was fixed in https://github.com/mate-desktop/mate-panel/commit/4d20a0003798eea253fed6ea8f02cec5da6ffb2d, but it only needed fixing when LC_TIME and LANGUAGE are different.

In the code introduced by this commit `env_language != env_lc_time` doesn't properly compare the strings. It compares the pointers, and thus it always returns TRUE.

This fixes the comparison so the actual values of the strings are compared and we don't apply the format fix when we don't need to.

The comparison was also duplicated in the original code, it's factorized here into a `use_lctime` variable.

Ref: https://github.com/linuxmint/cinnamon-desktop/pull/142.

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -428,6 +428,7 @@ get_updated_timeformat (ClockData *cd)
         char       *clock_format;
         const gchar *env_language;
         const gchar *env_lc_time;
+        gboolean     use_lctime;
 
         /* Override LANGUAGE with the LC_TIME environment variable
          * This is needed for gettext to fetch our clock format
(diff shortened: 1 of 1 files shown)


applets/clock/clock.c
```diff
@@ -428,14 +428,17 @@ get_updated_timeformat (ClockData *cd)
...
+        gboolean     use_lctime;
 
         /* Override LANGUAGE with the LC_TIME environment variable
          * This is needed for gettext to fetch our clock format
          * according to LC_TIME, and not according to the DE LANGUAGE.
          */
         env_language = g_getenv("LANGUAGE");
         env_lc_time = g_getenv("LC_TIME");
-        if (env_language != NULL && env_lc_time != NULL && env_language != env_lc_time) {
+        use_lctime = (env_language != NULL) && (env_lc_time != NULL) && (g_strcmp0 (env_language, env_lc_time) != 0);
```
cwendling: One of the `NULL` checks is redundant with `g_strcmp0()`, but that can be seen as an optimization.
Also, style would suggest parenthesizing the whole expression rather than the specific components.

clefebvre: The first two conditions (NULL checks) aren't there to make strcmp safe, they're part of the logic to not change the environment.

I had parenthesis around the entire thing initially. I changed that to make it clearer there are three logical conditions at play.

cwendling: > The first two conditions (NULL checks) aren't there to make strcmp safe, they're part of the logic to not change the environment.

I was about to explain why it would be the same with only one check, but I realized that indeed I was plain wrong, sorry.  So the logic LGTM.

For the parenthesis I don't find it better, but that really doesn't matter.

PR #1050, opened 2020-01-14
clock: fix unused warning

yetist:


sc0w:
@yetist for more readability, I think you can put G_GNUC_UNUSED at the right instead left

I mean, for example:

```
weather_tooltip (GtkWidget  *widget G_GNUC_UNUSED,
                 gint        x G_GNUC_UNUSED,
                 gint        y G_GNUC_UNUSED,
                 gboolean    keyboard_mode G_GNUC_UNUSED,
                 GtkTooltip *tooltip,
                 gpointer    data)
```

instead

```
weather_tooltip (G_GNUC_UNUSED GtkWidget  *widget,
                 G_GNUC_UNUSED gint        x,
                 G_GNUC_UNUSED gint        y,
                 G_GNUC_UNUSED gboolean    keyboard_mode,
                 GtkTooltip *tooltip,
                 gpointer    data)
```

and the same for other changes

raveit65:
I see those warnings all over the place in all repos since we use `MATE_COMPILE_WARNINGS(maximum)`
https://github.com/mate-desktop/mate-panel/blob/master/configure.ac#L35
The `-Wunused-parameter` warning is part of `-Wextra`
https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
```
-Wextra

    This enables some extra warning flags that are not enabled by -Wall. (This option used to be called -W. The older name is still supported, but the newer name is more descriptive.)

    -Wclobbered  
    -Wcast-function-type  
    -Wdeprecated-copy (C++ only) 
    -Wempty-body  
    -Wignored-qualifiers 
    -Wimplicit-fallthrough=3 
    -Wmissing-field-initializers  
    -Wmissing-parameter-type (C only)  
    -Wold-style-declaration (C only)  
    -Woverride-init  
    -Wsign-compare (C only) 
    -Wstring-compare 
    -Wredundant-move (only for C++)  
    -Wtype-limits  
    -Wuninitialized  
    -Wshift-negative-value (in C++03 and in C99 and newer)  
    -Wunused-parameter (only with -Wunused or -Wall) 
    -Wunused-but-set-parameter (only with -Wunused or -Wall)
```
Isn't it better and much simple to replace `-Wextra` with single warnings but without `-Wunused-parameter` at mate-common?
https://github.com/mate-desktop/mate-common/blob/d404d8b45c48c912ce80d683d4406018779b52d3/macros/mate-compiler-flags.m4#L44

I mean adding `G_GNUC_UNUSED` to all repos is a lot of work and a bit weird when using the warning at mate-common.

raveit65:
Another option might be to use `MATE_COMPILE_WARNINGS(yes)` as default.

sc0w:
I think is good to see `G_GNUC_UNUSED` to know which parameters are unused but needed

and I think is good to see the warnings to be the code more clear preventing new unused and unneeded parameters in the future.

For example, the warnings were useful for me in this commit in mate-terminal to avoid unused parameters:

https://github.com/mate-desktop/mate-terminal/commit/2aeda2e5225f48021d5b287552e15617a33b690e

raveit65:
Guys, you know that better than me.
Are all this warnings here and in other repos false positives and it makes sense to ignore them?
Or shouldn't they fixed in another way?

> and I think is good to see the warnings to be the code more clear preventing new unused and unneeded parameters in the future.

Makes sense.
`--enable-compile-warnings=maximum` as compiler flag for Travis Ci

> For example, the warnings were useful for me in this commit in mate-terminal to avoid unused parameters:
https://github.com/mate-desktop/mate-terminal/commit/2aeda2e5225f48021d5b287552e15617a33b690e

Agree, but this can be handle like this, or not?
Using `MATE_COMPILE_WARNINGS(yes)` in configue.ac in general and compiling with `--enable-compile-warnings=maximum`

I'd like to find a compromise for all repos where all members of the team are happy with.

lukefromdc:
I don't know enough about this to know what is correct

monsta:
This looks less readable to me... and the warning is minor, why did they include it in `-Wextra`? :confused: 

Would adding `-Wno-unused-parameter` after `-Wextra` work? The manual says that each warning option has such "no" switch to turn it off.

sc0w:
`G_GNUC_UNUSED` is used in a lot of more places into other `mate-desktop` repositories, why not here?

@rbuj 

> plus it has not been used correctly

why? it can be used at the left or right

@yetist 

there are more unused parameters in the modified files:

```
clock.c:2170:46: warning: unused parameter 'context' [-Wunused-parameter]
clock.c:2175:34: warning: unused parameter 'error' [-Wunused-parameter]
clock-location.c:663:36: warning: unused parameter 'info' [-Wunused-parameter]
```

rbuj:
@sc0w I don't agree to use this macro, which is only for GCC, but for declaring variables put always the name on the right, you'll make happier the C preprocessor and compiler and static code analizers like splint ;-)

sc0w:
@rbuj you give us these warnings (`-Wunused-parameter`) since the commit https://github.com/mate-desktop/mate-panel/commit/06b9aac1847161f027546574d2ce4aed7eacbebd

I like to see these warnings to detect unused parameters, if you don't like `G_GNUC_UNUSED` which is your suggestion instead?

rbuj:
Remove the unused parameter from the signature/call as you did on https://github.com/mate-desktop/mate-terminal/commit/2aeda2e5225f48021d5b287552e15617a33b690e

sc0w:
@rbuj I didn't test it, but probably some parameters are unused but needed, and the build will fail if removed.

rbuj:
@sc0w At first glance, most of them belong to cb that are no longer used in their bodies / calls. If you silence those warnings we'll always have long headers keeping unused args.

vkareh:
According to the [documentation](https://developer.gnome.org/glib/stable/glib-Miscellaneous-Macros.html#G-GNUC-UNUSED:CAPS):

> For functions, place the attribute after the declaration, just before the semicolon. For arguments, place the attribute at the beginning of the argument declaration.
```
void my_unused_function (G_GNUC_UNUSED gint unused_argument,
                         gint other_argument) G_GNUC_UNUSED;
```

That said, GNOME and GTK both use it on the right, and I wonder if we should standardize to what GNOME/GTK use.

But this is all bikeshedding...

I think the real issue here is that this patch seems like an overuse of `G_GNUC_UNUSED`. I'm not a big fan of silencing warnings for the sake of silencing warnings... Sounds like the kind of thing that I'd like silenced locally, but exposed in Travis, so build flags might make the most sense.

raveit65:
Travis CI is using now mate-common-1.24.1 with `-Wno-unused-parameter` for maximum compile warning level.
https://github.com/mate-desktop/mate-panel/commit/b74d05995bc516d3b07d7724267897222d3a559b

PR #1060, opened 2020-03-02
Do not collect the translation for Icon

rbuj:
Test:
```shell
$ find . -name *.mate-panel-applet -exec cat {} +
$ cat mate-panel/mate-panel.desktop
```

### DIFF
--- a/applets/clock/Makefile.am
+++ b/applets/clock/Makefile.am
@@ -136,7 +136,7 @@ $(applet_in_files): $(applet_in_files).in Makefile
 		$< > $@
 
 $(applet_DATA): $(applet_in_files)
-	$(AM_V_GEN) $(MSGFMT) --desktop --keyword Name --keyword Description --template $< -d $(top_srcdir)/po -o $@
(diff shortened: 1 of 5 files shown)


applets/clock/Makefile.am
```diff
@@ -136,7 +136,7 @@ $(applet_in_files): $(applet_in_files).in Makefile
 		$< > $@
 
 $(applet_DATA): $(applet_in_files)
-	$(AM_V_GEN) $(MSGFMT) --desktop --keyword Name --keyword Description --template $< -d $(top_srcdir)/po -o $@
+	$(AM_V_GEN) $(MSGFMT) --desktop --keyword= --keyword=Name --keyword=Description --template $< -d $(top_srcdir)/po -o $@
```
sc0w: `--keyword= --keyword=Name`

typo?

rbuj: `--keyword=` do not to use the default keywords which include `Icon`

PR #1064, opened 2020-03-16
notification_area: Add do-not-disturb toggle menu item

rbuj:
closes https://github.com/mate-desktop/mate-notification-daemon/issues/166

VirtualBox_11 Fedora Rawhide_16_03_2020_15_35_33

### DIFF
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -58,6 +58,7 @@ struct _NaTrayAppletPrivate
   NAPreferencesDialog     *dialog;
   GtkBuilder              *builder;
 
+  GSettings               *notification_settings;
   GSettings               *settings;
   gint                     min_icon_size;
 
@@ -300,7 +301,6 @@ static const GtkActionEntry menu_actions [] = {
 	  G_CALLBACK (about_cb) }
 };
 
-
 static void
 na_tray_applet_realize (GtkWidget *widget)
 {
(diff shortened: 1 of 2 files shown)


raveit65:
I am not sure if this is really the right place for adding a menu entry.
This will gave us another runtime dependency to m-n-d.
What happens if m-n-d isn't installed?
I would prefer a little panel-applet which can open mate-notification-properties, and with a do-not-disturb-toggle-button (middle-click).
Any way, this is not a blocker by me, when the team like this implementation.

rbuj:
ping

raveit65:
Again,
> This will gave us another runtime dependency to m-n-d.
> What happens if m-n-d isn't installed?

rbuj:
@raveit65 do you mean to move org.mate.NotificationDaemon.gschema.xml.in into mate-settings-daemon?

raveit65:
Or better mate-desktop?
mate-panel depends already on mate-desktop, and mate-desktop includes a lot of other schemas.
https://github.com/mate-desktop/mate-panel/blob/master/configure.ac#L56
I know from people the don't use m-n-d because of conflicts with other notification-daemons, if several desktops are installed.
A MATE desktop without mate-deskop is impossible, so this might be the best place to move the schema.
It's clear that mate-panel with this PR will crash when org.mate.NotificationDaemon.gschema isn't installed.
So, adding mate-desktop dependency to m-n-d is not a big issue.
For some reasons current m-n-d has already build requires of mate-desktop-devel at travis, but i think this is wrong.
@mate-desktop/core-team 
Thoughts?

lukefromdc:
Any way to use a runtime check to see if the schema is installed or
whether m-n-d is installed? If not moving the schema to mate-desktop
sounds like a good fix too, and that also is lighter code with one
less runtime check. Would be best if the toggle is greyed out or not
shown when m-n-d is not installed. 

On 4/22/2020 at 10:00 AM, "raveit65"  wrote:
	Or better mate-desktop?
 mate-panel depends already on mate-desktop, and mate-desktop includes
a lot of other schemas.
 https://github.com/mate-desktop/mate-panel/blob/master/configure.ac#L56
 I know from people the don't use m-n-d because of conflicts with
other notification-daemons, if several desktops are installed.
 A MATE desktop without mate-deskop is impossible, so this might be
the best place to move the schema.
 It's clear that mate-panel with this PR will crash when
org.mate.NotificationDaemon.gschema isn't installed.
 So, adding mate-desktop dependency to m-n-d is not a big issue.
 For some reasons current m-n-d has already build requires of
mate-desktop-devel at travis, but i think this is wrong.
 @mate-desktop/core-team
 Thoughts?

lukefromdc:
Any updates on this? Is is ready to merge? This is a nice feature to have

raveit65:
This isn't ready to merge, see my comments.
Or do you want another extra dependency for mate-panel?
Mate-panel will crash if the gsettings key isn't installed.
This is minimal work to do.

Beside from this, it's better done in an new extra new applet for desktop notifications, with an enable/disable button and a menu entry to open notification preferences, IHMO.

rbuj:
It will be implemented in a new applet to prevent extra dependencies ASAP.

raveit65:
Cool, i think m-n-d is the right place for this applet.

kbsali:
Hi @rbuj , did you have a chance to work on an alternative solution?
I liked your initial proposal a lot ! :) 

I have installed a fresh copy of manjaro mate and I'm back with... 2 custom launchers to turn on/off "do not disturb"! :(

raveit65:
https://github.com/mate-desktop/mate-notification-daemon/pull/186
is merged.

PR #1066, opened 2020-03-19
Generate missing png icons from svg files

rbuj:
Before:

Captura de pantalla a 2020-03-19 11-22-18

After:

Captura de pantalla a 2020-03-19 12-38-47


sc0w:
I can't reproduce the issue with 1.24, and the icon looks better here

rbuj:
As you can see on caja, there are 4 svg files that are not rendered properly on Fedora 32 beta up to date:

VirtualBox_10 Fedora 32_19_03_2020_14_00_11

VirtualBox_10 Fedora 32_19_03_2020_14_00_33

rbuj:
Ubuntu MATE 20.04:

VirtualBox_Ubuntu 20 04_19_03_2020_14_17_42

raveit65:
I can confirm the issue with fedora 32 and MATE 1.24.
But not with fedora 30 and MATE 1.24, that's weird.....

raveit65:
Ok, adding the 256 px icons fixes the problem in dialogs.
But the issue still exists in add-to-panel dialog and the force-quit icon fails to display as panel icon.
Screenshot at 2020-03-25 18-28-56
Maybe re-generating the icons for all sizes helps?

rbuj:
I added the missing icons for Add To Panel dialog.
eyes & cpufreq: PR https://github.com/mate-desktop/mate-applets/pull/447

raveit65:
Thanks for quick reaction, adding 64 px icons fixes the missing icons in add-to-panel dialog.
But the force-quit icon fails to display itself as panel icon when panel size is < 32 px
Maybe we need to regenerate all sizes?
I am still wondering why this is needed for f32 and ubuntu 20.04.
There must be a change with a underlying program, maybe glib2 or gtk+?
With f30 and MATE 1.24 all is fine.
Mate 1.24 for f30 can be downloaded from copr repo https://copr.fedorainfracloud.org/coprs/raveit65/MATE-1.23.x/builds/
I will test mate-applets PR tomorrow.

raveit65:
It looks like there is a problem with `gdk-pixbuf-thumbnailer`.
After fresh installation in qemu-kvm with virtual-viewer. Browsing file-system with caja from terminal:
```
sudo caja
** (gdk-pixbuf-thumbnailer:1791): WARNING **: 13:13:38.910: Could not thumbnail 'file:///usr/share/icons/hicolor/scalable/apps/mate-panel-workspace-switcher.svg': XML parse error: error code=99 (3) in (null):4:65: xmlns:s: 'http://inkscape.sourceforge.net/DTD/s odipodi-0.dtd' is not a valid URI
```
Fixed by installing new generated SVGs from PR. This fixes also broken icons from dialogs with using non-hidpi resolution (1920x1080).

When using HIDPI resolution i noticed the broken icons in add-to-panel dialog and broken force-quit icon in panel.
But all those issues are fixed when installing new generated SVGs from this PR.
The other new icon sizes aren't needed to fix the problems in my test.
So, the issue is probably  caused by a change in `gdk-pixbuf-thumbnailer` in f32 and ubuntu 20.04.
@rbuj 
What do you think?
Should we only replace the SVGs to avoid blow up package size?

I will test now mate-applet PR.

rbuj:
Some SVG files couldn't be rendered because of an xml parse error:
```
XML parse error: error code=99 (3) in (null):4:65: xmlns:s: 'http://inkscape.sourceforge.net/DTD/s odipodi-0.dtd' is not a valid URI
```
Test: open `icons/scalable/mate-panel-window-menu.svg` using eom.
Screenshot at 2020-03-26 14-20-11

Including rendered png files at a certain size, it can't be considered as blow up package size, since they are loaded from a lookup table creted by `update-icon-cache` command. This replaces the rendering with searching and loading, which reduces the processing.

raveit65:
Agree, can you add icon-sizes for 48px and 96px to fix broken force-quit icon in panel please, without using SVGs?
Normal panel size use 48 px sicon size with HIDPI, and 96px icon-size for panel-size > 48px.

rbuj:
@raveit65 done.

PR #1068, opened 2020-04-06
build: Do not ship mate-panel-test-applets

rbuj:


lukefromdc:
Why do we want to stop building mate-panel-test-applets? (first reply did not get through)

rbuj:
Because it's a developer-oriented tool, not end-users. You can still build and use `mate-panel-test-applets` from your local build folder. You don't really need to install `mate-panel-test-applets` on your system to use it.

raveit65:
But than a developer needs to install it by hand instead of using a distro package.
Sounds for me like unnecessary extra work.
And there are people like me who use always the package-manager to install their RPM/APT's.
With this PR i need to build 2 versions, otherwise i can't use the test-applet.
I don't see the advantage of removing the applet from normal installation.

rbuj:
OK, I close the PR. From my point of view, it's only useful for debugging the applets using this launcher.

PR #1082, opened 2020-05-14
Change Back buton behavior

tamplan:
Hi,

Back button is hidden by default.
Selecting Add application launcher, the button is shown.
Add button icon is changed to "list-add" icon if user go back.

Regards

### DIFF
--- a/mate-panel/panel-addto.c
+++ b/mate-panel/panel-addto.c
@@ -863,6 +863,8 @@ panel_addto_dialog_response (GtkWidget *widget_dialog,
 		/* This response only happens when we're showing the
 		 * application list and the user wants to go back to the
 		 * applet list. */
+		gtk_button_set_image (GTK_BUTTON (dialog->add_button),
+				      gtk_image_new_from_icon_name ("list-add", GTK_ICON_SIZE_BUTTON));
 		panel_addto_present_applets (dialog);
 		break;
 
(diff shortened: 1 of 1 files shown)


raveit65:
We're showing the button when in application list and the user wants to go back to the applet list.
What is wrong with this behaviour and which behaviour you want to add?

tamplan:
Hi,
Just explain the behavior before in my previous post with this PR:
- Back button is hidden by default.
- Selecting Add application launcher, the button is shown.
- Add button icon is changed to "list-add" icon if user go back. (here there is a bug without this PR).

Regards

raveit65:
Do you really think i didn't read you first post?
This doesn't explain you motivation to change current behaviour!

tamplan:
Ok,
Let's try this:
- open mate-panel to add one applet
- choose a launcher
- press previous button
- the add button icon is now incorrect (unchanged)

First motivation is to hide an unactive button (previous button).

Please try this PR by yourself and give me feedback with the experience because my english seem to be unclear.

Regards.

raveit65:
@mate-desktop/core-team 
Maybe someone other from team, can handle this PR?

lukefromdc:
Current behavior on master is the back button inactive but visible most of the time, and can be used to go back from the list of application launchers that can be added. The "add" button normally shows a "+" sign except when selecting the "application launcher" menuitem, which changes it to the "go forward" arrow as clicking on it brings up the menu of addable applications. When that menu comes up, we get the "+" sign(list-add) again. On selecting "back" nothing is selected so both "back" and "add" are greyed out (disabled) but shown as in my opinion they should be.

lukefromdc:
The difference with this is hiding rather than greying out the back button when not in the list of applications. This is a less consistant UI, though there is never anything to go back to from the toplevel dialog. I can accept this either way, it is a very minor difference

raveit65:
> I will leave the decision on whether to make this UI change (very minor) to others.

I am undecided too.

selectiveduplicate:
I prefer the way it is now, showing greyed out instead of hiding completely; there's also the matter of consistency as @lukefromdc said. Just my opinion.

lukefromdc:
Closing as others seem happy with the UI as it stands, as do I. Can be reopened if others disagree, but we've got a LOT of PR's open now

PR #1095, opened 2020-06-20
Use '#include <config.h>' instead of '#include "config.h"'

rbuj:
The config.h file is not in the same folder that uses it, the preprocessor searches in -I folders, there is only one config.h file.

### DIFF
--- a/applets/clock/clock-utils.c
+++ b/applets/clock/clock-utils.c
@@ -24,7 +24,7 @@
  * Most of the original code comes from clock.c
  */
 
-#include "config.h"
+#include <config.h>
 
 #ifdef HAVE_LANGINFO_H
 #include <langinfo.h>
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -30,7 +30,7 @@
  *      Stefano Karapetsas
  */
 
-#include "config.h"
+#include <config.h>
 
 #include <stdio.h>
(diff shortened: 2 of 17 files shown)


sc0w:
I think every new change needs explanations, why this change?

https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Configuration-Headers.html

> With the appropriate -I option, you can use ‘#include <config.h>’. Actually, it's a good habit to use it, because in the rare case when the source directory contains another config.h, the build directory should be searched first.

rbuj:
```
$ ./autogen.sh --prefix=/usr && find . -name config.h
```

sc0w:
I remind you this commit https://github.com/mate-desktop/caja/commit/e50549cf502f988ff214a2714a8cdd6fb6892244 which you aren't agree and it was reverted later with https://github.com/mate-desktop/caja/commit/857eda84cab854b949eca40f7ace78d0a7f63d11

rbuj:
https://github.com/mate-desktop/caja/pull/1204#pullrequestreview-225186151

Code to avoid:
```
#include "../eel/eel-glib-extensions.h"
#include "../eel/eel-stock-dialogs.h"
```

sc0w:
and you are doing now the inverse

commit message must change

rbuj:
I changed the PR title but I didn't update the commit title.

Issue #1085, opened 2020-05-27
high memory usage - mate-panel


sandeep-raman:
#### Expected behaviour
mate-panel does not consume 13G of memory

#### Actual behaviour
mate-panel consumes 13G memory over two to three days of uptime

#### Steps to reproduce the behaviour
Nothing specific. The system runs on mate-desktop-environment and used for day-to-day activities like browsing, chat.

#### MATE general version
MATE Desktop Environment - 1.24.0

#### Package version
mate-panel-1.24.0-4.fc32.x86_64

#### Linux Distribution
Fedora release 32 (Thirty Two)
5.6.13-300.fc32.x86_64

#### Additional information about memory usage
free -h
              total        used        free      shared  buff/cache   available
Mem:           30Gi        27Gi       1.5Gi       319Mi       1.3Gi       2.1Gi
Swap:         4.4Gi        44Mi       4.3Gi

![image](https://user-images.githubusercontent.com/2987825/83002494-1fe55980-9fdb-11ea-9199-40f83625b58f.png)

sandeep-raman:
There seems to be a downstream report for this issue and building an in-process applet seems to solve the problem. https://bugzilla.redhat.com/show_bug.cgi?id=1830885

raveit65:
Re-opened because the panel leaks memory when build applets in-process.
This needs to be fixed here at MATE upstream.
In-process is needed to run the panel in a wayland compositor.
But it seems that na-tray applet leaks memory especially when wine+ emule-applet is running, see rhbz.
I can confirm this high memory consumption.
Usually the panel starts with 60MB and end up with 180-300 MB after ~12hours when i shutdown my box.
So, the panel has definitely a leak when build in-process.
I will switch back to out-of-process build for f32 (no wayland support) to work around the problem.
@mate-desktop/core-team 
This is really a no-go for wayland.
On user in rhbz reported about 14 GB memory consumption!!!

cwendling:
Agreed we should definitely fix this.  Unfortunately I don't think it's only the in-process/out-of-process case, because I'm running out-of-process currently and I see the issue from time to time (probably not as badly, but mate-panel grows over time, sometimes to problematic levels).
I'm trying a Valgrind run I'll leave on for a couple of days or so, and see if I can gather any useful info.

raveit65:
I updated fedora 32 mate-panel build to use out-process-applets.
https://bodhi.fedoraproject.org/updates/FEDORA-2020-9b86f77580
[...]

PR #1099, opened 2020-06-20
build: HAVE_X11/HAVE_WAYLAND should be used only when necessary

rbuj:


raveit65:
Honestly, i do not agree with breaking building mate-panel....
Everything was fine before.
Better check build log at travis.....

wmww:
Being able to independently enable/disable building with X11 and Wayland is useful. Both features require additional dependencies (Wayland requires `gtk-layer-shell` and X11 requires a number of things). If you only need support for one or the other, you should be able to only build support for that platform and not have to install the dependencies for the other one. I believe this PR would make that impossible.

If you did this to fix some specific problem you have then we might be able to figure out a better way, if it's just intended as code cleanup, I also recommend closing.

rbuj:
The files below use HAVE_X11:
```
$ git grep X11 | egrep "#include" | cut -d':' -f1 | sort -u
applets/notification_area/system-tray/na-tray-child.c
applets/notification_area/system-tray/na-tray-manager.c
libmate-panel-applet/mate-panel-applet.c
mate-panel/panel-action-protocol.c
mate-panel/panel-background-monitor.c
mate-panel/panel-force-quit.c
mate-panel/panel-multimonitor.c
mate-panel/panel-xutils.c
mate-panel/panel-xutils.h
mate-panel/xstuff.c
```
The files below use GDK_WINDOWING_X11:
```
$ git grep gdk_x11 | cut -d':' -f1 | sort -u
applets/clock/clock.c
applets/notification_area/system-tray/fixedtip.c
applets/notification_area/system-tray/na-tray.c
applets/notification_area/system-tray/na-tray-child.c
applets/notification_area/system-tray/na-tray-manager.c
applets/notification_area/testtray.c
applets/wncklet/showdesktop.c
applets/wncklet/window-list.c
applets/wncklet/wncklet.c
libmate-panel-applet/mate-panel-applet.c
mate-panel/libmate-panel-applet-private/panel-applet-container.c
mate-panel/panel-action-protocol.c
mate-panel/panel-applet-frame.c
mate-panel/panel-background-monitor.c
mate-panel/panel-force-quit.c
mate-panel/panel-layout.c
mate-panel/panel-profile.c
mate-panel/panel-run-dialog.c
mate-panel/panel-session.c
mate-panel/panel-struts.c
mate-panel/panel-xutils.c
mate-panel/xstuff.c
```
HAVE_X11 is not equal to GDK_WINDOWING_X11.

wmww:
You are correct that they are not the same. As I said in the [other thread](https://github.com/mate-desktop/mate-submodules/pull/11#discussion_r443161647):

> To clarify, `HAVE_X11` **guarantees** `GDK_WINDOWING_X11` is true. If `HAVE_X11` is true ever true and `GDK_WINDOWING_X11` is not then the build system is broken. `HAVE_X11`/`HAVE_WAYLAND` are useful because it allows us to disable one platform or the other even if GTK supports it. We might want to do that for testing, or to build without our additional dependencies required for that platform.

If we were making it consistent we should probably make everything use `HAVE_X11`/`HAVE_WAYLAND`, but as long as it's not causing any problems I think there's more important stuff to do. Specifically in cases where no 3rd party dependencies are involved, the two can be safely used interchangeably. When platform-specific 3rd party dependencies are being used then the `HAVE_*` macros are preferable because they can be turned off.

wmww:
`GDK_WINDOWING_*` is if **GTK** supports a platform and `HAVE_*` is if the **MATE** app supports a platform. MATE can not support a platform unless GTK does, but should not be required to support **all** platforms supported by the GTK installed. Therefor, the `HAVE_*` macros are preferable.

rbuj:
You are misusing macros, which makes it difficult to maintain the code, as looking at the code I do not know what functions are used, gdk_x11_* , x11 or both.

Why do you use HAVE_WINDOW_PREVIEWS in wncklet applet?
```
$ pkg-config --libs  gtk+-3.0 libwnck-3.0 mate-desktop-2.0
-lwnck-3 -lmate-desktop-2 -lgtk-3 -lgdk-3 -lpangocairo-1.0 -lpango-1.0 -lharfbuzz -latk-1.0 -lcairo-gobject -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lgobject-2.0 -lglib-2.0 -lstartup-notification-1 
$ git grep -A 1 HAVE_WINDOW_PREVIEWS | grep include
applets/wncklet/window-list.c-#include <gdk/gdkx.h>
```

@wmww  Why do you do a double check using ENABLE_X11 in Makefile.am and using HAVE_X11 in source files?
```
$ git grep "#error file should only" 
applets/notification_area/main.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/notification_area/system-tray/fixedtip.c:#error file should only be compiled when GDK_WINDOWING_X11 is enabled
applets/notification_area/system-tray/na-tray-child.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/notification_area/system-tray/na-tray-child.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/notification_area/system-tray/na-tray-manager.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/notification_area/system-tray/na-tray.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/notification_area/system-tray/na-tray.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/wncklet/showdesktop.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
applets/wncklet/wncklet.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
libmate-panel-applet/panel-plug-private.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
libmate-panel-applet/panel-plug.c:#error file should only be included when GDK_WINDOWING_X11 is enabled
mate-panel/panel-action-protocol.c:#error file should only be built when GDK_WINDOWING_X11 is enabled
mate-panel/panel-action-protocol.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
mate-panel/panel-background-monitor.c:#error file should only be compiled when GDK_WINDOWING_X11 is enabled
mate-panel/panel-background-monitor.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
mate-panel/panel-force-quit.c:#error file should only be built when GDK_WINDOWING_X11 is enabled
mate-panel/panel-force-quit.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
mate-panel/panel-struts.c:#error file should only be compiled when GDK_WINDOWING_X11 is enabled
mate-panel/panel-struts.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
mate-panel/panel-xutils.c:#error file should only be built when GDK_WINDOWING_X11 is enabled
mate-panel/panel-xutils.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
mate-panel/wayland-backend.h:#error file should only be included when HAVE_WAYLAND is enabled
mate-panel/xstuff.c:#error file should only be built when GDK_WINDOWING_X11 is enabled
mate-panel/xstuff.h:#error file should only be included when GDK_WINDOWING_X11 is enabled
```

wmww:
> I do not know what functions are used, gdk_x11_* , x11 or both.

Does it matter enough that you need a macro to tell you which are being used? If you really care you can always just check the implementation. I think the important thing is to know if a function or section of code depends on something X11-specific, not what functions it uses to access that

> Why do you use HAVE_WINDOW_PREVIEWS in wncklet applet?

I have no idea. Probably because some specific dependency is needed to implement that

>  Why do you do a double check using ENABLE_X11/ENABLE_WAYLAND

Did that so it's obvious that it's safe to use platform-specific code without additional checks. Without the check we might get strange/unhelpful errors if the file is accidentally included in the wrong type of build. Developers might also assume that we're missing per-function X11 compatibility checks and try to add them, which would only add confusion.

wmww:
Unless there's an actual problem you're trying to solve, I don't think this is worth any of our time. There's a lot of legacy code in MATE, and most can and should be left alone until it becomes a problem. A far more productive line of work would be to port another app to Wayland.

raveit65:
It is absolute necessary to have an option to enable/disable x11/wayland.
Why?
Because we have a big problem with a very huge memleak when building applets in-process for wayland compositor, see 
https://github.com/mate-desktop/mate-panel/issues/1085
https://bugzilla.redhat.com/show_bug.cgi?id=1830885
It was the second time that i had to revert the switch to enable wayland for fedora, to avoid tons of bugreport from fedora users. First when f31 was released last year and now again with f32.
In result only unreasonable people wants to build mate-panel with both option enabled at the moment.

I think it makes more sense to fix this actual problem, which is a ***NO GO*** for mate-panel transition to wayland, instead wasting your/my/our energy with an unnecessary trouble about an unneeded code change, ihmo.
And you didn't gave any reason in this discussion why it is necessary to change a working code.

rbuj:
Now I'm totally confused, and I don't know what is the current criterion.
I close it for lack of consensus.

See https://github.com/mate-desktop/pluma/pull/532#issuecomment-613547831

PR #1104, opened 2020-06-24
panel-util: avoid X11-specific header

jbeich:
When building with `--disable-x11` (Wayland-only) one X11-related header appears to be included outside of `#ifdef HAVE_X11`. However, Gtk+ 3 only installs the header if built with `--enable-x11-backend`.

### DIFF
--- a/mate-panel/panel-util.c
+++ b/mate-panel/panel-util.c
@@ -28,7 +28,7 @@
 #include <glib/gstdio.h>
 #include <gio/gio.h>
 #include <gdk-pixbuf/gdk-pixbuf.h>
-#include <gdk/gdkx.h>
+#include <gdk/gdk.h>
 
 #define MATE_DESKTOP_USE_UNSTABLE_API
 #include <libmate-desktop/mate-desktop-utils.h>


raveit65:
Commit message is misleading. Why in gtk3?
MATE use only gtk3 since 1.20.

rbuj:
It shoud be `#include <gdk/gdk.h>` instead of `#include <gdk/gdkx.h>`. Eclipse's Organize Imports output:
```
<cut>
#include <gdk/gdkcairo.h>
#include <gdk/gdkcursor.h>
#include <gdk/gdkdisplay.h>
#include <gdk/gdktypes.h>
#include <gdk/gdkwindow.h>
<cut>

rbuj:
@jbeich  Please, change the subjects of the pull request and the commit message to make easier the tracking at first sight.

For instance, you can add the name of the modified file, since only one file is modified. Next, you can add a summary of the change that other developers can uniquely understand.

`panel-util: summary`

Try to keep commit message subject under 72 characters.

jbeich:
@rbuj, is it good enough now or do PR description and commit message have to match?

PR #1106, opened 2020-06-25
panel-multimonitor: Do not require limits.h macros

rbuj:


raveit65:
Can you explain why this `limits.h macros` aren't needed?
What is the advantage?
Please call reviewer for yourself, if you want that someone will be notified.

rbuj:
The values of the `INT_MIN` and `INT_MAX` macros are defined in `limits.h` header, but this file is not included, ie `#include <limits.h>` is missing. In addition, `min` and `max` variables are GdkPoint type, and y and x are gint type, as shown below:
```C
struct GdkPoint {
  gint x;
  gint y;
};
```

PR #1114, opened 2020-07-22
clock: Use single-linked lists for locations

rbuj:


rbuj:
All seek movements are forward, so it makes no sense to use a double-stranded list. Using a single-linked list instead of a double-linked list reduces the ram memory consumption by using a smaller data structure for each node, reducing the number of operations required to manage the node itself. Both lists use [megesort](https://en.wikipedia.org/wiki/Merge_sort) for sorting the nodes, O(n log n) and search is O(n).

I also think sorting and reversing operations can be merged:

```C
        /* Copy the existing list, so we can sort it nondestructively */
        node = g_slist_copy (cities); /* O(n) */
        node = g_slist_sort (node, sort_locations_by_time); /* O(n) */
        node = g_slist_reverse (node); /* O(n) */
```

```C
        /* Copy the existing list, so we can sort it nondestructively */
        node = g_slist_copy (cities); /* O(n) */
        node = g_slist_sort (node, sort_locations_by_time_descending); /* O(n) */
```

tbzatek:
Yeah, looks good, though the memory savings will be negligible in the big picture.

lukefromdc:
Is this ready to be merged?

lukefromdc:
I take it this is still cooking with the WIP?

rbuj:
Now it's ready ;)

I resolved the merge conflict, I edited the files clock-map.c and clock-map.h, and I replaced two functions which were recently added and they caused a segfault after adding/removing locations.

lukefromdc:
Just deconflicted and rebased again after 
https://github.com/mate-desktop/mate-panel/pull/1117 which appears to have been split out from this. Did not catch that before merging

lukefromdc:
During a final pre-merge test I got a crash, but the backtrace went through libwnck (and I have the hidpi PR version of that) so cannot be related to the clock in any way, thus cannot be related to this PR. Adding and removing locations works fine with this, same as in my last test.

Issue #955, opened 2019-04-15
panel context menu "reset panel" is unclear


alete89:
#### Expected behaviour
If I right click on one of many panels and read "Reset Panel", and then a dialog that says "Reset this panel?", I expect only THAT panel to be reset.

#### Actual behaviour
even when the dialog says "Reset this panel?", all panels are reset.
![panel](https://user-images.githubusercontent.com/4999277/56145957-babbd980-5f7b-11e9-9c1d-6fcfc557df51.png)
![Screenshot at 2019-04-15 12-40-14](https://user-images.githubusercontent.com/4999277/56145979-c7403200-5f7b-11e9-9759-f1d07fcf6e2b.png)

#### Steps to reproduce the behaviour
Right click any panel, select reset panel, confirm reset this panel => all your panels are reset.

#### MATE general version
1.20.4

#### Package version
debian 9 backports

#### Linux Distribution
debian stretch

Might be related with #844
I propose either change the labels and warn the user that all the panels are going to be reset.
Or better yet, be able to reset one panel at a time.

lukefromdc:
The underlying code that the GUI option to reset panels invokes does not support resetting one panel, it resets all of mate-panel. Also at reset the panel(s) are restored to an original default, which in all distros is for all of mate-panel. Multiple panels are still one instance of mate-panel.

Thus, the label should probably be changed to "Reset All Panels" or something along those lines.

PR #1116, opened 2020-07-23
Clarify meaning of panel reset option

lukefromdc:
Change "Reset Panel" to "Reset all Panels" in context mnenu and in reset confirmation dialog to make clear that ALL panels will be reset. Closes https://github.com/mate-desktop/mate-panel/issues/955


lukefromdc:
Note that I had to use the web interface for this,I was unable to push to ANY github repo due to a 403 error that I was not able to fix:
```
remote: Temporary clone tokens are read-only
fatal: unable to access  'https://github.com/mate-desktop/mate-panel.git/": The requested URL returned error: 403
```

rbuj:
cherry-picked to stable 1.24 branch

PR #1113, opened 2020-07-22
Memory leak fixes pt. 2

tbzatek:
Another bunch of memleak fixes found by compiling applets in-process. The fixes from #1112 are not included in this branch.

I'm still seeing increasing memory consumption about 1MB an hour but nothing dramatic otherwise.

PR #1118, opened 2020-07-24
clock: Avoid double release of error

### LINKED ISSUE #1115: clock: dismissing priviliges dialog for changing time zone in calendar crashes mate-panel w a segfault
#### Expected behaviour
When setting a new home location in the calendar's "locations" section in a different timezone, the "priviliges required" dialog can be dismissed by clicking the "cancel" button, cancelling the change of home location

#### Actual behaviour
When setting a new home location in the calendar's "locations" section in a different timezone, dismissing the "priviliges required" dialog crashes mate-panel with a segfault. In some but not all cases, entering your sudo password/passphrase to confirm the change will also trigger this crash.
```

0x00007ffff6f3f4bd in g_slice_alloc () from /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
```

Open the calendar, add at least  location outside your timezone. It will not have the "home" icon. Hover over the right side of the center of the location in the calendar, so the "set" button appears. Click it, and the "Priviliges are requires to change the system time zone" dialog appears. To ensure the crash, click "cancel" and the panel will crash everytime

#### MATE general version
1.25

#### Package version
Locally compiled mate-panel 1.25+git20200713

#### Linux Distribution
Debian Unstable last updated 7-13-2020. Note that I have locally compiled gtk 3.24.1 and libglib 2.64.3 and this looks like a glib problem unless the backtrace shows we are calling g_slice_alloc incorrectly.

None as this is a locally compiled package

rbuj:
closes #1115

### DIFF
--- a/applets/clock/set-timezone.c
+++ b/applets/clock/set-timezone.c
@@ -192,7 +192,6 @@ set_time_notify (GObject *source_object,
 		if (error != NULL) {
 			if (data->callback)
 				data->callback (data->data, error);
-			g_error_free (error);
 		} else {
 			if (data->callback)
 				data->callback (data->data, NULL);


rbuj:
fixed on #1113

tbzatek:
Ah sorry, didn't notice you posted a patch already. Great that we found it at the same place.

PR #1125, opened 2020-08-04
panel-util: Fix memory leak

rbuj:


rbuj:

PR #1129, opened 2020-08-04
Fix creating several launchers at once

cwendling:
When creating a new launcher, we try and find an available object ID by browsing the existing ones.  This however fails to take into account objects that were created but that didn't make it to the backend yet, leading to likely re-using the same ID when creating several objects roughly at the same time.  This happens e.g. when dropping several URIs together.

# Test

Drag 2 or more **non-image** files from Caja over to a panel.

## Actual behavior (without this patch)

Only one launcher gets created for the the last URI.

## Expected behavior (with this patch)
A launcher gets created for each URI.


rbuj:

PR #1130, opened 2020-08-04
Ask the user what to do when receiving an image drop

cwendling:
Instead of unconditionally setting the panel's background when an image file URI gets dropped on a panel, ask the user whether to create a launcher for that URI (like for non-image URIs) or to use the image as the panel's background.

A number of users have seen their panel's background changed unexpectedly due to unwanted drop of an image file over the panel, sometimes seeing this as a panel's bug (I personally know 2, at least a couple others have popped here and there).

Also, this behavior is inconsistent with dropping a URI pointing to any other type of file, where it would create a launcher for it.

Consequently, and as setting the panel's background doesn't seem like a so common task it ought to be super fast, ask the user what to do when receiving an image URI drop.

Also, when a drop consists of more than one URI, just create launchers without asking because it doesn't make sense to set the background more than once at a time.

### DIFF
--- a/mate-panel/panel.c
+++ b/mate-panel/panel.c
@@ -604,10 +604,12 @@ drop_urilist (PanelWidget *panel,
 	      char        *urilist)
 {
 	char     **uris;
+	guint      n_uris;
 	gboolean   success;
 	int        i;
 
 	uris = g_uri_list_extract_uris (urilist);
+	n_uris = g_strv_length (uris);
 
 	success = TRUE;
 	for (i = 0; uris[i]; i++) {
@@ -657,8 +659,8 @@ drop_urilist (PanelWidget *panel,
 			can_exec = g_file_info_get_attribute_boolean (info,
 								      G_FILE_ATTRIBUTE_ACCESS_CAN_EXECUTE);
 
-			if (mime &&
-			    g_str_has_prefix (mime, "image")) {
+			if (n_uris == 1 &&
+			    mime && g_str_has_prefix (mime, "image")) {
 				if (!set_background_image_from_uri (panel->toplevel, uri))
 					success = FALSE;
 			} else if (mime &&
--- a/mate-panel/panel.c
+++ b/mate-panel/panel.c
@@ -598,6 +598,109 @@ drop_caja_desktop_uri (PanelWidget *panel,
 	return success;
 }
 
+enum {
+	RESPONSE_SET_BACKGROUND,
+	RESPONSE_CREATE_LAUNCHER
+};
+
+static void
+image_drop_dialog_response (GtkWidget   *dialog,
+                            int          response,
+                            PanelWidget *panel)
+{
+	if (response == RESPONSE_SET_BACKGROUND) {
+		const gchar *uri = g_object_get_data (G_OBJECT (dialog), "uri");
+
+		set_background_image_from_uri (panel->toplevel, uri);
+	} else if (response == RESPONSE_CREATE_LAUNCHER) {
+		const gchar *uri = g_object_get_data (G_OBJECT (dialog), "uri");
+		gint pos = GPOINTER_TO_INT (g_object_get_data (G_OBJECT (dialog), "pos"));
+
+		drop_uri (panel, pos, uri, PANEL_ICON_UNKNOWN);
+	}
+
+	gtk_widget_destroy (dialog);
+}
+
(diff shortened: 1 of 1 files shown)


mate-panel/panel.c
```diff
@@ -657,8 +657,8 @@ drop_urilist (PanelWidget *panel,
 			can_exec = g_file_info_get_attribute_boolean (info,
 								      G_FILE_ATTRIBUTE_ACCESS_CAN_EXECUTE);
 
-			if (mime &&
-			    g_str_has_prefix (mime, "image")) {
+			if (uris[1] == NULL /* if there's only one URI */ &&
```
rbuj, line 763: It's better to use [g_strv_length](https://developer.gnome.org/glib/stable/glib-String-Utility-Functions.html#g-strv-length) instead of `uris[1] == NULL` for a NULL terminated string array. (readability)

mate-panel/panel.c
```diff
@@ -598,6 +598,109 @@ drop_caja_desktop_uri (PanelWidget *panel,
...
+
+	panel_toplevel_push_autohide_disabler (toplevel);
+
+	return dialog;
+}
+
+static void
+ask_about_image_drop (PanelWidget *panel,
+		      int          pos,
+		      const char  *uri)
```
rbuj, line 689: indent

mate-panel/panel.c
```diff
@@ -598,6 +598,109 @@ drop_caja_desktop_uri (PanelWidget *panel,
...
 
+enum {
+	RESPONSE_SET_BACKGROUND,
+	RESPONSE_CREATE_LAUNCHER
+};
+
+static void
+image_drop_dialog_response (GtkWidget   *dialog,
+			    int          response,
+			    PanelWidget *panel)
```
rbuj, line 609: indent

mate-panel/panel.c
```diff
@@ -598,6 +598,109 @@ drop_caja_desktop_uri (PanelWidget *panel,
...
+
+		drop_uri (panel, pos, uri, PANEL_ICON_UNKNOWN);
+	}
+
+	gtk_widget_destroy (dialog);
+}
+
+static GtkWidget *
+image_drop_dialog (PanelToplevel *toplevel,
+		   const gchar   *uri)
```
rbuj, line 627: indent

mate-panel/panel.c
cwendling: ?  This file consistently uses tabs of width 8, and spaces for alignment.  AFAICT, this aligns nice with those settings.

mate-panel/panel.c
cwendling: OK, changed to use it.  We could then user the length in the loop, but that's kinda out of scope.

mate-panel/panel.c
rbuj: We no longer use tabs for indentation. You can use N tabs and the spaces needed to align the N indent level. The indent level 0 has no tab. You can also expand indent tabs to 8 spaces per indent: by using the command `expand -i -t 8 file`.

sc0w: @rbuj  the indentation is ok, see https://github.com/mate-desktop/mate-panel/blob/eff98bfc06d0e55f9c375fe9ebd48328b5e1e913/mate-panel/panel.c

it respects old style

mate-panel/panel.c
sc0w: @rbuj  the indentation is ok, see https://github.com/mate-desktop/mate-panel/blob/eff98bfc06d0e55f9c375fe9ebd48328b5e1e913/mate-panel/panel.c

it respects old style

mate-panel/panel.c
sc0w: @rbuj  the indentation is ok, see https://github.com/mate-desktop/mate-panel/blob/eff98bfc06d0e55f9c375fe9ebd48328b5e1e913/mate-panel/panel.c

it respects old style

mate-panel/panel.c
cwendling: I updated the commit to use spaces for alignment, but kept the indentation using tabs to fit the file.  This should fix what you were complaining about, and it mostly fits the current style of this file.

If it's wanted to change the style, I think it's better suited in a separate PR that would only do that with zero functional changes.

PR #1131, opened 2020-08-04
wncklet: Hide window preview on non-visible windows

### LINKED ISSUE #1127: Window thumbnails sometimes show really glichy preview + 3 errors
#### Expected behaviour
Show a window thumbnail preview of a window when I point my cursor on the opened application.

#### Actual behaviour
As an example, I used Telegram. However it does happen with every application if I can recall it.
Screenshot at 2020-08-04 15-44-26
Screenshot at 2020-08-04 15-44-48
Screenshot at 2020-08-04 15-45-00

It happens always just by opening few tabs and switching to minimise window and stuff. I noticed that it does often happen on Firefox, MATE-Terminal, Konversation and Telegram.

#### MATE general version
1.24.0

#### Linux Distribution
Fedora 32

Sorry, I do not know how to fill other bars. And sorry if I repeat the same issue. I don't seem to find similar issue.

As I am writing this, I also encountered this error. I left the window list preference window open and put my mouse on the panel few times. Three errors at the same time.
Screenshot at 2020-08-04 15-57-03

vkareh:
When a window is not visible we cannot capture its thumbnail. We used to
store the thumbnail when a window was visible so that we could reuse it
when minimized, but this can cause visual glitches.

Fixes #1127

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -293,6 +293,11 @@ static gboolean applet_enter_notify_event (WnckTasklist *tl, GList *wnck_windows
 	if (wnck_window == NULL)
 		return FALSE;
 
+	/* Do not show preview if window is not visible nor in current workspace */
+	if (!wnck_window_is_visible_on_workspace (wnck_window,
+						  wnck_screen_get_active_workspace (wnck_screen_get_default ())))
+		return FALSE;
+
 	thumbnail = preview_window_thumbnail (wnck_window, tasklist);
 
 	if (thumbnail == NULL)


rbuj:
I'm sorry, I couldn't reproduce the issue using Fedora 32 + [Telegram 2.2 (snapd)](https://snapcraft.io/install/telegram-desktop/fedora), with and without software compositor enabled.

I don't know... maybe it's because of the graphics card driver.

lukefromdc:
Letting someone who has the glitches test this. If nobody does I will proceed to merge this, as hiding previews of minimized windows is rational anyway.

raveit65:
@rbuj 
Reproducing is very easy. Work with several workspaces, let say use 8.
1. Open firefox (or your favourite browser) at let's say workspace 2
2. Open thunderbird or your favourite email program at workspace 8
3. Open an email in thunderbird with web links, maybe a github notification mail
4. click on the link in mail to open it in your webbroser

The firefox preview at workspace 8 is distorted.
Screenshot workspace 8:
previews

rbuj:
Then it would be better to check the flag [WNCK_WINDOW_STATE_HIDDEN](https://developer.gnome.org/libwnck/stable/WnckWindow.html#WnckWindowState)

vkareh:
Good call @rbuj - I'll try it out with that flag

rbuj:
@vkareh i don't think the thumbnail should be updated if the window is below another (WNCK_WINDOW_STATE_BELOW), otherwise dark areas will come out.

vkareh:
@rbuj - can you reproduce the dark areas thing on below windows? I just tested that and it works:
1. open firefox and mate-calc on top
2. go to different workspace and trigger a URL open from terminal
3. navigate to original workspace (without activating window)
4. preview of below window shows correctly
Screenshot at 2020-08-05 07-47-13

raveit65:
Actual state of PR fixes the issue when target window is at different workspace.
And i can't reproduce black previews when window is below a mximized window in a quick test.

rbuj:
I think it prevents showing thumbnails from other workspaces. I don't know if you can keep the last thumbnail and only grab it when the window is not hidden. Note the code below may cause memory leaks, it must be read as a pseudocode :)
```C
static gboolean applet_enter_notify_event (WnckTasklist *tl, GList *wnck_windows, TasklistData *tasklist)
{
        GdkPixbuf *thumbnail;
        WnckWindow *wnck_window = NULL;
        int n_windows;

        if (!tasklist->show_window_thumbnails || wnck_windows == NULL)
                return FALSE;

        n_windows = g_list_length (wnck_windows);
        /* TODO: Display a list of stacked thumbnails for grouped windows. */
        if (n_windows == 1)
        {
                GList* l = wnck_windows;
                if (l != NULL)
                        wnck_window = (WnckWindow*)l->data;
        }

	if (wnck_window == NULL)
                return FALSE;

        /* Do not grab a new preview if window is not visible */
        if (wnck_window_get_state (wnck_window) & WNCK_WINDOW_STATE_HIDDEN)
        {
                if (tasklist->preview != NULL)
                        gtk_widget_show (tasklist->preview);
                return FALSE;
        }

	/* Take a new window screenshot */
        if (tasklist->preview != NULL)
        {
                gtk_widget_destroy (tasklist->preview);
                tasklist->preview = NULL;
        }
	thumbnail = preview_window_thumbnail (wnck_window, tasklist);

        if (thumbnail == NULL)
                return FALSE;

        /* Create window to display preview */
        tasklist->preview = gtk_window_new (GTK_WINDOW_POPUP);

        gtk_widget_set_app_paintable (tasklist->preview, TRUE);
        gtk_window_set_resizable (GTK_WINDOW (tasklist->preview), TRUE);

        preview_window_reposition (tasklist, thumbnail);

        gtk_widget_show (tasklist->preview);

        g_signal_connect_object (G_OBJECT (tasklist->preview), "draw", G_CALLBACK (preview_window_draw), thumbnail, 0);

        return FALSE;
}
```

vkareh:
The problem with showing windows from other workspaces is that the thumbnail cannot be generated, and we don't keep a preview for _each_ window, we only keep the latest viewed one. This is what is causing the original issue in #1127 

Your code is fine and it works (I did some tweaking on my side) but it doesn't fix the issue unless we _also_ skip the previews from other workspaces

vkareh:
okay, after testing various iterations, I don't see a consistent way of reusing thumbnails from memory without having a code path that leads to visual glitches...

For now, hiding thumbnails from windows that are minimized/hidden or in a different workspace works well in all scenarios and fixes the bug. We can attempt a different system for improving on other workspaces as a future enhancement... Thoughts?

vkareh:
Explored a bit and to get a thumbnail for a different workspace requires loading the compositor libraries. Not worth it for this applet, to be honest.

We can explore pre-loading the thumbnails into a list whenever the workspace switches, but with many workspaces and windows this can become a memory hog. I currently have 16 windows open across 6 workspaces (and today is a pretty normal day) and most of them are maximized... at HiDPI resolutions this would end up in the order of a few GB of memory... not worth it either...

raveit65:
I am using minimum 20 open windows on 8 workspace as a normal daily workload.
Default settings for panel is to show only windows of current workspaces in panel, right?
I never changed that for myself because seeing more than 20 open windows in panel drives me crazy :)
Not sure, but i am thinking that is the most common usage.
So, it might be alright to display only previews of current workspace.
When we decided to load the compositor to display previews from another workspaces we should add a warning to preferences.
Enable `show windows from all workspace with previews can lead up in huge memory consumption",
which might be Ok on systems > 16 - 32 GB memory.
@vkareh 
Can you send me a patch to test how much memory the panel will use with > 20 open windows?

rbuj:
@raveit65 do you mean something like that?
```C
        /* Do not show preview if window is not visible on the current workspace */
        if (! wnck_window_is_visible_on_workspace (wnck_window,
                                                   wnck_screen_get_active_workspace (wnck_screen_get_default ())))
        {
                return FALSE;
        }
```

raveit65:
For me it is Ok to display only previews of current workspaces and avoid to use more memory for this feature.
But this is a team decision.

offtopic on
It looks like that previous are blurred on HIDPI, or not?
offtopic off

vkareh:
@raveit65 
> Can you send me a patch to test how much memory the panel will use with > 20 open windows and previews for all workspaces enabled?

Adding the compositor code for this would be a lot of work. I started a bit, but decided to stop when I needed to change things in the build process... If you want to test _without_ the compositor code, then you can just put all the windows in the same workspace.

> It looks like that previous are blurred on HIDPI, or not?

Yes, as they use `GdkPixbuf` rather than `cairo_surface_t`. This can be fixed at some point.

@rbuj 
> wnck_window_is_visible_on_workspace

Good find - explored the code on that one as it does _exactly_ the same as my code... I'm replacing it

raveit65:
> Adding the compositor code for this would be a lot of work. I started a bit, but decided to stop when I needed to change things in the build process... If you want to test without the compositor code, then you can just put all the windows in the same workspace.

Np, i fine with minimal solution, and there are other things to do, eg. open libwnck PR.

vkareh:
@raveit65
> open libwnck PR

I know... this has been at the top of my priority list for non-work stuff. It's probably the main thing blocking me from doing anything else in MATE (apart from being busy with work...) :/

raveit65:
@vkareh @rbuj 
Do you think it is possible to preview windows from same workspace but from a second/third Monitor without visual glitches?

vkareh:
Yep! Never had an issue with that:
Screenshot at 2020-08-05 14-56-14

raveit65:
Btw. i can't reproduce any memory leak with current master + PR and using previews.
Memory consumption goes back to normal after viewing the preview here at my box.

raveit65:
@vkareh 
Sorry for that noise.
My fault, i was watching a video with chromium in full-screen modus on second monitor.
With normal chromium window it works well :)

rbuj:

PR #1133, opened 2020-08-04
Handle locked down settings better when dropping an image URI (#1130 follow-up)

cwendling:
Follow-up to #1130.

Don't propose actions that will fail anyway, and create a launcher automatically if it's the only option.  However, still ask for setting
the background to still avoid unintentional changes.

To test this, you need to [lock down](https://help.gnome.org/admin/system-admin-guide/stable/dconf-lockdown.html.en) the relevant settings (*/org/mate/panel/toplevels/top/background/image* and */org/mate/panel/general/object-id-list*).  **BEWARE** though: I somehow lost all my panel config by locking */org/mate/panel/general/object-id-list*, likely because I didn't hardcode a default value and dconf sudently reported an empty object-id-list, resulting in the panel removing a bunch of objects.  I partially recovered from a backup, but meh, still no fun.  If you wanna try, I highly recommend dumping the panel configuration first (`dconf dump /org/mate/panel/ > backup`) in order to easily restore it.  And still, test at your own risks.
Testing the background is less scary as you'll likely only lose the image path itself, and you might remember which background you used.

### DIFF
--- a/mate-panel/panel.c
+++ b/mate-panel/panel.c
@@ -414,16 +414,35 @@ panel_key_press_event (GtkWidget   *widget,
 }
 
 static gboolean
+can_set_background_image (PanelToplevel *toplevel,
+                          const gchar   *uri,
+                          gchar        **filename)
+{
+	gchar *image;
+
+	if (! panel_profile_background_key_is_writable (toplevel, "type") ||
+	    ! panel_profile_background_key_is_writable (toplevel, "image"))
+		return FALSE;
+
+	image = g_filename_from_uri (uri, NULL, NULL);
+	if (! image)
+		return FALSE;
+
+	if (filename)
+		*filename = image;
+	else
+		g_free (image);
+
+	return TRUE;
+}
+
+static gboolean
 set_background_image_from_uri (PanelToplevel *toplevel,
 			       const char    *uri)
 {
 	char *image;
 
-	if ( ! panel_profile_background_key_is_writable (toplevel, "type") ||
-	     ! panel_profile_background_key_is_writable (toplevel, "image"))
-		return FALSE;
-
-	if (!(image = g_filename_from_uri (uri, NULL, NULL)))
+	if (! can_set_background_image (toplevel, uri, &image))
 		return FALSE;
 
 	panel_profile_set_background_image (toplevel, image);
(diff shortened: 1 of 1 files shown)


mate-panel/panel.c
```diff
@@ -413,17 +413,36 @@ panel_key_press_event (GtkWidget   *widget,
...
+		return FALSE;
+
+	image = g_filename_from_uri (uri, NULL, NULL);
+	if (! image)
+		return FALSE;
+
+	if (filename)
+		*filename = image;
+	else
+		g_free (image);
```
rbuj, line 434: You don’t really need this if you use `*filename` directly.

mate-panel/panel.c
```diff
@@ -413,17 +413,36 @@ panel_key_press_event (GtkWidget   *widget,
 	return FALSE;
 }
 
+static gboolean
+can_set_background_image (PanelToplevel *toplevel,
```
rbuj, line 417: This function is called more than once time, after dropping an image, and after clicking on "Set Background Image" button.

cwendling: Yes, because I left the check in `set_background_image_from_uri()` itself, which is called by the dialog response handler.
I can get the check out of that function and guard the other call, but I reckon the overhead is not a big deal, and anyway the most likely "costly" call is needed in both instances (get a filename from URI).
I it's really a big deal, I can refactor the code not to do this double checking, but OTOH there can be an arbitrary amount of time between the 2 calls (as it waits for user input), so lockdown *could* have been enabled in the meantime which doesn't make this as redundant (unlikely though, I agree).

mate-panel/panel.c
cwendling: What do you mean?  I have one caller of this function that pass `NULL` as `filename` because it doesn't care what the value is, just whether the other checks succeeded.  Admittedly the ability to return the filename is a mere convenience and optimization not to re-compute the filename, but it seemed handy and fairly innocent to me as it's internal API anyway.

rbuj: Where did you set the filename's value in can_set_background_image?

cwendling: I don't, it's only an out variable.  But I use the value I may assign to that out variable to determine the return value of `can_set_background_image()`, regardless of whether that out variable is `NULL` or not.

muktupavels:
> memory leak
> 
> ```c
> if (filename)
>     *filename = image;
> ```

where?

rbuj:
Now I see that is feed at set_background_image_from_uri

rbuj:
in fact it is a conditional assignment if the pointer of the output variable is passed

rbuj:
> where?

I'm sorry for the noise :)

rbuj:
@muktupavels can you take a look at https://github.com/mate-desktop/marco/pull/633

PR #1126, opened 2020-08-04
launcher: Fix memory leak

rbuj:

PR #1134, opened 2020-08-05
launcher: Fix memory leak

cwendling:
Follow-up to #1126 to address [my review comment](https://github.com/mate-desktop/mate-panel/pull/1126#pullrequestreview-460780426), which is a very similar leak further down in the same file.


cwendling:
Indeed. Updated, which led to a slight refactoring.

cwendling:
…and while at it I found a couple more.

rbuj:

PR #1135, opened 2020-08-08
wncklet: Check the length of the pixel data for empty thumbnails

rbuj:


zhangn1985:
gdk_pixbuf_get_pixels ()

Returns

A pointer to the pixbuf's pixel data. Please see the section on image data for information about how the pixel data is stored in memory.

**This function will cause an implicit copy of the pixbuf data if the pixbuf was created from read-only data.** 

I guess it need  to free pixels, instead of leave return value unreferenced.

vkareh (approved):
looks good to me, thanks!

vkareh:
cherry-picked to 1.24: https://github.com/mate-desktop/mate-panel/commit/1fbed28fa0f30f7257a64f8ae5cfca5856aa00f0

PR #1143, opened 2020-09-14
Drop 'xau' from PKG_CHECK_MODULES call

wmww:
It appears this is no longer being used (if it ever was)

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -49,7 +49,6 @@ requires:
     - libsoup2.4-dev
     - libwnck-3-dev
     - libx11-dev
-    - libxau-dev
     - libxrandr-dev
     - lsb-release
     - make
@@ -104,7 +103,6 @@ requires:
     - libsoup2.4-dev
     - libwnck-3-dev
     - libx11-dev
-    - libxau-dev
     - libxrandr-dev
     - lsb-release
     - make
--- a/configure.ac
+++ b/configure.ac
@@ -213,7 +213,7 @@ AC_SUBST(GTK_LAYER_SHELL_LIBS)
 # Check if we have the X development libraries
 have_x11=no
 if test "x$enable_x11" != "xno"; then
-  PKG_CHECK_MODULES(X, x11 xau, have_x11=yes, [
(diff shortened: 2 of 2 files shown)


raveit65:
How can we be sure that libXau.so or .Xauthority-files aren't used by the panel?
Btw. How did you check this in panel code?

You can also clean up build requires (libxau-dev) in https://github.com/mate-desktop/mate-panel/blob/master/.build.yml

epgfm:
Looks like `libXau.so.6` was loaded by mate-panel `1.20.1-3ubuntu1`:
image

wmww:
I checked by looking for a `Xauth.h` #include, or `xau`/`Xau` anywhere in the code.The only place those letters appear is here (plus a generated Makefile as a result of this). I'm assuming it was loaded because of the thing I removed (not because the panel was actually using it).

raveit65:
@wmww 
Again, you can clean up build requires (libxau-dev) in https://github.com/mate-desktop/mate-panel/blob/master/.build.yml

raveit65:
I old and obsolete 1.20 branch we have a more complex check in configure.ac but none includes in files.
```
[rave@mother mate-panel]$ grep -nr Xau *
configure.ac:189:  AC_CHECK_LIB(Xau, XauFileName,, AC_MSG_ERROR([libXau not found]))
configure.ac:193:  X_LIBS="$X_PRE_LIBS $X_LIBS -lX11 -lXau $X_EXTRA_LIBS"
[rave@mother mate-panel]$ grep -nr Xauth.h *
[rave@mother mate-panel]$ 
``` 
So it seems it was never really use before wayland port.

raveit65 (approved):
LGTM,
builds and runs fine under X11

raveit65:
@wmww 
I will merge it when .build.yml is updated.

rbuj (approved):
LGTM. Legacy library,  see https://github.com/GNOME/gnome-panel/commit/f4e7e81152287d21a879993b631660f9ce9a3e45

PR #1146, opened 2020-09-28
Replace obsolete API and  add the font size setting for the menu

zhuyaliang:
Replace obsolete functions Use g_task... replace g_simple.... and adding support for menu font size modification in panel gsschema

### DIFF
--- a/data/org.mate.panel.menubar.gschema.xml.in
+++ b/data/org.mate.panel.menubar.gschema.xml.in
@@ -8,6 +8,14 @@
     <value nick="32px" value="32"/>
     <value nick="48px" value="48"/>
   </enum>
+  <enum id="org.mate.panel.menubar.font-size">
+    <value nick="default" value="-1"/>
+    <value nick="14px" value="14"/>
+    <value nick="15px" value="15"/>
+    <value nick="16px" value="16"/>
+    <value nick="17px" value="17"/>
+    <value nick="18px" value="18"/>
+  </enum>
 
   <schema id="org.mate.panel.menubar" path="/org/mate/panel/menubar/">
     <key name="show-applications" type="b">
@@ -34,6 +42,11 @@
       <default>'start-here'</default>
       <summary>Icon to show in menu bar</summary>
       <description>Set the theme icon name to use in menu bar.</description>
+    </key>
+    <key name="font-size" enum="org.mate.panel.menubar.font-size">
+      <default>'default'</default>
+      <summary>Menu bar font size</summary>
+      <description>Set the size of an font used in menu bar. The panel must be restarted for this to take effect.</description>
     </key>
     <key name="icon-size" enum="org.mate.panel.menubar.icon-size">
       <default>'default'</default>
--- a/data/org.mate.panel.menubar.gschema.xml.in
+++ b/data/org.mate.panel.menubar.gschema.xml.in
@@ -8,14 +8,6 @@
     <value nick="32px" value="32"/>
     <value nick="48px" value="48"/>
   </enum>
-  <enum id="org.mate.panel.menubar.font-size">
-    <value nick="default" value="-1"/>
-    <value nick="14px" value="14"/>
-    <value nick="15px" value="15"/>
-    <value nick="16px" value="16"/>
-    <value nick="17px" value="17"/>
-    <value nick="18px" value="18"/>
-  </enum>
 
   <schema id="org.mate.panel.menubar" path="/org/mate/panel/menubar/">
     <key name="show-applications" type="b">
@@ -42,11 +34,6 @@
       <default>'start-here'</default>
       <summary>Icon to show in menu bar</summary>
       <description>Set the theme icon name to use in menu bar.</description>
-    </key>
-    <key name="font-size" enum="org.mate.panel.menubar.font-size">
-      <default>'default'</default>
-      <summary>Menu bar font size</summary>
-      <description>Set the size of an font used in menu bar. The panel must be restarted for this to take effect.</description>
     </key>
     <key name="icon-size" enum="org.mate.panel.menubar.icon-size">
       <default>'default'</default>
--- a/mate-panel/libmate-panel-applet-private/panel-applet-container.c
+++ b/mate-panel/libmate-panel-applet-private/panel-applet-container.c
@@ -465,13 +465,13 @@ on_factory_appeared (GDBusConnection   *connection,
 		     GTask             *task)
 {
 	MatePanelAppletContainer *container;
-       	gchar                *object_path;
-       	AppletFactoryData    *data;
+    gchar                *object_path;
+    AppletFactoryData    *data;
 
-       	data = g_task_get_task_data (task);
+    data = g_task_get_task_data (task);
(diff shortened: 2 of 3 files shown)


zhuyaliang:
@raveit65 @yetist  Please check this request
In this request, I replaced the old API and added a change to the menu font size

lukefromdc (commented):
I just tested this and ran into two problems: 
The minor one is that changes in the menubar font size (the only font changed by this) are not applied until the panel is restarted.
The big one is that upon changing GTK themes (mine specifies a font so checked this) we get a panel crash and segfault(if started by the session) or a partial crash of some kind leaving the panel nonresponsive if started from the command line or from gdb. In both cases we get many lines of this error:
`(mate-panel:34647): GLib-GIO-CRITICAL **: 22:16:57.400: g_task_return_boolean: assertion '!task->ever_returned' failed`
This takes place with or without the menubar in the panel, and does NOT take place with master. Has to be in the second commit as the first commit only touches the panel menubar's code and the schema(which isn't the problem).

lukefromdc (commented):
This stops the crash on changing themes,  the g_free_error had to be doing something wrong as taking them out stops the crashes. I am NOT certain about this, but memory management SHOULT be OK with this:
https://developer.gnome.org/gio/stable/GTask.html#g-task-return-error

"Note that since the task takes ownership of error , and since the task may be completed before returning from g_task_return_error(), you cannot assume that error is still valid after calling this. Call g_error_copy() on the error if you need to keep a local copy as well."

Thus, I am thinking we were crashing by freeing error after handing it to g_task_return_error which still leaves the question of why we were getting to that code. My gues is this has to do with the g_dbus calls but I just don't know enough about that to speak with any confidence.

None the less, this works. We still don't update the text size until the panel is restarted, honestly I think we need to deal with that part in this PR, as otherwise users will think setting the text size is being entirely ignored and might give up and set it back without ever seeing the results they specified. Only discovery of this for an end user would be by accident.

lukefromdc:
Just found two other problems, which have to be related to the changes in panel-applet.c:
The fish won't render at all, and the trash icon never resizes from 16x16. Those both work fine with master. Also on one and only one restart of the panel the network-manager applet in indicator mode was also rendered undersize but that was not repeatable so may have been a one-off. Note that I have the tray, fish, clock, and wncklet all built in-process

zhuyaliang:
> None the less, this mostly works. We still don't update the text size until the panel is restarted however. I think we need to deal with that part in this PR, as otherwise users will think setting the text size is being entirely ignored and might give up and set it back without ever seeing the results they specified. Neither changing the panel height nor the GTK themes (to force rerendering the panel) applies the changed font size, only restarting the panel does. The real problem is that the only discovery of this panel restart requirement for an end user would be by accident.

In the description of setting the font size of the application menu, the user has been told that the panel needs to be restarted to take effect
2020-10-10 13-58-46屏幕截图

zhuyaliang:
> Thus, I am thinking we were crashing by freeing error after handing it to g_task_return_error which still leaves the question of why we were getting to that code. My gues is this has to do with the g_dbus calls but I just don't know enough about that to speak with any confidence.
```
static void
container_child_background_set (GObject      *source_object,
                GAsyncResult *res,
                gpointer      user_data)
{
    MatePanelAppletContainer *container = MATE_PANEL_APPLET_CONTAINER (source_object);
    MatePanelAppletFrameDBus *frame = MATE_PANEL_APPLET_FRAME_DBUS (user_data);

    mate_panel_applet_container_child_set_finish (container, res, NULL);

    frame->priv->bg_operation = NULL;
}
static void
mate_panel_applet_frame_dbus_change_background (MatePanelAppletFrame    *frame,
                       PanelBackgroundType  type)
{
    MatePanelAppletFrameDBus *dbus_frame = MATE_PANEL_APPLET_FRAME_DBUS (frame);
    MatePanelAppletFrameDBusPrivate *priv = dbus_frame->priv;
    char *bg_str;

    bg_str = _mate_panel_applet_frame_get_background_string (
            frame, PANEL_WIDGET (gtk_widget_get_parent (GTK_WIDGET (frame))), type);

    if (bg_str != NULL) {
        if (priv->bg_operation)
            mate_panel_applet_container_cancel_operation (priv->container, priv->bg_operation);

        priv->bg_operation = mate_panel_applet_container_child_set (priv->container,
                          "background",
                          g_variant_new_string (bg_str),
                          NULL,
                          container_child_background_set,
                          dbus_frame);

        g_free (bg_str);
    }
}
```
When the user calls ```mate_ panel_ applet_ container_ child_ set_finish ``` You can get this error when fails
such as
```
    if (!mate_panel_applet_container_child_set_finish (container, res, &error))
    {   
        g_warning ("%s\n", error->message);
        g_error_free (error);
    }   
```
So it should be the user who uses g_ error_ free

zhuyaliang:
> Just found two other problems, which have to be related to the changes in panel-applet.c:
> The fish won't render at all, and the trash icon never resizes from 16x16. Those both work fine with master. Also on one and only one restart of the panel the network-manager applet in indicator mode was also rendered undersize but that was not repeatable so may have been a one-off. Note that I have the tray, fish, clock, and wncklet all built in-process

I think this is a separate problem that existed before, and I will solve it in the next merge request

zhuyaliang:
@yetist In addition, the Travis CI build distro=“ archlinux:latest "There seems to be a problem

lukefromdc:
The fish and trash applets render and resize fine on master. 
OK, I see you document the panel restart issue in the gsettings key, that helps a lot as it can only be reset from where this is read, or from the command line(implying advanced user in latter case). There is also the fact that ONLY the menu bar font and not fonts used in any other applets such as the clock, CPU freq applet,, sensors applet are affected. Any setting other than default thus creates an inconsistancy. 

Note that since you are already using a style provider to change the font, it is trivial to apply it to every applet in the panel instead of just the menubar. You are applying the style class to the panel menubar, if you apply it to the entire panel all applets will be affected equally. You will want to apply it to 
`.mate-panel-menu-bar menu label`
but NOT to
`.mate-panel-menu-bar menubar menu menuitem label`
unless of course you want to also add the user-facing feature of changing the menu font size by another gsettings key.

I control menu fonts and enforce a bold font on the panel with this code in my UbuntuStudio_Legacy GTK theme
https://github.com/lukefromdc/gtk-theme-ubuntustudio-legacy
Note that the description inside my them has not been updated to reflect that we have renamed the panel menu button and menubar:
```

/*Font used on panel itself and "custom menu bar" */

.mate-panel-menu-bar  label, .gnome-panel-menu-bar  label{
	font-weight: bold;
}

.mate-panel-menu-bar menu label, .gnome-panel-menu-bar menu label{
	font-weight: normal;
}

/*Open up the main menu size a bit*/
/*This speeds up main menu navigation*/

.mate-panel-menu-bar menubar menu menuitem label,
.mate-panel-menu-bar button menu menuitem label,
.gnome-panel-menu-bar menubar menu menuitem label,
.gnome-panel-menu-bar button menu menuitem label{
	font-stretch:  Normal;
	font-size: 120%;
}

/*Smaller font on other menus so they don't look ballooned*/

.gnome-panel-menu-bar menuitem label,
.mate-panel-menu-bar menuitem label{
	font-stretch: condensed;
	font-size: 110%;
}

```

PR #1155, opened 2020-10-28
Drop background compositing

wmww:
There was previously a feature where if you had a transparent background with a non-compositing window manager, the panel would grab the desktop background and composite itself over top of it. This may have made sense when written (but tbh it probably didn't) and it definitely doesn't make sense now. This removes all that complexity, including the monitor that waited for background changes.

Advantages of merging this PR include: less code to maintain, less code to port to non-deprecated functions


mate-panel/panel-background.c
```diff
@@ -1019,13 +801,7 @@ panel_background_make_string (PanelBackground *background,
 
 	effective_type = panel_background_effective_type (background);
 
-#ifdef HAVE_X11
-	if (is_using_x11 () &&
-	    (effective_type == PANEL_BACK_IMAGE ||
-	     (effective_type == PANEL_BACK_COLOR &&
-	      background->has_alpha &&
-	      !gdk_window_check_composited_wm(background->window)))) {
```
wmww, line 1027: Note that originally disabling this bit of code on Wayland was a mistake. There's no reason not to report a pixmap background on Wayland, I just got confused with all the color compositing logic.

lukefromdc:
There is a problem with dropping this code: marco still supports running with or without compositing, and removing this code is guaranteed to break setting a custom background with color transparency OR an image that includes transparent areas/does not fill the rectangle 100%. 

Also, mate supports running with other window managers, such as IceWM etc that are not compositers.  This DOES remove a lot of code, but that code isn't broken and hasn't broken again since I originally got it working with GTK 3.20 after first getting it to work with GTK 3.14 and then having the GTK 3.20 transition break that years ago. GTK3 is now stable with GTK development focussed on GTK4.

This, I propose we keep this code unless and until it actually stops working due to an underlying change in another program. By that time we might be done with xorg anyway.

wmww:
IMO If a user wants transparency to work, they can install a compositing wm or external compositor. If they don't have a compositing wm, they should not expect apps to be transparent. It's not the panel's job to polyfill compositing, especially since transparency is a vanity feature anyway. It might also confuse users that the panel is transparent and nothing else is.

Of course, up to the maintainers, just my two cents.

lukefromdc:
This is basically a request for a feature removal, which we avoid

raveit65:
> There is a problem with dropping this code: marco still supports running with or without compositing, and removing this code is guaranteed to break setting a custom background with color transparency OR an image that includes transparent areas/does not fill the rectangle 100%.
> 
Well, the default is to run marco with compositor.
If a user do not run the compositor, for wat ever reason ever, i think it is OK not to have a fake-compositor feature of the panel.

> Also, mate supports running with other window managers, such as IceWM etc that are not compositers.

Really?
I don't think that using another WM than marco is supported by MATE.
We are a very small team and from my side i don't support another WM than marco.
I never heard that another MATE dev like to spend more of his limited spare time to support another WM in the last 8 years.
Other DEs like gnome support only `mutter` and they have a lot of more developers than MATE.

 This DOES remove a lot of code, but that code isn't broken and hasn't broken again since I originally got it working with GTK 3.20 after first getting it to work with GTK 3.14 and then having the GTK 3.20 transition break that years ago. GTK3 is now stable with GTK development focussed on GTK4.
> 
> This, I propose we keep this code unless and until it actually stops working due to an underlying change in another program. By that time we might be done with xorg anyway.

I don`t think that a lot of users are affected by removing fake-transparent support if marco is used without compositor.
So, i am fine with removing this code which makes the code simpler.

lukefromdc:
OK, reopening this

lukefromdc (approved):
This builds and runs as expected. If a partially transparent background is selected and the alpha slider moved, the background varies between the selected color and black rather than the selected color and the desktop when a non-compositing WM is in use. I tested over IceWM and got the expected behavior.

Note that when testing this OR master with an image background, the image is not rendered properly on some applets when hidpi is in use, notably the volume control, trash, sensors applet, and cpufreq applet (maybe all out-of-process applets)? Looks like image rendered twice height and cut off to fit but not sure. 

Panel_Issue


Again, this does NOT come from this PR and master has the same problem. I will have to poke around and find it sometime. Probably I just hadn't played with a user-set image background since my switch to a hidpi monitor so didn't find this before.

Anyway, approving as this does work

raveit65 (changes requested):
Transparency slider is still there under x11 when marco without compositor is used.
The result is that you can fade a chosen solid color (eg. `#FFFFFF` to black (non-color). o_O
Screenshot at 2020-11-11 12-03-22
Screenshot at 2020-11-11 12-02-20

I didn't test this with wayland because this needs already fixed.
Not sure if we run into another ifdef hell to disable the slider for x11 and non compositor case.

wmww:
> Not sure if we run into another ifdef hell to disable the slider for x11 and non compositor case.

That doesn't make sense, since we don't know if it's going to run on a compositing wm or not at compile time. Needs to be runtime logic. I'll look into that.

Wayland is like a compositing X11 wm in all ways relevant to this PR.

wmww:
Gee wiz it's impossible to check what does what because you've got 3 functions in the same file that do the *exact* same thing (one is a wrapper of the other, the third is an identical reimplementation).

wmww:
The slider should now be disabled on non-compositing window managers and it will render the selected color without transparency

lukefromdc:
I just retested this, and found one problem that may not have an easy fix: whether the slider is enabled or not depends not on the window manager currently being used, but rather the WM that was in use when the panel was started. Also any partially transparent color set up with a compositing WM remains in use (transparent to underlying black) if the WM is then switched to a noncompositing WM.

This can create problems if the user turns off compositiong in marco after having started their session with it turned on. The slider cannot now be moved, so the user must select a new background color if they were previously using a color background and don't want to keep the now darkened color. With the slider enabled unconditionally they can get out of this more easily.

We could let this go but we would sooner or later get bug reports for either or both of these conditions. To fix it, we would need to update the previously chosen color to 100% solid (which then kills the prior selection on returning to a composited WM), and update whether or not to allow the slider to be moved. Easier to keep the current code unless and until it breaks on something.

Wayland of course is always a compositor, so that problem won't exist there. We could choose not to build these files on Wayland-only builds

wmww:
Another option is to revert the last two commits. On non-compositors, transparency=black. Doesn't seem like unreasonable behavior to me. Yes, it might not look the best, but the user has already installed a non-default wm and set a non-default background, so they know how to fix it if they want.

raveit65:
I tested again, better....
The issue which @lukefromdc  described can be solved for the by restarting the panel.
But only for the panel background itself and applets from mate-panel package.
Then we run into another issue with applets from mate-applets, mate-power-manager and mate-media (volume-control panel applet).
screenshot after `killall mate-panel`
Screenshot at 2020-11-12 14-17-20

All those applets doesn't have a background color when the transparency slider isn't reset.
Looks like all those applets needs to be edited to follow this change.  -_-

mbkma:
@raveit65 to fix this I think the following is sufficient:

```diff
diff --git a/libmate-panel-applet/mate-panel-applet.c b/libmate-panel-applet/mate-panel-applet.c
index 0acd3dda..7b9f2b6b 100644
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1455,6 +1455,11 @@ mate_panel_applet_handle_background (MatePanelApplet *applet)
 
        type = mate_panel_applet_get_background (applet, &color, &pattern);
 
+       if (!gdk_screen_is_composited (gdk_screen_get_default ())) {
+               color.alpha = 1.;
+       }
+
+
        switch (type) {
        case PANEL_NO_BACKGROUND:
                g_signal_emit (G_OBJECT (applet),
```

wmww:
I'm assuming our goals are:
1. Work well in supported cases (transparent background OR non-compositing wm)
2. Never crash or have buggy/unreasonable behavior
3. Have as simple code as possible to achieve 1 and 2

IMO removing the last two commits would do that. "transparency on non-compositing wms == black" is simplest to implement and easy for users to understand. Keeping track of the wm, disabling UI controls depending on it and/or having inconsistent behavior when the window manager is changed all seem like features that will just confuse users and developers. Engineering time is better spent fixing any issues users have with compositing Marco, and porting to Wayland where we don't have these problems.

raveit65:
It looks like that it isn`t possible to do a ***clean*** remove this feature for X11 without having glitches for a supported case.
Why not disable it for wayland only?

I will test the proposal from @mbkma

raveit65:
@mbkma 
Great, this will reset the panel-background to solid color with non-compositor-WM after a restart of the panel.

wmww:
I don't consider a black background on non-compositing wms a glitch. It's defined, consistent and easy to understand behavior. It might not be what users want, but since they have the power to fix it in multiple ways, I don't see the issue.

It's already disabled on Wayland (thus why Wayland works). The point of this PR is to get it out of the code base because it's 800 LOC that (almost?) nobody is using today.

raveit65:
Please consider to test and add the proposal from @mbkma 
......it fixes all concerns by me ;)

raveit65:
> The point of this PR is to get it out of the code base because it's 800 LOC that (almost?) nobody is using today.

What makes you sure that nobody use it with X11 and an older pc system with poor graphic?
Not everyone in the world uses high-end computer systems like first world people like you and me.
Any way,
- I agree with removing the feature when only a few user will probably use it.
- I agree that it is acceptable that a few users have to restart the panel to restore solid color.
- But i don't agree to add bugs to MATE in x11, which is used by the majority of Mate users, only for an actual experimental state of Mate in wayland.

Again, please consider to add the proposal and i am fine with your PR.

wmww:
Done, and seems to work to me

raveit65:
@wmww 
I am still wondering why you add a link to another people in commit header message?
When i am browsing through git history i want to see quickly what the commit does and not who did proposed the commit.
Can you please rename the commit please before merging?
Links are additional information's which belongs in commit body.

<commit header>
one empty line to mark end of commit header
<commit body>

Thank you :)

raveit65 (approved):
It works fine now after latest changes.

raveit65:
@lukefromdc 
Any concern left or is this ready to go?

wmww:
> Links are additional information's which belongs in commit body

Done

lukefromdc:
OK on my end

PR #1159, opened 2020-11-16
Remove USE_MATE2_MACROS from autogen.sh (legacy)

rbuj:


lukefromdc (approved):
This configures and builds fine, and I didn't get any (very unexpected) runtime surprises either

PR #1162, opened 2020-11-18
panel-applets-manager: Boilerplate

rbuj:


lukefromdc (approved):
Test build and install went fine with the other two "boilerplate" commits also applied in th same build directory (I tested these sequentially that way). Worked fine with no surprises on restarting the panel.

Line 72 in mate-panel-applets.h, I don't know if that indent (does not look like it lines up the same way for a variable without *) is correct or not. We don't want to add new problems with this on top of all the others already there

raveit65 (approved):
Builds and runs fine.

PR #1164, opened 2020-11-24
clock: Set window hint to GDK_WINDOW_TYPE_HINT_MENU

knatsakis:
This makes the window appear properly when using [i3](https://i3wm.org/l) WM.

Without this patch, the clock window appears as a dock/panel window covering the full width of the screen:
2020-11-24-221653_7680x2160_scrot

According to these [docs](http://gtk.php.net/manual/en/html/gdk/gdk.enum.windowtypehint.html) a GDK_WINDOW_TYPE_HINT_MENU is more appropriate, as the current GDK_WINDOW_TYPE_HINT_DOCK is reserved for dock/panel windows.

### DIFF
--- a/applets/clock/calendar-window.c
+++ b/applets/clock/calendar-window.c
@@ -474,7 +474,7 @@ calendar_window_init (CalendarWindow *calwin)
 	calwin->priv = calendar_window_get_instance_private (calwin);
 
 	window = GTK_WINDOW (calwin);
-	gtk_window_set_type_hint (window, GDK_WINDOW_TYPE_HINT_DOCK);
+	gtk_window_set_type_hint (window, GDK_WINDOW_TYPE_HINT_MENU);
 	gtk_window_set_decorated (window, FALSE);
 	gtk_window_set_resizable (window, FALSE);
 	gtk_window_set_default_size (window, 337, -1);


raveit65:
Does the window stays open with this PR?

knatsakis:
> Does the window stays open with this PR?

Yes

raveit65:
I expect a little bit more info than a minimalistic `yes`, before i spend more spare time on this PR.
Doesn't this break actual behaviour?
1. first click on panel-button opens the calendar window, and it is in the foreground above other windows.
2. second click on panel-button closes the calendar window.

This is useful to work in other application where you need calendar informations, eg. text-editors/writers.

knatsakis:
Hey, sorry, I didn't realize you wanted more information :-). My apologies!

> Doesn't this break actual behaviour?
> 
> 1. first click on panel-button opens the calendar window, and it is in the foreground above other windows.
> 2. second click on panel-button closes the calendar window.

No, not as far as I can tell. It continues working as you describe in 1. and 2. I have tested it with Marco and i3wm.

> This is useful to work in other application where you need calendar informations, eg. text-editors/writers.

I agree.

Let me know if you want me to provide more details or anything else!

rbuj (changes requested):
The CalendarWindow is a panel window, so the current value is correct. There is no need to apply the change you proposed in your pull request. You should also keep in mind that window managers may have differences in the implementation of the desktop metaphor.

https://developer.gnome.org/gdk3/stable/gdk3-Windows.html#GdkWindowTypeHint

knatsakis:
> The CalendarWindow is a panel window, so the current value is correct. There is no need to apply the change you proposed in your pull request. You should also keep in mind that window managers may have differences in the implementation of the desktop metaphor.

I was of the impression that dock/panel windows are windows that can only be placed on a screen edge and reduce the available desktop area (e.g: the main mate-panel window). I don't see how a drop-down window like the calendar window fits this description.

rbuj:
The i3 window manager handles floating windows in another way.

muktupavels:
Calendar is not panel, is it? Or in case you think differently... Do preference window (or any other window created by panel) also has dock hint?

rbuj:
I just saw an initialization of CalendarWindow:
```
applets/clock/clock.c:805:        window = calendar_window_new (&cd->current_time,
```

rbuj:
Does GDK_WINDOW_TYPE_HINT_DOCK exclude popup windows? in such case, why isn't GDK_WINDOW_TYPE_HINT_POPUP_MENU a better approach?

rbuj:
Anyway, it's not a tear-off menu.

raveit65:
I tested both mentioned options and they are working for me with marco WM under x11.
Sadly they do not improve the situation under wayland. Here the calendar window is always detached from panel and opens in the middle of the screen.

vkareh:
With `GDK_WINDOW_TYPE_HINT_MENU` I can now drag the calendar window anywhere on my screen. Not good.

The only ones with prevent the window from being moved are:
* `GDK_WINDOW_TYPE_HINT_DOCK` reserved for docks and panels (?)
* `GDK_WINDOW_TYPE_HINT_SPLASHSCREEN` not appropriate, but seems to work
* `GDK_WINDOW_TYPE_HINT_DESKTOP` - shows up under all windows, not really appropriate anyway

I think either `GDK_WINDOW_TYPE_HINT_UTILITY` or `GDK_WINDOW_TYPE_HINT_POPUP_MENU` would be a more appropriate choice, but we need to find a way to make it non-movable.

Alternatively, maybe i3 is not doing the right thing respecting the default size of the window, but it might just be that by seeing the DOCK hint, it automatically sets it as a toplevel panel...

vkareh:
From the [i3 docs](https://github.com/i3/i3/blob/next/docs/wsbar#L17-L26):
```
== dock mode

You typically want to see the same workspace bar on every workspace on a
specific screen. Also, you don’t want to place the workspace bar somewhere
in your layout by hand. This is where dock mode comes in: When a program sets
the appropriate hint (_NET_WM_WINDOW_TYPE_DOCK), it will be managed in dock
mode by i3. That means it will be placed at the bottom or top of the screen
(while other edges of the screen are possible in the NetWM standard, this is
not yet implemented in i3), it will not overlap any other window and it will be
on every workspace for the specific screen it was placed on initially.
```

raveit65:
Confirmed, with `alt + left click` i can drag the window now.
Seems that is the reason why former gnome-panel-2.0 devs had chosen GDK_WINDOW_TYPE_HINT_DOCK.
I am pretty sure that was never changed after forking.

knatsakis:
> I think either `GDK_WINDOW_TYPE_HINT_UTILITY` or `GDK_WINDOW_TYPE_HINT_POPUP_MENU` would be a more appropriate choice, but we need to find a way to make it non-movable.

+1
This sounds like the most reasonable way to resolve this after all, to me as well.

rbuj:
In addition, marco and compiz keep `_NET_WM_WINDOW_TYPE_DOCK` on top of all other windows as the standard comments.

NetWM standard (aka Extended Window Manager Hints)
https://specifications.freedesktop.org/wm-spec/wm-spec-1.4.html

PR #1169, opened 2020-12-04
Fix solid color controls after disabling compositor

rbuj:
- Enable the compositor
- Open panel properties dialog
- Set the panel background to solid color and enable color transparency.
- Close panel properties dialog
- Disable the compositor
- Open panel properties dialog

Screenshot at 2020-12-04 23-14-41

### DIFF
--- a/mate-panel/Makefile.am
+++ b/mate-panel/Makefile.am
@@ -158,8 +158,7 @@ mate_panel_LDADD = \
 	$(DCONF_LIBS) \
 	$(XRANDR_LIBS) \
 	$(X_LIBS) \
-	$(WAYLAND_LIBS) \
-	-lm
+	$(WAYLAND_LIBS)
 
 if ENABLE_WAYLAND
 mate_panel_LDADD += \
--- a/mate-panel/button-widget.c
+++ b/mate-panel/button-widget.c
@@ -1,5 +1,4 @@
 #include <config.h>
-#include <math.h>
 #include <string.h>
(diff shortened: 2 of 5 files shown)


lukefromdc (approved):
System background case:
No effects as this code is not used.


Color background case:
Mostly works , and if the panel is started and the panel preferences is opened with a noncompositing WM in use, the slider is shown insensitive at 100% and the color is same as if set to 100% opacity with any window manager.

We still don't update the panel color if the window manager is changed to a noncompositing WM from a composioting WM without restarting the panel. In that case, the panel's color is the transparent color rendered over black until the panel is restarted or the color is changed, after which the expected color is shown. The reverse is true on switching to a composting WM. 

The slider renders as expected but color shown not properly applied until it is changed or the panel restarted.  Note that if the preferences dialog is kept open while the WM is changed between a compositing and a non-compositing WM it does not get updated either.

Fixing either of these would require more complex code to detect the change of WM in near-realtime, and might mean heavier code, exactly what anyone using a noncompositing WM is trying to avoid. Thus I am approving this as written, the choice being between this behavior and the immediately previous behavior.

Image background case:
Behaves as expected with this, no need to change or restart anything. Applied properly, with any transparent portions transparent with a compositing WM or over black with a noncompositing WM.

raveit65 (approved):
LGTM,
it's better than before to reset the dialog.
Builds and runs fine.

PR #1170, opened 2020-12-05
fish: Do not to include the math.h library to get the PI value

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.

PR #1171, opened 2020-12-05
cppcheck warning: The if condition is the same as the previous one

rbuj:


lukefromdc (approved):
Builds and runs fine, opening applet menus works same as before

vkareh (approved):
lgtm

mate-panel/applet.c
```diff
@@ -645,11 +645,10 @@ applet_show_menu (AppletInfo     *info,
 
 	panel_widget = mate_panel_applet_get_panel_widget (info);
 
-	if (info->menu == NULL)
+	if (info->menu == NULL) {
 		info->menu = mate_panel_applet_create_menu (info);
-
-	if (info->menu == NULL)
-		return;
+		g_return_if_fail (info->menu != NULL);
```
muktupavels, line 650: https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-if-fail
`If G_DISABLE_CHECKS is defined then the check is not performed. You should therefore not depend on any side effects of expr.`

With this change you now can end up calling `mate_panel_applet_menu_set_recurse (NULL, ...)`! Or you might end up with warning when applet simply has no menu to show.

Check `mate_panel_applet_create_menu`. If panels are locked down and there is no user_menu or its items are not enabled you will end up returning `NULL`...

rbuj: @muktupavels done. PR #1174

PR #1172, opened 2020-12-08
window-list: Get thumbnail size and scale in preview_window_thumbnail

rbuj:


applets/wncklet/window-list.c
```diff
@@ -404,9 +410,10 @@ static gboolean applet_enter_notify_event (WnckTasklist *tl, GList *wnck_windows
 	tasklist->preview = gtk_window_new (GTK_WINDOW_POPUP);
 
 	gtk_widget_set_app_paintable (tasklist->preview, TRUE);
-	gtk_window_set_resizable (GTK_WINDOW (tasklist->preview), TRUE);
+	gtk_window_set_default_size (GTK_WINDOW (tasklist->preview), thumbnail_width/thumbnail_scale, thumbnail_height/thumbnail_scale);
+	gtk_window_set_resizable (GTK_WINDOW (tasklist->preview), FALSE);
```
vkareh, line 414: For some reason setting this as `FALSE` prevents large windows from showing up as thumbnails.

rbuj: @vkareh done

vkareh:
This works well now.

But once you get `thumbnail_w|h`, every instance of it outside of `preview_window_thumbnail` you're dividing by `thumbnail_scale` and you're only using the `thumbnail_scale` value for that, why not skip passing it around and make sure that `thumbnail_w|h` are already scaled before you leave `preview_window_thumbnail`? This way you're passing around fewer values and the other functions no longer need to worry about scaling.

rbuj:
IMHO, I don't think there is much penalty for storing the actual size in pixels instead of storing the relative size considering the HiDPI scaling. The division between integers is very fast. The divisor is a small value: if it's 1 no operation is done, and if it's a power of two it performs a shift.

vkareh:
That's a good point, thanks!

vkareh (approved):
works well for me

lukefromdc (approved):
Builds fine, didn't create any runtime surprises. Thumbnails render same size on my setup for windows on the main and secondary monitor, though note that the window-scaling is 2 on both displays. Using xrandr to force the small monitor to window-scaling=1 just consumed too much CPU and gives a slightly laggy mouse on my setup.

Note that we get this deprecation warning on building,  but it comes from code in master (from the last commit to change this file) that is not changed by this PR. :
```

window-list.c: In function ‘preview_window_thumbnail’:
window-list.c:264:2: warning: ‘gdk_cairo_create’ is deprecated: Use 'gdk_window_begin_draw_frame() and gdk_drawing_context_get_cairo_context()' instead [-Wdeprecated-declarations]
  264 |  cairo_t *win_cr = gdk_cairo_create (window);
      |  ^~~~~~~

```

I am not sure how to fix that one

rbuj:
@lukefromdc
I think we can use gdk_window_create_similar_surface and gdk_cairo_set_source_window instead.
PR #1173

lukefromdc:
Final recheck worked fine, will merge this

PR #1174, opened 2020-12-11
applet: Avoid a NULL pointer dereference

rbuj:


raveit65:
@muktupavels 
Does this solve your concern from https://github.com/mate-desktop/mate-panel/pull/1171 ?

vkareh (approved):
LGTM

Builds and runs fine

lukefromdc (approved):
Builds and runs fine, didn't have any problems with applet menus in a quick test. All my normal applet menus came up fine

PR #1175, opened 2020-12-14
Signal name is 'value-changed' for GtkSpinButton

rbuj:
```C
  /**
   * GtkSpinButton::value-changed:
   * @spin_button: the object on which the signal was emitted
   *
   * The ::value-changed signal is emitted when the value represented by
   * @spinbutton changes. Also see the #GtkSpinButton::output signal.
   */
  spinbutton_signals[VALUE_CHANGED] =
    g_signal_new (I_("value-changed"),
                  G_TYPE_FROM_CLASS (gobject_class),
                  G_SIGNAL_RUN_LAST,
                  G_STRUCT_OFFSET (GtkSpinButtonClass, value_changed),
                  NULL, NULL,
                  NULL,
                  G_TYPE_NONE, 0);
```
https://github.com/GNOME/gtk/blob/daa5d5aeadbebd2d1e764e953b8b6f0d21fbfacf/gtk/gtkspinbutton.c#L493-L507

### DIFF
--- a/applets/fish/fish.c
+++ b/applets/fish/fish.c
@@ -467,7 +467,7 @@ static void display_preferences_dialog(GtkAction* action, FishApplet* fish)
 	gtk_spin_button_set_value (GTK_SPIN_BUTTON (fish->frames_spin),
 				   fish->n_frames);
 
-	g_signal_connect (fish->frames_spin, "value_changed",
+	g_signal_connect (fish->frames_spin, "value-changed",
 			  G_CALLBACK (n_frames_value_changed), fish);
 
 	setup_sensitivity (fish, builder,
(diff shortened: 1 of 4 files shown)


lukefromdc (approved):
Builds fine, and changing the values of spinbuttons in the panel-prefs dialog, window list prefs dialog, fish prefs dialog, and notification area prefs dialog all seem to be applied in realtime. 

Not an extensive test but no ugly surprises. When testing remember WM must be marco not compiz to change the number of workspaces from the switcher applet.

Note that with

cwendling:
Note: using `_` or `-` doesn't change behavior, although `-` is preferred.  See https://developer.gnome.org/gobject/stable/gobject-Signals.html#g-signal-new and https://developer.gnome.org/gobject/stable/gobject-GParamSpec.html#canonical-parameter-names.

I didn't test, but it looks OK. Unrelated changes (styles & means of accessing objects) makes this harder to review, but it seems good.

vkareh (approved):
works well for me

PR #1179, opened 2020-12-25
wncklet: condition is always true warning reported by cppcheck

rbuj:
```
applets/wncklet/workspace-switcher.c:212:6: style: Condition '!wm_name' is always true [knownConditionTrueFalse]
 if (!wm_name)
     ^
```


lukefromdc (approved):
Tested a rebase of this after merging the other two PRs. Running cppcheck-gi on this file now shows NO errors. Test builds with and without x11 support, both in-process and out all completed. Finished in-process build with x11 support works fine in my normal x11 session, and we don't use the workspace switcher in wayland yet. Will merge this.

EDIT: Travis build error is usual Fedora rate limiter, unrelated to this PR

PR #1183, opened 2020-12-26
panel-properties-dialog: show dialog on _present instead of _new

rbuj:
warning reported by cppcheck
```
mate-panel/panel-properties-dialog.c:941:9: style: Variable 'dialog' is assigned a value that is never used. [unreadVariable]
 dialog = panel_properties_dialog_new (toplevel, gui);
	^
```

### DIFF
--- a/mate-panel/panel-properties-dialog.c
+++ b/mate-panel/panel-properties-dialog.c
@@ -830,11 +830,11 @@ panel_properties_dialog_update_for_attached (PanelPropertiesDialog *dialog,
 }
 
 static PanelPropertiesDialog *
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Builds (with NO build warnings from this file)  and runs fine, and memory management looks OK here so far as I can tell (basic knowledge of it but not expert level). No problems with repeatedly bringing up the panel preferences dialog. This makes the code a bit more rationally organized

raveit65 (approved):
LGTM,
warning is gone and dialog opens like before.

PR #1186, opened 2020-12-27
Register SNI host only if SNI is enabled

ilya-fedin:
From time to time, I'm testing some third-party SNI implementations (just for the sake of knowledge) on my favorite DE ( ;) ) and observed that with `enable-sni-support = false`, default applet still registers SNI host and if SNI watcher of other tray implementation accepts registration from that applet, tray icons being duplicated. In practice, I discovered this with the xapp's applet.

### DIFF
--- a/applets/notification_area/na-grid.c
+++ b/applets/notification_area/na-grid.c
@@ -318,6 +318,7 @@ na_grid_realize (GtkWidget *widget)
   GdkScreen *screen;
   GtkOrientation orientation;
   NaHost *tray_host;
+  GSettings *settings = g_settings_new ("org.mate.panel");
 
   GTK_WIDGET_CLASS (na_grid_parent_class)->realize (widget);
 
@@ -330,7 +331,11 @@ na_grid_realize (GtkWidget *widget)
                           G_BINDING_DEFAULT);
 
   add_host (self, tray_host);
-  add_host (self, sn_host_v0_new ());
+
+  if (g_settings_get_boolean (settings, "enable-sni-support"))
+    add_host (self, sn_host_v0_new ());
+
+  g_object_unref (settings);
 }
 
 static void


lukefromdc:
How should this be tested?

ilya-fedin:
You can install xapp from your distro (on arch linux it's community/xapp, probably this won't work if you're using something with old packages like debian stable since this library got SNI support not long time ago), enable its sn-watcher by `dconf write /org/x/apps/statusicon/status-notifier-enabled-desktops '["MATE"]'` (it compares `XDG_CURRENT_DESKTOP` variable), disable MATE's SNI support by `dconf write /org/mate/panel/general/enable-sni-support false`, relogin (or kill and restart the notification-area applet) and add XApp's applet to a panel with the standard dialog. After that you should get two icons for each (or some, depends on can they register on the fly or not) SNI-capable application.

lukefromdc:
Thanks. I will leave that one for others here to work on, as missing one piece of this when it's time to revert could screw up my sessions and I work only on bare metal

rbuj:
Related #600

alexarnaud:
I wouldn't have the time to try this one. I'm focus on reporting [GTK4 accessibility issues](https://gitlab.gnome.org/GNOME/gtk/-/issues?scope=all&utf8=%E2%9C%93&state=opened&label_name[]=GTK4&label_name[]=8.%20Accessibility).

raveit65:
> I wouldn't have the time to try this one. I'm focusing on reportin

Thank you

alexarnaud:
> > I wouldn't have the time to try this one. I'm focusing on reporting ........
> 
> Thank you

I mean I may do not have the time to handle this due to a lot of GTK4 a11y issues to report but I would like to try it if I could.

 I totally trust @cwendling for it. FYI, xapp is available into the Debian repository.

ilya-fedin:
> xapp is available into the Debian repository.

It could be too old, SNI support was added this year

ilya-fedin:
> However, it _might_ be interesting to be able to disable the builtin watcher to use another one yet still have the host

According to the comment, it already handles the situation when there's a SNI watcher already and this comment was there even before the setting was added:
https://github.com/mate-desktop/mate-panel/blob/755d6e73771bddfd192e47ef593460e8540591ca/applets/notification_area/main.c#L78-L79

> It would probably be better if changing the setting would enable/disable the host dynamically instead of requiring a restart.

Oh, this would require writing code to start/stop host & watcher on the fly :(

raveit65:
I don't think that we should disable our SNI-Support in generally only because of xapps from linuxmint disto.
Because in most distros Mate installation is without xapps and most xapps are forks from atril, pluma and eom.
Only their bluetooth tray-applet (a frontend of gnome-bluetooth) seems useful for a Mate installation.
In the end it is a problem baked by linuxmint itself.
But i agree to to do that with dynamically check to enable/disable the support.

ilya-fedin:
> only because of xapps from linuxmint disto

Why only because of xapps? There's multiple others SNI implementations for MATE and there's also a setting, but it does only half of the work (disable the watcher, but doesn't disable the host) what is weird behavior

raveit65:
In my opinion there is `NotShowIn=MATE;` missing in this file.
```
[rave@mother ~]$ cat /etc/xdg/autostart/xapp-sn-watcher.desktop 
[Desktop Entry]
Type=Application
Name=xapp-sn-watcher
Comment=A service that provides the org.kde.StatusNotifierWatcher interface for XApps

Exec=/usr/libexec/xapps/sn-watcher/xapp-sn-watcher

X-GNOME-Autostart-Phase=Panel
```

raveit65:
Any way, i have disable it in my installations and i don't have any problems with double icons of SNI applets ;)
Do what you want.....

ilya-fedin:
> In my opinion there is `NotShowIn=MATE;` missing in this file.

It won't start on MATE by default due to a filter in gsettings, but the point of this PR is to make already present `enable-sni-support` setting work correctly, no matter what SNI implementation is used.

raveit65:
I am trusting @cwendling 
When he drives the PR it is fine for me.

ilya-fedin:
> I am trusting @cwendling
> When he drives the PR it is fine for me.

Looks like the PR is on neutral transmission

ilya-fedin:
More than a year has passed, is there a blocker?

cwendling:
> > It would probably be better if changing the setting would enable/disable the host dynamically instead of requiring a restart.
> 
> Oh, this would require writing code to start/stop host & watcher on the fly :(

Yeah… I don't think it'd be very hard, but given the watcher is already conditional and not monitioring the property it's probably not a must have, so this PR LGTM.

This said, for the host part it'd be a matter of something like this (untested):
```diff
diff --git a/applets/notification_area/na-grid.c b/applets/notification_area/na-grid.c
index dbd95d50..df41bde9 100644
--- a/applets/notification_area/na-grid.c
+++ b/applets/notification_area/na-grid.c
@@ -311,6 +311,32 @@ na_grid_draw (GtkWidget *grid,
   return TRUE;
 }
 
+static void
+setting_sni_support_notify (GSettings    *settings,
+                            const gchar  *key,
+                            NaGrid       *self)
+{
+  gpointer sn_host = NULL;
+
+  for (GSList *node = self->hosts; node && ! sn_host; node = node->next)
+  {
+    if (SN_IS_HOST_V0 (node->data))
+      sn_host = node->data;
+  }
+
+  /* we are already synchronized, don't do anything */
+  if (sn_host && g_settings_get_boolean (settings, key))
+    return;
+
+  if (! sn_host)
+    add_host (self, sn_host_v0_new ());
+  else
+  {
+    self->hosts = g_slist_remove (self->hosts; sn_host);
+    g_object_unref (sn_host);
+  }
+}
+
 static void
 na_grid_realize (GtkWidget *widget)
 {
@@ -330,7 +357,13 @@ na_grid_realize (GtkWidget *widget)
                           G_BINDING_DEFAULT);
 
   add_host (self, tray_host);
-  add_host (self, sn_host_v0_new ());
+
+  g_signal_connect (self->settings, "changed::enable-sni-support",
+                    G_CALLBACK (setting_sni_support_notify), self);
+  if (g_settings_get_boolean (self->settings, "enable-sni-support"))
+    add_host (self, sn_host_v0_new ());
 }
 
 static void
```

Something similar for the watcher might be as easy, *but* for the fact the watcher might not actually be needed in-house it already provided by some other process (the session or such).

Anyway, you might be right that it's unnecessary and for such a thing that requiring a restart is fine.  Maybe this should be highlighted in the setting description though?

applets/notification_area/na-grid.c
```diff
@@ -318,6 +318,8 @@ na_grid_realize (GtkWidget *widget)
   GdkScreen *screen;
   GtkOrientation orientation;
   NaHost *tray_host;
+  GSettings *settings;
+  settings = g_settings_new ("org.mate.panel");
```
cwendling, line 322:
```suggestion
  GSettings *settings = g_settings_new ("org.mate.panel");
```

applets/notification_area/na-grid.c
```diff
@@ -330,7 +332,11 @@ na_grid_realize (GtkWidget *widget)
                           G_BINDING_DEFAULT);
 
   add_host (self, tray_host);
-  add_host (self, sn_host_v0_new ());
+
+  if (g_settings_get_boolean (settings, "enable-sni-support") == TRUE)
```
cwendling, line 336:
```suggestion
  if (g_settings_get_boolean (settings, "enable-sni-support"))
```

cwendling (commented):
Not actually tested (and I unfortunately don't foresee when I'll be able to properly do so), but looks good and safe to me :+1: (minus my 2 cosmetic suggestions)

cwendling (approved):
Still not tested, but it seems simple and safe enough for me to LGTM it.

cwendling:
@ilya-fedin yeah it's not very important, and as you say it's probably only for power-users likely to think of restarting the various pieces at hand.  So don't worry about it here I'd say.

lukefromdc (approved):
Builds and runs fine. I can confirm this doesn't break enabling or disabling SNI support. While I don't have any Xapps etc installed, I tested both ways with nm-applet (network-manager-gnome aka network-manager applet) This applet can be started with --indicator to trigger SNI mode, and when in that mode falls back to a tray icon if SNI support is absent. Note that to re-enable SNI mode with this applet after it has been disabled the applet has to be restarted as well as the panel. This has has been said before would probably be expected. Two other approving reviews so will merge. Hopefully this will fix problems some users have had

cwendling:
thanks @lukefromdc!

PR #752, opened 2018-03-14
workspace-switcher: support i3 WM

raveit65:
Fixes https://github.com/mate-desktop/mate-panel/issues/711

PR #1188, opened 2021-01-02
workspace-switcher: support xmonad WM

4eUeP:
This is just a "copy" of [this pr](https://github.com/mate-desktop/mate-panel/pull/752/), which adds xmonad to the supported wm.

It seems to work well on my Debian testing. 
(Replacing system "wnck-applet" with the new built "wnck-applet", restart, then the option "Show workspace names in switcher" appears.)

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -58,6 +58,7 @@ typedef enum {
 	PAGER_WM_METACITY,
 	PAGER_WM_COMPIZ,
 	PAGER_WM_I3,
+	PAGER_WM_XMONAD,
 	PAGER_WM_UNKNOWN
 } PagerWM;
 
@@ -107,7 +108,8 @@ static void pager_update_wnck(PagerData* pager, WnckPager* wnck_pager)
 	if (pager->display_names && (
 		pager->wm == PAGER_WM_MARCO ||
 		pager->wm == PAGER_WM_METACITY ||
-		pager->wm == PAGER_WM_I3))
+		pager->wm == PAGER_WM_I3 ||
+		pager->wm == PAGER_WM_XMONAD))
 	{
 		display_mode = WNCK_PAGER_DISPLAY_NAME;
 	}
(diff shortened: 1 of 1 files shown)


4eUeP:
@raveit65 If you mean any programmatically way to test, I don't know because of lack of what mate-session really does. However, I have used this modified version of `wnck-applet` for a few months without problems. (just replace system `/usr/lib/mate-panel/wnck-applet` with the modified version)

Btw, for xmonad implementation: <https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/src/XMonad.Hooks.EwmhDesktops.html#setSupported>, you can find it `setWMName` to "xmonad"

muktupavels:
`xmonad --replace` will replace current window manager. After testing just do `marco --replace`.

4eUeP:
As a remark for what @muktupavels say,

Set xmonad for mate with default settings:

**Install xmonad (e.g. debian user)**

```
apt-get install xmonad libghc-xmonad-contrib-dev
```

**Start with the following `~/.xmonad/xmonad.hs`**

```
import XMonad
import XMonad.Config.Mate

main = xmonad mateConfig
```

ref: https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Config-Mate.html

**Compile the config file**

```
xmonad --recompile
```

**Change system wm**

change wm by `dconf` if `--replace` doesn't work

```
dconf write /org/mate/desktop/session/required_components/windowmanager "'xmonad'"
```

For some default key bindings: https://xmonad.org/manpage.html#default-keyboard-bindings
(the default mod key is `alt`)

raveit65:
That xnomad use --replace option was sufficient information.

raveit65:
Seems a basic xmonad.hs config is needed to avoid a complete empty desktop :)
Thanks.
I will test again.

raveit65:
Is this the expected behavior?
I would expect that workspaces are not grey out and switch from 8 to 9 workspaces when i start xmonad.
Bildschirmfoto zu 2021-06-01 20-33-23

Also, `marco --replace` doesn't have any effect when xmonad is running, in result i am not very happy to test a WM longer, which i definitely don't want to use for my desktop.

4eUeP:
> Is this the expected behavior?

Yes, the number of workspaces is controlled by `xmonad.hs` config file.

4eUeP:
Here is a simple example to set number of workspaces to 8:

```
import           XMonad
import           XMonad.Config.Mate

main = xmonad myMateConfig

myMateConfig = mateConfig
  { workspaces = myWorkspaces
  }

myWorkspaces = map show [1..8]
```

raveit65:
Ok, i have  8 workspaces now but workspace-switcher settings Gui shows 9 (grey out).
And marco --replace or metacity --replace doesn't kill xmonad.
Any way, i don't think that a few xmonad users want to switch back.
Also, from my side i do not support to use mate-desktop with xnomad WM.
So every user do that for his own risk.

raveit65 (approved):
I don't see any problem to add this code.

PR #1189, opened 2021-01-04
Revert "window-list: remove intermediate steps to get scaled thumbnail"

vkareh:
This reverts commit 08146a80fc1727cde7a6d02d8436f2c86c159b8c.

This commit breaks the thumbnail refreshing, causing the thumbnail to be always stuck in the first generated image of a window, rather than following its changes. Reverting it fixes it so that the thumbnail is always captured on the latest state of the window.

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -252,6 +252,7 @@ preview_window_thumbnail (WnckWindow   *wnck_window,
                           int          *thumbnail_scale)
 {
 	GdkWindow *window;
+	cairo_surface_t *screenshot;
 	cairo_surface_t *thumbnail;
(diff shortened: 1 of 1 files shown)


rbuj:
Without this pr the thumbnail is refreshed if the window is resized, therefore there is some way to force new thumbnail when necessary.

lukefromdc:
I can confirm the issue here

lukefromdc (approved):
Builds and runs fine, and I can confirm both that thumbnails are NOT updated on master at this point, and that they are updated again and work as expected with the PR

rbuj:
#1190 may be related to the issue about updating thumbnails

vkareh:
Fixed by #1190

PR #1190, opened 2021-01-05
wncklet: search the window before creating a new one

rbuj:


applets/wncklet/window-list.c
```diff
@@ -252,15 +252,18 @@ preview_window_thumbnail (WnckWindow   *wnck_window,
...
 	cairo_t *cr;
 	double ratio;
 	int width, height, scale;
 
-	window = gdk_x11_window_foreign_new_for_display (gdk_display_get_default (), wnck_window_get_xid (wnck_window));
-
-	if (window == NULL)
-		return NULL;
+	win = wnck_window_get_xid (wnck_window);
+	if ((window = gdk_x11_window_lookup_for_display (gdk_display_get_default (), win)) == NULL)
```
vkareh, line 262: This function is already called by `gdk_x11_window_foreign_new_for_display`: https://gitlab.gnome.org/GNOME/gtk/-/blob/3.24.24/gdk/x11/gdkwindow-x11.c#L1244-1245

I don't think this PR changes anything, it just forces the lookup function to be called twice.

rbuj: The thumbnail of a window has been updated the second time I hovered the mouse over the taskbar.

vkareh:
> The thumbnail of a window has been updated the second time I hovered the mouse over the taskbar.

Huh, I did not see that the first time I tried it, but I can see it now... very interesting!

So basically this forces a new reference to the window to be returned, rather than reusing the previous one?

vkareh:
LGTM

rbuj:
IMHO I think it's because of there is a `g_object_ref (win)` that is left over in `gdk_x11_window_foreign_new_for_display`:
https://gitlab.gnome.org/GNOME/gtk/-/blob/3.24.24/gdk/x11/gdkwindow-x11.c#L1307

vkareh:
Makes sense, thanks

PR #1193, opened 2021-01-09
Redundant NULL check before freeing a variable

rbuj:


raveit65 (approved):
LGTM,
builds and runs fine.
Would be nice to check this (redundant) by another reviewer.

lukefromdc:
At the moment I am on the road with limited Internet access, back end of the week.

lukefromdc (approved):
Builds and runs fine. No problem adding the fish to the panel, no problems adding or removing locations to the calendar, or resetting the home time zone from the calendar clock locations. No crash on closing the dialogs either as a problem with this might be expected to cause.

No issues setting a command for the fish to get text from (I used dmesg) either

PR #1195, opened 2021-01-23
mate-panel: remove mate-screensaver support

rbuj:
Test: Add `Lock Screen` to panel:

Screenshot at 2021-01-23 23-02-35

Related https://github.com/mate-desktop/mate-screensaver/issues/243

### DIFF
--- a/mate-panel/panel-util.c
+++ b/mate-panel/panel-util.c
@@ -375,54 +375,28 @@ panel_load_icon (GtkIconTheme  *icon_theme,
 	return surface;
 }
 
-static char* panel_lock_screen_action_get_command(const char* action)
+static char*
+panel_lock_screen_action_get_command (const char* action)
 {
-	char* command = NULL;
-	gboolean use_gscreensaver = FALSE;
-
-	if (panel_is_program_in_path("mate-screensaver-command") && panel_is_program_in_path("mate-screensaver-preferences"))
-	{
-		use_gscreensaver = TRUE;
-	}
-	else if (!panel_is_program_in_path("xscreensaver-command"))
-	{
-		return NULL;
-	}
+	char *command = NULL;
 
 	if (strcmp (action, "prefs") == 0)
 	{
-		if (use_gscreensaver)
-		{
-			command = g_strdup ("mate-screensaver-preferences");
-		}
-		else if (panel_is_program_in_path ("xscreensaver-demo"))
-		{
-			command = g_strdup ("xscreensaver-demo");
-		}
-		else
-		{
-			command = NULL;
-		}
+		if (!panel_is_program_in_path ("xscreensaver-demo"))
+			return NULL;
+
+		command = g_strdup ("xscreensaver-demo");
 	}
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Builds and runs fine, adding the applet and triggering xscreensaver from the panel applet works fine. Does in master too but no point supporting a screensaver we intend to drop for both copyright and security issues.

raveit65:
Why?
Only why one user complains an issue after 10 years.
I do not agree with using x-screensaver.

lukefromdc:
OK, we can consider it this way: it's not a copyright violation to support our existing screensaver, and xscreensaver does not work with our session if we do get a direct copyright complaint from the actual authors of xscreensaver(the only kind that counts). Thus xscreensaver becomes a MAYBE, we can support it but nothing should depend on it.  Distros can then decide for themselves.

I will not cast the deciding vote on this

raveit65:
Where did actual author of x-screensaver said that we violate copyright?

lukefromdc:
So far there is only that 3ed party blog post and the bug report. I don't know much about this sort of law and on such subjects as music and movies (owned by filthy rich publishers) am on the "other side of the house" so should not be the judge of this

raveit65:
Which 3ed party blog and which bugreport?
Are you sure that this aren't fake news?

lukefromdc:
https://github.com/mate-desktop/mate-screensaver/issues/243
and the post mostly about toolkit use and security at
https://www.jwz.org/blog/2021/01/i-told-you-so-2021-edition/

You should assume I have ZERO ability to evaluate the validity of the copyright claim, and zero ability to vouch for the posters

raveit65:
Honestly,
we should make a proper licensing of mate-screensaver instead of removing mate-screensaver.
As the author of this blog said that gnome-screensaver devs never did ask him.
This sounds for me that it is possible to get a proper licensing for mate-screensaver when someone ask him in this blog, if he really is the developer of x-screensaver.
For me is blindly removing m-s, without verifying that this blog is from x-screensaver devs, unnecessary work which should be avoid.

raveit65:
Git source of x-screensaver is located here https://github.com/Zygo/xscreensaver#readme , which pointed to 
This repo pointed to https://www.jwz.org/xscreensaver/
So this blog seems to be valid.
An email address is here https://www.jwz.org/xscreensaver/bugs.html
```
And finally, email your bug report to jwz@jwz.org. 
```
Can you please ask there for permission for proper licensing for mate-screensaver?

flexiondotorg:
I'll email Jamie.

cwendling:
I don't see why support should be removed here so long as mate-screensaver exists.  Even if mate-screensaver itself is infringing some copyrights, *calling* that program doesn't violate any license.  This PR doesn't remove any infringing code, so it's definitely not required.

And if anything, mate-screensaver itself should be fixed to resolve the issue, not every component that might call it should stop doing so.

lukefromdc:
OK to close these PR's?

raveit65:
> OK to close these PR's?

I'll do.

PR #1204, opened 2021-02-20
notification_area: fix memory leak and move XAyatana(New)Label case at the bottom

rbuj:

PR #1205, opened 2021-02-22
travis CI: use base-devel tag for archlinux  docker image

rbuj:
archlinux:base-devel has the [base-devel group](https://archlinux.org/groups/x86_64/base-devel/) installed
https://hub.docker.com/r/archlinux/archlinux

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -69,7 +69,7 @@ after_success:
      fi'
 
 env:
-  - DISTRO="archlinux:latest"
+  - DISTRO="archlinux:base-devel"
   - DISTRO="debian:testing"
   - DISTRO="fedora:latest"
   - DISTRO="ubuntu:rolling"


raveit65:
This doesn't fix the problem.
I have feeling we need to define a new build dependency for pacman.

rbuj:
> This doesn't fix the problem.
> I have feeling we need to define a new build dependency for pacman.

We need to downgrade glibc https://github.com/mate-desktop/mate-panel/pull/1204#issuecomment-783221837

raveit65:
For me this should be fixed by archlinux or by someone who is familiar with it, and can spend more of his spare time than me.
Really important is that builds for fedora and debian are running, for releasing and for extended build warnings.
And we see that our commits are running successfully.
The rest is cosmetics and building against 4 distros seems bit overload.
It's only a nice-to-have for me.

Issue #1138, opened 2020-08-26
Clock applet crashed once the first location was added


PSLLSP:
#### Expected behaviour

Just add location, do not crash.

#### Actual behaviour

Applet crashed, and dialog appeared with information that it crashed and offered an option to reload it. Reload worked.

#### Steps to reproduce the behaviour

Fresh install of Linux MINT. Add location to the clock applet on toolbar.

#### MATE general version

https://termbin.com/ub12

#### Linux Distribution

Linux Mint MATE 20.0

vrozkovec:
I second this. My clock applet crashes anytime I add location.
https://termbin.com/1cn2
[crash.tar.gz](https://github.com/mate-desktop/mate-panel/files/5819442/crash.tar.gz)

lukefromdc:
MATE 1.20 is now quite old, current release is 1.24

vrozkovec:
After updating Linux Mint to 20.1, MATE 1.24 came with it and the problem disappeared.

vrozkovec:
I was too quick. Still crashes:
https://termbin.com/bm7m
[crash.tar.gz](https://github.com/mate-desktop/mate-panel/files/5832412/crash.tar.gz)

cramaker:
This is still present in 1.24.1 and occurs after pressing 'OK' on the 'Choose Location' box. It's reproducible each time a location is added, not just the first location.

A stack trace is here if helpful: https://gist.github.com/cramaker/195465d25d3e705fb9ee36fb2cc4331b

lukefromdc:
I had this happen several times in an initially unrelated test yesterday, only to have the problem refuse to reappear in further tests in another session. As I have the panel built in-process, it led the whole panel to crash and repeatedly restart. Running mate-panel --replace brought it up with most or all of my in-process applet not being added. 

 Very strange. I was able to get this stacktrace before the problem quit and did not return. From the libgio references it looks like we are either hitting an issue in libglib or else feeding it bad data at some point. From my inconsistant results I wonder if a race condition (data not ready when a function is called maybe?) is involved:
```

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from mate-panel...
(No debugging symbols found in mate-panel)
(gdb) run
Starting program: /usr/bin/mate-panel 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[...]

PR #1207, opened 2021-02-28
clock: Add option to show all location's local time in display

GreenArchon:
Hi! Not sure if this is useful to enough people to warrant merging but I figured I'd check.

I have to work with multiple timezones, enough that clicking on the applet to get multiple local times was annoying... So I added an option to the clock applet to display the local time of all locations in the locations tabs of the settings.

image

This MR:

* Adds the logic to optionally display all locations' formatted time in the task bar
* Adds a `show-all-locations` boolean in the settings schema
* Add a tickbox to enable the option in the settings
    *    image

I'm not familiar with how translations are supposed to be done however.

## Tests

This was developed tested on mate-panel 1.24 (apt-get source in Ubuntu 20.04) then ported to master. Patch applied almost cleanly, with the exception that the `cd->locations` list was now a singly-linked list (GSList) instead of a double-linked one (GList).

Works fine, only issue is adding locations sometimes crashes the applet, but it also does that without these changes so I think that is #1138 and not me :)

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -81,6 +81,7 @@
 #define KEY_SHOW_TEMPERATURE        "show-temperature"
 #define KEY_CUSTOM_FORMAT        "custom-format"
 #define KEY_SHOW_WEEK                "show-week-numbers"
+#define KEY_SHOW_ALL_LOCATIONS    "show-all-locations"
 #define KEY_CITIES                "cities"
 #define KEY_TEMPERATURE_UNIT        "temperature-unit"
 #define KEY_SPEED_UNIT                "speed-unit"
@@ -148,6 +149,7 @@ struct _ClockData {
         gboolean     showseconds;
         gboolean     showdate;
         gboolean     showweek;
+        gboolean     show_all_locations;
         gboolean     show_weather;
         gboolean     show_temperature;
 
@@ -543,6 +545,7 @@ format_time (ClockData *cd)
         char *utf8;
 
         utf8 = NULL;
+        hour[0] = '\0';
 
         tm = localtime (&cd->current_time);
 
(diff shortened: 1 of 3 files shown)


lukefromdc:
This could cause the clock applet to expand to the point it takes up the entire panel (or more) if the user has enough locations configured, and if there are three or more timezones configured, displaying them without labelling them could also be confusing. I think the panel should only show the me for the timezone the computer is currently using, with the others shown in the calendar as is the case now

GreenArchon:
Well, technically there's a 255 bytes limit :grimacing: 

I'm really fine with you not merging - I pushed this in case it'd be useful to someone else than just me and a few coworkers (and, let's be honest, to try not to have to carry my own patches...).

One last suggestion, and no is a very valid answer! - I get not feature creeping the clock. Could this be a hidden option, settable only by dconf, a bit like custom strftime specifiers appear to be right now?

lukefromdc:
I will leave that choice to the rest of the team for now

zhuyaliang:
I think it's better for the panel to display only the time of the current region, and the time of other time zones should be displayed in the calendar。
2021-03-01 15-06-05屏幕截图

raveit65:
Yes, only the main time should be displayed in the panel.
You can't be in 3 citizens at the same time ;)

applets/clock/clock.c
```diff
@@ -561,19 +564,46 @@ format_time (ClockData *cd)
                         utf8 = g_strdup_printf ("@%3.2f", itime);
                 else
                         utf8 = g_strdup_printf ("@%3d", (unsigned int) itime);
-        } else if (cd->format == CLOCK_FORMAT_CUSTOM) {
-                char *timeformat = g_locale_from_utf8 (cd->custom_format, -1,
+        } else {
+                char *timeformat;
+                gboolean timeformat_dynamic = FALSE;
```
rbuj, line 569: I don't understand the meaning/purpose of the variable `timeformat_dynamic`.

rbuj (changes requested):
The clock applet should only show the current system clock on the panel.

GreenArchon:
Seeing the overall negative reactions I won't take more of your time on this. Thanks for spending the time to humor me!

PR #1208, opened 2021-03-01
Fix clock applet crash after adding time zone location

zhuyaliang:
Fix clock applet crash after adding City Location
#1138 
### test
```
... make pr ... make install
pkill -f /usr/libexec/clock-applet
pkill -f mate-panel

clock-applet->edit->location->add
```

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -2252,7 +2252,7 @@ location_start_element (GMarkupParseContext *context,
         if (current && clock_location_is_current_timezone (loc))
                 clock_location_make_current (loc, NULL, NULL, NULL);
 
-        data->cities = g_slist_append (data->cities, loc);
+        data->cities = g_slist_append (data->cities, g_object_ref (loc));
 }
 
 static GMarkupParser location_parser = {
@@ -2682,7 +2682,7 @@ run_prefs_edit_save (GtkButton *button, ClockData *cd)
                  */
                 clock_location_is_current (loc);
 
-                cd->locations = g_slist_append (cd->locations, loc);
+                cd->locations = g_slist_append (cd->locations, g_object_ref (loc));
         }
         g_free (name);
         g_free (city);
--- a/applets/clock/clock-location.c
+++ b/applets/clock/clock-location.c
@@ -92,7 +92,7 @@ clock_location_find_and_ref (GSList       *locations,
         }
 
         if (l != NULL)
-                return g_object_ref (CLOCK_LOCATION (l->data));
+                return CLOCK_LOCATION (l->data);
         else
                 return NULL;
 }


lukefromdc (approved):
Builds and runs fine, though most of the time I cannot invoke the crash this fixes I DID see it (in one session only) a few days ago. Needs testing by someone who reliably gets the crash to determine if this is fixed. 

Codewise I see this moves a g_object_ref to take place when appending a location to the list, which would fix this if the problem is the reference count reaching zero and the variable being freed before being used by adding to the list.

lukefromdc:
Just got a random segfault somewhere in the panel (clock in-process) with this running, could not re-invoke it by adding a city.  No idea the cause but again the libgio reference:
`
[ 1213.879695] mate-panel[6484]: segfault at 8 ip 00007f9a9ed453a2 sp 00007fffa35ec470 error 4 in libgio-2.0.so.0.6700.0[7f9a9ecd8000+109000]
`
Will let this run(restarted) for a while, I did not get random segfaults before.

zhuyaliang:
Code before change adding multiple cities in succession will result in segment error

rbuj:
With or without this PR, I haven't been able to replicate the segment violation either in a system with the latest MATE components installed from git (1), nor in a live media of Fedora 33 MATE Compiz spin up to date (2).

(1) https://gist.github.com/rbuj/bc62164be10a7290f5213c47711c7998#file-test-mate-desktop-sh
(2) https://www.dropbox.com/sh/madaacduuiffylj/AACGWh__uCPjb3QsNYfYyn34a?dl=0&lst=

zhuyaliang:
d
It's easy to reproduce here

lukefromdc:
I'm wondering if having the panel in-process is trapping or otherwise stopping the error most but not quite all of the time, as I've seen it but cannot duplicate it at will for testing

zhuyaliang:
> I'm wondering if having the panel in-process is trapping or otherwise stopping the error most but not quite all of the time, as I've seen it but cannot duplicate it at will for testing

I don't quite understand what you mean

lukefromdc:
The clock applet can be built "in-process" meaning it is a shared object loaded by the panel and running as part of the panel's process, or it can be built "out-of-process" (most applets only out of process) meaning it runs as a separate process. Any part of one process crashing takes all of that process down, while one process (perhaps an out of process applet) can crash without taking down another process(perhaps the panel that invoked it). 

Each method has its advantages and disadvantages. There are some differences in the codepaths, thus it is not unheard of for an issue to exist only for in-process or only for out-of-process builds of an applet. Only the clock, fish, wncklet, and tray support in-process builds at this time. Gnome-panel ported ALL their applets to in-process, note that only in-process applets can run under Wayland, as out-of-process applets rely on xembed (which is X-only) to display within the panel's window(yes, it's a window).

zhuyaliang:
A clock crash does not cause a panel crash，In a word, the problem of clock crash is solved

lukefromdc:
If the clock is built OUT OF PROCESS it can crash but not take down the panel, but cannot be used in Wayland due to Xembed. If the clock is built IN PROCESS it will take the panel down if it crashes-but is theoretically lighter and can be displayed in Wayland.

zhuyaliang:
@lukefromdc 
I built in-process applet according to your tips 
```Applets to build in-process:    wncklet clock fish notification-area```
Before using this PR：
By adding a new city will soon get a segment error, which causes the panel to restart.
After using this PR：
 I can add cities correctly

File: applets/clock/clock.c [hunk 1]
```diff
@@ -2682,7 +2682,7 @@ run_prefs_edit_save (GtkButton *button, ClockData *cd)
                  */
                 clock_location_is_current (loc);
 
-                cd->locations = g_slist_append (cd->locations, loc);
+                cd->locations = g_slist_append (cd->locations, g_object_ref (loc));
```
rbuj, line 2685: I don't think it fixes any errors, as it always creates a new reference to a newly created ClockLocation object, fact that leads to a memory leak.

zhuyaliang: First, the fact is that adding a city will lead to segment errors, which have disappeared through this PR segment

Second, the reason for the segment error is that ```g_object_unret```
Here the memory has been freed
```
static void
free_locations (ClockData *cd)
{
        if (cd->locations != NULL) {
                GSList *l;
                for (l = cd->locations; l; l = l->next)
                        g_object_unref (l->data);
                g_slist_free (cd->locations);
        }
        cd->locations = NULL;
}
```

rbuj: It not NULL here.

VirtualBox_9 Fedora 33_02_03_2021_13_02_30

zhuyaliang: You can try adding cities many times, and it does solve my problem，We can wait for anyone with a segment error to test it

rbuj: I think your problem comes from somewhere else.

lukefromdc:
Any updates on this? It seemed to fix the issue for one user, I wasn't able to trigger the bug at will for a definitive test

zhuyaliang:
> Any updates on this? It seemed to fix the issue for one user, I wasn't able to trigger the bug at will for a definitive test, even by adding multiple cities

Needs testing by someone who reliably gets the crash to determine. (the clock applet crashes after adding the city) to confirm whether this problem is fixed.Although this PR does fix the crash problem in me, it's not convincing,

raveit65:
I added a lot of cities to the clock-applet but i couldn't reproduce the crash, with building applets in-process or out-of-process.
Bildschirmfoto zu 2021-03-07 13-01-40
Your video description recall me at this already fixed report.
https://github.com/mate-desktop/mate-panel/issues/1115
This was fixed by https://github.com/mate-desktop/mate-panel/pull/1113
Exactly with this commit https://github.com/mate-desktop/mate-panel/pull/1113/commits/0da0eadae709f9529fd4af085cc14c62bc95d931  , ihmo.
Which version of mate-panel you have running?
Current master or an older 1.24.x release?

zhuyaliang:
@raveit65 Hi, I should have known the cause of the crash. I'll go to the company tomorrow to test it.I feel that the crash has something to do with the discovery version of the system. My colleagues and I use archlinux, and we all have a crash. I'm waiting to verify this problem in Ubuntu and fedora tomorrow

zhuyaliang:
@raveit65 @lukefromdc  @rbuj  I tested three Linux distributions, ```Fedaro, Ubuntu, and ArchLinux```.
I found that this problem is only normal on ```fedaro```. No matter how many cities are added, the clock applet will not crash
There are crash problems on both ```Ubuntu``` and ```ArchLinux```
This PR can solve the clock applet crash problem of Ubuntu and ArchLinux

rbuj (changes requested):
Please, provide a backtrace for your segfaults, installing all the related debug symbols for getting a suitable info, as well the versions of all packages displayed in the backtrace.

Note I couldn't trigger the segment violation either on debian or fedora.

You can also check that it isn't due because of your locale setting, by not using any locale,  i.e. using the locale en_US-UTF-8 or setting the environment variable LANG when launching clock-applet from command line.

You can also provide a brief description of your change, providing a tracking for the object reference counter, from object creation to destruction.

raveit65:
> I tested three Linux distributions, `Fedaro, Ubuntu, and ArchLinux`.
> I found that this problem is only normal on `fedaro`. No matter how many cities are added, the clock applet will not crash

Because fedora use mate-panel-1.24.1 which include the fixes for https://github.com/mate-desktop/mate-panel/issues/1115

> There are crash problems on both `Ubuntu` and `ArchLinux`

Which versions of mate-panel are used by Ubuntu and archlinux?
I am pretty sure it is mate-panel-1.24.0, which doesn't have the fix for https://github.com/mate-desktop/mate-panel/issues/1115 

Note, and older Ubuntu version, eg. 20.04 doesn't use latest version from  1.24 branch.
It is also possible that archlinux didn't update to latest mate-panel-version.

zhuyaliang:
## Ubuntu
```
Linux mouse 5.8.0-44-generic #50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
mate-panel-1.25.2
```
## ArchLinux

```
Linux ll 5.11.2-arch1-1 #1 SMP PREEMPT Fri, 26 Feb 2021 18:26:41 +0000 x86_64 GNU/Linux
mate-panel-1.25.2
```


https://user-images.githubusercontent.com/36913152/110566552-279d7580-818b-11eb-8a3a-9b153d5fccbe.mp4


https://user-images.githubusercontent.com/36913152/110566561-29ffcf80-818b-11eb-8fbb-2b357e69c4d3.mp4

zhuyaliang:
We don't have to rush to merge this pr. if other people encounter this problem, we'll pay attention to this pr.
@raveit65  I want to ask you a question. If this PR has been waiting, how can I initiate a new pull requests，I hope to get your help

raveit65:
>  I want to ask you a question. If this PR has been waiting, how can I initiate a new pull requests，

You can do a PR from a new branch.
Best is you do a PR always from a branch and not from your master

Can you please run the clock applet with gdb and post the backtrace of the crash here?

raveit65:
@zhuyaliang 
Can you check if you're not affected by this issue from libmateweather?
https://github.com/mate-desktop/libmateweather/issues/75
This one was fixed by https://github.com/mate-desktop/libmateweather/commit/00a0b332f1cb8f7ab0b93eea87ecd4a29180ca9f
with libmateweather-1.25.0 or 1.24.1
Fedora stable repos does have this fix.

zhuyaliang:
@raveit65 After updating ```libmateweather``` the crash disappeared.It seems that this has something to do with ```libmateweather```. I just solved the problem from another way.After closing this PR, I'll reduce some compilation warnings next

raveit65:
Cool, i am so happy that i could help you.
But this shows that some distros didn't updated to latest libmateweather :/

raveit65:
@zhuyaliang 
I guess we can close this now as issue was cased by not updating libmateweather in distros.

lukefromdc:
Note that when 
https://github.com/mate-desktop/libmateweather/pull/100 is used without this (or same change in https://github.com/mate-desktop/mate-panel/pull/1238), we get a crash, while when it is used with this there is no crash. Thus, it appears from this behavior that https://github.com/mate-desktop/libmateweather/pull/100 without this = crash from prematurely clearing these objects, this without  https://github.com/mate-desktop/libmateweather/pull/100 = memory leak from the added ref increment given that no crash resulted on old libmateweather without this. Thus the two changes appear to work but have to be used together.

In my tests of https://github.com/mate-desktop/libmateweather/pull/100 with mate-panel from current master, I got a crash on adding the second city. That crash did not occur with https://github.com/mate-desktop/mate-panel/pull/1238 also applied

PR #1211, opened 2021-03-11
Remove compilation warning  “g_completion....  is deprecated”

zhuyaliang:
Using ```G_GNUC_BEGIN_IGNORE_DEPRECATIONS and ```G_GNUC_END_IGNORE_DEPRECATIONS```  removing compilation warnings

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -252,10 +252,11 @@ panel_run_dialog_destroy (PanelRunDialog *dialog)
 	g_list_free (dialog->completion_items);
 	dialog->completion_items = NULL;
 
+	G_GNUC_BEGIN_IGNORE_DEPRECATIONS
 	if (dialog->completion)
 		g_completion_free (dialog->completion);
 	dialog->completion = NULL;
-
+	G_GNUC_END_IGNORE_DEPRECATIONS
 	panel_run_dialog_disconnect_pixmap (dialog);
 
 	g_free (dialog);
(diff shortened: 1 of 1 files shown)


rbuj:
I disagree with silencing compilation warnings, wrapping deprecations with G_GNUC_BEGIN_IGNORE_DEPRECATIONS, G_GNUC_END_IGNORE_DEPRECATIONS macros.

raveit65:
In general i prefer a replacement for a deprecation and not silencing them.
This makes  only sense for deprecation which we can't fix, eg. GtkAction warnings.
We did that in a few other repos because replacing GtkAction will remove all icons from menus.
That means we can't replace GtkAction before we switch to Gtk4.
So, only in this rare case it is Ok to silence them and when GtkAction warnings make build logs unreadable, ihmo.

zhuyaliang:
I think the same as @raveit65 Some warnings that we can't find a suitable solution to replace，I can silence this warning.

raveit65:
But warnings from this PR should be fixable.

zhuyaliang:
@raveit65 Sorry, after several days of hard work, I didn't find a replacement method of ```g_completion....```，If I have enough time in the future, I will continue to try to solve this problem

mbkma:
@zhuyaliang Maybe [GtkEntryCompletion](https://developer.gnome.org/gtk3/stable/GtkEntryCompletion.html) might be a good starting point.

zhuyaliang:
@mbkma I think the same as you think, and I've already started the work.It is believed that the replacement work can be completed in a short time

lukefromdc:
Can this be closed in favor of
https://github.com/mate-desktop/mate-panel/pull/1219 ?

PR #1212, opened 2021-03-11
remove deprecated gtk_container_set_focus_chain

zhuyaliang:
This function can be removed directly. This function has no effect and brings compilation warning

### DIFF
--- a/mate-panel/panel-ditem-editor.c
+++ b/mate-panel/panel-ditem-editor.c
@@ -582,6 +582,7 @@ panel_ditem_editor_make_ui (PanelDItemEditor *dialog)
 {
 	PanelDItemEditorPrivate *priv;
 	GtkWidget *dialog_vbox;
+	GtkWidget *box;
 
 	priv = dialog->priv;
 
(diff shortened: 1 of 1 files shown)


raveit65:
Why this function has no effect?
From https://developer.gnome.org/gtk3/stable/GtkContainer.html#gtk-container-set-focus-chain
```


gtk_container_set_focus_chain has been deprecated since version 3.24 and should not be used in newly-written code.

For overriding focus behavior, use the GtkWidgetClass::focus signal.
```

rbuj:
Have you checked whether the tab index problem exists when using a gtkbuilder file for defining the layout of the dialog, or even better, the whole dialog?

zhuyaliang:
@rbuj I don`t understand the meaning of ```gtkbuilder tab index```. Can you help me understand it

rbuj:
I mean to use glade for designing the graphical interface instead of coding it in C.

zhuyaliang:
@rbuj I'm not very good at using glade to build a new interface, and this change is very small, just adding a new ```box```

rbuj:
It changes the layout of the dialog, like the commit below:
- https://github.com/GNOME/gnome-panel/commit/f3b48b6a20022a14e7462de98b1200ea5291da5b

Why do you think setting 'focus order', aka 'tab index' and 'tab order', is not longer necessary after changing the layout?
- https://github.com/GNOME/gnome-panel/commit/22d00c42a29d3ba1ca888dfbbd96bfdc3e9a480d

zhuyaliang:
@rbuj To be honest, I don't know what the focus chain means here.I deleted this code for the first time，I just reduced compilation warnings without affecting usage.Can you explain the meaning of setting focus order?

rbuj:
> > It changes the layout of the dialog, like the commit below:
> > 
> > * [GNOME/gnome-panel@f3b48b6](https://github.com/GNOME/gnome-panel/commit/f3b48b6a20022a14e7462de98b1200ea5291da5b)
>
> To be honest, I don't know what the focus chain means here.I deleted this code for the first time，I just reduced compilation warnings without affecting usage.

I wonder why do you propose the change from other author when you don't understand it...

> Why do you think setting 'focus order', aka 'tab index' and 'tab order', is not longer necessary after changing the layout?
> 
> * [GNOME/gnome-panel@22d00c4](https://github.com/GNOME/gnome-panel/commit/22d00c42a29d3ba1ca888dfbbd96bfdc3e9a480d)

The tab order workaround may be obsolete as the original code comes from 2005 and it may be due because of a specific gtk version or gtk2, so it may not be required anymore.

> Can you explain the meaning of setting focus order?

https://en.wikipedia.org/wiki/Tabbing_navigation#Tab_order

zhuyaliang:
@rubj So are we going to close this PR?

rbuj:
> @rbuj I'm not very good at using glade to build a new interface, and this change is very small, just adding a new `box`

PR #1214

raveit65:
> So are we going to close this PR?

I merged the other one because it's just the same plus using a glade file for the dialog seems an improvement.
Thank you for your effort, you're welcome.

PR #1219, opened 2021-04-07
Replace deprecated GCompletion with GtkEntryCompletion 

zhuyaliang:
Using  ```gtk_entry_completion```  replaces that has been discarded ```g_ompletion```
By replacement, a lot of redundant code is removed

### test
```
make pr && make install
Shortcut key Alt + F2
```

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -88,7 +88,7 @@ typedef struct {
 	GHashTable       *dir_hash;
 	GList		 *possible_executables;
 	GList		 *completion_items;
-	GCompletion      *completion;
+	GtkEntryCompletion *completion;
 
 	int	          add_items_idle_id;
 	int		  find_command_idle_id;
@@ -251,10 +251,6 @@ panel_run_dialog_destroy (PanelRunDialog *dialog)
 		g_free (l->data);
 	g_list_free (dialog->completion_items);
 	dialog->completion_items = NULL;
-
-	if (dialog->completion)
-		g_completion_free (dialog->completion);
-	dialog->completion = NULL;
 
 	panel_run_dialog_disconnect_pixmap (dialog);
 
@@ -1418,6 +1414,33 @@ fill_executables (GList *possible_executables,
 	return list;
 }
 
+static GtkTreeModel *
+create_completion_model (GList *list)
+{
+	GtkListStore *store;
+	GtkTreeIter   iter;
+	GList        *l;
+
+	store = gtk_list_store_new (1, G_TYPE_STRING);
+	for (l = list; l; l = l->next)
+	{
+		gtk_list_store_append (store, &iter);
+		gtk_list_store_set (store, &iter, 0, l->data, -1);
+	}
+
+	return GTK_TREE_MODEL (store);
+}
+
+static void
+completion_add_items (GtkEntryCompletion *completion, GList *list)
+{
+	GtkTreeModel          *completion_model;
+
+	completion_model = create_completion_model (list);
+	gtk_entry_completion_set_model (completion, completion_model);
+	g_object_unref (completion_model);
+}
+
 static void
 panel_run_dialog_update_completion (PanelRunDialog *dialog,
 				    const char     *text)
@@ -1434,14 +1457,6 @@ panel_run_dialog_update_completion (PanelRunDialog *dialog,
 
 	list = NULL;
 	executables = NULL;
-
-	if (!dialog->completion) {
-		dialog->completion = g_completion_new (NULL);
-		dialog->possible_executables = fill_possible_executables ();
-		dialog->dir_hash = g_hash_table_new_full (g_str_hash,
-							  g_str_equal,
-							  g_free, NULL);
-	}
 
 	buf = g_path_get_basename (text);
 	prefix = buf[0];
@@ -1486,8 +1501,7 @@ panel_run_dialog_update_completion (PanelRunDialog *dialog,
 	if (list == NULL)
 		return;
 
-	g_completion_add_items (dialog->completion, list);
-
+	completion_add_items (dialog->completion, list);
 	dialog->completion_items = g_list_concat (dialog->completion_items,
 						  list);
 }
@@ -1499,7 +1513,6 @@ entry_event (GtkEditable    *entry,
 {
 	char             *prefix;
 	char             *nospace_prefix;
-	char             *nprefix;
 	char             *temp;
 	int               pos, tmp;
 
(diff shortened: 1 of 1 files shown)


muktupavels:
You could try to explain why code has become redundant...

This does not look like direct replacement to me. What functionality has been lost and/or gained?

For example, why do you think `fill_files_from` is redundant? Also I bet that with this change `enable-autocompletion` setting has become redundant too. Same with `enable-program-list`...

zhuyaliang:
OK, I'll recover the ```fill_files_from``` function

mate-panel/panel-run-dialog.c
```diff
@@ -1748,10 +1724,19 @@ panel_run_dialog_setup_entry (PanelRunDialog *dialog,
 	GtkWidget             *entry;
 
 	dialog->combobox = PANEL_GTK_BUILDER_GET (gui, "comboboxentry");
+	dialog->possible_executables = fill_possible_executables ();
+	dialog->dir_hash = g_hash_table_new_full (g_str_hash,
+							  g_str_equal,
+							  g_free, NULL);
```
raveit65, line 1730: Indents

zhuyaliang: Update indent

raveit65: See how it looks in browser now :/
https://github.com/mate-desktop/mate-panel/pull/1219/commits/b3a27410cb1d462fa3856c557dd29d74446038e0

mate-panel/panel-run-dialog.c
```diff
@@ -1748,10 +1724,19 @@ panel_run_dialog_setup_entry (PanelRunDialog *dialog,
 	GtkWidget             *entry;
 
 	dialog->combobox = PANEL_GTK_BUILDER_GET (gui, "comboboxentry");
+	dialog->possible_executables = fill_possible_executables ();
+	dialog->dir_hash = g_hash_table_new_full (g_str_hash,
+				  g_str_equal,
+				  g_free, NULL);
```
raveit65, line 1730: Indents issue still exists.

mate-panel/panel-run-dialog.c
zhuyaliang: Update again

lukefromdc (approved):
Builds and runs fine here, didn't get any problems using the run dialog, and didn't get any warnings in .xsession-errors from this either.

We even get a new feature: when partially typing a command, we get a drop-down box showing possible completions of that text, rather than a single "best guess." Different behavior but should be more useful

mbkma:
@zhuyaliang You should replace the commit message: Something like `Replace deprecated GCompletion with GtkEntryCompletion `instead of `Replace GtkEntryCompletion with GCompletion`

zhuyaliang:
Update commit message

mate-panel/panel-run-dialog.c
raveit65: Thx :)

raveit65:
Can you please fix your master branch and sync it with upstream before next PR?
I can't pull your branch locally.
```
[rave@mother mate-panel]$ LANG=C
[rave@mother mate-panel]$ git checkout -b zhuyaliang-entry_completion master
Switched to a new branch 'zhuyaliang-entry_completion'
[rave@mother mate-panel]$ git pull https://github.com/zhuyaliang/mate-panel.git entry_completion
From https://github.com/zhuyaliang/mate-panel
 * branch              entry_completion -> FETCH_HEAD
fatal: Not possible to fast-forward, aborting.
[rave@mother mate-panel]$
```
This commit, which isn't on upstream master, breaks everything https://github.com/zhuyaliang/mate-panel/commit/9c50b51c614f63577e53c83472182ec7fbfea475
Or delete you repo before next PR and fork it new.
Well. i was able to fork you repo locally and could create a patch for my rpm from your PR.
But , i can't do this unnecessary work every time ;)

raveit65:
Btw. commit message isn't updated.
Seems you renamed only PR name.

raveit65:
How to rename a commit https://www.atlassian.com/git/tutorials/rewriting-history

zhuyaliang:
@raveit65 Thank you for your help. I rename a commit.Next time PR, I will synchronize master

raveit65:
Hmm, i got complete weird  inconsistent behavior when i enable the application list or start the gui directly with a open application list.
When i type a word in search field the GUI use sometimes new auto completion and sometimes it start searching in the application list.
Can you confirm this?

lukefromdc:
I just retested with the applications list open, and what I get is both the panel applet's application list AND the gtk_entry_completion drop-down list displayed at once, with the gtk_entry_completion list displayed on top of the application list created by the applet.

I looked at 
https://developer.gnome.org/gtk3/stable/GtkEntryCompletion.html#gtk-entry-completion-set-popup-completion
and
https://developer.gnome.org/gtk3/stable/GtkEntryCompletion.html#GtkEntryCompletion--inline-completion
and it looks like we have the option of "inline completion" but if I turn popup completion OFF, than I don't get any completions listed at all. This code did NOT work:
```

	gtk_entry_completion_set_inline_selection (dialog->completion, TRUE);
	gtk_entry_completion_set_inline_completion (dialog->completion, TRUE);
	gtk_entry_completion_set_popup_completion (dialog->completion, FALSE);
```
nor did any variations I tried, don't know what I'm doing wrong with this,

lukefromdc:
Accidently closed trying to comment

lukefromdc:
A workaround for this would be to disable completions when the application list is open. Another problem is that the completion popup extends over other parts of the dialog. 

I think we need to get this to work inline without relying on the popup if this is to be merged. Otherwise we need another solution

raveit65:
Ihmo, we should only search in application-list when it is displayed, and the GtkEntry popup should be be disabled in this case.
When application-list is disabled it makes sense to use new auto-completion.
Or another clear separation between both functions.
Don't know if this is possible.....

zhuyaliang:
@raveit65 @lukefromdc I'm very sorry that I didn't reply in time because of some things at work these days. I will continue to improve this function in the weekend

lukefromdc:
Don't worry about having to be quickly responsive: we are not paid staffers
on a product that is sold to the end users. We
all work on this when we have time to, and as
we can see from the end results this works

zhuyaliang:
@lukefromdc @raveit65 My idea is to disable the pop-up function completely and only keep the automatic completion function

lukefromdc:
This now works the same as before as far as I can tell, looks good to bo

raveit65 (approved):
LGTM,
behavior is same like before.

PR #1220, opened 2021-04-13
Preparation wm strut area

raveit65:
This is a preparation for adding some commits from gnome-panel to fix https://github.com/mate-desktop/mate-panel/issues/303#issuecomment-808874551
This makes it better to clean apply those commits.
In the meanwhile i will prepare a PR for metacity.


vkareh (approved):
/lgtm

PR #1236, opened 2021-07-04
Use absolute path g_get_home_dir replaces "~/"

### LINKED ISSUE #1234: Alt+F2 or Run Application dialog incorrectly interprets ~ (tilde)
#### Expected behaviour
Running commands like `pluma ~/somefile` opens Pluma text-editor which shows contents of `~/somefile`.

#### Actual behaviour
Running command like `pluma ~/somefile` opens strange file name `~/~` in Pluma. Saving file is not possible because of wrong path `/home/user/~/somefile` used.

1. Press <kbd>Alt</kbd>+<kbd>F2</kbd> to open Run Application dialog
1. Type command like `pluma ~/somefile`
1. Press <kbd>Enter</kbd> or click *Run*
1. Wait for Pluma window to be opened

#### MATE general version
1.24.1

#### Package version
1.24.2-0ubuntu1

#### Linux Distribution
Ubuntu MATE 21.04

https://pad.lv/1933376

zhuyaliang:
Use g_get_home_dir function replace “~/”
Fix #1234

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -332,6 +332,9 @@ command_is_executable (const char   *command,
 		       char       ***argvp)
 {
 	gboolean   result;
+	GRegex    *regex = NULL;
+	gsize      argv_len;
+	g_autofree gchar *home_path = NULL;
 	char     **argv;
 	char      *path;
 	int        argc;
@@ -341,6 +344,18 @@ command_is_executable (const char   *command,
 	if (!result)
 		return FALSE;
 
+	regex = g_regex_new ("^~/", 0, 0, NULL);
+	argv_len = g_strv_length (argv);
+	home_path = g_build_filename (g_get_home_dir (), "/", NULL);
+	for (gsize i = 0; i < argv_len; i++)
+	{
+		gchar *tmp_argv = NULL;
+
+		tmp_argv = g_regex_replace_literal (regex, argv[i], -1, 0, home_path, 0, NULL);
+		g_free (argv[i]);
+		argv[i] = tmp_argv;
+	}
+	g_regex_unref (regex);
 	path = g_find_program_in_path (argv[0]);
 
 	if (!path) {


cwendling (changes requested):
It is not trivial to expand `~` correctly, and this simply replaces *all* occurrences of `~/` in the string, which could cause another set of issues. `~/` only expands in specific situations (basically, at the start of an unquoted word)
```terminal
$ echo ~/foo
/home/user/foo
$ echo /~/foo
/~/foo
$ echo ~/foo/~/foo
/home/user/foo/~/foo
$ echo '~/foo'
~/foo
$ echo \~/foo
~/foo
$ echo ~
/home/user
```
At the very least I think `~/` should only be replaced at the start of arguments, if at all.  However, short of actually parsing the command according to shell rules ourselves and performing that specific expansion, it is impossible to do 100% correctly.

That said, if really one wants shell expansions in this dialog, maybe the command should simply be passed to a shell, similar to `sh -c COMMAND` instead.

IMHO supporting this kind of features is a rabbit hole that is probably not worth falling into; but again if we *really* want this to feel like a shell, just use one to run the command and get other expansions like environment, whildcards and such.

mate-panel/panel-run-dialog.c
```diff
@@ -326,17 +326,34 @@ panel_run_dialog_set_icon (PanelRunDialog *dialog,
...
 		       char       ***argvp)
 {
 	gboolean   result;
+	g_autofree gchar *new_command = NULL;
 	char     **argv;
 	char      *path;
 	int        argc;
 
-	result = g_shell_parse_argv (command, &argc, &argv, NULL);
+	new_command = replace_command_string (command);
```
vkareh, line 355: Instead of replacing the entire command, I would iterate through argv[1..n] and use the `^~/` regex. Something like:
```
for (int i = 1; i < g_strv_length (argv); i++) {
  regex = g_regex_new ("^~/", 0, 0, NULL);
  path = g_build_filename (g_get_home_dir (), "/", NULL);
  argv[i] = g_regex_replace_literal (regex, argv[i], -1, 0, path, 0, NULL);
}
```

zhuyaliang: Good advice, but index can start from 0  ```int i = 0```
For example, run the following command
```~/pluma ~/some```

**Comment by zhuyaliang** [comment]
@cwendling Use wildcards to replace "~ /" at the beginning of a word

File: mate-panel/panel-run-dialog.c [hunk 2]
```diff
@@ -341,6 +343,13 @@ command_is_executable (const char   *command,
 	if (!result)
 		return FALSE;
 
+	for (int i = 0; i < g_strv_length (argv); i++)
+	{
+		regex = g_regex_new ("^~/", 0, 0, NULL);
```
cwendling, line 348: This could be outside the loop as it'll be the same for all iterations

cwendling (changes requested):
There are leaks to be fixed.  And **disclaimer**: I didn't test it yet.

In any case, this still suffers from not supporting escaping the `~` as the shell syntax that would do that has already been stripped. `pluma '~/foo'` will run `pluma /home/user/foo`, regardless of the quotes (same goes for other possible escapes).
It's better I agree (as I suggested before), but I still fear the cases where one would want a literal `~/` at the start of their command for whatever reason.  It probably is very rare, but there would be no workaround to actually do it.

BTW in all normal situations, the run dialog will have a working directory in the user's some, so `pluma foo` will already do the same as what you're trying to get `pluma ~/foo` to do.

mate-panel/panel-run-dialog.c
```diff
@@ -341,6 +343,13 @@ command_is_executable (const char   *command,
...
+		home_path = g_build_filename (g_get_home_dir (), "/", NULL);
```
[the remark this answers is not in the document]
zhuyaliang: Fix memory leak

mate-panel/panel-run-dialog.c
```diff
@@ -341,6 +343,13 @@ command_is_executable (const char   *command,
...
+		argv[i] = g_regex_replace_literal (regex, argv[i], -1, 0, home_path, 0, NULL);
```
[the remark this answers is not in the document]
zhuyaliang: Fix memory leak

mate-panel/panel-run-dialog.c
```diff
@@ -341,6 +342,18 @@ command_is_executable (const char   *command,
...
+	for (int i = 0; i < g_strv_length (argv); i++)
+	{
+		g_autofree gchar *home_path = NULL;
+		g_autofree gchar *tmp_argv = NULL;
+
+		regex = g_regex_new ("^~/", 0, 0, NULL);
+		home_path = g_build_filename (g_get_home_dir (), "/", NULL);
+		tmp_argv = g_regex_replace_literal (regex, argv[i], -1, 0, home_path, 0, NULL);
+		g_free (argv[i]);
+		argv[i] = g_strdup (tmp_argv);
```
cwendling, line 354: you could avoid the `g_strdup()` by not marking `tmp_argv` as `g_autofree` (or transfer ownership manually by assigning `NULL` to `tmp_argv` once it has been assigned to `argv[i]`).

mate-panel/panel-run-dialog.c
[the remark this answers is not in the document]
cwendling: The leak is fixed, but again you could simply move the call outside the loop as its result does not depend on the iteration we're in.  Doing this for all iterations has unnecessary overhead.

mate-panel/panel-run-dialog.c
```diff
@@ -341,6 +342,18 @@ command_is_executable (const char   *command,
 	if (!result)
 		return FALSE;
 
+	for (int i = 0; i < g_strv_length (argv); i++)
```
cwendling, line 345: This could be optimized as `for (gsize i = 0; argv[i] != NULL; i++)`.  `g_strv_length()` is a somewhat costly operation (it traverses the list to count items, [basically `gsize len = 0; while (argv[len]) len++; return len;`)](https://gitlab.gnome.org/GNOME/glib/-/blob/main/glib/gstrfuncs.c#L2949), and this calls it once per loop iteration.  As the length of `argv` doesn't change in the loop body, there's not need to re-compute it each time.
Alternatively, you could cache the length in a variable: `gsize argv_len = g_strv_length (argv); for (gsize i = 0; i < argv_len; i++)`.

mate-panel/panel-run-dialog.c
cwendling: …which would lower overhead.

cwendling (commented):
Leaks seem to be fixed, but there are a few simple optimizations that would be worth it (basically, avoiding duplicate work and unnecessary memory copy)

zhuyaliang:
@cwendling Your advice is very good, which has taught me a lot.

cwendling (commented):
There are still a couple of minor things that could be fixed, but no major issue I can see anymore. :+1:

I'm still not fond of the feature, but I guess the possible drawbacks I mentioned are probably minor enough that if the feature is useful to some they don't have to be a blocker.

I did **not** test this myself yet, though.  I plan on trying to do so, but I can't guarantee when I'll have time for it, so if others feel like, please go ahead :slightly_smiling_face:

zhuyaliang:
@cwendling I updated the code

raveit65:
@cwendling 
Can you please test. I'd like to do a release.

cwendling (approved):
LGTM, and a little testing shows it works as expected.  It does show the flaws I suggested like `pluma '~/foo'` and `pluma \~/foo` being treated as `pluma ~/foo` (no quotes), but apart from that it seems to behave as expected.

cwendling:
I still thinks it would make more sense to run the command in a shell if we really want such fancy features, but the current status works as advertised and doesn't seem to include bugs.

raveit65 (approved):
LGTM

PR #1251, opened 2021-08-30
window-list: restore grouping radio button selection on startup

### LINKED ISSUE #1250: Windows list problem: loss of config
#### Expected behaviour
When windows grouping is configured  to be 'Always', then it remains so until configured otherwise.

#### Actual behaviour
Windows grouping configuration is lost on each login. Instead of the configured 'Always' it comes up with 'Never'.
Additionally, I have a vertical panel, and the behavior of 'Never' is quite strange there: when starting new instances of an application, instead of getting new icons,  the icons are grouped, but the height of the group is increasing. And the strangest of all is that the heights of the other application icons in the window list grow in parallel. The operation of the windows list in a horizontal panel seems to be all right.

#### MATE general version
1.26.0 

#### Linux Distribution
Fedora 34

mbkma:
This commit should probably be in 1.26.1.
Fixes #1250

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -43,11 +43,11 @@
 #define WINDOW_LIST_PREVIEW_SCHEMA "org.mate.panel.applet.window-list-previews"
 #endif /* HAVE_WINDOW_PREVIEWS */
 
-enum TasklistGroupingType {
+typedef enum _TasklistGroupingType {
   TASKLIST_NEVER_GROUP,
   TASKLIST_AUTO_GROUP,
   TASKLIST_ALWAYS_GROUP
-};
+} TasklistGroupingType;
 
 typedef struct {
 	GtkWidget* applet;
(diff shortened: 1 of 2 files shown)


lukefromdc (commented):
I found that the issue could also be invoked simply by restarting mate-panel. With this commit, the window list menu's radio buttons now save your configuration, but it is not actually applied (only shown as applied in the applet's  right click menu) on restarting the panel.

In master the user's configuration is not saved in the menu either. Thus, we have the menu saving the configuration, but we still need to read it and apply it on starting the panel

lukefromdc (approved):
This now works as intended, looks good to go on my end

mbkma:
Please wait a moment with reviewing as I am restructuring this PR.

mbkma:
see #1252

raveit65:
I guess this branch can be deleted?

mbkma:
yes

PR #1223, opened 2021-04-30
window-list: use GtkNotebook, improve codebase

mbkma:
Similar to #1221 

window-list.ui: use GtkNotebook
Set some callback functions in the ui file.

window-list.c: use g_settings_* functions directly instead of using stored values
use g_settings_bind when possible

For tabs to spaces I can do another PR, when this one is merged.

Screenshot at 2021-05-01 17-35-33
Screenshot at 2021-05-01 17-35-45

PR #1252, opened 2021-09-01
Fix window list

### LINKED ISSUE #1250: Windows list problem: loss of config

mbkma:
I messed up somewhat with commit bcfc5255b55e9b3630ab1447defaa6f4edcfc65c: I did not notice that they used the `::changed` signal on purpose to make the properties dialog work even when changing the settings in dconf-editor for example.

This PR reverts that commit but retains the UI file changes (GtkNotebook) made in https://github.com/mate-desktop/mate-panel/pull/1223 and fixes the thumbnails checkbox (which had no effect before #1223 ).

Please test that all options are remembered after `mate-panel --replace` and that the properties dialog respond to changes made with dconf-editor.

Fixes #1250 (This should probably be in 1.26.1)

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -43,18 +43,26 @@
 #define WINDOW_LIST_PREVIEW_SCHEMA "org.mate.panel.applet.window-list-previews"
 #endif /* HAVE_WINDOW_PREVIEWS */
 
-enum TasklistGroupingType {
+typedef enum {
   TASKLIST_NEVER_GROUP,
   TASKLIST_AUTO_GROUP,
   TASKLIST_ALWAYS_GROUP
-};
+} TasklistGroupingType;
 
 typedef struct {
 	GtkWidget* applet;
 	GtkWidget* tasklist;
 #ifdef HAVE_WINDOW_PREVIEWS
 	GtkWidget* preview;
+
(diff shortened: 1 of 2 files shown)


lukefromdc (approved):
This looks good, and fixes two problems: 

1: the window grouping settings configured by the user are now saved across restarts of mate-panel (as they were always supposed to be)

2: Turning window thumbnails on or off is now applied immediately

raveit65 (approved):
All windows-list settingsare ther after a session restart

raveit65:
@vkareh 
Please can you do a review?
.....before i make another bogus release.

vkareh (approved):
works fine for me

PR #1272, opened 2021-11-28
Use g_clear_pointer() and g_clear_object()

rbuj:


lukefromdc (approved):
A build with all possible applets in-process on Debian Unstable completes and this runs without any obvious problems. Note that we still have some memory leaks in the panel, as RAM use slowly grew from 50 to 60B in maybe 10 min of testing and deliberate interaction with the panel and its included applets. Not a lot compared to memory leaks of the past (especially w in-process applets) but only 10 min or so of testing.

While I got two blocks on non-deprecation warnings on this build, they are identical to those in master and thus not affected by this PR.

This simplifies code quite a bit, taking advantage of functions that have been in glib quite a while. We take out some NULL checks but documentation on this states that the g_clear_object simply does nothing "if the reference is NULL" though also that "object_ptr must not be NULL"

Here is the documentation in question:

https://www.freedesktop.org/software/gstreamer-sdk/data/docs/2012.5/gobject/gobject-The-Base-Object-Type.html

g_clear_object ()

void                g_clear_object                      (volatile GObject **object_ptr);

Clears a reference to a GObject.

object_ptr must not be NULL.

If the reference is NULL then this function does nothing. Otherwise, the reference count of the object is decreased and the pointer is set to NULL.

This function is threadsafe and modifies the pointer atomically, using memory barriers where needed.

A macro is also included that allows this function to be used without pointer casts.

object_ptr :
	a pointer to a GObject reference

Since 2.28

applets/clock/clock-location.c
```diff
@@ -210,17 +210,14 @@ clock_location_finalize (GObject *g_obj)
                                               G_CALLBACK (network_changed),
                                               CLOCK_LOCATION (g_obj));
 
-        g_free (priv->name);
-        g_free (priv->city);
+        g_clear_pointer (&priv->name, g_free);
+        g_clear_pointer (&priv->city, g_free);
```
mbkma, line 214: What is the advantage of using `g_clear_pointer()` in this case? I know that the pointers are set to `NULL`, but do we need this here?

rbuj: avoid issues with inheritance:

> instance finalization function, should finish the finalization of the instance begun in dispose and chain up to the finalize method of the parent class.

lukefromdc:
We need to fix a conflict in panel-widget.c here. Other than that, what is the status of this?

applets/clock/clock-location.c
rbuj: Before freeing a variable you can check wether it's NULL, and free it when it isn't NULL, next you can assign the variable to NULL to avoid to free it again, either because it can call dispose several times or because it can call dispose before calling finalize.
```C
/**
 * g_clear_pointer: (skip)
 * @pp: (not nullable): a pointer to a variable, struct member etc. holding a
 *    pointer
 * @destroy: a function to which a gpointer can be passed, to destroy *@pp
 *
 * Clears a reference to a variable.
 *
 * @pp must not be %NULL.
 * 
 * If the reference is %NULL then this function does nothing.
 * Otherwise, the variable is destroyed using @destroy and the
 * pointer is set to %NULL.
 *
 * A macro is also included that allows this function to be used without
 * pointer casts. This will mask any warnings about incompatible function types
 * or calling conventions, so you must ensure that your @destroy function is
 * compatible with being called as `GDestroyNotify` using the standard calling
 * convention for the platform that GLib was compiled for; otherwise the program
 * will experience undefined behaviour.
 *
 * Since: 2.34
 **/
#undef g_clear_pointer
void
g_clear_pointer (gpointer      *pp,
                 GDestroyNotify destroy)
{
  gpointer _p;

  _p = *pp;
  if (_p)
    {
      *pp = NULL;
      destroy (_p);
    }
}
```

rbuj:
@lukefromdc done, I think current status is awaiting approval from  @mbkma and @lukefromdc, however more reviewers are always welcomed :-)

applets/clock/clock-location.c
mbkma: got it, thanks for the clarification

mbkma (approved):
Code looks good. I installed mate-panel with this PR applied and had no issues over the past week or so.

PR #1275, opened 2021-12-14
applets/wncklet/wayland-backend: Add basic icon support

Consolatis:
This commit uses GTK to find an icon with the same name as the wayland app_id.
A more complete implementation would need to search for .desktop files matching the app_id, get the icon name and finally use GTK to find the proper icon based on that name.

For a discussion about this topic see https://gitlab.freedesktop.org/wayland/wayland-protocols/-/issues/52

### DIFF
--- a/applets/wncklet/wayland-backend.c
+++ b/applets/wncklet/wayland-backend.c
@@ -26,6 +26,7 @@
 #endif
 
 #include <gdk/gdkwayland.h>
+#include <gio/gdesktopappinfo.h>
 
 #include "wayland-backend.h"
 #include "wayland-protocol/wlr-foreign-toplevel-management-unstable-v1-client.h"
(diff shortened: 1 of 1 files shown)


Consolatis:
Added "the more complete implementation" and now Pluma has an icon as well.
As written in the other PR this is my first time working with GTK / GIO so please verify the memory management is done correctly.

Specifcally: When I am doing a g_object_unref (G_OBJECT (icon)) at line 222 gtk_image_set_from_gicon() (actually likely one of its internal callbacks) complains about icon being not a GIcon. I was under the impression that the call should have increased the ref counter for that icon instance so we can unref it from our end.

applets/wncklet/wayland-backend.c
```diff
@@ -207,7 +209,27 @@ foreign_toplevel_handle_app_id (void *data,
 				struct zwlr_foreign_toplevel_handle_v1 *toplevel,
 				const char *app_id)
 {
-	/* ignore */
+	ToplevelTask *task = data;
+
+	gchar *app_id_lower = g_utf8_strdown (app_id, -1);
+	gchar *desktop_app_id = g_strdup_printf ("%s.desktop", app_id_lower);
```
wb9688, line 215: This is unfortunately not complete. In some cases, especially with X11 apps, the app ID is not the name of the .desktop file but the value of the `StartupWMClass` key inside it.

Consolatis: While this may be true it is not "best practice" to define the wayland app_id, at least based on the comment in xdg-shell.xml [1]:
> As a best practice, it is suggested to select app
> ID's that match the basename of the application's .desktop file.
> For example, "org.freedesktop.FooViewer" where the .desktop file is
> "org.freedesktop.FooViewer.desktop".

I think this issue should instead be handled at the xwayland bridge so the app_id it generates based on WmClass points to a valid .desktop basename (if one exists).

Doing a lowercase of the app_id seems not best practice either though. I just assumed it would be because the app_id I got for pluma was "Pluma" but the .desktop file for pluma is actually called pluma.desktop.

In any case this is better than not showing any icons at all and can be changed later on.
The  only downside of not parsing *all* desktop files to match based on StartupWMClass is that some icons may show up with the theme's default broken image.

[1] https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml#L654

wb9688: Yes, for native Wayland apps. There are however still X11-only apps that run using XWayland, and X11's `WM_CLASS` is of course a bit different from Wayland's `app_id`. I do however agree that only handling native Wayland apps' `app_id` properly is better than not handling icons at all on Wayland.

lukefromdc: It's been a year and a half, we can fix the rest of this later. Icons are now working with this

lukefromdc (approved):
Finally got wayfire to work again, thanks to current wayland being able to turn around and use x11 as its backend, providing wayland inside another terminal window. That bypasses the dbus-launch issue

Icons in window buttons are functional with this, which is a real step forward. Merging as this has zero effect on x11 operation and is a real step forward for using mate-panel in a wayland session even though we are a long way from a mate-wayland session.

We need to port everything in mate-applets to in-process or at least offer that as an option so the rest of the applets can be built to support wayland

PR #1276, opened 2021-12-14
applets/wncklet/wayland-backend: Add basic context menu

Consolatis:
Add a basic shared context menu.
I could only verify 'close' on sway because it doesn't support set_minimize or set_maximize but they should work just fine.

This is my first time doing anything with GTK so please verify g_new0 / gtk_widget_destroy / g_free is working as intended and it neither leaks memory nor tries to access free'd memory.

### DIFF
--- a/applets/wncklet/wayland-backend.c
+++ b/applets/wncklet/wayland-backend.c
@@ -34,8 +34,18 @@ static const int window_button_width = 140;
 
 typedef struct
 {
+	GtkWidget *menu;
+	GtkWidget *maximize;
+	GtkWidget *minimize;
+	GtkWidget *on_top;
+	GtkWidget *close;
+} ContextMenu;
+
+typedef struct
+{
 	GtkWidget *list;
 	GtkWidget *outer_box;
+	ContextMenu *context_menu;
 	struct zwlr_foreign_toplevel_manager_v1 *manager;
 } TasklistManager;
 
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
On a test build of this, the build and installation went fine, and things worked normally on Xorg.
I then tried to get the panel running on my (very old) wayfire install, which used to work but now errors out on trying to launch the panel. With or without the xorg mate-session running made no difference, I got this error from dbus on trying to run the panel under wayfire, which I doubt has anything to do with this PR:

`** (mate-panel:19455): WARNING **: 12:53:05.092: Cannot register the panel shell: Error spawning command line “dbus-launch --autolaunch=0caba25203a261b01af20e4e0000040f --binary-syntax --close-stderr”: Child process exited with code 1`

Thus I am not now able to test this, and I don't know how to debug that error from dbus or wayfire.

Consolatis:
The wayland implementation of mate panel (or really any other kind of panel) requires wlr-layer-shell support to work in a meaningful way like docking to the top / bottom.
The tasklist wncklet applet which this PR extends additionally requires wlr-foreign-toplevel-management.

It's quite possible your version of wayfire doesn't implement either of those.
I tested this PR with a recent version of sway (activate / close only) and a recent version of labwc (activate / maximize / minimize only).

lukefromdc:
layer-shell is used in the version of wayfire I have (build Jan 2021) but there might have been too many changes since then. No time at the moment to rebuild all of it from scratch

lukefromdc (approved):
I have finally gotten mate-panel to work in wayfire again. Still get the dbus-launch problem on starting in a separate vt on wayfire, but now wayfire can be run with x11 as a backend, giving you wayland inside the terminal, sort of like x11-wayland in reverse. In this mode, mate-panel will run, and I get a functional context menu with this.

For all that, the context menu renders, it renders in the right place, and maximize, unmaximize, minimize, and close all work. The option to close windows by middle click does not work, but that's separate from the context menus and is for another PR

PR #1277, opened 2021-12-15
clock: fix temperature_unit_changed and -Wsign-compare warnings

rbuj:
`if (oldvalue != cd->speed_unit)` should be `if (oldvalue != cd->temperature_unit)` in `temperature_unit_changed`

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1921,7 +1921,8 @@ format_changed (GSettings    *settings,
                 gchar        *key,
                 ClockData    *clock)
 {
-        int          new_format;
+        ClockFormat new_format;
+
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
A test build finished fine, but on testing broke displaying both the weather and the temperature(either one) in the panel applet button itself. Still showed up in the calendar display but not on the panel

rbuj:
You need to change the temperature unit using dconf-editor.

lukefromdc (approved):
The latest change brings back the weather and temperature display in the panel that didn't work in the first attempt. Changing the temperature units from the clock preferences->weather combo box also works as expected

PR #1282, opened 2021-12-17
clock: Define _clock_get_widget instead of using one sentence method

rbuj:


lukefromdc (approved):
Very simple change and works fine. We get only the old deprecations warnings on build.
Merging this, we have a ton of open PR's and this one is simple

PR #1283, opened 2021-12-17
Remove unnecessary cast in g_signal_connect* call

rbuj:
In addition, use dashes for signals: applet-added, applet-removed, applet-move, back-change, size-change

### DIFF
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -82,7 +82,7 @@ clock_location_tile_new (ClockLocation *loc,
 
         g_signal_connect (priv->weather_icon, "query-tooltip",
                           G_CALLBACK (weather_tooltip), this);
-        priv->location_weather_updated_id = g_signal_connect (G_OBJECT (loc), "weather-updated",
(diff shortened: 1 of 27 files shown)


lukefromdc (approved):
Builds and runs fine, I checked changing several preferences to check that signals work, notably changing timezones from the calendar by changing locations. Note that when changing time zones to test signals the time itself has to update in order for the change to be reflected in the clock applet, same as with master. If you set the time preferences to display seconds this change will be displayed almost immediately as displaying seconds forces an update every second. Same behavior as in master

raveit65:
@rbuj  
Can you please do a push to get an deployment for travis?

rbuj:
@raveit65 done

raveit65:
@rbuj 
Ohh, looks like we have conflict after merging last PRs.

rbuj:
@raveit65 done. In addition, use dashes for signals: applet-added, applet-removed, applet-move, back-change, size-change.

raveit65 (approved):
LGTM,
applet runs fine.

PR #1284, opened 2021-12-18
Use dashes instead of underscores in signal names

rbuj:


lukefromdc (approved):
This builds and runs fine, note that when changing time zones to test signals the time itself has to update in order for the change to be reflected in the clock applet, same as with master

raveit65:
Hmm, i guess this is needed because of glade changes in UI files?

rbuj:
Both cases are recognized by g_signal. The underscores are also valid, but they are no longer used in new code, as well as you said, glade uses dashes.

raveit65 (approved):
LGTM

PR #1290, opened 2022-01-04
notification_area: fix build warning -Wbad-function-cast

rbuj:
```
testtray.c:89:49: warning: cast from function call of type 'gint' (aka 'int') to non-matching type 'GtkOrientation' [-Wbad-function-cast]
  GtkOrientation orientation = (GtkOrientation) gtk_combo_box_get_active (combo);
```
https://docs.gtk.org/gtk3/method.ComboBox.get_active.html

### DIFF
--- a/applets/notification_area/testtray.c
+++ b/applets/notification_area/testtray.c
@@ -86,12 +86,16 @@ tray_removed_cb (GtkContainer *box, GtkWidget *icon, TrayData *data)
 
 static void orientation_changed_cb (GtkComboBox *combo, TrayData *data)
 {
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
testtray.c builds clean w no warnings at all with this applied and my normal build settings. On installing and restarting the panel, the tray works normally and orientation changes horizontal to vertical are applied on dragging the tray between them.

I cannot as easily test the tray in mate-panel-test-applets however, as all in-process applets segfault on that test, which works only for out of process applets and my normal builds are in-process

mbkma (approved):
LGTM. GtkOrientation is an enum with positive values only, so if `gtk_combo_box_get_active` returns `-1` the conversion does not work.

PR #1293, opened 2022-01-26
wncklet: do not exceed workspace number when setting n_rows

rbuj:


raveit65:
@rbuj
Can you please explain what commit fixes?

rbuj:
The user can`t specify more workspaces than available when setting their layout.

raveit65:
You mean when set up the workspaces in a  panel.layout file here https://github.com/mate-desktop/mate-panel/tree/master/data/ ?

lukefromdc (approved):
This builds and runs fine. I tested this both with compiz-reloaded and with marco. With march it works as expected, the maximum number of rows becomes the number of workspaces.

With compiz setting the number of rows does not really work with master, as compiz renders its own line or grid of workspaces always in the first row. This PR enforces one row with compiz and what is shown there is configurable from CCSM for both vertical and horizontal workspaces, with master you can set multiple rows but all but the top row will be empty.

One issue that remains affects users switching from marco to compiz: If on marco you have two or more rows set up, they remain configured this way on a switch to compiz, with only the top row used. With this PR you can reduce the number of rows to one and then cannot return it to the non-working (in compiz) multirow configuration without first switching back to marco.
Not sure it's possible to fix this and save a user's marco configuration of this on switching to compiz though.

raveit65:
@rbuj 
I can't reproduce the issue with work-space-switcher at my bottom panel.
Workspace-switcher works as expected here.

Steps to reproduce the issue please.

rbuj:
Try setting a number of columns or rows greater than the number of workspaces.

raveit65:
Got it, issue confirmed.

raveit65 (approved):
LGTM

PR #1303, opened 2022-03-15
clock: cast from function call of type 'gdouble' to non-matching type

rbuj:
```
git submodule init
git submodule update --recursive --remote
CFLAGS="-g -O0 -Wconversion -Wunused-macros -Wunused-parameter" CC=clang ./autogen.sh --prefix=/usr --enable-debug --enable-compile-warnings=maximum && make &> make.log
```
Fix the build warning below:
```
clock.c:1825:23: warning: cast from function call of type 'gdouble' (aka 'double') to non-matching type 'int' [-Wbad-function-cast]
        value = (int) gtk_adjustment_get_value (adj);
                      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1819,11 +1819,11 @@ output_cb (GtkSpinButton *spin,
 {
         GtkAdjustment *adj;
         gchar *text;
-        int value;
+        gdouble value;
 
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Clock works fine with this applied. Simple change too, will merge this

PR #1304, opened 2022-03-16
wncklet: Condition 'sdd->wnck_screen!=NULL' is always false

rbuj:
```
cppcheck --enable=all -q .
```
Fx the cppcheck warning below:
```
applets/wncklet/showdesktop.c:376:23: style: Condition 'sdd->wnck_screen!=NULL' is always false [knownConditionTrueFalse]
 if (sdd->wnck_screen != NULL)
                      ^
applets/wncklet/showdesktop.c:369:21: note: Assignment 'sdd->wnck_screen=NULL', assigned value is 0
 sdd->wnck_screen = NULL;
                    ^
applets/wncklet/showdesktop.c:376:23: note: Condition 'sdd->wnck_screen!=NULL' is always false
 if (sdd->wnck_screen != NULL)
                      ^
```

### DIFF
--- a/applets/wncklet/showdesktop.c
+++ b/applets/wncklet/showdesktop.c
@@ -370,15 +370,20 @@ static void show_desktop_applet_realized(MatePanelApplet* applet, gpointer data)
 
 #ifdef HAVE_X11
 	if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
-		sdd->wnck_screen = wnck_screen_get(gdk_x11_screen_get_screen_number (screen));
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
showdesktop.c builds with only deprecation warnings with this applied and default build warnings. Adding and removing the show desktop applet works fine and once added it works as intended. 

Using cppcheck-gui with this applied and all warning options turned on (buttons pressed) I got "no errors found" for this file. Warnings are fixed, will merge due to the huge backlog.

PR #1305, opened 2022-03-16
panel-multimonitor: Condition '*monitors_ret>0' is always false

rbuj:
```
cppcheck --enable=all -q .
```
Fix the cppcheck warning below:
```
mate-panel/panel-multimonitor.c:255:27: style: Condition '*monitors_ret>0' is always false [knownConditionTrueFalse]
 if (res && *monitors_ret > 0)
                          ^
mate-panel/panel-multimonitor.c:245:18: note: Assignment '*monitors_ret=0', assigned value is 0
 *monitors_ret = 0;
                 ^
mate-panel/panel-multimonitor.c:255:27: note: Condition '*monitors_ret>0' is always false
 if (res && *monitors_ret > 0)
                          ^
```

### DIFF
--- a/mate-panel/panel-multimonitor.c
+++ b/mate-panel/panel-multimonitor.c
@@ -240,20 +240,22 @@ static void
 panel_multimonitor_get_raw_monitors (int           *monitors_ret,
 				     GdkRectangle **geometries_ret)
 {
-	gboolean res = FALSE;
-
 	*monitors_ret = 0;
 	*geometries_ret = NULL;
 
 #ifdef HAVE_X11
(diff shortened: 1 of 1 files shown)


mbkma (approved):
LGTM, `*monitors_ret` is only modified if `HAVE_X11`and `HAVE_RANDR` are true.

lukefromdc (approved):
Builds and runs fine. Could not check with two monitors as my secondary monitor is already packed away for the upcoming move, but I didn't see anything in the code being changed limiting it to the case of more than one monitor. Will merge as this already has another approving review

PR #1309, opened 2022-04-27
notification_area: incorrect place for setting the local variable

rbuj:


aromatic-dev (approved):

cwendling:
Do we have style rules for this?  I don't have an opinion on this change as it doesn't change anything technically, but I wonder if there's something somewhere that actually tells us it's "incorrect" in the context of mate-panel

rbuj:
@cwendling It' is not a good habit to set a variable long before it is used...

cwendling:
it's not a good habit to leave variables uninitialized either, take your pick :)

rbuj:
@cwendling what are you saying? This variable is initialized before it is used.. since you don't get a compilation error nor runtime exception.

cwendling:
it's not, but it's not good practice to have a variable uninitialized because then it's very easy to modify the code in a way that uses the variable before it's initialized, and a C compiler will at the very best warn you (but is not mandated to at all).

Some people will tell you to use C99 and declare the variable inside the function body where it is initialized instead of at the top of the scope so we both win (it's never in an uninitialized state, and is initialized as late as possible).  But I'll then raise the fact that it somewhat hinders readability, and more importantly that it becomes messy if the variable is assigned more than once (the first one has the declaration, but the second just sets it and it's trickier do locate the matching declaration), to what some people will suggest that all variables should be constant (and declared and assigned exactly once), and so on.

All I'm trying to point out is that it's kind of a preference thing, there is not one perfect answer.

Yet again, I'm not against what you're suggesting here, and you could argue your suggestion better matches the code around it, which is a valid point.

cwendling:
it doesn't have to, but if it's the proper way to obtain the initial value, yes

rbuj:
I think you mean C89.

cwendling:
> Initial value? An initial value for the variable settings could be NULL.

It could.  Not very useful but at least not undefined.

> IMHO, by calling g_settings_new function before setting the other variables you're actually obfuscating the code.

I wouldn't go so far as say obfuscating, but indeed as here the value is used once it's fair to set it close to the usage.

> In addition, you aren't following the style of the existing code, where usually the variables are initialized using values or casts.

I'm not sure what this means?  Everything is always initialized with an rvalue of some kind, and I don't see what casts have to do with the matter at hand?  A cast is (basically) syntax for converting the type of an rvalue, so you ought to have an rvalue to cast.

> I think you mean C89.

No, C89 does not allow declarations but at the start of a scope.  C99 introduced this.

---

Anyhow, we're diving deep down a rabbit hole here with a question of style; I just wanted to know whether there was actual written guidance or such, but either way I'm find with it.

rbuj:
@muktupavels done, I just added a comment on the PR.

muktupavels:
How info in PR will help someone who will look at git history?

Also if you are fixing style, then why are not you changing `NaGrid *self = NA_GRID (widget);` part?

rbuj:
Another MATE team member told me to do not write the description in the commit.
`NA_GRID (widget)` is a pointer cast which is used shortly after.

muktupavels:
Oh, then follow MATE rules. :)

aromatic-dev:
You _could_ change the commit message to something like "Move value assignment of local variable to a better place.", but anyway, this commit doesn't make the code _worse_, so ... can we merge this?

raveit65:
> Another MATE team member told me to do not write the description in the commit.

Nope, i told you not to write a book in commit body.
```
If
write a book
than
use pull regress description
else
write a short description what commit does in commit body
````
....haha

**Comment by rbuj** [comment]
Ok, no problem, I thought the title was quite clear enough for such small change, however if you think I need to write a more accurate description, I'll do it, even when I think it isn't required. Do you want to change the title and/or to add a brief description? Do you have any suggestions? I think I have given enough explanations of the change. Maybe it could even be written that way, without mixing different styles, as Albert's comment pointed out
```C
static void
na_grid_realize (GtkWidget *widget)
{
  NaGrid *self = NA_GRID (widget);
  GdkScreen *screen = gtk_widget_get_screen (GTK_WIDGET (self)); 
  GtkOrientation = gtk_orientable_get_orientation (GTK_ORIENTABLE (self));
  NaHost *tray_host = na_tray_new_for_screen (screen, orientation);
  GSettings *settings = g_settings_new ("org.mate.panel"); 

  GTK_WIDGET_CLASS (na_grid_parent_class)->realize (widget);

  /* Instantiate the hosts now we have a screen */
  g_object_bind_property (self, "orientation",
                          tray_host, "orientation",
                          G_BINDING_DEFAULT);

  add_host (self, tray_host);

  if (g_settings_get_boolean (settings, "enable-sni-support"))
    add_host (self, sn_host_v0_new ());

  g_object_unref (settings);
}
```
or
```C
static void
na_grid_realize (GtkWidget *widget)
{
  NaGrid *self;
  GdkScreen *screen;
  GtkOrientation orientation;
  NaHost *tray_host;
  GSettings *settings;

  self = NA_GRID (widget);
  GTK_WIDGET_CLASS (na_grid_parent_class)->realize (widget);

  /* Instantiate the hosts now we have a screen */
  screen = gtk_widget_get_screen (GTK_WIDGET (self));
  orientation = gtk_orientable_get_orientation (GTK_ORIENTABLE (self));
  tray_host = na_tray_new_for_screen (screen, orientation);
  g_object_bind_property (self, "orientation",
                          tray_host, "orientation",
                          G_BINDING_DEFAULT);

  add_host (self, tray_host);

  settings = g_settings_new ("org.mate.panel");
  if (g_settings_get_boolean (settings, "enable-sni-support"))
    add_host (self, sn_host_v0_new ());
  g_object_unref (settings);
}
```

yetist (approved):
LGTM

cwendling (approved):
Not that I really care for this change, but it's fine

lukefromdc (approved):
Builds and runs fine, I'll leave the style decisions to the others as per prior discussion here

Issue #426, opened 2016-04-11
Feature Request: ‘Sticky-Center’ property for panel items  [$120]


epgfm:
It would be nice to be able to "snap" panel items to a relative position at the center of an horizontal panel layout.
I have no clue how hard such a feature would be to implement and I understand this might not seem like something very important, but I feel like it would add quality to the overall interface.
Not having the "sticky-center" makes it dicey to locate items, especially variable width items such as clocks at the dead center of the screen.

I've described further what I mean in the Ubuntu MATE forums [here](https://ubuntu-mate.community/t/mate-panel-feature-request-sticky-center-property-for-panel-items/4913).

And here's a couple of screenshots showing what's doable under the old "fallback" desktop:
http://i.imgur.com/doQF6KE.jpg
http://i.imgur.com/8mevJG4.jpg

In this layout the two system monitors, the two network folders and the clock are all "Sticky-Center".

Cheers

Edit: a downstream bug report for this issue:

https://bugs.launchpad.net/ubuntu-mate/+bug/1708649

##

FennFelis:
U find a way to center items in mate-panel?
I have this trouble:
[http://imgur.com/xzhYWyR](url)
[http://imgur.com/xzhYWyR](url)

The clock are moved when I resize my screen. I want to put in the center of screen independently of screen size, there is any solution?

epgfm:
Not really. You can set the position of an item via gsettings but you have to know the exact width and it's of no help for variable-width items.
I've just posted a [50$ bounty on this](https://www.bountysource.com/issues/32817997-feature-request-sticky-center-property-for-panel-items), feel free to add to it.

Michael-AU:
This has been reported before as early as 2012 (https://github.com/mate-desktop/mate-panel/issues/62) and 2014 (https://github.com/mate-desktop/mate-panel/issues/183). Note in the 2014 report the then main developer had an offer of help but suggested it would not be incorporated. Now the developers want money to do it. I know Mate is a small project with very few devs but to ignore an offer of help and then 2-3 years later ask for money to do what would have been done  for free is a [gobsmack](https://www.urbandictionary.com/define.php?term=gobsmack) moment for FOSS.

DoktorStein:
Surely a very sad moment for **FOSS**.

Toromino:
Is there anyone working on this? :)

thesquash:
[...]

PR #1310, opened 2022-05-06
Add "center-stick" capability applets on the panel

thesquash:
Conventionally, the applets on a MATE Panel are positioned relative to the
left edge of the panel (if the panel is horizontal) or the top edge (if the
panel is vertical).  There has also been some (buggy) support for positioning
of applets relative to the right (or bottom) edge of the panel, so that
applets on the right side of the panel will stay on the right side even if
the user changes screen resolutions or if the panel changes size for some
other reason.

However, many users want to also place applets at or near the center of their
panel(s).  There is no such conventional support for positioning applets
relative to the center of the panel, so users have positioned applets near
the center of the panel -- but the position recorded is relative to the left
side of the panel.  As such, the applets will almost certainly shift over
to the left or right slightly if the panel is ever resized, and the user will
have to reposition all those centered applets yet again.  This is especially
frustrating if the user switches monitors on a regular basis!

This patch radically revamps the MATE Panel's positioning framework, and
deprecates the original "right-stick" feature.  To replace the right-stick
feature, this patch instead associates an "edge relativity" setting with
each and every panel applet:  An applet can be relative to the start (left/
top), end (right/bottom), or center of the panel.  This setting can be
changed using DConf/GSettings, using a custom panel layout file, or even
by simply dragging the applet to the appropriate place on the panel.
(Conventionally, applets are not even right-stuck automatically even when the
user drags the applet over to the far right of the panel!)  As a bonus, when
the user drags an applet across the center of the panel, the applet will
temporarily "stick" to the very center of the panel, to allow the user to
very precisely align any applet they wish.

Fixes:  #426

### DIFF
--- a/mate-panel/applet.c
+++ b/mate-panel/applet.c
@@ -766,12 +766,12 @@ mate_panel_applet_destroy (GtkWidget  *widget,
 }
 
 typedef struct {
-	char            *id;
-	PanelObjectType  type;
-	char            *toplevel_id;
-	int              position;
-	guint            right_stick : 1;
-	guint            locked : 1;
+	char                      *id;
+	PanelObjectType            type;
+	char                      *toplevel_id;
+	int                        position;
+	PanelObjectEdgeRelativity  edge_relativity;
+	guint                      locked : 1;
 } MatePanelAppletToLoad;
 
 /* Each time those lists get both empty,
@@ -894,9 +894,14 @@ mate_panel_applet_load_idle_handler (gpointer dummy)
 
 	panel_widget = panel_toplevel_get_panel_widget (toplevel);
 
-	if (applet->right_stick) {
-		if (!panel_widget->packed)
-			applet->position = panel_widget->size - applet->position;
+	if (applet->edge_relativity == PANEL_EDGE_CENTER ||
+	    applet->edge_relativity == PANEL_EDGE_END) {
+		if (!panel_widget->packed) {
+			if (applet->edge_relativity == PANEL_EDGE_CENTER)
+				applet->position = panel_widget->size/2 + applet->position;
+			else if (applet->edge_relativity == PANEL_EDGE_END)
+				applet->position = panel_widget->size - applet->position;
+		}
 		else
 			applet->position = -1;
 	}
@@ -971,12 +976,12 @@ mate_panel_applet_load_idle_handler (gpointer dummy)
 }
 
 void
-mate_panel_applet_queue_applet_to_load (const char      *id,
-				   PanelObjectType  type,
-				   const char      *toplevel_id,
-				   int              position,
-				   gboolean         right_stick,
-				   gboolean         locked)
+mate_panel_applet_queue_applet_to_load (const char                *id,
+					PanelObjectType            type,
+					const char                *toplevel_id,
+					int                        position,
+					PanelObjectEdgeRelativity  edge_relativity,
+					gboolean                   locked)
 {
 	MatePanelAppletToLoad *applet;
 
@@ -987,12 +992,12 @@ mate_panel_applet_queue_applet_to_load (const char      *id,
 
 	applet = g_new0 (MatePanelAppletToLoad, 1);
 
-	applet->id          = g_strdup (id);
-	applet->type        = type;
-	applet->toplevel_id = g_strdup (toplevel_id);
-	applet->position    = position;
-	applet->right_stick = right_stick != FALSE;
-	applet->locked      = locked != FALSE;
+	applet->id              = g_strdup (id);
+	applet->type            = type;
+	applet->toplevel_id     = g_strdup (toplevel_id);
+	applet->position        = position;
+	applet->edge_relativity = edge_relativity;
+	applet->locked          = locked != FALSE;
 
 	mate_panel_applets_to_load = g_slist_prepend (mate_panel_applets_to_load, applet);
 }
@@ -1005,8 +1010,8 @@ mate_panel_applet_compare (const MatePanelAppletToLoad *a,
 
 	if ((c = strcmp (a->toplevel_id, b->toplevel_id)))
 		return c;
-	else if (a->right_stick != b->right_stick)
-		return b->right_stick ? -1 : 1;
+	else if (a->edge_relativity != b->edge_relativity)
+		return a->edge_relativity - b->edge_relativity;
 	else
 		return a->position - b->position;
 }
@@ -1087,12 +1092,12 @@ mate_panel_applet_save_position (AppletInfo *applet_info,
 			    const char *id,
 			    gboolean    immediate)
 {
-	PanelWidget       *panel_widget;
-	const char        *toplevel_id;
-	char              *old_toplevel_id;
-	gboolean           right_stick;
-	gboolean           locked;
-	int                position;
+	PanelWidget               *panel_widget;
+	const char                *toplevel_id;
+	char                      *old_toplevel_id;
+	PanelObjectEdgeRelativity  edge_relativity;
+	gboolean                   locked;
+	int                        position;
 
 	g_return_if_fail (applet_info != NULL);
 
@@ -1124,7 +1129,7 @@ mate_panel_applet_save_position (AppletInfo *applet_info,
 	g_free (old_toplevel_id);
 
 	/* Note: changing some properties of the panel that may not be locked down
-	   (e.g. background) can change the state of the "panel_right_stick" and
+	   (e.g. background) can change the state of the "relative-to-edge" and
 	   "position" properties of an applet that may in fact be locked down.
 	   So check if these are writable before attempting to write them */
 
@@ -1135,18 +1140,32 @@ mate_panel_applet_save_position (AppletInfo *applet_info,
 	if (locked) {
 		/* Until position calculations are refactored to fix the issue of the panel applets
 		   getting reordered on resolution changes...
-		   .. don't save position/right-stick on locked applets */
+		   .. don't save position/relative-to-edge on locked applets */
 		return;
 	}
 
-	right_stick = panel_is_applet_right_stick (applet_info->widget) ? 1 : 0;
+	edge_relativity = panel_determine_applet_edge_relativity (applet_info->widget);
+	if (g_settings_is_writable (applet_info->settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY) &&
+	    g_settings_get_enum (applet_info->settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY) != edge_relativity)
+		g_settings_set_enum (applet_info->settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY, edge_relativity);
+
+	/*
+	 * If the deprecated `panel-right-stick` setting is set, unset it.
+	 * See the `panel_profile_load_object ()` function in the file
+	 * `panel-profile.c` for a lengthy comment detailing why this is
+	 * done.
+	 */
 	if (g_settings_is_writable (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY) &&
-	    (g_settings_get_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY) ? 1 : 0) != right_stick)
-		g_settings_set_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY, right_stick);
+	    g_settings_get_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY))
+		g_settings_set_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY, FALSE);
 
 	position = mate_panel_applet_get_position (applet_info);
-	if (right_stick && !panel_widget->packed)
-		position = panel_widget->size - position;
+	if (!panel_widget->packed) {
+		if (edge_relativity == PANEL_EDGE_CENTER)
+			position -= panel_widget->size/2;
+		else if (edge_relativity == PANEL_EDGE_END)
+			position = panel_widget->size - position;
+	}
 
 	if (g_settings_is_writable (applet_info->settings, PANEL_OBJECT_POSITION_KEY) &&
 	    g_settings_get_int (applet_info->settings, PANEL_OBJECT_POSITION_KEY) != position)
@@ -1363,7 +1382,7 @@ mate_panel_applet_can_freely_move (AppletInfo *applet)
 	if (!g_settings_is_writable (applet->settings, PANEL_OBJECT_TOPLEVEL_ID_KEY))
 		return FALSE;
 
-	if (!g_settings_is_writable (applet->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY))
+	if (!g_settings_is_writable (applet->settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY))
 		return FALSE;
 
 	return TRUE;
--- a/mate-panel/panel-profile.c
+++ b/mate-panel/panel-profile.c
@@ -1183,10 +1183,9 @@ panel_profile_destroy_toplevel (const char *id)
 }
 
 char *
-panel_profile_prepare_object_with_id (PanelObjectType  object_type,
-				      const char      *toplevel_id,
-				      int              position,
-				      gboolean         right_stick)
+panel_profile_prepare_object_with_id (PanelObjectType            object_type,
+				      const char                *toplevel_id,
+				      int                        position)
 {
 	PanelGSettingsKeyType  key_type;
 	char              *id;
@@ -1203,7 +1202,6 @@ panel_profile_prepare_object_with_id (PanelObjectType  object_type,
 	g_settings_set_enum (settings, PANEL_OBJECT_TYPE_KEY, object_type);
 	g_settings_set_string (settings, PANEL_OBJECT_TOPLEVEL_ID_KEY, toplevel_id);
 	g_settings_set_int (settings, PANEL_OBJECT_POSITION_KEY, position);
-	g_settings_set_boolean (settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY, right_stick);
 
 	/* Force writing the settings in order to reserve the object ID *now*,
 	 * so that a later call to panel_profile_find_new_id() won't find the same
@@ -1217,15 +1215,13 @@ panel_profile_prepare_object_with_id (PanelObjectType  object_type,
 }
 
 char *
-panel_profile_prepare_object (PanelObjectType  object_type,
-			      PanelToplevel   *toplevel,
-			      int              position,
-			      gboolean         right_stick)
+panel_profile_prepare_object (PanelObjectType            object_type,
+			      PanelToplevel             *toplevel,
+			      int                        position)
 {
 	return panel_profile_prepare_object_with_id (object_type,
 						     panel_profile_get_toplevel_id (toplevel),
-						     position,
-						     right_stick);
+						     position);
 }
 
 void
@@ -1243,13 +1239,13 @@ panel_profile_delete_object (AppletInfo *applet_info)
 static void
 panel_profile_load_object (char *id)
 {
-	PanelObjectType  object_type;
-	char            *object_path;
-	char            *toplevel_id;
-	int              position;
-	gboolean         right_stick;
-	gboolean         locked;
-	GSettings       *settings;
+	PanelObjectType            object_type;
+	char                      *object_path;
+	char                      *toplevel_id;
+	int                        position;
+	PanelObjectEdgeRelativity  edge_relativity;
+	gboolean                   locked;
+	GSettings                 *settings;
 
 	object_path = g_strdup_printf (PANEL_OBJECT_PATH "%s/", id);
 	settings = g_settings_new_with_path (PANEL_OBJECT_SCHEMA, object_path);
@@ -1257,14 +1253,53 @@ panel_profile_load_object (char *id)
 	object_type = g_settings_get_enum (settings, PANEL_OBJECT_TYPE_KEY);
 	position = g_settings_get_int (settings, PANEL_OBJECT_POSITION_KEY);
 	toplevel_id = g_settings_get_string (settings, PANEL_OBJECT_TOPLEVEL_ID_KEY);
-	right_stick = g_settings_get_boolean (settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY);
+	edge_relativity = g_settings_get_enum (settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY);
 	locked = g_settings_get_boolean (settings, PANEL_OBJECT_LOCKED_KEY);
+
+	/* If a panel layout uses the deprecated 'panel-right-stick' property,
+	 * interpret the position of the applet relative to the applet's left
+	 * side, rather than its right like we'd do for end-relative objects.
+	 * The position of an end-relative applet is relative to the *right*
+	 * side of the applet, which makes more sense, IMHO.  The difference
+	 * between these two positioning schemes is on the order of the width
+	 * of the applet; without this backwards-compatible safeguard, many of
+	 * the user's manually-placed right-stuck applets may be moved far out
+	 * of place when they upgrade the panel to this version.
+	 *
+	 * The `panel-right-stick` property will, unfortunately, remain set,
+	 * unless and until the user explicitly moves the applet under the new
+	 * version of the panel.  Keeping the `panel-right-stick` property set
+	 * is necessary since we can't know the equivalent 'edge-relative'
+	 * position to use in the place of the old, 'right-stuck' position;
+	 * after all, we have no idea of the size of the applet when it was
+	 * first positioned on the panel.  (Some applets dynamically change
+	 * size; sure, chances are right-stuck applets will be forced /
+	 * constrained to the very far right side of the panel, but who knows,
+	 * somebody might have set the positions of their applets using a
+	 * layout file or directly via GSettings.)
+	 *
+	 * If the user moves the applets, of course, the applets will be
+	 * repositioned edge-relatively, which is good -- unless the user ever
+	 * downgrades the panel, in which case the applets will be in the wrong
+	 * places.  However, there's a limit to what you can do -- hopefully
+	 * the user will only downgrade the panel if they know what they are
+	 * doing, and the user will have a completely screwed-up panel if they
+	 * *center*-stuck applets and then downgraded.  So the point of this
+	 * hack is to ensure the user experiences a seamless (or nearly
+	 * seamless) transition forward, but if the user moves backward a few
+	 * versions, then we really can't help them.
+	 */
+	if (g_settings_get_boolean (settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY)) {
+		edge_relativity = PANEL_EDGE_END;
+		g_settings_set_enum (settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY, edge_relativity);
+	}
+
 
 	mate_panel_applet_queue_applet_to_load (id,
 					   object_type,
 					   toplevel_id,
 					   position,
-					   right_stick,
+					   edge_relativity,
 					   locked);
 
 	g_free (toplevel_id);
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -2898,8 +2898,8 @@ panel_toplevel_move_resize_window (PanelToplevel *toplevel,
 {
 	GtkWidget *widget;
 
-	GList *list;
-	gboolean stick;
+/*	GList *list;
+	PanelObjectEdgeRelativity edge_relativity;*/
 
 	widget = GTK_WIDGET (toplevel);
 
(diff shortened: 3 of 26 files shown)


zhuyaliang:
I think this function is very good. It can set the position of panel applet from three relative positions: start, center and end.

mate-panel/panel-profile.c
```diff
@@ -1258,13 +1259,25 @@ panel_profile_load_object (char *id)
...
+	 * set the new, more flexible 'relative-to-edge' property to the
+	 * equivalent setting, then disable the 'panel-right-stick' property
+	 * so it doesn't interfere in the future.
+	 */
+	if (right_stick) {
+		edge_relativity = PANEL_EDGE_END;
+		g_settings_set_enum (settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY, edge_relativity);
+		g_settings_set_boolean (settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY, FALSE);
+	}
+
```
zhuyaliang, line 1275: ```right_stick``` can be completely removed

mate-panel/applet.c
```diff
@@ -1135,18 +1141,23 @@ mate_panel_applet_save_position (AppletInfo *applet_info,
...
+	    g_settings_get_enum (applet_info->settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY) != edge_relativity)
+		g_settings_set_enum (applet_info->settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY, edge_relativity);
 
 	position = mate_panel_applet_get_position (applet_info);
-	if (right_stick && !panel_widget->packed)
-		position = panel_widget->size - position;
+	size = mate_panel_applet_get_size (applet_info);
+	if (!panel_widget->packed) {
+		if (edge_relativity == PANEL_EDGE_CENTER)
+			position -= panel_widget->size/2 - size/2;
```
vkareh, line 1157: probably cast to `int` here too

mate-panel/panel-profile.c
vkareh: nah, I think it's good to leave it there for now. Otherwise an upgrade will cause panel objects to shift without this block in place.

mate-panel/panel-toplevel.c
```diff
@@ -2931,15 +2931,23 @@ panel_toplevel_move_resize_window (PanelToplevel *toplevel,
...
+				int size = mate_panel_applet_get_size (info);
+				int toplevel_size;
+
+				if (toplevel->priv->orientation & PANEL_HORIZONTAL_MASK)
+					toplevel_size = toplevel->priv->geometry.width;
+				else
+					toplevel_size = toplevel->priv->geometry.height;
+
+				if (edge_relativity == PANEL_EDGE_CENTER)
+					ad->pos = toplevel_size/2 - size/2 + position;
```
vkareh, line 2948: cast to `int`

mate-panel/panel-widget.c
```diff
@@ -623,36 +626,92 @@ get_applet_list_pos (PanelWidget *panel,
...
+PanelObjectEdgeRelativity
+panel_widget_determine_applet_edge_relativity (PanelWidget *panel_widget,
+					       GtkWidget   *widget)
 {
-	AppletData *applet;
+	AppletData *applet_in_question;
+	gboolean    applet_is_centered = FALSE;
+
+	g_return_val_if_fail (PANEL_IS_WIDGET (panel_widget), PANEL_EDGE_START);
+	g_return_val_if_fail (GTK_IS_WIDGET (widget), PANEL_EDGE_START);
```
vkareh, line 643: I don't understand this change. You're trying to determine whether the widget is actually a widget, but you're comparing against an enum. I suspect this change is a mistake.

mate-panel/panel-widget.c
```diff
@@ -623,36 +626,92 @@ get_applet_list_pos (PanelWidget *panel,
...
-	g_return_val_if_fail (GTK_IS_WIDGET (widget), FALSE);
+			if (applet->pos + applet->cells < end_pos)
+				break;
 
-	applet = g_object_get_data (G_OBJECT (widget), MATE_PANEL_APPLET_DATA);
-	if (applet) {
-		GList *applet_list, *l;
-		int    end_pos = -1;
+			if (applet->pos <= panel_widget->size/2 &&
+			    applet->pos + applet->cells > panel_widget->size/2)
```
vkareh, line 667: cast to `int`

mate-panel/panel-widget.c
```diff
@@ -623,36 +626,92 @@ get_applet_list_pos (PanelWidget *panel,
...
 
-			if (end_pos != -1 && applet->pos != end_pos)
+			if (applet->pos > start_pos)
 				break;
 
-			end_pos = applet->pos + applet->cells;
-			if (end_pos >= panel_widget->size)
-				return TRUE;
+			if (applet->pos <= panel_widget->size/2 &&
+			    applet->pos + applet->cells > panel_widget->size/2)
```
vkareh, line 696: cast to `int`

mate-panel/panel-widget.c
vkareh: I guess you're not assigning anything, so nevermind on this one

mate-panel/panel-profile.c
zhuyaliang: @vkareh I have to mean to completely remove the content about ```right t-stick```.Now that we have chosen the new arrangement of panel applets, we can completely abandon the previous arrangement. When the user chooses to upgrade the panel, the user will face a new arrangement.
All other settings about ```right-stick``` have been removed from this PR，If we accept this PR, we will also remove the ```right-stick``` code later
In my test, I directly deleted all the code of ```right-stick``` and found no problems.
Of course, if any problem is found in the follow-up test, I will agree with your suggestion

mate-panel/panel-toplevel.c
thesquash: Ummm...  Aren't we only working with integers here?

 - `toplevel_size`:  Declared `int` at `mate-panel/panel-toplevel.c` @ 2940.
 - `size`:  Declared `int` at `mate-panel/panel-toplevel.c` @ 2939.
 - `position`:  Declared `int` at `mate-panel/panel-toplevel.c` @ 2938.

I see literally no reason to include an explicit cast.  Am I missing something here?

mate-panel/panel-widget.c
thesquash: That's not what `g_return_val_if_fail ()` does.  It tests if its first parameter (say `PANEL_IS_WIDGET (panel_widget)`) is true (nonzero, in this case non-`NULL`), and if it is not true (i.e., it is `NULL`), then immediately terminate the current function, returning the value specified in the second argument to the function (actually `g_return_val_if_fail ()` is a pre-processor macro).

If you don't believe me, here's the official GLib documentation about the macro (in the code itself, at https://github.com/GNOME/glib/blob/main/glib/gmessages.h, lines 588-611 as I write this):

```
/**
 * g_return_val_if_fail:
 * @expr: the expression to check
 * @val: the value to return from the current function
 *       if the expression is not true
 *
 * Verifies that the expression @expr, usually representing a precondition,
 * evaluates to %TRUE. If the function does not return a value, use
 * g_return_if_fail() instead.
 *
 * If @expr evaluates to %FALSE, the current function should be considered to
 * have undefined behaviour (a programmer error). The only correct solution
 * to such an error is to change the module that is calling the current
 * function, so that it avoids this incorrect call.
 *
 * To make this undefined behaviour visible, if @expr evaluates to %FALSE,
 * the result is usually that a critical message is logged and @val is
 * returned from the current function.
 *
 * If `G_DISABLE_CHECKS` is defined then the check is not performed.  You
 * should therefore not depend on any side effects of @expr.
 *
 * See g_return_if_fail() for guidance on how to debug failure of this check.
 */
```

(By the way, the version of `g_return_val_if_fail ()` defined right below that comment is for if GLib was compiled with sanity checks disabled -- that version is a dummy macro that does nothing anyway; furhter below, at line 649, the real version used in most builds of GLib is defined:

```
#define g_return_val_if_fail(expr, val) \
  G_STMT_START { \
    if (G_LIKELY (expr)) \
      { } \
    else \
      { \
        g_return_if_fail_warning (G_LOG_DOMAIN, \
                                  G_STRFUNC, \
                                  #expr); \
        return (val); \
      } \
  } G_STMT_END
```

Not to be arrogant, but let's break that down:  The macro tests for the likely case that the first parameter [`expr`] evaluates to true or a nonzero / non-null value:  In that case, nothing happens [hence the empty `{}` brackets].  However, if that test fails and `expr` evaluates to false, zero or NULL, then the `else` block is run, a warning is printed on the terminal [or into `.xsession-errors`, more likely], and the second parameter [`val`] is returned.)

So, maybe you have a point, but I don't see what's wrong.  I'd appreciate to hear from you on this matter.  Thank you.

mate-panel/panel-profile.c
thesquash: I have to agree with @zhuyaliang on this one.  That function in which the right-stick testing code appears is only run when an applet's state needs to be *saved*.  An applet's state needs to be saved only when it changes (e.g., the user moved the applet on the panel).  By the time the user moves the applet, the position has been recalculated and so has the edge relativity, so in truth this code shouldn't even work at all.

I'll remove the right-stick property entirely and this code, but if someone insists I'll add the code back and try to make it actually do the job (better than before, that is).

thesquash:
Also, has anybody actually tried applying this patch / PR and using the patched version of the panel yet?  Because if not, I wish somebody would.  Discussing code is one thing; testing it is another.

Thanks in advance, as always.  I appreciate the help.

lukefromdc (commented):
I got segfaults trying to run this-and got them with both in process and out of process builds or wncklet.

With an in-process build I got this backtracee:
```
0x00007ffff71ccf75 in g_type_check_instance_is_fundamentally_a ()
   from /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
(gdb) bt
#0  0x00007ffff71ccf75 in g_type_check_instance_is_fundamentally_a ()
    at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#1  0x000055555558a9e9 in mate_panel_applet_get_size ()
#2  0x00007ffff34439d1 in show_desktop_applet_fill ()
    at /usr/lib/x86_64-linux-gnu/mate-panel/libwnck-applet.so
#3  0x00007ffff343c8a5 in wncklet_factory ()
    at /usr/lib/x86_64-linux-gnu/mate-panel/libwnck-applet.so
#4  0x00007ffff05cb372 in mate_panel_applet_marshal_BOOLEAN__STRING ()
    at /usr/lib/x86_64-linux-gnu/libmate-panel-applet-4.so.1
#5  0x00007ffff71ab4ef in g_closure_invoke ()
    at /usr/lib/x86_64-linux-gnu/libgobject-2.0.so.0
#6  0x00007ffff05cf1da in mate_panel_applet_setup ()
    at /usr/lib/x86_64-linux-gnu/libmate-panel-applet-4.so.1
#7  0x00007ffff05d096e in mate_panel_applet_get_applet_widget ()
    at /usr/lib/x86_64-linux-gnu/libmate-panel-applet-4.so.1
#8  0x00005555555da134 in mate_panel_applets_manager_dbus_get_applet_widget ()
#9  0x000055555559fe82 in mate_panel_applets_manager_get_applet_widget ()
#10 0x00005555555db685 in panel_applet_container_setup ()
#11 0x00005555555dc0c6 in on_proxy_appeared ()
#12 0x00007ffff72a7de9 in  () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#13 0x00007ffff72a892b in  () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#14 0x00007ffff730d8fc in  () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#15 0x00007ffff72a7de9 in  () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
--Type <RET> for more, q to quit, c to continue without paging--c
#16 0x00007ffff72a7e29 in  () at /usr/lib/x86_64-linux-gnu/libgio-2.0.so.0
#17 0x00007ffff70b5874 in g_main_context_dispatch () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#18 0x00007ffff70b5bd8 in  () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#19 0x00007ffff70b5eb3 in g_main_loop_run () at /usr/lib/x86_64-linux-gnu/libglib-2.0.so.0
#20 0x00007ffff783f5f5 in gtk_main () at /usr/lib/x86_64-linux-gnu/libgtk-3.so.0
#21 0x00005555555791ea in main ()
(gdb
```)

**Comment by lukefromdc** [comment]
Note that I am running a very customized panel configuration, breaking this sort of thing is exactly what we have to worry about here

**Comment by thesquash** [comment]
No, I'm simply an idiot who uses link-time optimization, so I was able to get away with defining a (semi-public) function with the same name as another (public) function.  It took me all day to figure this out, though, so I think I got punishment enough for the mistake.

The problem lay within `mate-panel/applet.c`.  On line 1365, I defined a function called `mate_panel_applet_get_size ()`, which accepted an `AppletInfo *` and returned the number of "cells" an applet used on the panel.  I chose the name because similar "helper" functions in that file were named similarly, and like the other helper functions (e.g. `mate_panel_applet_get_position ()`), this function could not be made private / `static` because it was used in other source files, including `mate-panel/panel-toplevel.c`.

Unfortunately, the name `mate_panel_applet_get_size ()` was already used by a function in `libmate-panel-applet`.  That function does pretty much the same thing, except that it takes a `MatePanelApplet *` parameter as input instead of an `AppletInfo *`.  When the dynamic loader resolves all the symbols used by all the libraries, the panel executable itself, and the applets, it confused the two functions, and one got loaded under all circumstances, even when the other should have been called.

I did not notice this at first because I often compile applications with link-time optimization for code size reasons.  The link-time optimizer in GCC is smart enough to realize that the two functions are not the same, and my version of the function is "privatized" (made non-global and thus invisible to the dynamic loader).  Thus when compiled with LTO, my bug does not show up.  But for everybody else with some sanity left, they see a crashing panel as the wrong function with the wrong input is called.

**TL;DR**:  It's fixed now, and I've put a sticky note on my monitor:  "Do NOT use LTO when testing new pull requests!"  :-)

**Review by lukefromdc — COMMENTED** [review]
The last round of changes stopped the segfaults. One minor issue remains: the last two applets on my panel (trash and shutdown) did not come up where intended, in the two farthest right positions. Note that I dragged them there by hand and locked them to the panel, not sure the code that invokes. The trash applet disapeared and the shutdown applet appeared where (and possibly thus OVER) the "show desktop" applet which is second from left in my layout.

Nothing in my config was permanently corrupted, rolling back to master put the applets back in their expected positions,

I will leave the judgement to others as to whether or not it is OK for users to have to reposition (or if lost, re-add) some applets after a panel update. Note that this behavior (even if once -only on panel update) resembes a very old GNOME 2/MATE bug seen by some users where applets can appear and disappear. 

If this can be fixed too than I will post my formal approval of this PR.

**Comment by thesquash** [comment]
To prevent the unexpected panel applet movement that @lukefromdc profiled above, I added the (deprecated) `panel-right-stick` property back into the code, but implemented the backward-compatibility *correctly* this time.

I also updated the "Ubuntu" panel layout very slightly:  One panel applet (the shutdown icon, I think) was supposed to be positioned at the far right of the top panel.  It was positioned correctly when the panel positioned the *left* side of applets relative to the *right* side of the panel (`panel-right-stick`), but is positioned with a gap between the right of the applet and the right of the panel when my center-stick patch is applied.  (My center-stick patch also changes the semantics of right-stuck applets slightly; now the *right* side of the applet is positioned relative to the right side of the panel, which makes more sense than the former behavior, at least IMHO.)

I hope this latest change does the trick!

**Comment by thesquash** [comment]
I'm puzzled.  The Debian test build passed, but the Fedora one didn't.  Still, it appears both containers successfully *compiled* my changes.

Is this another instance of the build testing system being broken?

**Comment by zhuyaliang** [comment]
> 我很困惑。Debian测试构建通过了，但是Fedora没有通过，但是看起来两个容器都成功了_编译_我的改变
> 
> 这是构建测试系统被破坏的另一个实例吗？

Then compile it again. I started the compilation check for you

**Comment by lukefromdc** [comment]
Just retested, still got moving applets some of the time but not all of the time. Space issues may have triggered it in my second test, in which the clock (3ed from right) disappeared. Also got some moving applets on rollback this time around too, but this would be extremelty uncommon for normal users as it would require a  major version rollback.  Again, not a blocker to me so long as users are aware of this, remember we are on 1.27 so just not cherrypicking would mean users only see this if they update a distro to a new version,  compile MATE  locally and move up a major version, or get a major version update on a rolling release

**Comment by zhuyaliang** [comment]
@thesquash Why did you add the ```right-stick``` code

**Comment by thesquash** [comment]
> still got moving applets some of the time but not all of the time. Space issues may have triggered it in my second test, in which the clock (3ed from right) disappeared.

It's unacceptable for this PR to move applets unexpectedly, so this can't be ignored.  However, any chance of showing me a screenshot of your panel layout under normal circumstances, and one of the panel after loading my patch, and one when you try to move stuff around but applets disappear?  Because I'm working totally blindly right now -- if I could replicate your panel layout (or at least come *somewhere* close to it) then I could probably fix all these bugs.  As it is, though, I can't reproduce what you're talking about.

> Why did you add the `right-stick` code

I thought I already explained this above, but let's say I didn't:  For backward compatibility (or actually, at least forward compatibility).  In the first revision of the PR, I included code to make applets end-relative if the right-stick property was set on the applet, but the code did not quite work as intended.  Then I removed the old right-stick code entirely, and @lukefromdc said his right-stuck panel applets jumped over to the left unexpectedly.  So now I've added the right-stick compatibility code back in, albeit in a more functional form, so that people's applets won't jump to the other side of the panel; at most they'll now move a few pixels, since the right side of an end-relative applet is positioned relative to the end of the panel, whereas right-sticking positioned the *left* side of an applet relative to the right side of the panel!

So, in other words, maybe eventually we can remove the right-stick property, but IMHO it should be kept around for a stable release or two at least.  Do users really read release notes anyway?  No.  Distributors read release notes.  And distributors almost *never* have been known to, say, display a dialog box upon system upgrade that says "your panel applets will be rearranged upon the next login, be prepared for that to happen!"  So really, it's our job to implement some level of backward compatibility so we don't get a flood of irate users come along and create issue after issue describing this "bug".

I said the preceding text as both a programmer and a user myself, and I think most users would agree.

**Comment by lukefromdc** [comment]
I am on the roar now, might be 36 hours or so.before I can send a panel screenshot but will do so. It's a single panel on the bottom, last tjree applets left to right are clock, trash, and logout, usually with no extra space between them.


Overall on that box is menubutton, show desktop, a bunch of app launchers, window list on the left. aftet space set aside for thr windoe list ai have netspeed, inhibit, force-quit (order of those two sometimes reversed when moved by accident), workspace switcher, hardware monitor, cpufreq, tray, volume control applet,clock, trash, and logout.

On 5/13/2022 at 3:20 PM, "Gordon N. Squash" ***@***.***> wrote:
>
>> still got moving applets some of the time but not all of the 
>time. Space issues may have triggered it in my second test, in 
>which the clock (3ed from right) disappeared.
>
>It's unacceptable for this PR to move applets unexpectedly, so 
>this can't be ignored.  However, any chance of showing me a 
>screenshot of your panel layout under normal circumstances, and 
>one of the panel after loading my patch, and one when you try to 
>move stuff around but applets disappear?  Because I'm working 
>totally blindly right now -- if I could replicate your panel 
>layout (or at least come *somewhere* close to it) then I could 
>probably fix all these bugs.  As it is, though, I can't reproduce 
>what you're talking about.
>
>> Why did you add the `right-stick` code
>
>I thought I already explained this above, but let's say I didn't:  
>For backward compatibility (or actually, at least forward 
>compatibility).  In the first revision of the PR, I included code 
>to make applets end-relative if the right-stick property was set 
>on the applet, but the code did not quite work as intended.  Then 
>I removed the old right-stick code entirely, and @lukefromdc said 
>his right-stuck panel applets jumped over to the left 
>unexpectedly.  So now I've added the right-stick compatibility 
>code back in, albeit in a more functional form, so that people's 
>applets won't jump to the other side of the panel; at most they'll 
>now move a few pixels, since the right side of an end-relative 
>applet is positioned relative to the end of the panel, whereas 
>right-sticking positioned the *left* side of an applet relative to 
>the right side of the panel!
>
>So, in other words, maybe eventually we can remove the right-stick 
>property, but IMHO it should be kept around for a stable release 
>or two at least.  Do users really read release notes anyway?  No.  
>Distributors read release notes.  And distributors almost *never* 
>have been known to, say, display a dialog box upon system upgrade 
>that says "your panel applets will be rearranged upon the next 
>login, be prepared for that to happen!"  So really, it's our job 
>to implement some level of backward compatibility so we don't get 
>a flood of irate users come along and create issue after issue 
>describing this "bug".
>
>I said the preceding text as both a programmer and a user myself, 
>and I think most users would agree.
>
>-- 
>Reply to this email directly or view it on GitHub:
>https://github.com/mate-desktop/mate-panel/pull/1310#issuecomment-
>1126370627
>You are receiving this because you were mentioned.
>
>Message ID: <mate-desktop/mate-
***@***.***>

**Comment by lukefromdc** [comment]
Here is a scaled-down screenshot of my panel:
Panel

**Comment by thesquash** [comment]
I couldn't replicate exactly the behavior you encountered with applets disappearing, but I did notice the moving applets, and I fixed that and some other bugs that I should have noticed earlier.  The only "disappearing" applet I noticed was the notification area applet, which randomly appeared as a 1 pixel wide line on the panel if no system tray icons were loaded; focusing the panel with Ctrl-Alt-Tab and then Tab-ing around to the notification area resized the notification area and shows the handle.  Everything else works as intended, as far as I can tell (which may not be worth much!).  I hope I fixed the wonkiness this time.

**Review by lukefromdc — COMMENTED** [review]
Still getting applets moving (both in-process and out of process) on changeover to a build of the panel with this applied. No corruption of the config on rolling back to master at least, but the problem is still here on a crowded single bottom panel

My normal panel on the laptop:
Panel_Normal

Panel with this applied, same machine:
Panel_after_PR

**Comment by thesquash** [comment]
OK, I know this sounds a bit unreasonable, but I'd like your entire panel applet position database:

```
dconf dump /org/mate/panel/objects/
```

The reason I want this is because I'm not convinced that *any* of your applets are actually right-stuck anyway -- the right-sticking behavior was actually broken before I made my patch.  You might therefore just be witnessing the results of applets getting rearranged because the way applets are positioned post-application of this patch is different from how they are positioned normally.  (I'll say it again:  The left side of applets, even right-stuck applets, were positioned relative to one end of the panel [depending on whether right-sticking was enabled]; my patch positions the right side of right-stuck applets relative to the right of the panel, the center of centered applets relative to the center of the panel, and the left side of everything else relative to the left side of the panel.)

When you provide the dump, I'll check it over.  If we determine that the problem's just caused by the new positioning policy I described above, sorry but there's nothing I think I can do about it.  I really think my method is more logical, and doing otherwise would be an injustice to users.

Unless someone else can think of a way around the problem...

**Comment by lukefromdc** [comment]
That will be seveeal days away, as I am about to depart on a 1,000 mile road trip

**Comment by lukefromdc** [comment]
I manually position applets and have never intentiomally tried to invoke a "right stick" though I do lock applets to the panel. So far, every change to this has only had a minor effect on this behavior.

Had this w both machines, but I normally copy hidden/config files along with OS installs from machine to machine

I will not block this or any other PR because I am.often on the road (due to not having reliable housing) and thus cannot guarantee availability for retests at.any given time. Not like things were for me back in 2016.

**Comment by thesquash** [comment]
All right -- apply this patch, and nobody should have unexpectedly-moving panel applets anymore.  And if you don't move any applets, downgrading should work too.  But once you position some applets anew with this patch applied, do not downgrade, because the new applets will wind up in the wrong places.  Still, I think, short of finding any real *bugs* now, that this patch has as good behavior as any.

Please review this round of patches.

**Review by lukefromdc — APPROVED** [review]
This now works, no applets moved on upgrading to this in my last round of tests. Even downgrading to current master only moved applets if (as you described) I moved applets manually while running the new code. Sometimes I even got away with moving applets and then downgrading, but not always. Since no distro will be pushing downgrades barring a severe bug somewhere (presumably) else, this now sems good to go.

**Comment by thesquash** [comment]
Wow!  I might have actually done something right -- good to hear!

**Comment by lukefromdc** [comment]
OK, we now need to know if these changes satisfy  @vkareh and we need another review

File: mate-panel/panel-profile.c [hunk 7]
```diff
@@ -1243,28 +1239,69 @@ panel_profile_delete_object (AppletInfo *applet_info)
...
-	char            *toplevel_id;
-	int              position;
-	gboolean         right_stick;
-	gboolean         locked;
-	GSettings       *settings;
+	PanelObjectType            object_type;
+	char                      *object_path;
+	char                      *toplevel_id;
+	int                        position;
+	gboolean                   right_stuck;
```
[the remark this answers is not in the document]
rbuj:
```diff
diff --git a/mate-panel/panel-profile.c b/mate-panel/panel-profile.c
index 0de37eee..05da882e 100644
--- a/mate-panel/panel-profile.c
+++ b/mate-panel/panel-profile.c
@@ -1243,7 +1243,6 @@ panel_profile_load_object (char *id)
        char                      *object_path;
        char                      *toplevel_id;
        int                        position;
-       gboolean                   right_stuck;
        PanelObjectEdgeRelativity  edge_relativity;
        gboolean                   locked;
        GSettings                 *settings;
@@ -1254,7 +1253,6 @@ panel_profile_load_object (char *id)
        object_type = g_settings_get_enum (settings, PANEL_OBJECT_TYPE_KEY);
        position = g_settings_get_int (settings, PANEL_OBJECT_POSITION_KEY);
        toplevel_id = g_settings_get_string (settings, PANEL_OBJECT_TOPLEVEL_ID_KEY);
-       right_stuck = g_settings_get_boolean (settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY);
        edge_relativity = g_settings_get_enum (settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY);
        locked = g_settings_get_boolean (settings, PANEL_OBJECT_LOCKED_KEY);
 
@@ -1291,7 +1289,7 @@ panel_profile_load_object (char *id)
         * seamless) transition forward, but if the user moves backward a few
         * versions, then we really can't help them.
         */
-       if (right_stuck) {
+       if (g_settings_get_boolean (settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY)) {
                edge_relativity = PANEL_EDGE_END;
                g_settings_set_enum (settings, PANEL_OBJECT_RELATIVE_TO_EDGE_KEY, edge_relativity);
        }
```

mate-panel/panel-toplevel.c
```diff
@@ -2920,7 +2920,7 @@ panel_toplevel_move_resize_window (PanelToplevel *toplevel,
 				   toplevel->priv->geometry.width,
 				   toplevel->priv->geometry.height);
 
-	if (resize || move) {
+/*	if (resize || move) {
```
rbuj, line 2923: Does this commented code block mean it's still in progress?

mate-panel/panel-widget.c
```diff
@@ -2093,16 +2253,49 @@ panel_widget_applet_move_to_cursor (PanelWidget *panel)
...
+		      NULL);
+
+	if ((moveby < 0 && pos > panel->size/2 - ad->cells/2 &&
+	     pos+moveby <= panel->size/2 - ad->cells/2 &&
+	     pos+moveby >= panel->size/2 - ad->cells/2 - drag_threshold*SNAP_TOLERANCE_FACTOR) ||
+	    (moveby > 0 && pos < panel->size/2 - ad->cells/2 &&
+	     pos+moveby >= panel->size/2 - ad->cells/2 &&
+	     pos+moveby <= panel->size/2 - ad->cells/2 + drag_threshold*SNAP_TOLERANCE_FACTOR))
+		moveby = (panel->size/2 - ad->cells/2) - pos;
+	else if (pos == panel->size/2 - ad->cells/2)
```
rbuj, line 2280: you can use a variable to store the value of `panel->size/2 - ad->cells/2`

mate-panel/panel.c
```diff
@@ -1415,7 +1415,7 @@ panel_is_applet_right_stick (GtkWidget *applet)
 	if (!panel_toplevel_get_expand (panel_widget->toplevel))
 		return FALSE;
```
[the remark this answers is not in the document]
rbuj: Same for previous lines, you can't return a boolean value from now:
```
	g_return_val_if_fail (GTK_IS_WIDGET (applet), FALSE);
```
```
	g_return_val_if_fail (PANEL_IS_WIDGET (parent), FALSE);
```

mate-panel/applet.c
```diff
@@ -1135,18 +1140,32 @@ mate_panel_applet_save_position (AppletInfo *applet_info,
...
+	/*
+	 * If the deprecated `panel-right-stick` setting is set, unset it.
+	 * See the `panel_profile_load_object ()` function in the file
+	 * `panel-profile.c` for a lengthy comment detailing why this is
+	 * done.
+	 */
 	if (g_settings_is_writable (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY) &&
-	    (g_settings_get_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY) ? 1 : 0) != right_stick)
-		g_settings_set_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY, right_stick);
+	    g_settings_get_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY) != FALSE)
```
rbuj, line 1159: please, use
```
g_settings_get_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY)
```
instead of
```
g_settings_get_boolean (applet_info->settings, PANEL_OBJECT_PANEL_RIGHT_STICK_KEY) != FALSE)
```
for the if-clause

thesquash:
Gosh -- I know I actually found (and fixed) those things at one point.  I guess I never committed my changes though.  Great catches, though -- I'm highly appreciative for your consistently good eye for programming mistakes, even ones that don't cause unexpected behavior.  Thumbs up!

lukefromdc:
Just retested, no icons moved. Looks good to go on this end. Do we have other problems still outstanding or is this good to go?

rbuj (approved):
LGTM

vkareh (approved):
LGTM

vkareh:
Can we merge? Something with docker build and fedora is bugging travis... but seems to be the case in all PRs

PR #1311, opened 2022-05-31
applet: SMALL_ICON_SIZE macro is not used [-Wunused-macros]

rbuj:


lukefromdc (approved):
Works fine, build finishes and runs same as before. SMALL_ICON_SIZE does not appear anywhere else in the file. This looks good to go on this end

lukefromdc:
Not sure what the "line buffering" issue in the Fedora build in Travis is, note that the Debian build finished

lukefromdc:
Merging as this worked in my test, and `SMALL_ICON_SIZE` does not appear anywhere else in the source of mate-panel after being defined in applet.c and never used or mentioned again there

PR #1312, opened 2022-06-01
build warning: AC_OUTPUT should be used without arguments

rbuj:


cwendling (approved):
LGTM, but it'd be good to have all tests passing (although it doesn't look related)

PR #1317, opened 2022-06-28
Fix two leaks

cwendling:
Fix a leak in the clock applet (`gtk_style_context_get()` return values are owned by the caller), and in the no-background code path for all applets ([`gdk_window_set_background_pattern()`](https://docs.gtk.org/gdk3/method.Window.set_background_pattern.html) gets its own reference to the given pattern, so we need to release our own).

IMO the commits make more sense split as they are here, so I'd advise a rebase (rather than a squash).

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -368,7 +368,7 @@ calculate_minimum_height (GtkWidget        *widget,
 {
         GtkStateFlags    state;
         GtkStyleContext *style_context;
-        const PangoFontDescription *font_desc;
+        PangoFontDescription *font_desc;
         GtkBorder        padding;
         PangoContext     *pango_context;
         PangoFontMetrics *metrics;
@@ -389,6 +389,7 @@ calculate_minimum_height (GtkWidget        *widget,
         descent = pango_font_metrics_get_descent (metrics);
 
         pango_font_metrics_unref (metrics);
+        pango_font_description_free (font_desc);
 
         gtk_style_context_get_padding (style_context, state, &padding);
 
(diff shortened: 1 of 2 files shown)


zhuyaliang (commented):
Will there be any impact
[line 1550](https://github.com/mate-desktop/mate-panel/blob/master/libmate-panel-applet/mate-panel-applet.c#L1550)

cwendling:
> Will there be any impact [line 1550](https://github.com/mate-desktop/mate-panel/blob/master/libmate-panel-applet/mate-panel-applet.c#L1550)

No, it's unrelated. The pattern here was created another way anyway -- and in any case the change I make does not destroy the pattern per se, it just hands it over to GDK without leaking a reference to it (so it prevents a leak when GDK releases its own reference, e.g. when replacing the pattern again).

lukefromdc (approved):
I noticed the mate-panel-applet.c change only affects out of process builds, so I tested this both ways, had no problems with the clock in process or out of process-and didn't notice problems in a quick test with any other applet. Note that (so far) there will almost always be one or more other out of proces applets in use unless running under Wayland

The second commit fixes an old mistake of mine (leaving out `cairo_pattern_destroy` and thus leaking the pattern) and renames the variable` pattern` to `transparent `, which makes more sense here. The variable name change why its a 3 liner not a one liner.

Other commit is simple too. Note that since the font can be changed at any time by the user while the panel is running, the approach prior to the PR of getting the minimum height from the font description does not strike me as a good place to declare the font description as a constant: it may change at any time.  will merge

PR #1320, opened 2022-07-25
configure.ac: The macro 'AC_HELP_STRING' is obsolete

rbuj:


lukefromdc (approved):
configuration and ./configure --help work fine with this, and the new code has been around well over a decade so will merge this

PR #1323, opened 2022-08-15
Fix center- and right-sticking of expanding applets

thesquash:
Fix center- and right-sticking of expanding applets

In a prior commit (38e00280e9d17282717595a05411736a308804c0), I added support to stick applets to the center of the panel, so that no matter the width of the applet or the panel, the applet would always stay at the center of the panel.  That commit also added an "end-relative" mode, which is similar to the panel's old "right-stick" feature, except end-relative mode positions the applet's right edge relative to the panel's right edge, whereas right-sticking positioned the applet's left edge relative to the right edge of the panel.  The advantage of end-relative positioning is that the applet can resize without moving itself even further to the right than the user intended, whereas if positioned so that there is empty space between the applet and the edge of the panel, right-stuck applets would change the size of that gap (or close it entirely) when the applet resized.

Unfortunately, I have since discovered that my original implementation of the center-stick and end-relative positioning of applets was inadequate. Some applets can accept any amount of space between a minimum and their "preferred" size -- if such an applet cannot be allocated its preferred size (perhaps because panel space is scarce), then the applet will be allocated as much space as is available, and the applet will "flex" itself to compensate.  Some examples of these "expanding" applets include the Window List (a component of the WNCKlet included in this package) and the MATE Dock Applet (a third-party panel applet from the Ubuntu MATE project).  When writing my original center-stick code, I did not properly take expanding applets into account, and with my original commit, such applets were off-center -- usually way off-center since the code seemed to position such applets based on their minimum size, rather than their actual allocated size.

This commit adds center-sticking support for expanding applets.  It also works with end-relative expanding applets -- such applets will expand towards the left, thus keeping their right-relative position intact.

Fixes: #1324.

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -649,8 +649,8 @@ panel_widget_determine_applet_edge_relativity (PanelWidget *panel_widget,
 		int         start_pos, end_pos;
 
 		applet_list = g_list_find (panel_widget->applet_list, applet_in_question);
-		start_pos = applet_in_question->pos + applet_in_question->cells;
-		end_pos = applet_in_question->pos;
+		start_pos = applet_in_question->constrained;
+		end_pos = applet_in_question->constrained;
 
 		/*
 		 * Start by moving towards the start of the panel, checking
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Builds fine, didn't notice any new warnings. Runs without issue and fixes the issue wity expanding center-stuck applets

lukefromdc:
Ideally we get another tester before merging this, if not we can merge this

thesquash:
Yeah, I'm willing to wait as long as necessary for this to be well-tested.  I'm just hoping that nobody decides to produce a new release of the MATE Panel in the meantime.  (I find that unlikely, though.)

lukefromdc:
If nobody test this I will merge it on the grounds that I should not block you from cleaning up your own mess. 

Since the existing code makes the panel so easy to screw up this is "merge the fix or revert the original comnmit to merge both together later"  territory

Another release COULD be put out by someone at any time, though distros only use the even major number releases so far as I know and 1.28 seems a ways off

Anyone else up for testing this?

PR #1335, opened 2022-11-10
panel-menu-button: comment out wayland debugging message

lukefromdc:
leave the code in place so it can be uncommented and used for further work on wayland


lukefromdc:
The panel position messages from the menu button show up in normal user's ~/xsession-errors files without this. It was included in initial Wayland support work but should not be active in production builds. Messages easily re-enabled locally if needed again by any dev

lukefromdc:
Passes checks, nobody has reviewed it in nearly two months. This just comments out debugging messages that may have been leftovers from wayland development and at any rate are spam in normal user's logs, so will proceed to merge this and take responsibility for it. Very simple change and if we ever have to revert it (or devs working on wayland suppport need to revert it locally) should be no problem.

PR #1337, opened 2022-12-02
windows-list.ui: lower gtk3 required version

raveit65:
Using gtk+-3.24 prevent starting the windows-list-preferences window in rhel8. See https://bugzilla.redhat.com/show_bug.cgi?id=2149747
Using gtk+-3.22 fixes the issue.
New version is checked with glade

### DIFF
--- a/applets/wncklet/window-list.ui
+++ b/applets/wncklet/window-list.ui
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <!-- Generated with glade 3.38.2 -->
 <interface>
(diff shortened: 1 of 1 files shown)


mbkma (approved):
LGTM. No new errors show up in Glade; in fact the lowest possible version would be 3.16 before deprecation warnings are shown.

raveit65:
Should i use 3.16 ? 
But i think rhel8 is distro with lowest gtk+ which use MATE 1.26.

mbkma:
I think 3.22 is fine :)

raveit65:
Ok, i will cherry-pick to 1.26 branch

PR #1348, opened 2023-01-07
wncklet: fix incorrect preview positions for multiple monitors

balazs-endresz:
I've just noticed that the preview positions were incorrect when the secondary display was above or to the left of the primary one. 

This PR will ensure that we have some space left for the padding only on one side. It takes some effort to come up with a config where the preview is even slightly off screen (lot of open windows, no other widgets on the panel), so even this is rarely necessary. Otherwise this would be a lot more difficult to calculate properly for multiple displays near every edge.

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -333,7 +333,6 @@ static int find_offset(GList *list, int target)
 }
 
 #define PREVIEW_PADDING 5
-#define CONSTRAIN(x, min, max) x < min ? min : (x > max ? max : x)
 static void
 preview_window_reposition (WnckTasklist    *tl,
                            TasklistData    *tasklist,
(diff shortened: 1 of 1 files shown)


lukefromdc (commented):
I cannot really test this (though it builds and runs fine here w no visible behavior changes) because I am using a 4K monitor, when I set window-scaling=1 I have tons of leftover space on the panel, we have excluded centering previews from the scaled case due to bugs, and my second monitor is packed away somewhere I cannot easily get to it right now.

That said, I didn't get any visible (meaning visible to me) changes in behavior with this, or any new build warnings. No idea from this test if this can push a preview partially off the edge of a screen.

balazs-endresz:
Here's a couple of screenshots where you can see that it can be cut off only on one side now with this PR. I added code to prevent this from happening in the recent commits, but that wasn't really necessary for the centering.

But the thing this PR fixes is when the preview was completely misplaced with multiple monitors. Now I tested that with various different arrangements in the display settings. 

left-still-has-padding
right-cut-off-slightly

zhuyaliang:
How can I test the problem of recurrence

balazs-endresz:
Put the secondary display above or to the left of the primary one in the display settings. With this PR the preview positions should be correct. You can also try placing it in all four corners to be sure.

Also, note that this commit is included in https://github.com/mate-desktop/mate-panel/pull/1349

zhuyaliang:
@balazs-endresz I can't test this PR right now, let others see it

raveit65:
https://github.com/mate-desktop/mate-panel/pull/1349 is merged

PR #1349, opened 2023-01-11
wncklet: center previews when scaling is enabled

balazs-endresz:
Turns out the off-by-one issue when scaling is enabled can be easily fixed by getting the pointer position as a double: https://docs.gtk.org/gdk3/method.Device.get_position_double.html 

This PR also includes https://github.com/mate-desktop/mate-panel/pull/1348 to make testing easier but let me know if you'd prefer otherwise.

I've tested this with scaling, grouping, and multiple monitors in various positions.

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -320,7 +320,7 @@ static int g_int_compare(gconstpointer a, gconstpointer b)
 	return 1;
 }
 
-static int find_offset(GList *list, int target)
+static int find_offset(GList *list, gdouble target)
 {
 	GList *node = list;
 	while (node != NULL) {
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Tested this file with both changes applied and a 4K monitor: No problems now with off-position window previews even if I enable grouping and no matter where I put the mouse on the window list buttons.

lukefromdc:
Hoping we can get a second reviewer to have a look at this. It's much nicer looking over here, but I am one user testing on one distro, with one monitor.

zhuyaliang:
I am interested in this. I will test it later

zhuyaliang (approved):

raveit65 (approved):
I can confirm that scaled previews are always centered with 2 x 4k monitors.

Issue #962, opened 2019-05-08
No option to disable middle-click to close in window list


haltman-at:
#### Summary

I can't find any way to disable the middle-click-to-close behavior of the window list.  I'm using a touchpad that frequently generates spurious middle clicks, so it is is *seriously* important to me that I disable this; this feature could *easily* cause me to lose data when I don't have a mouse handy (or am in a situation where a mouse would be seriously uncomfortable to use).

#### Expected behaviour

When I open the window list preferences, there should be an option to turn this off.

#### Actual behaviour

There is no such option.  Neither does there seem to be an option for it in `gsettings`.

#### Steps to reproduce the behaviour

To see that there is no such option in the window list preferences: Open the window list preferences.

To see that there is no such option in `gsettings`: run `gsettings list-keys org.mate.panel.applet.window-list:/`.

#### MATE general version

1.20.1

#### Linux Distribution

Linux Mint 19.1

#### Remark

It is more generally true that if you're introducing an easy new way to do something potentially destructive, you should almost certainly include an option to disable it.  This particular example is quite dangerous to anyone using an ASUS F-series, whose touchpads, as mentioned, tend to generate spurious middle-clicks.

nsoggia:
...And when you are used to cycle through grouped windows using the middle mouse button you may also close a non-grouped window by mistake.

I just closed a ssh session by grabbing my mouse in the wrong way, so I wrote a patch that I hope will be included in the next release.

Initially I added the new option at the bottom of the preferences window, but then I decided that it had a loose relationship to window grouping options and I also liked the idea of leaving the mouse scrolling option at the bottom, where people are used to find it.

**The job in not over yet.**
I need help from all the translators because I had to add a couple of labels and I was able to take care only of the Italian .po.

[middle_button.patch](https://github.com/mate-desktop/mate-panel/files/10661852/middle_button.patch.txt) 🎉

lukefromdc:
all of the .po files should be entirely left alone, the translators handle that themselves

nsoggia:
Ok. do you prefer a pull request?
(I  will `head -n -18 middle_button.patch.txt` first)

lukefromdc:
A PR is much easier to test and work with

raveit65:
PR is merged.

PR #1353, opened 2023-02-11
applets/wncklet/window-list: Add option to disable middle-click to close in window list

nsoggia:
Add a check button in the preferences context menu to select whether a middle mouse click in the window list should close the window or not.
Addresses issue #962.


lukefromdc (approved):
This works as intended, and the new checkbox in the tasklist preferences menu looks good with the others

lukefromdc:
Calling for a second review if I can get one

zhuyaliang (approved):
There is no problem with the code. I accept this function

PR #1355, opened 2023-02-18
tell dconf-edit about relocatable schemas

dnk:
support for dconf-editor relocatable schemas

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -2393,17 +2393,7 @@ show_week_changed (GSettings    *settings,
 static void
 setup_gsettings (ClockData *cd)
 {
-        gint format;
-        gchar *custom_format;
-
         cd->settings = mate_panel_applet_settings_new (MATE_PANEL_APPLET (cd->applet), CLOCK_SCHEMA);
-
-        /* hack to allow users to set custom format in dconf-editor */
-        format = g_settings_get_enum (cd->settings, KEY_FORMAT);
-        custom_format = g_settings_get_string (cd->settings, KEY_CUSTOM_FORMAT);
-        g_settings_set_enum (cd->settings, KEY_FORMAT, format);
-        g_settings_set_string (cd->settings, KEY_CUSTOM_FORMAT, custom_format);
-        g_free (custom_format);
 
         g_signal_connect (cd->settings, "changed::" KEY_FORMAT, G_CALLBACK (format_changed), cd);
         g_signal_connect (cd->settings, "changed::" KEY_SHOW_SECONDS, G_CALLBACK (show_seconds_changed), cd);
--- a/libmate-panel-applet/mate-panel-applet-gsettings.c
+++ b/libmate-panel-applet/mate-panel-applet-gsettings.c
@@ -34,7 +34,9 @@ add_to_dict (GVariant *dict, const gchar *schema, const gchar *path)
 {
     GVariantIter iter;
     GVariantBuilder builder;
-    gboolean already_contains;
+    gboolean is_schema_found;
+    gboolean is_incorrect_schema;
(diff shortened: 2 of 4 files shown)


dnk:
here result of applying relocatable schema
image

zhuyaliang:
@dnk I don't quite understand the meaning of the commit message. Can you supplement the information and compress the submission

dnk:
@zhuyaliang these changes is evolution of https://github.com/mate-desktop/mate-desktop/issues/300
dconf-editor did not use gsettings and cannot get information about relocatable schemas. which cause inability to view/edit mate-applets' preferences (path /org/mate/panel/objects/<object-id>/prefs) in dconf-editor.

dconf-editor have ca.desrt.dconf-editor.Settings:relocatable-schemas-user-paths to specify users' relocatable schemas

>gsettings describe ca.desrt.dconf-editor.Settings relocatable-schemas-user-paths
>
>A dictionary that maps schema IDs with path specifications. It is used to allow the user to >associate a relocatable schema to certain paths. Path specifications may contain wildcards in >the form of empty segments (e.g. /ca/desrt/dconf-editor//), defining possibly multiple paths. >The same schema ID may be associated with multiple path specifications.`

this pull request update ca.desrt.dconf-editor.Settings:relocatable-schemas-user-paths to allow dconf-editor to use information about users' relocatable schema and to show it properly

zhuyaliang:
@dnk Thank you for your explanation. I can test it now

lukefromdc (commented):
Builds and runs without any obvious problems, but I'm not the best reviewer for this onel

This cannot be tested for function with older versions of dconf-editor as they do not support relocatable schemas. I normally keep dconf-editor locked to version 3.18 as I have difficulty with the UI on newer versions. Testing with 3.36 this appeared to work, the name of the schema came up, and as a simple test I was able to lock the clock applet to the panel from it but wasn't able to find the custom font prefs from https://github.com/mate-desktop/mate-panel/pull/1358
which I had build with this.

Assuming that's a problem with the other PR or just an inability of myself to navigate the newer dconf-editor UI to find it's changes so approving this PR as its functionality seems to work in newer versions of dconf-editor.

At some point I might need to try reverting all the UI changes from dconf-editor but keeping the backend updates in a local repo, but that might be unresolvable without applying commits by hand after all these years. Might be easier to just backport relocatable schema support to the older dconf, though so far I have been able to get by without it

L-U-T-i:
@lukefromdc
> wasn't able to find the custom font prefs from #1358

I've rebuilt mate-panel with both PRs (#1355 and #1358), and I can find both settings "custom-font" and "use-custom-font" in "/ org / mate / panel /objects / clock / prefs" (dconf-editor-3.38.3-3.el9.x86_64).

lukefromdc:
I'm not opposing merge of this, but it needs another reviewer. I find the UI of GNOME's current dconf-editor nearly unusable
so I use a very old version that has the GNOME 2 UI but does not support relocatable schemas. Therefore, I could have
encountered another issue unique to me even when testing with the later version of dconf-editor.

dnk:
these changes went into libmate-panel-applet-4.so
to see result it is need to relaunch applets with updated libmate-panel-applet-4.so

lukefromdc:
I have the clock built in-process, so relaunching the panel relaunches the clock too

raveit65 (approved):
LGTM,
the relocatable schema name is now visible in dconf-editor.
I recall in the beginning of MATE when gsettings port was done by Stefano that there was a problem with relocatable schemas.
Good to remove the hack to use such schemas.

raveit65:
This will also detcect old non usable forgotten settings in relocatable-schemas which can be removed.
So +1 for merging it.

Issue #1090, opened 2020-06-11
Feature Request: Custom formatting for date on vertical panel


ellishg:
#### Expected behaviour
With a little configuration, it is possible to show the date and time in a nice way on a vertical panel in Cinnamon. https://github.com/linuxmint/cinnamon/issues/8554#issuecomment-491566445
![panel](https://user-images.githubusercontent.com/8961915/84353143-758d3a80-ab73-11ea-9673-5c9595d3d2f9.png)

#### Actual behaviour
When using a vertical panel with the clock applet, the date and time are hard to read because they are vertical.
![panel](https://user-images.githubusercontent.com/8961915/84352698-a9b42b80-ab72-11ea-8d20-dfdb6b080775.png)

Can there be a setting to customize the date/time format? Or could the default format for the data/time be changed when the panel is vertical?

PR #1358, opened 2023-02-19
custom font to render clock applet

dnk:
use custom font to render clock applet.
configuration is available via gsettings/dconf-editor

example how it looks with vertical panel (panel width: 44, font:8 "DejaVu Sans", format: '24-hour'):
image

can be helpful for https://github.com/mate-desktop/mate-panel/issues/1090 and https://github.com/mate-desktop/mate-panel/issues/157

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -80,6 +80,8 @@
 #define KEY_SHOW_WEATHER        "show-weather"
 #define KEY_SHOW_TEMPERATURE        "show-temperature"
 #define KEY_CUSTOM_FORMAT        "custom-format"
+#define KEY_USE_CUSTOM_FONT      "use-custom-font"
+#define KEY_CUSTOM_FONT          "custom-font"
 #define KEY_SHOW_WEEK                "show-week-numbers"
 #define KEY_CITIES                "cities"
 #define KEY_TEMPERATURE_UNIT        "temperature-unit"
@@ -147,6 +149,8 @@ struct _ClockData {
         /* preferences */
         ClockFormat  format;
         char        *custom_format;
+        char        *custom_font;
+        gboolean     use_custom_font;
         gboolean     showseconds;
         gboolean     showdate;
         gboolean     showweek;
@@ -184,6 +188,7 @@ struct _ClockData {
         gboolean   can_handle_format_12;
 
         GSettings *settings;
+        GtkCssProvider * custom_font_css_provider;
 
         const gchar *weather_icon_name;
 };
@@ -197,6 +202,7 @@ static int clock_numbers = 0;
 static void  update_clock (ClockData * cd);
 static void  update_tooltip (ClockData * cd);
 static void  update_panel_weather (ClockData *cd);
+static gboolean update_custom_font (ClockData *cd);
 static int   clock_timeout_callback (gpointer data);
 static float get_itime    (time_t current_time);
 
@@ -773,6 +779,8 @@ destroy_clock (GtkWidget * widget, ClockData *cd)
 
         g_free (cd->custom_format);
 
+        g_free (cd->custom_font);
+
         free_locations (cd);
 
         if (cd->location_tiles)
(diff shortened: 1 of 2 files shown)


applets/clock/clock.c
```diff
@@ -147,6 +149,8 @@ struct _ClockData {
         /* preferences */
         ClockFormat  format;
         char        *custom_format;
+        char        *custom_font;
```
zhuyaliang, line 152: ```custom_font``` not free memory when the clock is destroyed

applets/clock/clock.c
```diff
@@ -2370,6 +2379,68 @@ custom_format_changed (GSettings    *settings,
...
+        if (cd->use_custom_font && cd->custom_font != NULL && cd->custom_font[0] != '\0') {
+                GtkCssProvider *provider = gtk_css_provider_new ();
+                gchar *css = g_strdup_printf (
+                        "#clock-applet-button {\n"
+                        "  font: %s"
+                        "}", cd->custom_font);
+                if (gtk_css_provider_load_from_data (provider, css, -1, NULL)) {
+                        gtk_style_context_add_provider (gtk_widget_get_style_context (cd->panel_button),
+                                        GTK_STYLE_PROVIDER (provider),
+                                        GTK_STYLE_PROVIDER_PRIORITY_APPLICATION+1);
```
zhuyaliang, line 2404: Add space

zhuyaliang (approved):
The code format is good, the compilation runs well, and no memory leak is found

lukefromdc (commented):
I got this to build and run fine, but on killing the panel (so it could restart) and running my older version dconf-editor 3.18 I was not able to find any references to the clock schema (a relocatable one?) so wondering if this may depend on 
https://github.com/mate-desktop/mate-panel/pull/1355
to actually work. Will apply that on top of this change and see what happens

lukefromdc:
The changes here did not show up in dconf-editor after applying 
https://github.com/mate-desktop/mate-panel/pull/1355
rebuilding, restarting the pane, replacing my normal dconf-editor 3.18 with 3.36 (for relocatable schema support) and then opening dconf-editor. 

Due to the relocatable schema issue, this might be more useful if custom formats can be applied from the clock preferences dialog as well

L-U-T-i:
Why do we need "use-custom-font" setting?

If "custom-font" is left default ("Use default value" ON), custom font is not in use. If we turn it OFF and enter custom font, we use that one instead.

I find this setting redundant, and even potentially misleading (as one is on top of the window, and another one at the bottom - so one can easily miss the bottom "use-custom-font" setting, believing custom font is chosen already by setting "custom-font").

I support the idea of custom font through the clock preferences dialog (possibly with a selectable list of fonts installed instead of text window).

applets/clock/clock.c
```diff
@@ -1319,17 +1327,17 @@ clock_update_text_gravity (GtkWidget *label)
...
         gtk_css_provider_load_from_data (provider,
                                          "#clock-applet-button {\n"
-                                         "padding-top: 0px;\n"
-                                         "padding-bottom: 0px;\n"
-                                         "padding-left: 4px;\n"
-                                         "padding-right: 4px;\n"
+                                         "  padding-top: 0px;\n"
+                                         "  padding-bottom: 0px;\n"
+                                         "  padding-left: 0px;\n"
+                                         "  padding-right: 0px;\n"
```
raveit65, line 1340: Why removing the padding for vertical panels?
I can't recall but there was a reason why we use the padding, many years ago.
For me it looks nicer with a padding. Using 2px seems OK for me.
Any way i don't want to block it.
But this needs to be moved in a single commit.
Otherwise you will never find the change in git history, when user will blame us about.

raveit65: And why adding the 2 blanks before padding-......?

raveit65:
It works on my box but having 2 gsettings keys for the same setting seems to be overloaded.
I am sure most people will forget to enable the setting explicit after choosing the font.

applets/clock/clock.c
raveit65: last time that i ask........

lukefromdc:
What is the status of this?

lukefromdc:
We can reopen this if folks want to work on it further, closing as it's been almost two years with no updates

Issue #1354, opened 2023-02-11
[Feature Request] Run Application Dialog: launch the nth search hit by pressing alt+[number of hit]


mattarroz:
#### Expected behaviour
Next to each search hit, ranged right, is a hint to the short cut (maybe alt+[number] or something similar) to launch the corresponding application. 

#### Actual behaviour
No such hint or shortcut exists. I would not need this feature for applications that can be run by typing a single word. However, if you want to run an application that is only available through the mate menu, you need to type the actual command the menu item refers to, not the name of the menu item. So for example, if you create a menu item named "example" that runs the command /some/complicated/path/to/some/binary, you cannot type "example" to run /some/complicated/path/to/some/binary. You need to either grab the mouse and click on "example" (this is where the shortcut would definitely help) or type /some/complicated/path/to/some/binary.

#### Steps to reproduce the behaviour
N/A

#### MATE general version
1.24.1

#### Package version
1.24.1-2

#### Linux Distribution
Debian 11

#### Link to bugreport of your Distribution (requirement)
N/A

====
If you give me some hints, I may be able to implement the desired feature myself and create a PR.

PR #1359, opened 2023-02-26
panel-run-dialog: Add accelerator keys for "list of known applications"

mattarroz:
Following up on https://github.com/mate-desktop/mate-panel/issues/1354, this PR adds shortcuts to each search hit:

image

Pressing the shortcut key selects the corresponding program, which can then be run by pressing enter. This makes it possible to leverage the fuzzy search to run applications without touching the mouse or pressing tab several times to select the list first and then the desired item.

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -2138,7 +2138,7 @@ panel_run_dialog_present (GdkScreen *screen,
 			  guint32    activate_time)
 {
 	GtkBuilder *gui;
-	accelerator_keys_to_tree_iter_map = g_hash_table_new_full(g_direct_hash, g_direct_equal, NULL, NULL);
+	accelerator_keys_to_tree_iter_map = g_hash_table_new_full (g_direct_hash, g_direct_equal, NULL, NULL);
 
 	if (panel_lockdown_get_disable_command_line ())
 		return;
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -775,7 +775,7 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 		} else {
 			gtk_list_store_set (dialog->program_list_store,
 								&iter,
-								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+								COLUMN_ACCELERATOR_MASK, 0,
 								COLUMN_ACCELERATOR_KEY_VALUE, 0,
 								-1);
 		}
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -766,18 +766,18 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 		gtk_tree_model_get (model, &iter, COLUMN_VISIBLE, &visible, -1);
 		if (visible && visible_program_idx < G_N_ELEMENTS (accelerator_key_mapping)) {
 			gtk_list_store_set (dialog->program_list_store,
-								&iter,
-								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
-								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
-								-1);
+					    &iter,
+					    COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
+					    COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
+					    -1);
 			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[visible_program_idx].key_id), GINT_TO_POINTER(visible_program_idx));
 			visible_program_idx++;
 		} else {
 			gtk_list_store_set (dialog->program_list_store,
-								&iter,
-								COLUMN_ACCELERATOR_MASK, 0,
-								COLUMN_ACCELERATOR_KEY_VALUE, 0,
-								-1);
+					    &iter,
+					    COLUMN_ACCELERATOR_MASK, 0,
+					    COLUMN_ACCELERATOR_KEY_VALUE, 0,
+					    -1);
 		}
 
 		g_free (exec);
@@ -982,17 +982,17 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
 				    COLUMN_VISIBLE,   TRUE,
 				    -1);
 		if (i < G_N_ELEMENTS (accelerator_key_mapping)) {
-			gtk_list_store_set(dialog->program_list_store, &iter,
-								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
-								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
-								-1);
+			gtk_list_store_set (dialog->program_list_store, &iter,
+					    COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
+					    COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
+					    -1);
 			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[i].key_id), GINT_TO_POINTER(i));
 			i++;
 		} else {
-			gtk_list_store_set(dialog->program_list_store, &iter,
-								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
-								COLUMN_ACCELERATOR_KEY_VALUE, 0,
-								-1);
+			gtk_list_store_set (dialog->program_list_store, &iter,
+					    COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+					    COLUMN_ACCELERATOR_KEY_VALUE, 0,
+					    -1);
 		}
 	}
 	g_slist_free_full (all_applications, matemenu_tree_item_unref);
@@ -1707,18 +1707,18 @@ combobox_changed (GtkComboBox    *combobox,
 			for (valid = gtk_tree_model_get_iter_first (model, &iter); valid; valid = gtk_tree_model_iter_next (model, &iter)) {
 				if (i < G_N_ELEMENTS (accelerator_key_mapping)) {
 					gtk_list_store_set (GTK_LIST_STORE (GTK_TREE_MODEL (dialog->program_list_store)), &iter,
-										COLUMN_VISIBLE, TRUE,
-										COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
-										COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
-										-1);
+							    COLUMN_VISIBLE, TRUE,
+							    COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
+							    COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
+							    -1);
 					g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER (accelerator_key_mapping[i].key_id), GINT_TO_POINTER(i));
 					i++;
 				} else {
 					gtk_list_store_set (GTK_LIST_STORE (GTK_TREE_MODEL (dialog->program_list_store)), &iter,
-										COLUMN_VISIBLE, TRUE,
-										COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
-										COLUMN_ACCELERATOR_KEY_VALUE, 0,
-										-1);
+							    COLUMN_VISIBLE, TRUE,
+							    COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+							    COLUMN_ACCELERATOR_KEY_VALUE, 0,
+							    -1);
 				}
 			}
 
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -39,6 +39,7 @@
 #include <unistd.h>
 
 #include <glib/gi18n.h>
+#include <glib.h>
 #include <gio/gio.h>
 #include <gdk/gdkkeysyms.h>
 #include <matemenu-tree.h>
@@ -106,12 +107,36 @@ typedef struct {
 enum {
 	COLUMN_GICON,
 	COLUMN_NAME,
+	COLUMN_ACCELERATOR_MASK,
+	COLUMN_ACCELERATOR_KEY_VALUE,
 	COLUMN_COMMENT,
 	COLUMN_PATH,
 	COLUMN_EXEC,
 	COLUMN_VISIBLE,
 	NUM_COLUMNS
 };
+
+typedef struct {
+        gint list_item_idx;
+        GdkModifierType modifier;
+        guint key_id;
+} AcceleratorKeyMapping;
+
+const AcceleratorKeyMapping accelerator_key_mapping[] =
+{
+        {0, GDK_MOD1_MASK, GDK_KEY_1 },
+        {1, GDK_MOD1_MASK, GDK_KEY_2 },
+        {2, GDK_MOD1_MASK, GDK_KEY_3 },
+        {3, GDK_MOD1_MASK, GDK_KEY_4 },
+        {4, GDK_MOD1_MASK, GDK_KEY_5 },
+        {5, GDK_MOD1_MASK, GDK_KEY_6 },
+        {6, GDK_MOD1_MASK, GDK_KEY_7 },
+        {7, GDK_MOD1_MASK, GDK_KEY_8 },
+        {8, GDK_MOD1_MASK, GDK_KEY_9 },
+        {9, GDK_MOD1_MASK, GDK_KEY_0 },
+};
+
+static GHashTable *accelerator_keys_to_tree_iter_map = NULL;
 
 static PanelRunDialog *static_dialog = NULL;
 
@@ -237,6 +262,10 @@ panel_run_dialog_destroy (PanelRunDialog *dialog)
 	if (dialog->dir_hash)
 		g_hash_table_destroy (dialog->dir_hash);
 	dialog->dir_hash = NULL;
+
+	if (accelerator_keys_to_tree_iter_map)
+		g_hash_table_destroy (accelerator_keys_to_tree_iter_map);
+	accelerator_keys_to_tree_iter_map = NULL;
 
 	for (l = dialog->possible_executables; l; l = l->next)
 		g_free (l->data);
@@ -664,19 +693,6 @@ fuzzy_command_match (const char *cmd1,
 }
 
 static gboolean
-panel_run_dialog_make_all_list_visible (GtkTreeModel *model,
-					GtkTreePath  *path,
-					GtkTreeIter  *iter,
-					gpointer      data)
-{
-	gtk_list_store_set (GTK_LIST_STORE (model), iter,
-			    COLUMN_VISIBLE, TRUE,
-			    -1);
-
-	return FALSE;
-}
-
-static gboolean
 panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 {
 	GtkTreeIter   iter;
@@ -686,6 +702,7 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 	GIcon        *found_icon;
 	char         *found_name;
 	gboolean      fuzzy;
+	gint          visible_program_idx = 0;
 
 	model = GTK_TREE_MODEL (dialog->program_list_store);
 	path = gtk_tree_path_new_first ();
@@ -704,12 +721,14 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 	found_icon = NULL;
 	found_name = NULL;
 	fuzzy = FALSE;
+	g_hash_table_remove_all (accelerator_keys_to_tree_iter_map);
 
 	do {
 		char *exec = NULL;
 		GIcon *icon = NULL;
 		char *name = NULL;
 		char *comment = NULL;
+		gboolean visible = FALSE;
 
 		gtk_tree_model_get (model, &iter,
 				    COLUMN_EXEC,      &exec,
@@ -742,6 +761,23 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 					    &iter,
 					    COLUMN_VISIBLE, FALSE,
 					    -1);
+		}
+
+		gtk_tree_model_get (model, &iter, COLUMN_VISIBLE, &visible, -1);
+		if (visible && visible_program_idx < G_N_ELEMENTS (accelerator_key_mapping)) {
+			gtk_list_store_set (dialog->program_list_store,
+								&iter,
+								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
+								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
+								-1);
+			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[visible_program_idx].key_id), GINT_TO_POINTER(visible_program_idx));
+			visible_program_idx++;
+		} else {
+			gtk_list_store_set (dialog->program_list_store,
+								&iter,
+								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+								COLUMN_ACCELERATOR_KEY_VALUE, 0,
+								-1);
 		}
 
 		g_free (exec);
@@ -895,6 +931,8 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
 	dialog->program_list_store = gtk_list_store_new (NUM_COLUMNS,
 							 G_TYPE_ICON,
 							 G_TYPE_STRING,
+							 G_TYPE_INT, // For accelerator modifier mask
+							 G_TYPE_UINT, // For accelerator key value
 							 G_TYPE_STRING,
 							 G_TYPE_STRING,
 							 G_TYPE_STRING,
@@ -922,6 +960,9 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
 		}
 	}
 
+	gint i = 0;
+	g_hash_table_remove_all (accelerator_keys_to_tree_iter_map);
+
 	for (l = all_applications; l; l = l->next) {
 		MateMenuTreeEntry *entry = l->data;
 		GtkTreeIter    iter;
@@ -940,6 +981,19 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
 				    COLUMN_PATH,      matemenu_tree_entry_get_desktop_file_path (entry),
 				    COLUMN_VISIBLE,   TRUE,
 				    -1);
+		if (i < G_N_ELEMENTS (accelerator_key_mapping)) {
+			gtk_list_store_set(dialog->program_list_store, &iter,
+								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
+								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
+								-1);
+			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[i].key_id), GINT_TO_POINTER(i));
+			i++;
+		} else {
+			gtk_list_store_set(dialog->program_list_store, &iter,
+								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+								COLUMN_ACCELERATOR_KEY_VALUE, 0,
+								-1);
+		}
 	}
 	g_slist_free_full (all_applications, matemenu_tree_item_unref);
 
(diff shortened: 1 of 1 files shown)


mate-panel/panel-run-dialog.c
```diff
@@ -1962,6 +2030,20 @@ panel_run_dialog_new (GdkScreen  *screen,
 	g_signal_connect_swapped (dialog->run_dialog, "destroy",
 				  G_CALLBACK (panel_run_dialog_destroy), dialog);
 
+
+        for (int i = 0; accelerator_key_mapping[i].list_item_idx != -1; i++) {
+              GClosure* closure_key = g_cclosure_new(
```
mattarroz, line 2035: Do I need to manually destroy the objects created by g_cclosure_new and gtk_accel_group_new?

cwendling: * `g_cclosure_new()`: No, because [`gtk_accel_group_new()`](https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gtk/gtkaccelgroup.c#L627) [sinks](https://docs.gtk.org/gobject/method.Closure.sink.html) the `GClosure` reference.
* `gtk_accel_group_new()`: I'm less positive, but it *seems* to have to be unreffed manually: it's not a `GInitiallyUnowned`, and without additional info in the docs it should mean you have to manage your reference manually, and it seems the code indeed adds its own reference when attaching it to the window.

This said, shouldn't you create only *one* `GtkAccelGroup`, and connect all the keys on it?  After all, it's not called a *group* for nothing :wink:

mate-panel/panel-run-dialog.c
```diff
@@ -106,13 +106,37 @@ typedef struct {
...
+        {1, GDK_MOD1_MASK, GDK_KEY_2 },
+        {2, GDK_MOD1_MASK, GDK_KEY_3 },
+        {3, GDK_MOD1_MASK, GDK_KEY_4 },
+        {4, GDK_MOD1_MASK, GDK_KEY_5 },
+        {5, GDK_MOD1_MASK, GDK_KEY_6 },
+        {6, GDK_MOD1_MASK, GDK_KEY_7 },
+        {7, GDK_MOD1_MASK, GDK_KEY_8 },
+        {8, GDK_MOD1_MASK, GDK_KEY_9 },
+        {9, GDK_MOD1_MASK, GDK_KEY_0 },
+        {-1, GDK_MOD1_MASK, 0 }
```
cwendling, line 137: You could remove this last entry, and use `G_N_ELEMENTS(accelerator_key_mapping)` to know the array length.

mate-panel/panel-run-dialog.c
```diff
@@ -729,18 +754,26 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 			gtk_list_store_set (dialog->program_list_store,
 					    &iter,
 					    COLUMN_VISIBLE, TRUE,
+					    COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
+					    COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
 					    -1);
+                        if (accelerator_key_mapping[visible_program_idx].list_item_idx != -1) visible_program_idx++;
```
cwendling, line 760: Doesn't this mean all entries after the 9th will have the 10th accelerator?  I don't see anything that prevents setting the accelerator if it's past the last one.

I think I would actually move this outside those tests, and just have a flag "this is visible", and depending on that set or not the accel.  Something like this pseudo-code, although possibly optimizing the visibility check with a local variable (not that it matters that much though):
```c
if (visible_program_idx < G_N_ELEMENTS(accelerator_key_mapping)) {
	gboolean visible = FALSE;

	gtk_tree_model_get(model, &iter, COLUMN_VISIBLE, &visible, -1);
	if (visible) {
		gtk_list_store_set (dialog->program_list_store,
				    &iter,
				    COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
				    COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
				    -1);
		visible_program_idx++;
	}
}

File: mate-panel/panel-run-dialog.c [hunk 4]
```diff
@@ -932,14 +968,17 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
...
+                gtk_list_store_set (dialog->program_list_store, &iter,
+                                   COLUMN_GICON,     gicon,
+                                   COLUMN_NAME,      g_app_info_get_display_name(G_APP_INFO(ginfo)),
+                                   COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
+                                   COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
+                                   COLUMN_COMMENT,   g_app_info_get_description(G_APP_INFO(ginfo)),
+                                   COLUMN_EXEC,      g_app_info_get_commandline(G_APP_INFO(ginfo)),
+                                   COLUMN_PATH,      matemenu_tree_entry_get_desktop_file_path (entry),
+                                   COLUMN_VISIBLE,   TRUE,
+                                   -1);
```
cwendling, line 980: Indentation has a problem here, it doesn't align properly it seems

mate-panel/panel-run-dialog.c
```diff
@@ -967,9 +1006,18 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
 	gtk_tree_view_column_set_attributes (column, renderer,
                                              "text", COLUMN_NAME,
                                              NULL);
+	gtk_tree_view_append_column (GTK_TREE_VIEW (dialog->program_list), column);
 
+	renderer = gtk_cell_renderer_accel_new ();
+	g_object_set (renderer, "accel-mode", GTK_CELL_RENDERER_ACCEL_MODE_GTK,
+	             "editable", FALSE, NULL);
+
+	column = gtk_tree_view_column_new_with_attributes ("", renderer,
```
cwendling, line 1015: You should set a title to the column even if it's not visible, it helps accessibility tools present that properly

mate-panel/panel-run-dialog.c
```diff
@@ -967,9 +1006,18 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
...
+	                                                  "accel-mods", COLUMN_ACCELERATOR_MASK, "accel-key",
+	                                                  COLUMN_ACCELERATOR_KEY_VALUE, NULL);
 	gtk_tree_view_append_column (GTK_TREE_VIEW (dialog->program_list), column);
 
+
```
cwendling, line 1020: no need for that extra line

mate-panel/panel-run-dialog.c
```diff
@@ -1945,6 +1993,26 @@ key_press_event (GtkWidget    *run_dialog,
 	return FALSE;
 }
 
+static void panel_run_dialog_accelerator_key_pressed(GtkAccelGroup *accel_group, GObject *acceleratable,
+                                               guint keyval, GdkModifierType modifier, PanelRunDialog *dialog)
```
cwendling, line 1997: style: specifiers and return type on its own line, and each argument on one line with both type and name aligned

mate-panel/panel-run-dialog.c
```diff
@@ -1945,6 +1993,26 @@ key_press_event (GtkWidget    *run_dialog,
...
+                if (accelerator_key_mapping[i].key_id == keyval) {
+                        entry_number = accelerator_key_mapping[i].list_item_idx;
+                        break;
+                }
+        }
+        if (entry_number == -1) return;
+        GtkTreePath *path = gtk_tree_path_new_from_indices(entry_number, -1);
+        gtk_tree_view_set_cursor(GTK_TREE_VIEW(dialog->program_list), path, NULL, FALSE);
+        gtk_widget_grab_focus (dialog->program_list);
+        gtk_tree_path_free(path);
```
cwendling, line 2013: I am ever so slightly worried that the match is not done on the accelerator actually displayed, although indeed that should match.  Maybe that's good enough though and looking up the rows would be a bit more annoying, not sure what makes more sense.

cwendling (changes requested):
I don't really know what I think of the feature and I didn't test it, but I can see it possibly be handy. Here a few comments on the code anyway.

mate-panel/panel-run-dialog.c
mattarroz: I factored the gtk_accel_group_new call outside the loop. I could not find gtk_accel_group_free in the documentation, so I guess it's not a memory leak.

mate-panel/panel-run-dialog.c
mattarroz: Fixed

mate-panel/panel-run-dialog.c
mattarroz: Have a look again, I tried to fix it like you suggested. The entries after the last one have the "deactivated" shortcut label.

mate-panel/panel-run-dialog.c
mattarroz: I'm sorry my editor was not properly setup. I created a clang-format file to fix these issues. If you want, I can share it in some form.

mate-panel/panel-run-dialog.c
mattarroz: Added the title "Shortcut".

mate-panel/panel-run-dialog.c
mattarroz: Removed the line.

mate-panel/panel-run-dialog.c
mattarroz: Have a look again, I tried to fix the style.

mate-panel/panel-run-dialog.c
mattarroz: Used a hashmap to obtain the proper entry. Unless I'm mistaken that should also reduce the computational complexity.

mattarroz:
To showcase the potential of this feature in conjunction with the already existing fuzzy search, consider this further proposal:

image

Pressing 'Alt+2' now selects the second item, corresponding to the command `google-chrome "google.de/search?q=mate%20i"`. In consequence, you could open a new browser tab with the search results without touching the mouse, no matter if the browser is already running/is visible on your current workspace or not. This could of course be done for any combination of website and browser.

lukefromdc:
That would have to work with ANY browser. Lots of folks using MATE who would never install Google Chrome

raveit65:
The keyboard shortcut name (alt+1, alt+2, etc.) doesn't display very well in the entries.
Sometimes it works and sometimes not. Mostly i have to switch the gtk theme to make it visible.
I am using only MATE themes. Very weird.
The keyboard shortcuts itself are working well. I can switch to the search result list and between the entries there.

mattarroz:
> The keyboard shortcut name (alt+1, alt+2, etc.) doesn't display very well in the entries of the search result list. Sometimes it works and sometimes not. Mostly i have to switch the gtk theme to make it visible. I am using only MATE themes. Very weird. The keyboard shortcuts itself are working well. I can switch to the search result list and between the entries there.

That should not happen. Can you provide more details on how to reproduce the bug?

I tried all Mate themes on Debian 11 (Vanilla Mate 1.24 patched with this PR) and could not reproduce it. I was never able to make the window so small that you could not read the shortcuts. I also tried setting the font size to 40 to no avail. In your case, are you able to enlarge the window so that the shortcuts become readable? Are you able to make the window arbitrarily small?

raveit65:
Ok, i am using the hottest shit. ;)
Just released fedora 38 with Mate-1.27. All packages are based of my official fedora rpm packages.
I made a little video because it explains more than 1000 words.
I started with Menta-Blue (our default theme) and the alt-f2 shortcut.
Later i switched to our Blue-Submarine theme to make the alt-[1,2,3....] font visible.
https://user-images.githubusercontent.com/961604/232328048-accfd1a8-bf4a-47e5-9a20-549ce8709434.mp4

raveit65:
Ok, the font is there when i resize the window. I guess this is because i am using HDPI (3840x2160) resolution.
It seems the default wuindow size is different.


https://user-images.githubusercontent.com/961604/232329202-ea42ff2c-2121-4dc5-bf0d-546346710514.mp4

raveit65:
Maybe it's possible to place the font on the left side of the column?

mattarroz:
> Maybe it's possible to place the font on the left side of the column?

I can reproduce the effect by adding a very long menu entry to the mate menu. Then the vertical scrollbar appears, hiding the shortcut keys.

I now made the shortcut keys a "sticky column" on the right side, so they cannot be hidden by long program names. What are your thoughts on this idea? Would be cool if you could test it :-)

raveit65:
Yeah, i will test it.
I investigate some test for myself, I inspected the GUI with gtk3-inspector and the problem is definitely the first column which expand for some reason.
Gtk_TreeView_0

Gtk_TreeView

Going deeper in tree i found a setting to hardcode the column width of the first columm.
Gtk_TreeView_Columm0
In setting page of Gtk_TreeView_Column
Gtk_TreeView_Columm1
max-width = 450-460 works fine here.
With max-width = 453 the horizontal scrollbar disappear.
What do you think?

In general i think we have a HIDPI scaling issue here as the issue exits only on systems with gtk_scale = 2.0 or gtk_scale > 1.0. The fact that changing the gtk3 theme or in other words mate-setting-daemon is involved indicates a scaling issue.
@vkareh 
Any idea?

vkareh:
@mattarroz the new design with double scroll bars is visually jarring:
Screenshot at 2023-04-26 11-22-18

I prefer the previous approach, just making sure that the shortcut text is left-aligned.

vkareh:
@mattarroz regarding the google search, I don't think that's a good idea. It's far too opinionated. For it to be viable as part of core MATE it would need to be configurable to use _any_ application with _any_ parameter, so that I can use, for example, Firefox with DDG (instead of Chrome with Google).

That configurability should extend to other applications too, so that I can also configure it to use the `mate-dictionary` app and feed the search term to it, for example. Without that configurability it's no good, as it only addresses a single limited use case.

If this is the way we want to go, I suggest a completely different PR to review/discuss.

raveit65:
@vkareh 
The Gtk_TreeView is below the scrolled_window
Bildschirmfoto zu 2023-04-26 19-11-25

My screenshots are without latest commit.

I don't see the mentioned "google search"  in code change.
Only in the showcase screenshot.
So this really a future PR.

@mattarroz  A second scrolled window doesn't look nice, ihmo.
Personal i can live with that issue because i like this new function very much an i will use it for myself and maybe for fedora. This makes it very easy to use the dialog only with keyboard for me.

raveit65:
This code snip-set makes it better.
```
 mate-panel/panel-run-dialog.c | 5 ++++-
 1 file changed, 4 insertions(+), 1 deletion(-)

diff --git a/mate-panel/panel-run-dialog.c b/mate-panel/panel-run-dialog.c
index 47972c2e..730ab211 100644
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -1017,11 +1017,14 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
                                              NULL);
 
 	renderer = gtk_cell_renderer_text_new ();
-	gtk_tree_view_column_pack_start (column, renderer, TRUE);
+	gtk_tree_view_column_pack_start (column, renderer, FALSE);
 	gtk_tree_view_column_set_attributes (column, renderer,
                                              "text", COLUMN_NAME,
                                              NULL);
 
+	gtk_tree_view_column_set_sizing (column, 
+	                                 GTK_TREE_VIEW_COLUMN_AUTOSIZE);
+
 	gtk_tree_view_append_column (GTK_TREE_VIEW (dialog->program_list), column);
 
 	renderer = gtk_cell_renderer_accel_new ();
-- 
```
Ie. typing thunderbird.
typing_thunderbird
typing_thunderbird_2
After typing the shortcut is visible like it should.

raveit65:
And the real cause of this weird behaviour on my system is openjdk. Thank you, for adding such a long name to desktop file. After scrolling in my default application list.
Bildschirmfoto zu 2023-04-26 20-47-48

But my code suggestion is needed to resize columns during typing.

lukefromdc:
I certainly agree that a search function would have to be both browser and search engine agnostic. Under no circumstances would I be OK with effectively making Google part of MATE, and any other search engine (or any other browser) would also be problematic for somebody somewhere. Thus these must be user selected and any defaults must be privacy safe and certainly not Google or anything else supported by user data and ads

lukefromdc (approved):
Builds and runs fine, looks like this in my theme as it currently stands. There is a problem area in the partially filled dialog in my theme, and I will fix that in my theme if this is merged. BlackMATE, Menta, and Adwaita-dark do not have this problem and render correctly so no problem with the PR itself:

Completely filled dialog, my theme:

Completely_Filled_Dialog

Partially filled dialog, my theme

Partially_Filled_Dialog

Partially filled dialog, BlackMATE

Partially_Filled_Dialog_BlackMATE

lukefromdc:
We also have an issue in all themes: The accelerator treeview (which is a DIFFERENT treeview widget than the program list treeview) does not get the same height per row as the program list treeview. This occurs in every theme, here it is in Menta:
Completely_Filled_Dialog_Menta

lukefromdc (commented):
Changing my review to a comment due to the row misalignment in the UI

lukefromdc:
I just checked vanilla GTK 3.24.34 to ensure the row misalignment wasn't specific to the gtk-classic patches (which I use), the problem was less severe but still occurred. The sure way to fix this and possibly also lighter code would be to use a single treeview with  two columns rather than attempting to force two different treeviews to line up

raveit65:
@mattarroz 
I think you should remove the "sticky column" from last commit as it caused issues with themes.
Themes are different. It can work with one theme, but not with another.
Instead you should add the proposed code snip-set which managed the 1 column width better, ihmo.

@lukefromdc 
I think it is a problem with different css padding/font wight settings for treeview.view and treeview.view:disabled in themes.
It is very uncommon that you see 2 treeviews side by side with different status settings for fonts in a gui.

mattarroz:
> @mattarroz regarding the google search, I don't think that's a good idea. It's far too opinionated. For it to be viable as part of core MATE it would need to be configurable to use _any_ application with _any_ parameter, so that I can use, for example, Firefox with DDG (instead of Chrome with Google).
> 
> That configurability should extend to other applications too, so that I can also configure it to use the `mate-dictionary` app and feed the search term to it, for example. Without that configurability it's no good, as it only addresses a single limited use case.
> 
> If this is the way we want to go, I suggest a completely different PR to review/discuss.

My comment was apparently very easy to misunderstand, sorry. I wanted to highlight the usefulness of the current PR/feature (the shortcut keys) for a potential future feature (the websearch). I did not want introduce another feature into this PR. I agree 100% this future PR needs to be thought through properly and did not mean to introduce Google or any vendor into the Mate desktop. I only need google-chrome for work purposes, BTW.

mattarroz:
@raveit65 I reverted the last commit and added your patch.

raveit65:
@cwendling 
Can you please resolve all your conversations if they are fixed, to get this ready for merging?

vkareh:
This works well.

small nitpicks: can you remove the `Disabled` and just set an empty string or whatever when there are more than 10 results? Finally, can you squash/fixup the commits so that this is a bit more atomic before merging?

raveit65:
You can use the `git rebase -i <last-commit-hash_before_PR` for removing or squashing the unneeded/revert commits.

mattarroz:
> This works well.
> 
> small nitpicks: can you remove the `Disabled` and just set an empty string or whatever when there are more than 10 results? Finally, can you squash/fixup the commits so that this is a bit more atomic before merging?

I squashed all commits. I only left the individual commits while the review is still ongoing so it's easier to follow the development. As for the `Disabled` entries, they are automatically generated by [Gtk.CellRendererAccel](https://docs.gtk.org/gtk3/class.CellRendererAccel.html). I did not find a way to disable the `Disabled` entries in the documentation ;-) If someone knows how to do it, I would be willing to implement it.

vkareh (approved):
LGTM

mate-panel/panel-run-dialog.c
```diff
@@ -1962,6 +2063,19 @@ panel_run_dialog_new (GdkScreen  *screen,
 	g_signal_connect_swapped (dialog->run_dialog, "destroy",
 				  G_CALLBACK (panel_run_dialog_destroy), dialog);
 
+	GtkAccelGroup* accel_group = gtk_accel_group_new();
```
cwendling, line 2066: looking at my previous review, shouldn't this be unreffed after adding it to the window? I'd thing there is a leak here

mate-panel/panel-run-dialog.c
```diff
@@ -744,6 +763,23 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
...
+								&iter,
+								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
+								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
+								-1);
+			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[visible_program_idx].key_id), GINT_TO_POINTER(visible_program_idx));
+			visible_program_idx++;
+		} else {
+			gtk_list_store_set (dialog->program_list_store,
+								&iter,
+								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
```
cwendling, line 778: shouldn't the modifier be `0`? It doesn't actually matter though.

cwendling (changes requested):

mate-panel/panel-run-dialog.c
cwendling: note to clarify: this is just a suggestion, as this is hidden anyway, it really doesn't matter, it just made my eyebrows rise a tad when I saw it.

mate-panel/panel-run-dialog.c
```diff
@@ -744,6 +763,23 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
 					    -1);
 		}
 
+		gtk_tree_model_get (model, &iter, COLUMN_VISIBLE, &visible, -1);
+		if (visible && visible_program_idx < G_N_ELEMENTS (accelerator_key_mapping)) {
+			gtk_list_store_set (dialog->program_list_store,
+								&iter,
+								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[visible_program_idx].modifier,
+								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[visible_program_idx].key_id,
+								-1);
```
raveit65, line 772: Formatting issue. Too much tabs are used in this block. Tab size seems 8 in this file.

mate-panel/panel-run-dialog.c
```diff
@@ -744,6 +763,23 @@ panel_run_dialog_find_command_idle (PanelRunDialog *dialog)
...
+			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[visible_program_idx].key_id), GINT_TO_POINTER(visible_program_idx));
+			visible_program_idx++;
+		} else {
+			gtk_list_store_set (dialog->program_list_store,
+								&iter,
+								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+								COLUMN_ACCELERATOR_KEY_VALUE, 0,
+								-1);
```
raveit65, line 780: Same formatting issue here.

mate-panel/panel-run-dialog.c
```diff
@@ -940,6 +981,19 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
 				    COLUMN_PATH,      matemenu_tree_entry_get_desktop_file_path (entry),
 				    COLUMN_VISIBLE,   TRUE,
 				    -1);
+		if (i < G_N_ELEMENTS (accelerator_key_mapping)) {
+			gtk_list_store_set(dialog->program_list_store, &iter,
+								COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
+								COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
+								-1);
```
raveit65, line 988: Same formatting issue here.

raveit65: And a blank is missing after gtk_list_store_set.

mate-panel/panel-run-dialog.c
```diff
@@ -940,6 +981,19 @@ panel_run_dialog_add_items_idle (PanelRunDialog *dialog)
...
+			g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER(accelerator_key_mapping[i].key_id), GINT_TO_POINTER(i));
+			i++;
+		} else {
+			gtk_list_store_set(dialog->program_list_store, &iter,
+								COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+								COLUMN_ACCELERATOR_KEY_VALUE, 0,
+								-1);
```
raveit65, line 995: Same formatting issue here and missing blank after gtk_list_store_set.

mate-panel/panel-run-dialog.c
```diff
@@ -1630,12 +1696,31 @@ combobox_changed (GtkComboBox    *combobox,
...
+
+			g_hash_table_remove_all (accelerator_keys_to_tree_iter_map);
+			model = GTK_TREE_MODEL (dialog->program_list_store);
+			for (valid = gtk_tree_model_get_iter_first (model, &iter); valid; valid = gtk_tree_model_iter_next (model, &iter)) {
+				if (i < G_N_ELEMENTS (accelerator_key_mapping)) {
+					gtk_list_store_set (GTK_LIST_STORE (GTK_TREE_MODEL (dialog->program_list_store)), &iter,
+										COLUMN_VISIBLE, TRUE,
+										COLUMN_ACCELERATOR_MASK, (gint)accelerator_key_mapping[i].modifier,
+										COLUMN_ACCELERATOR_KEY_VALUE, accelerator_key_mapping[i].key_id,
+										-1);
```
raveit65, line 1713: Same formatting issue here.

mate-panel/panel-run-dialog.c
```diff
@@ -1630,12 +1696,31 @@ combobox_changed (GtkComboBox    *combobox,
...
+					g_hash_table_insert (accelerator_keys_to_tree_iter_map, GUINT_TO_POINTER (accelerator_key_mapping[i].key_id), GINT_TO_POINTER(i));
+					i++;
+				} else {
+					gtk_list_store_set (GTK_LIST_STORE (GTK_TREE_MODEL (dialog->program_list_store)), &iter,
+										COLUMN_VISIBLE, TRUE,
+										COLUMN_ACCELERATOR_MASK, (gint)GDK_MOD1_MASK,
+										COLUMN_ACCELERATOR_KEY_VALUE, 0,
+										-1);
```
raveit65, line 1721: Same formatting issue here.

lukefromdc:
According to https://docs.gtk.org/gtk3/class.CellRendererAccel.html
the accelerators can be made editable. This would allow users to reset the keys used from inside the dialog, and my guess is this would remove the disabled atribute as well, though I am not certain about that.

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 6cff6f3d2e9fd695399328e5597ac7bfbad2e2ab)

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 6cff6f3d2e9fd695399328e5597ac7bfbad2e2ab)

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 6cff6f3d2e9fd695399328e5597ac7bfbad2e2ab)

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 6cff6f3d2e9fd695399328e5597ac7bfbad2e2ab)

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 6cff6f3d2e9fd695399328e5597ac7bfbad2e2ab)

mate-panel/panel-run-dialog.c
mattarroz: All formatting issues were fixed using https://gist.github.com/mattarroz/6d32243ece7445efc7e71ea6e6202b4a

Maybe we should consider adding a .clang-format file to the source tree?

(fixup 6cff6f3d2e9fd695399328e5597ac7bfbad2e2ab)

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 851f23981817a9a29670f6f31aed59cb115a9ee8)

mate-panel/panel-run-dialog.c
mattarroz: Fixed (fixup 5dbd153e35a034b233494b6ba7e7377b8d6af537).

However, `valgrind --leak-check=full ./mate-panel --run-dialog` still shows lots of leaks, but I guess this is normal?

mattarroz:
> According to https://docs.gtk.org/gtk3/class.CellRendererAccel.html the accelerators can be made editable. This would allow users to reset the keys used from inside the dialog, and my guess is this would remove the disabled atribute as well, though I am not certain about that.

Making the keybindings editable would be nice, but it requires more code changes than setting the corresponding attribute on the cell renderer. I think it should be left to a future PR.

mate-panel/panel-run-dialog.c
raveit65: How does this work?

raveit65 (approved):
LGTM

lukefromdc:
Commit carrying the request for a change replaced by a force-push, three approving reviews. No way to mark it resolved it here as that commit no longer exists, so force-merging. No unresolved conversations showing on the currently visible commits, so presuming this a limitation of git itself. NOT dismissing the review so as to preserve the conversation.

mate-panel/panel-run-dialog.c
cwendling: @mattarroz some will be normal (there's a lot of one-shot "leaks" of persistent data in e.g. freetype, some in GTK and others, but they are not really "leaks", just annoying in the reports -- you might be lucky using GTK and GLib's suppression files), but some might still be bugs.  Yet, so long as your changes don't introduce new issues you're fine for this PR.

PR #1361, opened 2023-03-18
Clock Applet: Update date/time and weather after resume from sleep mode

### LINKED ISSUE #481: Clock doesn't get updated when opening notebook display lid
On notebook, when closing the lid and then opening it after a while, clock is stuck on date and time the lid was closed at and doesn't get update until clicking it. E.g. when closing notebook in the evening at 23:30 PM and opening it in the morning, instead of showing new date and time, it is stuck at yesterday's date and time. Currently at 1.14.0 version, but issue existed in previous versions.
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/37931764-clock-doesn-t-get-updated-when-opening-notebook-display-lid?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

thesquash:
Before this commit, the Clock Applet was unaware of when the system resumed from sleep mode (e.g. suspend or hibernate).  This meant that, if the user suspended the system and then woke it up some time later, the clock might not show the real current time, but instead would continue to show the time at which the system was suspended for up to a minute after the system woke up.  This could be confusing to the user if they peered up at the clock immediately after, say, opening their laptop lid.

This commit connects to ConsoleKit or (if available) systemd-logind via D-Bus, and listens for a signal that indicates when the system wakes up from sleep mode.  When the system does wake up, the Clock Applet updates the date/time immediately, and also updates all the weather/temperature readings if possible.

Note that if logind is missing, ConsoleKit2 is required for this feature to work; the original ConsoleKit lacked the signal required to implement this feature.  This should not be a problem since it is unlikely that anyone would even consider using the original ConsoleKit in this day and age.

Fixes #481.

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -85,6 +85,16 @@
 #define KEY_TEMPERATURE_UNIT        "temperature-unit"
 #define KEY_SPEED_UNIT                "speed-unit"
 
+/* For watching for when the system resumes from sleep mode (e.g. suspend)
+ * and updating the clock as soon as that happens. */
+#define LOGIND_RUNNING() (access("/run/systemd/seats/", F_OK) >= 0)
+#define SYSTEMD_LOGIND_SERVICE   "org.freedesktop.login1"
+#define SYSTEMD_LOGIND_PATH      "/org/freedesktop/login1"
+#define SYSTEMD_LOGIND_INTERFACE "org.freedesktop.login1.Manager"
+#define CK_SERVICE               "org.freedesktop.ConsoleKit"
+#define CK_MANAGER_PATH          "/org/freedesktop/ConsoleKit/Manager"
+#define CK_MANAGER_INTERFACE     "org.freedesktop.ConsoleKit.Manager"
+
 #define _clock_get_widget(x,y) (GTK_WIDGET (gtk_builder_get_object ((x)->builder, (y))))
 
 enum {
@@ -186,6 +196,8 @@ struct _ClockData {
         GSettings *settings;
 
         const gchar *weather_icon_name;
+
+        GDBusProxy *system_manager_proxy;
 };
 
 /* Used to count the number of clock instances. It's there to know when we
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
I can confirm that this works. Note that as the machine comes back out of suspend, updating the time is one of the last things that happens, coming at about the same time you get the mouse back but it does work, and time will now auto-update after resume without having to click on it.

Test was on Debian Unstable with systemd

lukefromdc:
What happens with this if someone not using systemd has the old version of ConsoleKit installed? Does the clock just fail to update on resume or does the applet crash?

thesquash:
In the event that the MATE Panel with this commit applied is used with the original ConsoleKit (or a sufficiently old pre-release version of ConsoleKit2), the Clock Applet will act as it used to:  It will not automatically update when the computer comes out of suspend mode, but will otherwise work as expected.  I was very careful to implement this feature in such a way that the Clock Applet will fall back to its old behavior if no system manager (`logind` or ConsoleKit2) can be contacted or if the login manager doesn't emit the required signal upon exiting sleep mode.

And regarding your original comment:  Yes, it's true, the Clock Applet might not instantaneously update when you resume the computer from sleep mode, but that is only because the Linux kernel freezes all processes before entering sleep mode, and unfreezes them only after *all* steps in the resume process have been executed.  That's why the mouse pointer is frozen initially too:  The X server is a process like any other, and the kernel freezes it to make handling sleep mode easier.  Imagine saving the state of all applications, *while the applications were still actively running and changing their state!*

**TL;DR:**  Neither issue you mentioned is even a bug or an issue.  If somebody wants to test my claims about the original ConsoleKit, though, feel free.  *I* tried it (as improbable as that sounds), but I know I make mistakes, and I want others to find my mistakes.  If I didn't make mistakes, I would have just merged this changed myself.

Thanks as always.

lukefromdc:
Hoping we can get another review on this

zhuyaliang (approved):
No problem with the code

PR #1362, opened 2023-03-21
panel-run-dialog: Fix infinite loop in remove_parameters

mattarroz:
When a '%' character contained in the `exec` parameter of `remove_parameters` is followed by a character different from U, F, N, D, f, u, d, n, m, i, c, k, or v; an infinite loop causes the dialog to freeze. 

To reproduce this bug, create a menu item with name 'bla' and the command `firefox-esr "%"`. Then open the run dialog and search for 'bla'. After clicking on the 'bla' item, the dialog freezes.

I am not quite sure if the purpose of the function is to remove a sole '%' too. This would make it impossible to run the exact 'bla' item. Anyway, the infinite loop is in the mentioned test case is gone.

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -37,6 +37,7 @@
 #include <errno.h>
 #include <sys/types.h>
 #include <unistd.h>
+#include <regex.h>
 
 #include <glib/gi18n.h>
 #include <gio/gio.h>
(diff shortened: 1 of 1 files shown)


lukefromdc:
What is the status of this? Is it still being worked on?

mattarroz:
> What is the status of this? Is it still being worked on?

Currently not working on it. I am unsure what the original purpose of this (buggy) function is. That is why I am concerned that my fix does not meet those specifications.

lukefromdc:
Closing for now, can reopen if you decide to start working on it again

PR #1370, opened 2023-05-21
Wayland, hidpi: do not scale buttons in wayland

lukefromdc:
Otherwise they are rendered double-sized and offset. Wayland apparently handles this for us

### DIFF
--- a/mate-panel/button-widget.c
+++ b/mate-panel/button-widget.c
@@ -5,6 +5,7 @@
 #include <gdk/gdkkeysyms.h>
 #include <gtk/gtk.h>
 #include <gdk/gdk.h>
+#include <gdk/gdkx.h>
 
 #include "button-widget.h"
 #include "panel-widget.h"
@@ -157,6 +158,8 @@ button_widget_unset_surfaces (ButtonWidget *button)
 static void
 button_widget_reload_surface (ButtonWidget *button)
 {
+    GdkDisplay *display;
+    gint scale;
 	button_widget_unset_surfaces (button);
 
 	if (button->priv->size <= 1 || button->priv->icon_theme == NULL)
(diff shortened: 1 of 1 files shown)


lukefromdc:
Note that the code style in this file is an ugly mix of tabs and spaces, I had to do my best with that unless we want to convert it to all spaces (in a separate commit) now

mate-panel/button-widget.c
```diff
@@ -370,10 +387,21 @@ button_widget_draw (GtkWidget *widget,
...
-	w = cairo_image_surface_get_width (button_widget->priv->surface) / scale;
-	h = cairo_image_surface_get_height (button_widget->priv->surface) / scale;
-	x = off + (width - w) / 2;
-	y = off + (height - h) / 2;
+	display = gdk_display_get_default ();
+	if  (GDK_IS_X11_DISPLAY (display))
+	{
+		w = cairo_image_surface_get_width (button_widget->priv->surface) / scale;
+		h = cairo_image_surface_get_height (button_widget->priv->surface) / scale;
+		x = off + (width - w) / 2;
```
zhuyaliang, line 395: Can this be placed outside of conditional judgment? This can simplify the code

lukefromdc: No it cannot be: x11 and Wayland treat this differently, with wayland scaling by default and x11 requiring explicit scaling. Otherwise you get buttons offset and cropped in either Wayland or x11, depending on which code block you enable unconditionally.

zhuyaliang: What I mean is that the values of x and y can be placed outside the conditional judgment
```
x = off + (width - w) / 2;
y = off + (height - h) / 2;
```
Of course, it doesn't have much impact, so I agree with this PR

zhuyaliang (approved):
Running well

PR #1372, opened 2023-06-07
Wayland, Tray applet: allow in-process applet to load, using SNI only

lukefromdc:
*We cannot load the legacy system tray in Wayland as that requires xembed. *We CAN load the status-notifier (SNI support)
*Wayland needs us to explicitly set gtk_button_set_always_show_image to TRUE 
*Wayland needs us to explicitly set gtk_image_menu_item_set_always_show_image to TRUE

### DIFF
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -494,11 +494,6 @@ applet_factory (MatePanelApplet *applet,
         strcmp (iid, "SystemTrayApplet") == 0))
     return FALSE;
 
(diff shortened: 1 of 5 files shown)


lukefromdc:
This makes the notification area useful in a wayland session. In particular it brings the ability to use the network manager applet in indicator mode with` nm-applet --indicator` which FULLY works, with the menus rendering in the right place and fully functional with button and menu icons turned on fulltime. The latter as they are needed for the notification icon to appear (it's a button), and also for the menu icons to appear.

Without mate-session-manager running (we need a wayland session down the road), gtk defaults to not rendering icons in buttons or menus, which means indicator icons won't appear at all. That took me over a week to find

PR #1377, opened 2023-06-27
Clock applet/wayland: position calendar window same as in x11

lukefromdc:


lukefromdc:
Had to clean up indents, but could NOT find the cause of repeated travis build failures. Mate-panel with wayland enabled already requires gtk-layer-shell, I made similar revisions to the clock's makefile.am as in wncklet, but while my local build found the included symbols from gtk-layer-shell.h the travis builds did not.

What am I missing here?

raveit65:
Not sure about, but it seems that clang complains about a compiler error.
https://app.travis-ci.com/github/mate-desktop/mate-panel/jobs/604956193#L3881
```
clang: error: linker command failed with exit code 1 (use -v to see invocation)

make[5]: *** [Makefile:822: clock-applet] Error 1
```
In CI we are using clang to compile the packages.
https://app.travis-ci.com/github/mate-desktop/mate-panel/jobs/604956193#L1467
`compiler:                      /usr/share/clang/scan-build-14/bin/../libexec/ccc-analyzer`
I guess at home you're using `gcc`. For this reason you didn't run in this compiler error.
Here the compiler use this options
https://app.travis-ci.com/github/mate-desktop/mate-panel/jobs/604956193#L1485
```
+ scan-build -enable-checker deadcode.DeadStores -enable-checker alpha.deadcode.UnreachableCode -enable-checker alpha.core.CastSize -enable-checker alpha.core.CastToStruct -enable-checker alpha.core.IdenticalExpr -enable-checker alpha.core.SizeofPtr -enable-checker alpha.security.ArrayBoundV2 -enable-checker alpha.security.MallocOverflow -enable-checker alpha.security.ReturnPtrRange -enable-checker alpha.unix.SimpleStream -enable-checker alpha.unix.cstring.BufferOverlap -enable-checker alpha.unix.cstring.NotNullTerminated -enable-checker alpha.unix.cstring.OutOfBounds -enable-checker alpha.core.FixedAddr -enable-checker security.insecureAPI.strcpy --keep-cc --use-cc=clang --use-c++=clang++ -o html-report make -j 2
```

But, i 've no idea to fix it. Sorry

cwendling:
The reason is simple: clock is not built in-process (https://app.travis-ci.com/github/mate-desktop/mate-panel/jobs/604956193#L1472), yet the Wayland flags are only set in such a case (see diff), which kind of makes sense as out-of-process doesn't work there.  So I guess you need an extra check somewhere only to enable wayland support for in-process applets.

applets/clock/clock.c
```diff
@@ -901,8 +906,74 @@ position_calendar_popup (ClockData *cd)
         int             i, n;
         gboolean        found_monitor = FALSE;
 
+
         if (!GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
-                return;
+        {
```
cwendling, line 911: This should be behind `GDK_IS_WAYLAND_DISPLAY` rather.  And not inside a `HAVE_X11` section, should it?

lukefromdc: It's actually a NOT X11 selector., but I can try switching that over, would be an needed if ever build without x11 at all

lukefromdc: GDK_IS_WAYLAND_DISPLAY worked fine, will push

applets/clock/clock.c
```diff
@@ -901,8 +906,74 @@ position_calendar_popup (ClockData *cd)
         int             i, n;
         gboolean        found_monitor = FALSE;
 
-        if (!GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
-                return;
+
+        if (GDK_IS_WAYLAND_DISPLAY (gdk_display_get_default ()))
```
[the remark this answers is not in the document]
cwendling: and if the X11 and Wayland code don't share much logic, you could have helpers for each so the body of that one is just the selector calling the right one… just a thought, because it gets big.

raveit65:
I guess we need `--with-in-process-applets=all` configure option in build.yml. https://github.com/mate-desktop/mate-panel/blob/master/.build.yml#L143

raveit65:
Ok, but how you will test the code of this PR if not build in-process.
And yes, using x11 with in-process or out-of-process should be possible.

Offtopic on:
Since i am using mate-media sound applet build in-process on my production daily used system, the panel crashed very often. That's the downside if all applets are build in-process.
There are some open issues from media-applet which needs to be fixed before using the applet build -in-process.
Also, the sound part from mate-settings-daemon is involved.
Currently i have the feeling my MATE desktop isn't stable anymore.
Well, i will open a report when i have actual stack-traces.
So i am thinking it might be a good option to have a configure option --with-in-process-applets for the other applet packages.
offtopc off

cwendling:
> Ok, but how you will test the code of this PR if not build in-process. And yes, using x11 with in-process or out-of-process should be possible.

I mean having more test jobs, one doing the in-process, the other the out-of-process.  And they should both succeed, if all is well

> Offtopic on: Since i am using mate-media sound applet build in-process on my production daily used system, the panel crashed very often. That's the downside if all applets are build in-process. There are some open issues from media-applet which needs to be fixed before using the applet build in-process. Also, the sound part from mate-settings-daemon is involved. Currently i have the feeling my MATE desktop isn't stable anymore. Well, i will open a report when i have actual stack-traces. So i am thinking it might be a good option to have a configure option --with-in-process-applets for the other applet packages. offtopc off

Yeah that's the risk, more code is in the same process, more risk of bugs… and possibly the applets have more bug in-process because the environment is not exactly the same and they do things they shouldn't.

Another solution would be to create a whole new out-of-process mechanism, like a custom protocol (DBus I guess) for applets to expose what they need.  The downside is that it's more complex, and either weird or a lot more restricted as to what an applet could do (I don't see how e.g. the system monitor applet would work but if it uploads a frame -- still doable, but a lot more tricky).
But the advantage is obviously that less "random" (3rd party) code runs in the panel process, so less likely to mess things up.  that's just a thought though, because it's complex and I'm not sure if it's worth it.

lukefromdc:
More help on these would be a benefit, as my understanding of the build system is still quite limited. For this one we still support building in or out of process (the c code differences are minimal as can be seen in my mate-applets work), so I will try to put wayland enablement behind an in-process switch if that has already been done for wncket to give me a template for it.

raveit65:
Before this PR it was possible to build all applets (including clock)  in-process or out-of-process with this configure flag `--with-in-process-applets=all/none`.
In my understanding you don't need change anything in clock applet code to build in process or out-of-process.
The configure flag should do the job, ihmo.

raveit65:
See https://github.com/mate-desktop/mate-panel/blob/master/applets/clock/Makefile.am#L58
```
if CLOCK_INPROCESS
.
.
.
.
```

raveit65:
Hmm, now build fails in the `make distcheck` part in afterscripts.
https://github.com/mate-desktop/mate-panel/blob/master/.build.yml#L164
Normal compile part works fine.

raveit65:
Ok, i found the `DISTCHECK_CONFIGURE_FLAGS` https://github.com/mate-desktop/mate-panel/blob/master/Makefile.am#L14
```
DISTCHECK_CONFIGURE_FLAGS = \
	--enable-gtk-doc \
	--disable-introspection \
	--enable-compile-warnings=no \
	CFLAGS='-Wno-deprecated-declarations'
```

lukefromdc:
I just completely split the x11 and wayland code paths, limiting the wayland part to in-process builds. Got local successful builds out of process, in-process with wayland disabled,. and in-process with wayland enabled

lukefromdc:
I found out of process builds broken prior to that change, not sure when in this work I broke that but it's fixed now. The last problem was a closing bracket inside instead of outside the` #endif` statements to limit the wayland code to wayland/in-process builds

lukefromdc:
At the moment, building a new out of process mechanism is beyond my capacity

raveit65:
Found out that the build failure in `make distcheck` part exists also in master branch.
In both branches (**master and wayland-calendar**) this command failed to run.
`./autogen.sh --with-in-process-applets=all --enable-x11 --enable-wayland && make && make distcheck`
But this command runs fine.
`./autogen.sh --with-in-process-applets=none --enable-x11 --enable-wayland && make && make distcheck`

When adding `--with-in-process-applets=all` to `distcheck`
```
DISTCHECK_CONFIGURE_FLAGS = \
	--enable-gtk-doc \
	--disable-introspection \
	--enable-compile-warnings=no \
	--with-in-process-applets=all \
	CFLAGS='-Wno-deprecated-declarations'
```
both previous commands are running fine :)
I think it is OK to add the `--with-in-process-applets=all` only to `distcheck` as the main usage of `distcheck` is to build our tarballs for releasing.
And we see here in CI that the new code compiles fine with both options.
Or fixing the distcheck error for both build cases, which seems complicate.
What do you think?

cwendling:
@raveit65 distcheck failing usually means files are missing in the generated tarball.  So I don't think changing the flags is teh way to go, but we should find out which files are missing, and include them (either using the `dist_` prefix, or adding them to `EXTRA_DIST` or alike)

raveit65:
@cwendling 
https://app.travis-ci.com/github/mate-desktop/mate-panel/jobs/605019674#L4129
I agree, but `org.mate.panel.applet.ClockAppletFactory.service.in` is missing for distcheck (without in-process) only when compiling (with in-process) before distcheck was done.
That means distcheck have to detect which option was used before during the normal `make` command.
Honestly, i have no idea how should that work :)
Also, i am in worried that we see the problem with the other applets too.

Anyway, i will check if a tarball generated with in-process does work for a out-of-process build,
to see if files are missing.

raveit65:
Ok, you're right. My idea doesn't work.
With in-process in distcheck and generating a tarball with `./autogen.sh --with-in-process-applets=all --enable-x11 --enable-wayland && make && make distcheck`, building from the tarball with `./configure --with-in-process-applets=none --enable-x11 --enable-wayland && make` fails :\

cwendling:
Yeah I'm not surprised.  I'll try and take a look in a bit

cwendling:
@raveit65 see #1378

lukefromdc:
This work can take all the time it needs to get it right. My hope would be to have an basic wayland session that looks and feels like the x11 session available in MATE 1.28, and maybe have one distros can treat as a default desktop by 1.32 or so, similar to the GTK 3 transition except for never dropping x11 suppport. 

I believe we need to be ready no matter what comes out of the x11 vs wayland contest over the years.

raveit65:
> This work can take all the time it needs to get it right. My hope would be to have an basic wayland session that looks and feels like the x11 session available in MATE 1.28, and maybe have one distros can treat as a default desktop by 1.32 or so, similar to the GTK 3 transition except for never dropping x11 suppport.
> 
> I believe we need to be ready no matter what comes out of the x11 vs wayland contest over the years.

When we have a session file for MATE-Layer-Shell which makes it possible to start the session from desktop-manger login, similar what we use for x11, i am happy to start with both sessions for 1.28 in fedora. Same what gnome did with fc15 in May 24, 2011 ;)

lukefromdc:
That is exactly my objective for this work: For MATE to have a desktop session in wayland just as GNOME and KDE do. I believe that will be necessary going forward-and it's looking like wayland supports lighter code and may itself be lighter as well though correct me if I am wrong about that. The kludges necessary for wayland (and there are many) still seem less numerous than the kludges needed in x11 where scaling and desktop resizing create so many hassles

applets/clock/clock.c
```diff
@@ -987,6 +996,80 @@ position_calendar_popup (ClockData *cd)
...
+    if (GDK_IS_WAYLAND_DISPLAY (gdk_display_get_default ()))
+    {
+        GtkWindow *window;
+        GdkWindow *panelwin;
+        GtkWidget *toplevel;
+        int  x, y, w, h, panel_w, panel_h;
+
+        /*Get the calendar window dimensions*/
+        window = (GTK_WINDOW (cd->calendar_popup));
+        gtk_window_get_size(window, &w, &h);
```
raveit65, line 1014: Can you please update code format? .......using a blank before a bracket?
This looks very wild in several places here :)

lukefromdc: No problem, will get on that now

lukefromdc: Done

raveit65: thx

raveit65 (approved):
LGTM

applets/clock/clock.c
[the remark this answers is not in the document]
lukefromdc: `#ifdef HAVE_X11` compile time selectors have been added to the x11 portion of the code as well as a runtime 
`if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))` selector. Therefore, if the panel is built with only x11 or with only wayland support (allowed since the initial wayland commits a while ago), only the relevent code is built. In builds supporting both, only the x11 or wayland code is selected at runtime, so this has been implemented.

Thanks for your work and your help with the wayland support, I appreciate it

applets/clock/clock.c
lukefromdc: We now have both x11 and wayland ifdefs as well as the runtime selectors, as advised

lukefromdc:
Does anyone else want to further review this or should I proceed to merge it?

applets/clock/clock.c
```diff
@@ -51,6 +51,13 @@
 #include <gdk/gdkkeysyms.h>
 #include <gio/gio.h>
 
+#ifdef CLOCK_INPROCESS
+#ifdef HAVE_WAYLAND
```
cwendling, line 55: could be `#if defined(CLOCK_INPROCESS) && defined(HAVE_WAYLAND)` to only have one block instead of nesting two

applets/clock/clock.c
```diff
@@ -889,6 +896,8 @@ static void
 position_calendar_popup (ClockData *cd)
 {
 #ifdef HAVE_X11
```
cwendling, line 898: Ideally this would be `#if defined(HAVE_X11) && defined(GDK_WINDOWING_X11)`
Maybe a bit nitpicky though

applets/clock/clock.c
```diff
@@ -987,6 +996,80 @@ position_calendar_popup (ClockData *cd)
 
         gtk_window_move (GTK_WINDOW (cd->calendar_popup), x, y);
         gtk_window_set_gravity (GTK_WINDOW (cd->calendar_popup), gravity);
+    }
+#endif
+    /*Only build wayland support when building in process*/
+#ifdef CLOCK_INPROCESS
+#ifdef HAVE_WAYLAND
```
cwendling, line 1003: Same here, ideally `#if defined(CLOCK_INPROCESS) && defined(HAVE_WAYLAND) && defined(GDK_WINDOWING_WAYLAND)`

applets/clock/clock.c
```diff
@@ -987,6 +996,80 @@ position_calendar_popup (ClockData *cd)
...
+
```
cwendling, line 1004: nitpick: this blank line should be between the X11/Wayland blocks, e.g. be line 1001

applets/clock/clock.c
```diff
@@ -987,6 +996,80 @@ position_calendar_popup (ClockData *cd)
...
+        toplevel = gtk_widget_get_toplevel (cd->applet);
+        panelwin = gtk_widget_get_window (toplevel);
+        gdk_window_get_geometry (panelwin, NULL, NULL, &panel_w, &panel_h);
+
+        /*Set up GTK Layer Shell*/
+        gtk_layer_init_for_window (window);
+        gtk_layer_set_layer (window, GTK_LAYER_SHELL_LAYER_TOP);
+
+        switch (cd->orient) {
+            case MATE_PANEL_APPLET_ORIENT_RIGHT:
```
cwendling, line 1029: style: case label indent

applets/clock/clock.c
```diff
@@ -987,6 +996,80 @@ position_calendar_popup (ClockData *cd)
...
+            gtk_layer_set_anchor (window,GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
```
cwendling, line 1030: style: space after comma (there are many of those)

lukefromdc: Done, I am assuming you mean use a space after each comma as in the rest of the file. That got missed probably on first cut and paste that generated one of these lines, then copied into all the others

applets/clock/clock.c
lukefromdc: Caught and fixed

applets/clock/clock.c
lukefromdc: selectors updated and they worked on a new build

applets/clock/clock.c
```diff
@@ -51,6 +51,12 @@
 #include <gdk/gdkkeysyms.h>
 #include <gio/gio.h>
 
+#if defined CLOCK_INPROCESS && defined HAVE_WAYLAND
```
cwendling, line 54: style: as you want, but I'm used to `defined(x)` rather than `defined x` and you use the former below

applets/clock/clock.c
```diff
@@ -888,7 +894,9 @@ create_calendar (ClockData *cd)
 static void
 position_calendar_popup (ClockData *cd)
 {
-#ifdef HAVE_X11
+#if defined(HAVE_X11) && defined(GDK_WINDOWING_X11)
+    if (GDK_IS_X11_DISPLAY (gdk_display_get_default ()))
+    {
```
cwendling, line 898: Remove lines 912-914 below, as they are now redundant

applets/clock/clock.c
```diff
@@ -987,6 +995,78 @@ position_calendar_popup (ClockData *cd)
...
+        GdkWindow *panelwin;
+        GtkWidget *toplevel;
+        int  x, y, w, h, panel_w, panel_h;
+
+        /*Get the calendar window dimensions*/
+        window = (GTK_WINDOW (cd->calendar_popup));
+        gtk_window_get_size (window, &w, &h);
+        /*Find the position of the applet*/
+        gdk_window_get_origin (gtk_widget_get_window (cd->panel_button),
+                                                                      &x, &y);
```
cwendling, line 1014: style: alignment

applets/clock/clock.c
```diff
@@ -987,6 +995,78 @@ position_calendar_popup (ClockData *cd)
...
+                {
+                    gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
+                    gtk_layer_set_margin (window, GTK_LAYER_SHELL_EDGE_LEFT, x);
+                }
+            else
+                gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
+            break;
+        }
+    return;
+	}
```
cwendling, line 1068: Indentation should be 4 spaces, not 1 tab

applets/clock/clock.c
```diff
@@ -51,6 +51,12 @@
 #include <gdk/gdkkeysyms.h>
 #include <gio/gio.h>
 
+#if defined CLOCK_INPROCESS && defined HAVE_WAYLAND
+
```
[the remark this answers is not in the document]
lukefromdc: removed

cwendling (changes requested):
clock.c actually currently use 8 spaces for indentation, the changes here adds 4-spaces stuff.  I don't know whether we care or not, I'll leave that to others.

otherwise apart from the small comments (only important-ish one is the duplicate `GDK_IS_X11_DISPLAY`) looks good -- but again, not tested beyond compiling

lukefromdc:
The problem with using 8 spaces is twofold: One is that some of this code is deeply nested and takes a lot of space on screen when 8 spaces are used. The other is that as originally written the file used 8 spaces some places, 4 in others. In most new code in MATE we use four spaces. Some of the old code is 4 spaces, some of it is 8 spaces, some is tabs, and worst of all some of it is a mix of tags and spaces.

I'll take out that tab but prefer to keep the 4 spaces unless there is NO other 4 space indented code anywhere in that file.

raveit65:
When i found time i will fix the tab-size dilemma in a new PR, or replace them with spaces.
I did that already for some files in mate-panel or caja a while ago, but we have a lot of files with mixed tap-size.
2 or 3 different tab-sizes in a file isn't nice when you reading code. 
For that reason i prefer using only spaces.
Anyway, using 8 spaces for a tab is too large.

cwendling:
> […] All the same, on closer examination all the rest of this particular file is 8 space indents. Will change this

Discrepancies from 8-spaces I can see is this (ignoring your changes):
```diff
diff --git a/applets/clock/clock.c b/applets/clock/clock.c
index 4b203e1f..30fea391 100644
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -317,16 +317,16 @@ clock_set_timeout (ClockData *cd,
                         timeouttime = ((999 - itime_ms % 1000) * 86.4) / 100 + 1;
                 }
         } else {
-                 struct timeval tv;
+                struct timeval tv;
 
                 gettimeofday (&tv, NULL);
-                 timeouttime = (G_USEC_PER_SEC - tv.tv_usec)/1000+20;
+                timeouttime = (G_USEC_PER_SEC - tv.tv_usec)/1000+20;
 
                 /* timeout of one minute if we don't care about the seconds */
-                 if (cd->format != CLOCK_FORMAT_UNIX &&
+                if (cd->format != CLOCK_FORMAT_UNIX &&
                     !cd->showseconds &&
                     (!cd->set_time_window || !gtk_widget_get_visible (cd->set_time_window)))
-                         timeouttime += 1000 * (59 - now % 60);
+                        timeouttime += 1000 * (59 - now % 60);
          }
 
         cd->timeout = g_timeout_add (timeouttime,
@@ -452,7 +452,7 @@ get_updated_timeformat (ClockData *cd)
         use_lctime = (env_language != NULL) && (env_lc_time != NULL) && (g_strcmp0 (env_language, env_lc_time) != 0);
 
         if (use_lctime) {
-            g_setenv("LANGUAGE", env_lc_time, TRUE);
+                g_setenv("LANGUAGE", env_lc_time, TRUE);
         }
 
         if (cd->format == CLOCK_FORMAT_12)
@@ -497,7 +497,7 @@ get_updated_timeformat (ClockData *cd)
 
         /* Set back LANGUAGE the way it was before */
         if (use_lctime) {
-            g_setenv("LANGUAGE", env_language, TRUE);
+                g_setenv("LANGUAGE", env_language, TRUE);
         }
 
         result = g_locale_from_utf8 (clock_format, -1, NULL, NULL, NULL);
@@ -707,7 +707,7 @@ update_tooltip (ClockData * cd)
          * This can prevent problems with OpenGL on some drivers */
         old_tip = gtk_widget_get_tooltip_text (cd->panel_button);
         if (g_strcmp0 (old_tip, tip))
-            gtk_widget_set_tooltip_text (cd->panel_button, tip);
+                gtk_widget_set_tooltip_text (cd->panel_button, tip);
 
         g_free (old_tip);
         if (!cd->showdate)
@@ -2518,9 +2518,9 @@ parse_gsettings_cities (ClockData *cd, gchar **values)
         context = g_markup_parse_context_new (&location_parser, 0, &data, NULL);
 
         if (values) {
-            for (i = 0; values[i]; i++) {
-                    g_markup_parse_context_parse (context, values[i], strlen(values[i]), NULL);
-            }
+                for (i = 0; values[i]; i++) {
+                        g_markup_parse_context_parse (context, values[i], strlen(values[i]), NULL);
+                }
         }
 
         g_markup_parse_context_free (context);
@@ -2957,13 +2957,13 @@ location_changed (GObject *object, GParamSpec *param, ClockData *cd)
 static void
 location_name_changed (GObject *object, ClockData *cd)
 {
-    location_update_ok_sensitivity (cd);
+        location_update_ok_sensitivity (cd);
 }
 
 static void
 location_timezone_changed (GObject *object, GParamSpec *param, ClockData *cd)
 {
-    location_update_ok_sensitivity (cd);
+        location_update_ok_sensitivity (cd);
 }
 
 static void
```

So pretty much consistent but for very specific places.

> Anyway, using 8 spaces for a tab is too large.

Agreed, but I think patches should be consistent, and if indentation is to be changed that it's the *only* thing a patch changes.  Actually the PR here is ever so slightly confusing because of this: it adds an outer block (`if (display is X11) {…}`) but doesn't change indentation of the body.  Things like that are not a big issue, but can make understanding changes harder and reading the resulting file a tad harder as well.

lukefromdc:
I respaced it to 8 spaces-but saved an older version of the file in case people didn't like the results.

applets/clock/clock.c
lukefromdc: Well, they are both stated to be valid code

lukefromdc:
Question for both reviewers (note that respacing a lot of lines is time consuming): stay at a consistant 8 spaces or revert back to 4 spaces? I still have the old version of that file so can revert to 4 spaces.

cwendling:
> Can also be easily reverted if you all want to save that for a later PR

Yes *please*, this makes the review very painful because you have to at best ignore the given commit, and then verify it separately.  IMO this should really be another PR.

lukefromdc:
Done, the 4 space reformatting is gone.

lukefromdc:
rebased after https://github.com/mate-desktop/mate-panel/commit/5a7f770082568453f3a5f271a274156f08352101

applets/clock/clock.c
```diff
@@ -888,105 +893,181 @@ create_calendar (ClockData *cd)
...
+                        }
+                        else
+                                gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
 
-        gtk_window_move (GTK_WINDOW (cd->calendar_popup), x, y);
-        gtk_window_set_gravity (GTK_WINDOW (cd->calendar_popup), gravity);
+                        break;
+                }
+                return;
+        }
```
raveit65, line 1070: Line 1070 starts with a tab.

lukefromdc:
Thanks, I'll clean that up too when I get home. Pluma does NOT do a good job of flagging this, and not spoting the little things can make little problems that grow up to become big problems.

raveit65:
@lukefromdc 
Pluma got an option to show tabs and spaces, very useful. Otherwise i wouldn't see that in resulting patch for rpm building.
pluma_tab_spaces

Btw. i have wayfire running on bare metal, i can start testing all the wayland PRs now.
Sadly in qemu VM, wayfire or gtk-layer-shell has glitches. Nothing fixed since my first tests with mate-panel some years ago. That means i have to logout from my regular x11 session, not comfortable but ok.

lukefromdc:
I can switch to VT2 /3/whatever and run dbus-launch wayfire from there. You need dbus-launch for a lot of the MATE programs

raveit65:
Thanks for the tip.
Most applets from mate-panel and mate-applets are working more or less. I will post more details in mate-appelts PR.
But the icons from panel-launchers are blurred in wayland.
wayland:
mate-panel_wayfire_blurred_panel_launcher

compare to my x11 panel (HIDPI):
Bildschirmfoto zu 2023-07-02 15-54-25

raveit65:
Clock-applet works fine under wayfire. The window is exact at the same position like in X11 and behavior is the same.
Good job.

applets/clock/clock.c
lukefromdc: No tab found here, select in in browser here or in the file and you can move the cursor to each position, with a tab it jumps. That is not true with tabs. Not sure what's going on with that

lukefromdc:
thanks for the tip about showing tabs and spaces, that will save me time and errors. Here's what I got, wonder if you were looking at an older version or something?

clock-c-spaces

raveit65:
I downloaded the resuting patch from PR with wget and i see the tab. Try
`wget https://patch-diff.githubusercontent.com/raw/mate-desktop/mate-panel/pull/1377.patch`

lukefromdc:
Only thing I can try is to redo the spaces there and force-push again. Strange, I just cloned the repo to the laptop from github though, this is odd

lukefromdc:
When I retyped the spaces, git gui showed no change at all. Force-pushed anyway, if you still get a tab I have no idea what's causing that or how to fix it

applets/clock/clock.c
raveit65: This is weird, when look in clock.c local in your branch i see no tab.
If there is really a mysterious tab after merging we can fix it later.
No need to spend time on it.

raveit65:
Forget this weird tab issue.

raveit65 (approved):
LGTM

applets/clock/clock.c
```diff
@@ -888,105 +893,181 @@ create_calendar (ClockData *cd)
...
+                        break;
+                case MATE_PANEL_APPLET_ORIENT_UP:
+                        gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_BOTTOM, TRUE);
+                        if (x < (panel_w -w))
+                        {
+                                gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
+                                gtk_layer_set_margin (window, GTK_LAYER_SHELL_EDGE_LEFT, x);
+                        }
+                        else
+                                gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
```
cwendling, line 1065: Won't you have a problem similar to [the one in mate-media] when moving and re-orienting panels? (not tested though)

applets/clock/clock.c
```diff
@@ -888,105 +893,181 @@ create_calendar (ClockData *cd)
...
+                gdk_window_get_geometry (panelwin, NULL, NULL, &panel_w, &panel_h);
+
+                /*Set up GTK Layer Shell*/
+                gtk_layer_init_for_window (window);
+                gtk_layer_set_layer (window, GTK_LAYER_SHELL_LAYER_TOP);
+
+                switch (cd->orient) {
+                case MATE_PANEL_APPLET_ORIENT_RIGHT:
+                        gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
+                        if (y < (panel_h -h))
```
cwendling, line 1026: style: space around the `-` binary operator

applets/clock/clock.c
```diff
@@ -888,105 +893,181 @@ create_calendar (ClockData *cd)
...
+                        }
+                        else
+                                gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_BOTTOM, TRUE);
 
-                break;
-        }
+                        break;
+                case MATE_PANEL_APPLET_ORIENT_DOWN:
+                        gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_TOP, TRUE);
+                        if (x < (panel_w -w))
```
cwendling, line 1048: style: space around the `-` binary operator

applets/clock/clock.c
```diff
@@ -888,105 +893,181 @@ create_calendar (ClockData *cd)
...
+                                gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_LEFT, TRUE);
+                                gtk_layer_set_margin (window, GTK_LAYER_SHELL_EDGE_LEFT, x);
+                        }
+                        else
+                                gtk_layer_set_anchor(window, GTK_LAYER_SHELL_EDGE_RIGHT, TRUE);
+
+                        break;
+                case MATE_PANEL_APPLET_ORIENT_UP:
+                        gtk_layer_set_anchor (window, GTK_LAYER_SHELL_EDGE_BOTTOM, TRUE);
+                        if (x < (panel_w -w))
```
cwendling, line 1059: style: space around the `-` binary operator

cwendling (commented):
Minor nitpicks but possibly for the orientation thing.  Again, I didn't actually run this under Wayland myself though.

lukefromdc:
I tested moving panels containing the clock, and moving the clock applet around, and didn't ever have a problem with the calendar getting "stretched" between the new and old location. This would be expected behavior if the calendar window is made anew each time it is shown. If it is not, adding the same fix from mate-media would ensure against problems down the road.

applets/clock/clock.c
lukefromdc: I tested for that specifically and had no trouble, but will apply the fix from mate-media in case someone else gets different behavior. Not sure if the calendar window is made anew each time (the volume slider is just hidden and shown) or not

lukefromdc: Setting the unused anchors to FALSE for code durability/bug prevention reasons

applets/clock/clock.c
lukefromdc: Found and fixed

applets/clock/clock.c
lukefromdc: found and fixed again

applets/clock/clock.c
lukefromdc: On further testing, I found that there was indeed a "stretching" issue-but only if the applet was moved or the panel orientation changed while the calendar is showing. No change I tried stopped it in once situation without creating it in another situation, so that will need more experimentation. Leaving that part of the code as it was for now, but applying the style fixes for readability.

lukefromdc:
Latest commit cleans up style. The "stretching" issue on orientation change turns out to exist if and only if the clock applet is moved while the calendar is showing in wayland, implying a new calendar window is being made each time it is shown. Simple fixes for that didn't work. Not sure how big a problem that is but I'd like to find a fix

applets/clock/clock.c
lukefromdc: Just found the top style fix mentioned here, missed it when fixing the others

lukefromdc:
Latest commit ensure the calendar window can never be "stretched" across the screen no matter how you move the panel or applet even with the calendar window showing. When the calendar window is closed before moving the applet or panel there has never been a problem with incorrect rendering. 

While the calendar window can sometimes render in the wrong position along the panel if the applet or panel are moved with the calendar window showing, it will always stay adjacent to the same panel the applet is on, and never go offscreen. We do get a runtime warning if and only if the panel or applet is moved with the calendar window showing, as this causes gtk-layer-shell settings to be changed on a widget that is mapped. Not sure how to fix that without destroying and remaking the calendar window. Running `gtk_widget_unrealize` before setting anchors and `gtk_widget_realize()` afterwards sometimes gives clean results, but sometimes causes a segfault. The behavior here though is considerably better than in x11, this is a case the calendar window probably was NOT designed to deal with

If the panel carrying the clock or the clock applet is moved with the calendar window showing in x11 (x11 codepath should be unchanged here) it always stays in the previous location. If just the applet is moved in x11 with the calendar showing, the calendar window  moves with the applet but does not change orientation. That means moving the clock from a bottom to a top panel with the calendar showing in x11 reliably moves the calendar offscreen, with the visible edge over the applet showing it still exists.

Fixing that in x11 is a problem for another PR, the wayland behavior is now better than the x11 behavior.

applets/clock/clock.c
lukefromdc: Those have been changed to use the parentheses

applets/clock/clock.c
lukefromdc: Moved that code back to line up with the first letter in the code inside the opening parenthesis, as is common elsewhere in the file

lukefromdc:
A couple of conversations on outdated code cannot be resolved and will have to be bypassed no matter what we do

cwendling (approved):
Still not tested on Wayland, but LGTM now

PR #1378, opened 2023-06-28
Fix distribution of applet service files

cwendling:
Fix distribution of the applet service files sources when building from an in-process-configured build.  This fixes `make distcheck` and tarballs in builds configured with `--with-in-process-applets=all` (or actually any in-process applet).

The issue was that the .service.in file was referenced indirectly in EXTRA_DIST, and the variable listed there was only filled in a conditional.  Automake is not able to resolve this and see it still should distribute the reference of that variable, even if set conditionally.  Fix this by always setting the variable properly, even if it's only used for out-of-process builds.


raveit65 (approved):
LGTM

PR #1388, opened 2023-07-20
wayland: Fix overflowing window list and ensure it can be themed

lukefromdc:
*Allows window buttons to dynamically resize themselves as the window list fills up instead of pushing all the other applets to the end of the panel. 

*add the #tasklist-button widget name to tasklist buttons in wayland, it was only being applied in x11. This makes it much easier to theme


lukefromdc:
Note that it is still possible to overflow the window list with 20+ buttons, but we don't yet have a  way to implement per-workspace anything in wayland that I know of.  While it is not difficult to hard-limit the number of buttons, that would cause users to minimize windows and not be able to restore them, which is worse. Before this as few as 5 open windows would start pushing the other applets to the panel edges as button width in wayland was fixed.

raveit65:
Confirmed the weird issue.

raveit65 (approved):
LGTM,
fixes the issue and applet works fine now in wayland.

raveit65:
Easy change, here we go.

PR #1389, opened 2023-07-21
Wayland: Fix blurry rendering of panel launcher icons

lukefromdc:
Icons present in the icon theme are now handled by gtk_icon_theme_load_surface() so they render clear and sharp. This is basically all of the icons that will be in the menus and be draggable to the panel, and anything from /usr/share/applications dragged to the panel as well.

A custom icon (I use one for firefox runnng in RAM) not present in the icon theme and instead loaded from a file outside the normal icon directories is not findable by gtk_icon_theme_load_surface(), so the previous code was retained as a fallback to prevent such icons from disappearing. Such custom icons thus won't render crisply but all other will.

Very clear and sharp in x11 too

### DIFF
--- a/mate-panel/button-widget.c
+++ b/mate-panel/button-widget.c
@@ -33,6 +33,7 @@ struct _ButtonWidgetPrivate {
 	guint             ignore_leave  : 1;
 	guint             arrow         : 1;
 	guint             dnd_highlight : 1;
+	gboolean          needs_move;
 };
 
 static void button_widget_icon_theme_changed (ButtonWidget *button);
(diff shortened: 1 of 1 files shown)


mate-panel/button-widget.c
```diff
@@ -158,50 +159,67 @@ button_widget_unset_surfaces (ButtonWidget *button)
...
-		if  (GDK_IS_X11_DISPLAY (display))
+		GtkIconTheme *icon_theme = gtk_icon_theme_get_default();
+		button->priv->surface =
+			gtk_icon_theme_load_surface (icon_theme,
+				button->priv->filename,
+				button->priv->size,
+				scale,
+				NULL,
+				GTK_ICON_LOOKUP_FORCE_SIZE | GTK_ICON_LOOKUP_FORCE_SVG,
+				NULL);
```
raveit65, line 184: Indents
```suggestion
			gtk_icon_theme_load_surface (icon_theme,
										button->priv->filename,
										button->priv->size,
										scale,
										NULL,
										GTK_ICON_LOOKUP_FORCE_SIZE | GTK_ICON_LOOKUP_FORCE_SVG,
										NULL);
```

raveit65: Ok, forget this sugestions.
Indents are not correct with code change. But using tabs in github suggestion function is a mess and looks weird.

mate-panel/button-widget.c
```diff
@@ -158,50 +159,67 @@ button_widget_unset_surfaces (ButtonWidget *button)
...
 		{
-		    button->priv->surface =
-			    panel_load_icon (button->priv->icon_theme,
+			button->priv->surface =
+				panel_load_icon (button->priv->icon_theme,
 					 button->priv->filename,
 					 button->priv->size * scale,
 					 (button->priv->orientation & PANEL_VERTICAL_MASK)   ? button->priv->size * scale : -1,
 					 (button->priv->orientation & PANEL_HORIZONTAL_MASK) ? button->priv->size * scale: -1,
 					 &error);
```
raveit65, line 195: Indents

mate-panel/button-widget.c
```diff
@@ -158,50 +159,67 @@ button_widget_unset_surfaces (ButtonWidget *button)
...
-							       NULL,
-							       GTK_ICON_LOOKUP_FORCE_SVG | GTK_ICON_LOOKUP_USE_BUILTIN,
-							       NULL);
+			button->priv->surface =
+				gtk_icon_theme_load_surface (icon_theme,
+					"image-missing",
+					GTK_ICON_SIZE_BUTTON,
+					scale,
+					NULL,
+					GTK_ICON_LOOKUP_FORCE_SVG | GTK_ICON_LOOKUP_USE_BUILTIN,
```
raveit65, line 221: Indents

raveit65:
This works great. The icons from launchers and panel-action-buttons looking real sharp with HIDPI now, like in X11.
Nice job.
So, the fix was to use `gtk_icon_theme_load_surface` for loading the icons directly from theme? 

@mate-desktop/core-team 
Can this be tested/reviewed as this affects also x11?

mate-panel/button-widget.c
raveit65, line 195: Indents

mate-panel/button-widget.c
```diff
@@ -158,50 +159,67 @@ button_widget_unset_surfaces (ButtonWidget *button)
...
-					 &error);
+			/*fallback to catch the case of custom icons not in x11*/
+			button->priv->needs_move = TRUE;
+			button->priv->surface =
+				panel_load_icon (button->priv->icon_theme,
+					button->priv->filename,
+					button->priv->size * scale,
+					(button->priv->orientation & PANEL_VERTICAL_MASK)   ? button->priv->size  : -1,
+					(button->priv->orientation & PANEL_HORIZONTAL_MASK) ? button->priv->size  : -1,
+					&error);
```
raveit65, line 207: Indents

mate-panel/button-widget.c
```diff
@@ -158,50 +159,67 @@ button_widget_unset_surfaces (ButtonWidget *button)
...
-							       GTK_ICON_LOOKUP_FORCE_SVG | GTK_ICON_LOOKUP_USE_BUILTIN,
-							       NULL);
+			button->priv->surface =
+				gtk_icon_theme_load_surface (icon_theme,
+					"image-missing",
+					GTK_ICON_SIZE_BUTTON,
+					scale,
+					NULL,
+					GTK_ICON_LOOKUP_FORCE_SVG | GTK_ICON_LOOKUP_USE_BUILTIN,
+					NULL);
```
raveit65, line 222: Indents

raveit65:
offtopic on
It isn't a bad idea to add `/usr/libexec/polkit-mate-authentication-agent-` to autostart in wayfire.ini.
This solves opening/using apps which requires privileged rights, ie. cpufreq-applet, etc.
offtopic=off

mate-panel/button-widget.c
lukefromdc: That whole damned file is infested with tabs, and still worse tabs and spaces everywhere. Maybe a separate PR to reformat it to spaces? To get the formatting to where it is on the new code, I had to have Pluma display it with 2 space tabs, 4 space tabs, and 8 space tabs. What a mess leftover from long ago.

I try to avoid changing just a few tabs to spaces when adding new code, because it then looks nasty anytime it is opened with a different tab width

mate-panel/button-widget.c
lukefromdc: I spent an hour last night trying to figure out the best indent format for this. Previous code added when icons were cleared up for x11 mixed tabs and spaces. The previous code didn't provide a very good template, though I could use the same tabs and space mix it used if desired.

I can reformat all of this to spaces only in a separate PR  if needed

mate-panel/button-widget.c
lukefromdc: Same as above: rest of file is a tabby mess and needs reformatting

lukefromdc:
gtk_icon_theme_load_surface indeed fixes this sort of thing and simply too, so long as you do not have the case of loading an icon from outside the icon theme entirely by it's filename. My guess is GTK coded in in such a way that the behind the scenes pixbuf (yes, this is used) is always created at final size and never scaled up from a smaller size.

Supporting icons from user-selected random sources does put some spaghetti in the code, but it doesn't have to be used in any other case. A good way to load from a filename at final size and ensure it never has to be scaled up would allow it's removal without losing a feature.

raveit65:
Np, we can fix formatting for this file in another PR

lukefromdc:
Is this ready to go?

raveit65:
It affects also x11, or not?
In this case i would wait for another review with more coding skills than me.
Any way, if you think this code change is stable-as-a-fuck than merge it :smiley:

raveit65 (approved):
LGTM,
whole the file needs a code-style fix.
I will do this when it is merged.

lukefromdc:
OK, code is merged. File is ready for formatting

lukefromdc:
There should be no reason for trouble in x11, as we either use GTK directly (for all the normal icons), or we fall back to the previous code (for user-selected icons).

PR #1390, opened 2023-07-22
button-widget.c: replace tabs with 4 spaces

lukefromdc:
*and reformat as necessary

### DIFF
--- a/mate-panel/button-widget.c
+++ b/mate-panel/button-widget.c
@@ -57,12 +57,12 @@ G_DEFINE_TYPE_WITH_PRIVATE (ButtonWidget, button_widget, GTK_TYPE_BUTTON)
 static void
 do_colorshift (cairo_surface_t *dest, cairo_surface_t *src, int shift)
 {
-    gint i, j;
-    gint width, height, has_alpha, srcrowstride, destrowstride;
+    gint    i, j;
+    gint    width, height, has_alpha, srcrowstride, destrowstride;
     guchar *target_pixels;
     guchar *original_pixels;
-    int val;
-    guchar r,g,b;
+    int     val;
+    guchar  r,g,b;
 
     has_alpha = cairo_surface_get_content (src) != CAIRO_CONTENT_COLOR;
     width = cairo_image_surface_get_width (src);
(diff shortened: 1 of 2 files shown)


lukefromdc:
One possible way of doing this, I am open to other options as well. Objective is easier to read and maintain code.
This is a reformat of white space only. Should be NO changes to built binary and NO change to anything but a tab, a space, or in a few cases a blank line or an opening bracket position.

A test built and run gave no surprises and identical runtime behavior in both x11 and Wayland, ANY difference would have indicated a typo somewhere. We need to make sure no typos get in with this sort of reformat, note that GNOME did this with an automated tool some years ago, using a 2 spaces instead of tabs format. Most new code in MATE we use four spaces so that's what I used here.  Original tabs appear to have been intended for 8 spaces but little further work was needed to line everything up with 4 spaces, which saves space and matches most of our newer work.

lukefromdc:
@raveit65 , if you want to do this yourself in a different way, feel free to either force-push to this branch, or open another PR and I will close this one in  preference to it as you are probably our best formatter. Had some time so decided this might save you some.

raveit65:
Cool, i will take a look at it later and push changes for myself to this branch.

raveit65:
Seems like the file was a mix of tab-size=4/8. o_0 No wonder that a fine formatting was impossible.
Using spaces everywhere and 4 spaces for indents will terminate this confusion.

raveit65:
I'd like to take another look and do a runtime test before review and merging.

mate-panel/button-widget.c
```diff
@@ -57,876 +57,873 @@ G_DEFINE_TYPE_WITH_PRIVATE (ButtonWidget, button_widget, GTK_TYPE_BUTTON)
...
 
-	if (button->priv->surface_hc)
-		cairo_surface_destroy (button->priv->surface_hc);
-	button->priv->surface_hc = NULL;
+    button->priv->surface = NULL;
+
+    if (button->priv->surface_hc)
+        cairo_surface_destroy (button->priv->surface_hc);
+
+    button->priv->surface_hc = NULL;
```
raveit65, line 158: @mate-desktop/core-team 
Not sure about adding a blank line after the if clauses.
`button->priv->surface = NULL;`
and
`button->priv->surface_hc = NULL;`
doesn't belongs to the if clauses?

lukefromdc:
Those are single-line if clauses w no brackets, thus the blank lines beneath

mate-panel/button-widget.c
raveit65: good than adding a blank line is OK.

raveit65:
I tested that change with x11 and wayland. Everything works as expected.
Please review.
Commits can be squashed in one during merge.

lukefromdc:
I am on the road now away from my desktop, will look it over when I get home. Will also build and install, testing in both x11 and wayland

lukefromdc (commented):
This is really nice work you added, and with the overall reformatting the build warnings we used to get about misleading indents at the `if` statements are gone

mate-panel/button-widget.c
cwendling: If indentation is correct either way is fine, but I don't think a blank line is a bad idea, so I'm happy with it.

Issue #1156, opened 2020-11-05
Dragging panel broken on Wayland


wmww:
Moving the panel to a different screen edge by dragging is glitchy and broken in different ways on both Mir and wlroots. It would be theoretically possible to hack together a working implementation, but it would be a lot of platform-specific work for a non-essential feature. Probably a better idea to just disable panel dragging on Wayland. X11 should not be effected.

lukefromdc:
I never knew this was possible at all. I cannot do it by right, middle, or left click and hold in x11 on my setup

wmww:
Strange. I just tested on X11 and the "feature" doesn't seem to exist, however I certainly didn't add it specifically for Wayland so... idk. On Sway when you left-click-and-drag on the panel it always snaps to the top (orientation changes to "top" in settings as well, so it's not just a Sway bug). On Mir it has slightly different but equally glitchy behavior.

raveit65:
The only way to move a panel is to change orientation with gsettings key, or in preferences when the panel is expand.
Drag&Drop with the mouse was never possible.

raveit65:
Opps, sorry i was wrong.
It is possible when panel is un-expand and you need to use the grip area at left/right or top/bottom side.
But the default is to expand the panel.

raveit65:
Confirmed, with wayfire session it isn#t possible to move the panel when using the grip area of an un-expand panel.

raveit65:
For me it is OK to move panels via gsettings key or with preferences dialog.

raveit65:
Issue still exists with actual Mate and wafire session.

lukefromdc:
Confirm, this can be a real hassle. We should disable that function on wayland. This is where the "mouse-dodging" behavior on the panel caused by applets attempting to use a button release event (which is not working at least in wayfire) must have come from

lukefromdc:
See
https://github.com/mate-desktop/mate-panel/pull/1391

raveit65:
> Confirmed, with wayfire session it isn't possible to move the panel when using the grip area of an un-expand panel.

Opps, my fault. It is possible to move the panel to left/right edge and back to bottom edge.
Well, the unwanted move to the top is a bit annoying.

lukefromdc:
Since we have new complaints about this issue at
https://github.com/mate-desktop/mate-panel/issues/1477
I have brought back what was 
https://github.com/mate-desktop/mate-panel/pull/1391
as the new PR
[...]

PR #1391, opened 2023-07-24
wayland: stop panel "dodging" mouse and moving to top

lukefromdc:
*Panel grabs in wayland cause the whole panel to jump to the top, so disable them when not in x11

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -4220,6 +4220,10 @@ panel_toplevel_finalize (GObject *object)
 static void
 panel_toplevel_class_init (PanelToplevelClass *klass)
 {
+	GdkDisplay *display;
+
+	display = gdk_screen_get_display (gdk_screen_get_default());
+
 	GObjectClass      *gobject_class   = (GObjectClass      *) klass;
 	GtkWidgetClass    *widget_class    = (GtkWidgetClass    *) klass;
 	GtkContainerClass *container_class = (GtkContainerClass *) klass;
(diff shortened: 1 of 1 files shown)


lukefromdc:
Fixes
https://github.com/mate-desktop/mate-panel/issues/1156

lukefromdc:
Note that the pointer still changes to a hand, I've not figured out how to stop that, input is welcome on how to remove that as well so users don't think the panel is draggable

raveit65:
Ok, i learned that it is possible to move the panel in wayland. In report i thought grabbing the panel in general was broken.
In my test it is possible to move the panel to the right or left screen edge if i do not release my trackball button. Should be the same with a standard mouse.
Also, it is possible to move the panel back to bottom. Well, this is a bit harder but it works.
So, i am thinking it is fixable and i am not sure that this feature should be disabled for wayland in general in this early stage. I mean we are far a way from a production wayland session.
Minimum you should add a FIXME to code when you want to disable it.

mate-panel/panel-toplevel.c
```diff
@@ -4220,6 +4220,10 @@ panel_toplevel_finalize (GObject *object)
 static void
 panel_toplevel_class_init (PanelToplevelClass *klass)
 {
+	GdkDisplay *display;
+
+	display = gdk_screen_get_display(gdk_screen_get_default());
```
raveit65, line 4225: formating...
```suggestion
	display = gdk_screen_get_display (gdk_screen_get_default());
```

lukefromdc:
Should we close this since it is possible (though buggy) to grab and move the panel in wayland as master now stands? We can leave this issue as an outstanding bug at this stage if needed.

lukefromdc:
Fixed the formatting in case folks want to play with this. One possible fix would be to limit panel grabbing in wayland to a non-expanded panel, not sure how this is being done in x11 but this would be same behavior if the pointer change to a hand can be suppressed on the expanded panel.

On an expanded panel in x11, a left click, middle click, scroll, or hover on empty space are all simply ignored

mate-panel/panel-toplevel.c
lukefromdc: Format fixed in case we want to keep working on this

raveit65:
> Should we close this since it is possible (though buggy) to grab and move the panel in wayland as master now stands? We can leave this issue as an outstanding bug at this stage if needed.

Agree.

lukefromdc:
Keeping a local backup for the moment though the work was simple. We don't want to break unexpanded panel handling in Wayland as that would constitute a feature removal if Xorg ever goes away. We are a long way from feature parity in wayland but also have a while to get there.

Issue #1380, opened 2023-07-04
wayland: mate-panel needs restart after adding notification-area applet


Sunderland93:
#### Expected behaviour
`notification-area` applet should appears right after pressing Add button

#### Actual behaviour
After adding `notification-area` applet it does not appear in the panel, until panel is restarted (e.g. relogin to session)

#### Steps to reproduce the behaviour
1) build mate-panel with `--with-in-process-applets=all` option
2) run mate-panel on any wlroots compositor (I use LabWC)
3) add notification-area applet to the panel

#### MATE general version
1.26

#### Package version
`master`

#### Linux Distribution
Debian 12

#### Link to bugreport of your Distribution (requirement)
n/a

lukefromdc:
First of all, THANK YOU for testing and finding this. In my own testing I wasn't able to duplicate this on adding the applet, but DID get the same problem some but not all of the time on removing the applet. I got this warning message on removing the notification area in wayland, implying one or more signals not being properly disconnected:

`(mate-panel:11125): notification-area-applet-CRITICAL **: 13:26:03.140: na_grid_set_min_icon_size: assertion 'NA_IS_GRID (grid)' failed
`

Sunderland93:
Yes, I have this message too when I delete applet from the panel.

raveit65:
In my installation the panel crashes when i remove te notification area. But gdb gives me a stacktrace.
```
[sb-wayland@mother ~]$ gdb mate-panel
GNU gdb (GDB) Fedora Linux 13.2-2.fc38
<cut>
Reading symbols from mate-panel...
Reading symbols from /usr/lib/debug/usr/bin/mate-panel-1.27.1-2.fc38.x86_64.debug...
(gdb) run
Starting program: /usr/bin/mate-panel 

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.fedoraproject.org/>
Enable debuginfod for this session? (y or [n]) y
Debuginfod has been enabled.
To make this setting permanent, add 'set debuginfod enabled on' to .gdbinit.
Downloading separate debug info for system-supplied DSO at 0x7ffff7fc8000
Downloading separate debug info for /lib64/libSM.so.6                          
Downloading separate debug info for /lib64/libICE.so.6                         
<cut>

Thread 1 "mate-panel" received signal SIGSEGV, Segmentation fault.
0x00007ffff78b1709 in gtk_spin_button_set_value (spin_button=0xd1, value=24)
[...]

PR #1393, opened 2023-07-24
notification-area: stop warning on removal

lukefromdc:
*Use automatic signal handling for changing icon size so this code doesn't get called when the applet is removed

### DIFF
--- a/applets/notification_area/main.c
+++ b/applets/notification_area/main.c
@@ -139,7 +139,7 @@ static void
 setup_gsettings (NaTrayApplet *applet)
 {
   applet->priv->settings = mate_panel_applet_settings_new (MATE_PANEL_APPLET (applet), NA_TRAY_SCHEMA);
-  g_signal_connect (applet->priv->settings, "changed::" KEY_MIN_ICON_SIZE, G_CALLBACK (gsettings_changed_min_icon_size), applet);
+  g_signal_connect_object (G_OBJECT(applet->priv->settings), "changed::" KEY_MIN_ICON_SIZE, G_CALLBACK (gsettings_changed_min_icon_size), applet, 0);
 }
 
 static void


lukefromdc:
Fixes https://github.com/mate-desktop/mate-panel/issues/1380

raveit65:
I can confirm that it fixes the crashes/warnings on x11/wayland.
But what is the different between g_signal_connect and g_signal_connect_object in our case?
This fix needs to go in stable branch too, so we're lucky that no one remove this applet normally :o))
@mate-desktop/core-team 
Please test it is reproducible with stable branch.

raveit65 (approved):
LGTM,
it fixes crashes in fedora and (only) warnings in debian.

lukefromdc:
The difference is that g_signal_connect_object() automatically manages signal disconnection

lukefromdc:
Just confirmed the warning also occurs on 1.26 when removing the notification area applet

cwendling:
> The difference is that `g_signal_connect_object()` automatically manages signal disconnection

To clarify: [`g_signal_connect()`](https://docs.gtk.org/gobject/func.signal_connect.html) automatically disconnects when the object that emits the signal is destroyed (here, the settings instance). [`g_signal_connect_object()`](https://docs.gtk.org/gobject/func.signal_connect_object.html) assumes the `data` argument is also a `GObject`, and tracks its lifetime as well.

raveit65:
> > The difference is that `g_signal_connect_object()` automatically manages signal disconnection
> 
> To clarify: [`g_signal_connect()`](https://docs.gtk.org/gobject/func.signal_connect.html) automatically disconnects when the object that emits the signal is destroyed (here, the settings instance). [`g_signal_connect_object()`](https://docs.gtk.org/gobject/func.signal_connect_object.html) assumes the `data` argument is also a `GObject`, and tracks its lifetime as well.

So, you approve the PR?

cwendling:
@raveit65 I couldn't test it yet, but the code looks good, yes :+1:

PR #1397, opened 2023-07-30
wayland-tasklist: really set maximum label width

lukefromdc:
*Otherwise taskbuttons for programs using URLs or filenames in the window name can balloon across the whole taskbar

### DIFF
--- a/applets/wncklet/wayland-backend.c
+++ b/applets/wncklet/wayland-backend.c
@@ -31,6 +31,9 @@
 #include "wayland-backend.h"
 #include "wayland-protocol/wlr-foreign-toplevel-management-unstable-v1-client.h"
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
To test: create a new directory in caja with a very long filename, one that wraps over three lines or so.

Start a wayland session with mate-panel, then open that caja directory with no other window list buttons (or one other window list button if running from terminal) on the panel. Thewindow list button should be about the same size it is in x11/

Note that fonts are coming up larger in wayfire (and probably will for a lot of wayland users as they are computed differently, so
the value of` TASKLIST_TEXT_MAX_WIDTH ` had to be reduced from the 25 chars used in wnck-tasklist.

Users will most commonly need with with browsers, where websites can send an arbitrarily long filename to the window list

raveit65 (approved):
LGTM

PR #1400, opened 2023-08-09
libmate-panel-applet: don't use panel plug to find the screen

lukefromdc:
* There is only one screen in GTK 3.22 or later
* Finding the screen from that used by GtkPanelPlug is out of process only
* Any in-process applet calling mate_panel_applet_request_focus needs this to avoid a segfault
* One known user: dictionary applet

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -618,7 +618,7 @@ mate_panel_applet_request_focus (MatePanelApplet	 *applet,
 	g_return_if_fail (MATE_PANEL_IS_APPLET (applet));
 
 	priv    = mate_panel_applet_get_instance_private (applet);
-	screen  = gtk_window_get_screen (GTK_WINDOW (priv->plug));
+	screen  = gdk_screen_get_default(); /*There is only one screen since GTK 3.22*/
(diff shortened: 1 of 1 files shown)


lukefromdc:
This is needed to allow building the dictionary applet in-process on x11. Note that the out of process dictionary applet in mate-utils from git master with any recent version of mate-panel seems to be broken at least in Debian Unstable. The dialog box containing the results never shows up. With an in-process port and build it works, and in-process opens the way for making it work in wayland, though for that we will need a wayland codepath either in this same function or replacing it in the dictionary applet

lukefromdc:
Needed for https://github.com/mate-desktop/mate-utils/pull/353

raveit65:
Hmm, i used this commit but i can't add dictionary applet to the panel when build-in-process.

lukefromdc:
Found out what is different with the dialog in your case: With current git master the dictionary applet on x11 usually works if the panel is on the top. With the panel on the bottom the dialog does not come up. For convenience that was with this PR for mate-panel still applied, had no trouble adding the applet.

We need to find out why you had trouble adding the out of process applet. Assuming you restarted the panel with each test build of the applet changing from in-process to out of process or vice-versa

raveit65 (approved):
Works for me with x11 and wayland.

PR #1405, opened 2023-10-01
Hidpi x11: center panel force-quit popup window

lukefromdc:
Center the force-quit dialog on x11 on hidpi displays. Stop it from running off the bottom right corner

### DIFF
--- a/mate-panel/panel-force-quit.c
+++ b/mate-panel/panel-force-quit.c
@@ -58,7 +58,7 @@ display_popup_window (GdkScreen *screen)
 	GtkWidget     *image;
 	GtkWidget     *frame;
 	GtkWidget     *label;
-	int            screen_width, screen_height;
+	int            screen_width, screen_height, scale;
 	GtkAllocation  allocation;
 
 	retval = gtk_window_new (GTK_WINDOW_POPUP);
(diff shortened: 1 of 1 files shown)


raveit65:
I see no different with the placement of the right click popup when applet  is at top panel with hidpi.
What is the problem?
What is the goal?

raveit65 (approved):
lgtm

lukefromdc:
Thanks, right click menu was fine, but the left click popup window was what was off to the left.

raveit65:
Yeap, in a second test i saw that the popup was out of monitor area without PR.
With  2 monitors the popup is exactly between the monitors or in the middle of entire resolution , which is a bit weird.
But it is like it should now.

lukefromdc:
Still a slight offset with two dissimilar monitors (a 4K and a 1080p) but not by much. Fix for that would be to stop using GdkScreen (there is only one and it spans all monitors) and get the monitor the panel is on and center it on that using
`gdk_monitor_get_geometry ()`

lukefromdc (commented):
Builds and runs fine, though on my localepurged machines, trying to use translations included in any one package ( via env variable and prior to next run of localepurge) does not work. 

We need a review by someone with any non-english language installed

lukefromdc:
See follow-on PR 
https://github.com/mate-desktop/mate-panel/pull/1407
which fixes the multimonitor case by using the monitor (one monitor) and not the screen (covers all monitor)
to center the popup and has the further advantage that gdk_monitor_get_geometry() returns already scaled pixels

PR #1406, opened 2023-10-01
Wayland: Fix force-quit unsupported message dialog

lukefromdc:
*For some reason "widget" is not being properly passed to this, so use NULL for the parent window
*This makes the "force-quit only available in x11" dialog work

### DIFF
--- a/mate-panel/panel-action-button.c
+++ b/mate-panel/panel-action-button.c
@@ -471,7 +471,7 @@ panel_action_force_quit (GtkWidget *widget)
 	}
 #endif
 	flags = GTK_DIALOG_DESTROY_WITH_PARENT;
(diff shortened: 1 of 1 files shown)


raveit65:
That works, but is there no way to disable force-quit in add-to-panel dialog?

raveit65 (approved):
Does what it promise.

lukefromdc:
Disabling anything from appearing in the add to panel dialog won't prevent it from appearing on the panel if previously added in an x11 session. This problem also effects the logout button since that seems difficult to implement in wayland without a special logout manager program(these do exist but using one for this would mean depending on in in wayland builds. For the that, I tested just killing the compositor, but that did not allow SDDM to be shown again, giving a black screen on that VT instead.

PR #1407, opened 2023-10-03
Force-quit/x11: Fix misalignment of popup with multimonitors

lukefromdc:
*Compute the size of the monitor we are actually showing the popup on *by getting the monitor it is shown on and running gdk_monitor_get_geometry()

*This returns application (scaled) pixels not device pixels and apparently so does gtk_widget_get_allocation()
*Thus we no longer have to compute the scale ourselves and the popup window centers on the panel's monitor
*Previously we got the rectangle covering ALL monitors and centered on that

### DIFF
--- a/mate-panel/panel-force-quit.c
+++ b/mate-panel/panel-force-quit.c
@@ -58,8 +58,12 @@ display_popup_window (GdkScreen *screen)
 	GtkWidget     *image;
 	GtkWidget     *frame;
 	GtkWidget     *label;
-	int            screen_width, screen_height, scale;
+	int            monitor_width, monitor_height;
 	GtkAllocation  allocation;
+	GdkWindow      *window;
+	GdkDisplay     *display;
+	GdkMonitor     *monitor;
+	GdkRectangle    geometry;
 
 	retval = gtk_window_new (GTK_WINDOW_POPUP);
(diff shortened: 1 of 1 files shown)


lukefromdc:
Details:
`gdk_monitor_get_geometry()` returns application (scaled) pixels rather than device
pixels.

Testing with g_message commands temporarily added to the code and changing the 
window scaling showed that the value returned by gtk_widget_get_allocation()
was not changing with window scaling = 1 or  window scaling = 2 either
 
https://docs.gtk.org/gtk3/method.Widget.get_allocation.html
states that ` gtk_widget_get_allocation()` returns "adjusted allocation" which although
not documented here as such appears to also  be in application pixels

Therefore, the use of `gdk_monitor_get_geometry()` replaces
having to compute scaled pixels ourselves when using
`WidthOfScreen (gdk_x11_screen_get_xscreen (screen));`
and 
`HeightOfScreen (gdk_x11_screen_get_xscreen (screen));`
which return device pixels and in the multimonitor case give the dimensions of the 
smallest rectangle that can be drawn around all monitors combined.

raveit65 (approved):
Works fine,
popup is now in the middle of the screen with a 2 monitor setup.

zhuyaliang (approved):

PR #1409, opened 2023-10-06
replace deprecated gtk-image-menu-item*

raveit65:


applets/notification_area/status-notifier/sn-dbus-menu-item.c
```diff
@@ -203,9 +204,9 @@ sn_dbus_menu_item_new (GVariant *props)
               cairo_surface_destroy (surface);
             }
 
-          item->item = gtk_image_menu_item_new ();
+          item->item = mate_image_menu_item_new ();
           gtk_image_menu_item_set_always_show_image (GTK_IMAGE_MENU_ITEM (item->item), TRUE);
```
zhuyaliang, line 208: If ```mate_image_menu_item_set_image```  is used, there is no need to use ```gtk_image_menu_item_set_always_show_image```again

mate-panel/menu.c
```diff
@@ -964,7 +965,7 @@ panel_image_menu_item_new (void)
 {
 	GtkWidget *menuitem;
 
-	menuitem = gtk_image_menu_item_new ();
+	menuitem = mate_image_menu_item_new ();
 	gtk_image_menu_item_set_always_show_image (GTK_IMAGE_MENU_ITEM (menuitem),
```
zhuyaliang, line 969: Remove this line

lukefromdc (approved):
This builds and runs without issue here-and the menus-have-icons gsettings preference is still respected.

Just a straightforward replacement of deprecated gtk_image_menu_item with our own mate_image_menu_item , note that if for any reason we are ever forced to deal with GTK4 this and similar PR's elsewhere will restore what would otherwise be a lost feature. In the meantime by maintaining it ourselves we ensure it does not bitrot

lukefromdc (commented):
Update: on running mate-panel from terminal with this applied, I just got many iterations of these warnings:

```
gtk_image_menu_item_set_always_show_image: assertion 'GTK_IS_IMAGE_MENU_ITEM (image_menu_item)' failed

(mate-panel:46673): GLib-GObject-CRITICAL **: 19:07:45.983: invalid cast from 'MateImageMenuItem' to 'GtkImageMenuItem'
```

looks like we need to remove `gtk_image_menu_item_set_always_show_image` and when that is removed from menu.c at line 969 and from sn-dbus-menu-item.c at line 208 the terminal warnings are stopped

applets/notification_area/status-notifier/sn-dbus-menu-item.c
lukefromdc: Two issues here: using `gtk_image_menu_item_set_always_show_image` with mate_image_menu_item is (presumably) invalid and throws runtime warnings when the panel is run from terminal.

Also note that we have
` if (GTK_IS_IMAGE_MENU_ITEM (item->item))`
at line 441, I would expect it to always test FALSE

though for some reason I still got indicator menu item icons

mate-panel/menu.c
raveit65: done.

applets/notification_area/status-notifier/sn-dbus-menu-item.c
raveit65: Looks like i need to change `if (GTK_IS_IMAGE_MENU_ITEM (item->item))` to `if (MATE_IS_IMAGE_MENU_ITEM (item->item))` at line 441 and 450?

raveit65:
With last push i removed `gtk_image_menu_item_set_always_show_image` and runtime warnings are gone.

zhuyaliang (approved):

applets/notification_area/status-notifier/sn-dbus-menu-item.c
raveit65: @zhuyaliang see previous post ^^^^

raveit65: sorry, browser cache wasn't updated.

raveit65: Using `if (MATE_IS_IMAGE_MENU_ITEM (item->item))` with last push.

lukefromdc:
the `GTK_IS_IMAGE_MENU_ITEM` macro seems to work fine, and we already use it in mate-desktop. With this in use the runtime warnings are gone and the entire status notfier built with no warnings at all with my usual build defaults.

Looks good to go here

lukefromdc:
Turns out we lost the Places and System icons in the main menus and had a couple leftover deprecation warnings about `gtk_image_menu_item` 
That came from still defining these two menu items with GtkImageMenuItem and was easily fixed:

https://github.com/mate-desktop/mate-panel/pull/1410

raveit65:
Opps, i am using the classic panel menu in result i didn't saw that.
Thanks.

PR #1413, opened 2023-10-24
status-notifier: Show AttentionIcon when Status is NeedsAttention 

### LINKED ISSUE #1412: Attention icon doesn't shown for StatusNotifier
Hello!

After adding StatusNotifier support to Liferea there was issue reported for the Mate Desktop in Fedora 38 [1].
In the Mate Desktop under Fedora 38 normal icon is always shown instead of attention icon.

I created test case to demonstrate this issue [2].

[1] https://github.com/lwindolf/liferea/issues/1311#issuecomment-1770940601
[2] https://github.com/nE0sIghT/appindicator-testcase

Reproducible: Always

#### Expected behaviour
When indicator icon is set to `IndicatorStatus.ATTENTION` Mate should show attention icon instead of normal icon in the "tray".

#### Actual behaviour
In the Mate Desktop normal icon always shown in "tray" instead of attention icon

```
git clone https://github.com/nE0sIghT/appindicator-testcase
cd appindicator-testcase
python3 testcase.py
```

#### MATE general version
1.26.2

#### Package version
mate-panel.x86_64 1.26.3-1.fc38

#### Linux Distribution
Fedora 38

https://bugzilla.redhat.com/show_bug.cgi?id=2245224

cwendling:
Fixes #1412.

This PR also includes 2 unrelated typo fixes extracted from gnome-panel.  I can move remove them from this PR if wanted, as they are unrelated (and not necessary for anything but not having typos in the code).

### DIFF
--- a/applets/notification_area/status-notifier/sn-dbus-menu-item.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu-item.c
@@ -253,7 +253,7 @@ sn_dbus_menu_item_new (GVariant *props)
 }
 
 void
-sn_dubs_menu_item_free (gpointer data)
+sn_dbus_menu_item_free (gpointer data)
 {
   SnDBusMenuItem *item;
 
--- a/applets/notification_area/status-notifier/sn-dbus-menu-item.c
+++ b/applets/notification_area/status-notifier/sn-dbus-menu-item.c
@@ -21,7 +21,7 @@
 #include "sn-dbus-menu-item.h"
 
 static GdkPixbuf *
-pxibuf_new (GVariant *variant)
+pixbuf_new (GVariant *variant)
 {
   gsize length;
   const guchar *data;
@@ -150,7 +150,7 @@ sn_dbus_menu_item_new (GVariant *props)
       else if (g_strcmp0 (prop, "icon-name") == 0)
         item->icon_name = g_variant_dup_string (value, NULL);
       else if (g_strcmp0 (prop, "icon-data") == 0)
-        item->icon_data = pxibuf_new (value);
+        item->icon_data = pixbuf_new (value);
       else if (g_strcmp0 (prop, "label") == 0)
         item->label = g_variant_dup_string (value, NULL);
       else if (g_strcmp0 (prop, "shortcut") == 0)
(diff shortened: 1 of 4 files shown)


cwendling:
The attention icon from #1413 is rendered a bit ugly (scaled up), but that's orthogonal to the changes here.  Possibly we should separately avoid scaling up the icon if it's "close enough" to the expected size.

zhuyaliang (approved):
Agree to merge

lukefromdc (commented):
Builds clean, runs fine on my setup, but not sure how to test for function given I don't think I have anything installed that uses an "attention" icon, though it's true I could and not know it

cwendling:
@lukefromdc simplest is to use the test case from the OP of #1412.

lukefromdc (approved):
Running the test case proved this works: icon changes occur only with this PR in place. On my setup, all the icons from the test case come out way too small, but all my other tray and SNI icons are unchanged in rendering. Might be a test case issue, or we may need to force icon sizes when the provided icon is more than say, a 2px mismatch but that may be something to worry about  for another PR, if for no more reason than to make any later bisects easier

cwendling:
> Running the test case proved this works: icon changes occur only with this PR in place.

The for the test, merged.

> On my setup, all the icons from the test case come out way too small, but all my other tray and SNI icons are unchanged in rendering. Might be a test case issue, or we may need to force icon sizes when the provided icon is more than say, a 2px mismatch but that may be something to worry about for another PR, if for no more reason than to make any later bisects easier

Yeah this is orthogonal to the AttentionIcon support, which uses the exact same code to render the icons as the normal icon (just the source is changed when needed).

However, them being too small is kind of odd… or maybe you've got HiDPI and we have a HiDPI issue in the code path using a file path instead of an icon name?  That's possible as I doubt many people tried it, worth investigating separately.

However, it's a fairly tricky situation, because for me, with a 26px tall panel, I'd rather have the 16x16 icon (the attention one from the test) at it's real size than scaled up to 24x24 which makes it very ugly and blurry.  So I was thinking on my end not to scale up an icon if it's some percentage of the expected size (well, here it's about 33%, but for 16 to 24 it still looks better)… or maybe scale to a close enough size that is likely to result in less blurring, e.g. usually ×2 is cleaner than ×2.1 -- but "good" values likely depends on the scaling algorithm.

Anyway, this again should be discussed separately :)

lukefromdc:
Thanks, found and used it in my last test

lukefromdc:
Yes I am using a 4K monitor and window-scaling=2

cwendling:
> Yes I am using a 4K monitor and window-scaling=2

OK, could be that this part of the code is not handling that well, we'd have to investigate (in another PR)

PR #1375, opened 2023-06-26
libmate-panel-applet/Makefile.am: Fix build with gettext 0.22

listout:
With gettext-0.22 we are getting build errors like: /usr/bin/msgfmt --desktop --keyword Name --keyword Description --template org.mate.panel.TestApplet.mate-panel-applet.desktop.in -d ../po -o org.mate.panel.TestApplet.mate-panel-applet /usr/bin/msgfmt: no input file should be given if --desktop and -d are specified Try '/usr/bin/msgfmt --help' for more information. make[3]: *** [Makefile:1035: org.mate.panel.TestApplet.mate-panel-applet] Error 1 make[3]: *** Waiting for unfinished jobs....

Bug: https://bugs.gentoo.org/908877

PR #1415, opened 2023-11-07
WIP: Enable a github actions build

cwendling:


cwendling:
OK, so.  @mate-desktop/core-team do we want something like this?  I hear that we have Travis credits issues, so maybe we can unload some of this to GitHub Actions?  I don't know much of anything about it, but it seems to work, and seems to allow a lot of things.
I'm not advertising anything, but had a fairly good experience with *using* it in Universal-CTags and Geany (I did *not* setup it there).

lukefromdc:
I don't know anything about it, but if we can do this inside Github it will certainlt help our travis situation

raveit65:
From which travis-CI issues did you heard exactly?
For me everything is working.
With travis-CI we also do releases and upload them to our web-site, we display ccpcheck results to gh-pages, and it works for stable and master branch with CI builds for rpm- and deb-based distros.
Do you really want to create a such configuration for all repositories and dismiss a working CI configuration?
We can merge this because it exists, but i don't want spend any spare time for an ihmo unnecessary CI rewrite for the whole organization.
But if all other want to switch....go ahead.

raveit65:
Btw. i reduced the travis CI build time about 50% 2 weeks ago.

cwendling:
> From which travis-CI issues did you heard exactly?

Build credits, I think it was you mentioning we're needing to ask for renewal all the time… I'm happy if it's not an issue (anymore) though.

> For me everything is working.

For me as well, minus the mentioned credits issue, which I might not be up-to-date about.

> With travis-CI we also do releases and upload them to our web-site, we display ccpcheck results to gh-pages, and it works for stable and master branch with CI builds for rpm- and deb-based distros.

Yeah I know it's complex, and this here is just a first attempt *in case* it can be useful.

> Do you really want to create a such configuration for all repositories and dismiss a working CI configuration?

No, I don't.  But I certainly would even less want CI to stop working one day out of the blue because we run out of credits or alike.  But again, if you tell me I'm mistaken and it's not an issue, I just lost a bit of time and it's alright.

> We can merge this because it exists, but i don't want spend any spare time for an ihmo unnecessary CI rewrite for the whole organization.

Only if it has actual value, because maintenance has a cost.  Feel free to dump this, I'm not emotional about it :smile:

raveit65:
> Build credits, I think it was you mentioning we're needing to ask for renewal all the time… I'm happy if it's not an issue (anymore) though.

I will explain it again , (hopefully last time)
- We are sponsored in general by travis as open-source projekt.
- Every time they gave us 25000 credits.
- When our credits begin running low again a person from the team needs to write a simple one liner mail and ask again for 25000 credits
- Usually after 1 hour credits are updated.

It is understandable that travis-Ci wants to control the sponsored build time. I don't think that is an issue unless the team is too lazy to write an email with one sentence.
I could reduce builds for every push with disabling `build pushed pull request`. Now we use only `build pushed branches`.
That means we use only 2 builds (fedora+debian) for every push, before it was 4 builds.
For example when you do a force push 15 times in a PR it will trigger 30 builds of the code.
It is always possible to temporary disable travis-Ci to avoid this excessive usage in a PR.
grafik


> > We can merge this because it exists, but i don't want spend any spare time for an ihmo unnecessary CI rewrite for the whole organization.

> Only if it has actual value, because maintenance has a cost. Feel free to dump this, I'm not emotional about it 😄

No, only you can dump your ideas. I don't do that.

Any way, beside from my concerns and the question who like to do this rewrite for every repository,
is the the github-runner really for free and without any limits when it is used for 48 repositories?
I wrote a CI configuration for gitlab (with releasing und uploading to a download-server) and the gitlab-runner isn't free.

Maybe it make sense to replace debian build from travis-ci with a github-actions build to reduce build time again,
but i recommend not to replace fedora build as it it use for several other things.
I don't want to fall back in a time where we create releases by hand.

I agree with you not to use ccpscheck builds.... who needs this?
`debian:testing` docker tag was used because stable debian do not use current stable Mate packages.
For ubuntu i would select `ubuntu:rolling` tag which use newest and stable package releases. When using LTS you can run in the same problem as with debian.

PS: I disabled travis-CI temporary for mate-panel in case you want to work on this PR.

cwendling:
> It is understandable that travis-Ci wants to control the sponsored build time. I don't think that is an issue unless the team is too lazy to write an email with one sentence.

It's (probably) OK as long as we have somebody
- noticing
- knowing the procedure.  Is it written down anywhere, and can non-core people do it? (e.g. if we all disappear one day…)

> I could reduce builds for every push with disabling build pushed pull request.

Well, that's kind of one of my concerns: we're disabling things because we can't afford them.  No, we shouldn't waste resources, but we shouldn't be enticed to lower QA either.

> Now we use only build pushed branches.

IIUC, this means that we don't run CI on pull requests made outside of the mate-desktop repository, right?  This is an issue IMO, as it limits the checks we do on non-member PRs.

> That means we use only 2 builds (fedora+debian) for every push, before it was 4 builds.
> For example when you do a force push 15 times in a PR it will trigger 30 builds of the code.
> It is always possible to temporarily disable travis-Ci to avoid this excessive usage in a PR.

One other option, if I understand the settings correctly, could be to keep building PRs, but either
* only build pushes to master (if we can configure that)
* stop pushing PR branches to the main repo and do that on personal forks instead.

> is the github-runner really for free and without any limits when it is used for 48 repositories?

It (currently) [is free](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) for public repositories: *"GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners.*" However, we of course can't be sure what the future holds.

> Maybe it make sense to replace debian build from travis-ci with a github-actions build to reduce build time again,
but i recommend not to replace fedora build as it it use for several other things.

Well, if it makes sense to unload part of it only, sure.

> I agree with you not to use ccpscheck builds.... who needs this?

Did I say anything about disabling it?  Admittedly I don't use it, but I have just moved it to a separate job in order to run it concurrently, and have the results easier to reach.  If nobody uses it maybe we can remove it though, as it's never a blocking check, so the only value is if people actually look at the results.

> `debian:testing` docker tag was used because stable debian do not use current stable Mate packages.
For ubuntu i would select `ubuntu:rolling` tag which use newest and stable package releases. When using LTS you can run in the same problem as with debian. See https://hub.docker.com/_/ubuntu/

For now I used a plain runner without adding a docker layer, and selected "ubuntu-latest", which currently is Ubuntu 22.04.3.  But docker images should be fairly easy to use as well, universal-ctags seems to make use of a lot of them.

> PS: I disabled travis-CI temporarily for mate-panel in case you want to work on this PR.

You shouldn't have to do this I think, I specifically made it on my fork because I remembered you telling me we didn't build those.  And I indeed don't see Travis in the checks list.

cwendling:
BTW: I will **not** be pursuing this *unless* it's something the team desires.  As @raveit65 suggests, getting this working for all repositories is a significant undertaking, and I don't find this fun enough to do it if nobody else cares.
I care about CI health in general, not which runners use it or anything.

cwendling:
One last thing to think about, regarding CI in general:  it's useful for CI to be *fast*, so developers don't get frustrated or distracted.  If you have CI results within a minute or two, it makes development faster in case an issue is found: the author might not have switched to another task, or not went to sleep, or whatever.  I'm not saying that our CI is too slow for convenience ATM (it's pretty OK), but that making it faster (even if it means shorter jobs with a little more overall overhead) can be beneficial.

Personal experience with CI results within a minute was a lot nicer than another experience where results would generally take an hour or more :)  Admittedly the hour-long ones caught more issues in general (because of a huge project I was less familiar with), which didn't help frustration at waiting times, but it also meant that by the time the results came in I was doing something entirely different, and sometimes even went to sleep 'til next morning.

Just some thoughts.

raveit65:
> > It is understandable that travis-Ci wants to control the sponsored build time. I don't think that is an issue unless the team is too lazy to write an email with one sentence.
> 
> It's (probably) OK as long as we have somebody
> 
>     * noticing
> 
>     * knowing the procedure.  Is it written down anywhere, and can non-core people do it? (e.g. if we all disappear one day…)
> 
It is written here and in several other reports. Writing docs is a big job, volunteers are wanted....
> 
> > I could reduce builds for every push with disabling build pushed pull request.
> 
> Well, that's kind of one of my concerns: we're disabling things because we can't afford them. No, we shouldn't waste resources, but we shouldn't be enticed to lower QA either.
> 
No, this doesn't have any impact, `build pushed pull request` means a build between PR- and master-branch will be triggered.

> > Now we use only build pushed branches.
> 
> IIUC, this means that we don't run CI on pull requests made outside of the mate-desktop repository, right? This is an issue IMO, as it limits the checks we do on non-member PRs.
> 
No, `build pushed branches` means a build will be triggered for every push on every branch including PR-branches.

> > That means we use only 2 builds (fedora+debian) for every push, before it was 4 builds.
> > For example when you do a force push 15 times in a PR it will trigger 30 builds of the code.
> > It is always possible to temporarily disable travis-Ci to avoid this excessive usage in a PR.
> 
> One other option, if I understand the settings correctly, could be to keep building PRs, but either
> 
>     * only build pushes to master (if we can configure that)
> 
>     * stop pushing PR branches to the main repo and do that on personal forks instead.
> 
There is only one setting possible for master/stable and PR branches possible as far as i know from travis docs.
> 
> > is the github-runner really for free and without any limits when it is used for 48 repositories?
> 
> It (currently) [is free](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions) for public repositories: _"GitHub Actions usage is free for standard GitHub-hosted runners in public repositories, and for self-hosted runners._" However, we of course can't be sure what the future holds.
> 
> > Maybe it make sense to replace debian build from travis-ci with a github-actions build to reduce build time again,
> > but i recommend not to replace fedora build as it it use for several other things.
> 
> Well, if it makes sense to unload part of it only, sure.
> 
> > I agree with you not to use ccpscheck builds.... who needs this?
> 
> Did I say anything about disabling it? Admittedly I don't use it, but I have just moved it to a separate job in order to run it concurrently, and have the results easier to reach. If nobody uses it maybe we can remove it though, as it's never a blocking check, so the only value is if people actually look at the results.
> 
I read your comment in commit :-) 
`Do we need the real build for cppcheck run?  I don't think so`
https://github.com/mate-desktop/mate-panel/pull/1415/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R129
> > `debian:testing` docker tag was used because stable debian do not use current stable Mate packages.
> > For ubuntu i would select `ubuntu:rolling` tag which use newest and stable package releases. When using LTS you can run in the same problem as with debian. See https://hub.docker.com/_/ubuntu/
> 
> For now I used a plain runner without adding a docker layer, and selected "ubuntu-latest", which currently is Ubuntu 22.04.3. But docker images should be fairly easy to use as well, universal-ctags seems to make use of a lot of them.
> 
As i said it is possible that LTS releases do not use latest stable Mate releases after we made a new major release and than you need to build a mate dependency from git and not from docker image repositories.
Also, it make sense to see build issues with new updated dependencies in CI.
...only a hint not a concern :-)

> > PS: I disabled travis-CI temporarily for mate-panel in case you want to work on this PR.
> 
> You shouldn't have to do this I think,
>
Why you want want to test a github-CI configuration with travis-CI?
They are no code changes in mate-panel. You only need to see that your CI configuration works with github-CI in this case.
It only eats sponsored build time.
Travis can be enabled again when PR is ready.
>
> I specifically made it on my fork because I remembered you telling me we didn't build those. And I indeed don't see Travis in the checks list.

??? Which checklist.
For seeing the travis site from screenshot you need a travis account. Mandatory for Mate developers as long we use travis-ci :-)

cwendling:
> I read your comment in commit :-)
> `Do we need the real build for cppcheck run? I don't think so`
> https://github.com/mate-desktop/mate-panel/pull/1415/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721R129

Oh :)  What I meant there was wondering whether the cppcheck run does need to be run *after* a successful build, or running it on a plain git clone enough.  I guessed it was OK with a plain git clone, just adding the library dependencies to help with external declarations.  OTOH, maybe it *does* matter for generated files?  Not sure.
But then again, if nobody looks at it, we could just drop it.

> Why you want want to test a github-CI configuration with travis-CI?

I don't, I just meant that I didn't think it was necessary to do anything for Travis not to run on this PR (see below).

> ??? Which checklist.

Normally any check results (including Travis) show up in GitHub PR UI, both in the merge area "All checks have passed", and then "show all checks", or in the [*Checks* tab](https://github.com/mate-desktop/mate-panel/pull/1415/checks).

> No, this doesn't have any impact, build pushed pull request means a build between PR- and master-branch will be triggered.
> […]
> No, build pushed branches means a build will be triggered for every push on every branch including PR-branches.

Are you *sure*?  I don't see *any* build for e.g. this PR in https://app.travis-ci.com/github/mate-desktop/mate-panel/builds

From my understanding, this only works on the *repo on which Travis is set up*, e.g. for branches in *mate-desktop/mate-panel*, not any fork of it.  So yes, it works for PR *if we push the PR branch to the target repository*, but not if we push it to a fork.

Here (push to fork), it lists the 2 GitHub Actions, no Travis.  OTOH, on [1413](https://github.com/mate-desktop/mate-panel/pull/1413/checks) (push to repo branch) it shows *Travis CI - Branch*.  And on an older push-to-fork, [we only have *Travis CI - Pull Request*](https://github.com/mate-desktop/mate-panel/pull/1375/checks), whereas on a same-era push-to-repo [we have both PR and Branch](https://github.com/mate-desktop/mate-panel/pull/1372/checks).

raveit65:
> Are you _sure_? I don't see _any_ build for e.g. this PR in https://app.travis-ci.com/github/mate-desktop/mate-panel/builds
> 
> From my understanding, this only works on the _repo on which Travis is set up_, e.g. for branches in _mate-desktop/mate-panel_, not any fork of it. So yes, it works for PR _if we push the PR branch to the target repository_, but not if we push it to a fork.
> 
> Here (push to fork), it lists the 2 GitHub Actions, no Travis. OTOH, on [1413](https://github.com/mate-desktop/mate-panel/pull/1413/checks) (push to repo branch) it shows _Travis CI - Branch_. And on an older push-to-fork, [we only have _Travis CI - Pull Request_](https://github.com/mate-desktop/mate-panel/pull/1375/checks), whereas on a same-era push-to-repo [we have both PR and Branch](https://github.com/mate-desktop/mate-panel/pull/1372/checks).

Ohh, you're right, I tested this only with internal PR branches before i disabled `build pushed branches`, of course we need this for PRs from external contributors.
That means the only option is to exclude debian build from travis-ci to avoid **4 builds** for every push at travis and include it to github-actions-ci. 
So, this is a good reason for your PR :-) I saw you included debian build already.
I will start with reviewing and helping if needed.
Can you comment out debian in https://github.com/mate-desktop/mate-panel/blob/master/.travis.yml#L73 with an extra commit in this PR for testing please? I can't commit to  your branch :-)
I will enable  `build pushed branches` than. Or do you have access to the switch from my previous screenshot? I am not sure if all people from core-team can do that.

Maybe it make sense to include archlinux build to github-actions? Archlinux was part of the origin travis config and disabled when travis-ci wasn't free anymore a while ago. Build dependencies are listed in .build.yml.

.github/workflows/build.yml
```diff
@@ -0,0 +1,222 @@
...
+    gobject-introspection
+  # TODO
+  SCAN_BUILD_DEPS: |
+    clang
+    clang-tools
+
+jobs:
+  build:
+    name: Build on Ubuntu
+    runs-on: ubuntu-latest
```
raveit65, line 63: ubuntu-latest is the tag of the github-runner as i understand docs, right? https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources.
Maybe using unbuntu:rolling as container for building? Same as you did for debian-testing.
In ubuntu:rolling we will see a new mayor mate release much quicker than in LTS, if they do updating LTS.

.github/workflows/build.yml
```diff
@@ -0,0 +1,222 @@
...
+      - name: Build mate-desktop
+        run: make -C mate-desktop -j ${{ env.JOBS }}
+
+      - name: Install mate-desktop
+        run: make -C mate-desktop -j ${{ env.JOBS }} install
+
+      - name: Configure
+        run: |
+          cd mate-panel
+          NOCONFIGURE=1 ./autogen.sh
```
raveit65, line 118: Can we use `--enable-compile-warnings=maximum` when building on master branch? Same as we did for travis. https://github.com/mate-desktop/mate-panel/blob/master/.build.yml#L175
For stable branches we used normal compilor warn flags.

.github/workflows/build.yml
```diff
@@ -0,0 +1,222 @@
...
+          { ./configure || { cat config.log; exit 1; } ; }
```
raveit65, line 119: Can we use `--enable-compile-warnings=maximum` when building on master branch? Same as we did for travis. https://github.com/mate-desktop/mate-panel/blob/master/.build.yml#L175
For stable branches we used normal compilor warn flags.

cwendling:
> Ohh, you're right, I tested this only with internal PR branches before i disabled `build pushed branches`, of course we need this for PRs from external contributors. That means the only option is to exclude debian build from travis-ci to avoid **4 builds** for every push at travis

Well, not, there's an alternative: we could ask everybody (including core team) to create PRs on *their fork*, so only the Travis PR build runs on PRs.  It requires all of us to play along, but that's another option.

> and include it to github-actions-ci. So, this is a good reason for your PR :-)

> I saw you included debian build already.

Yeah, I figured I could try and see how containers worked.  Seemed easy enough, minus the fact that apparently we're root already, and sudo isn't available.  Fine, but one has to know :)

> I will start with reviewing and helping if needed.

Thanks!

> Can you comment out debian in https://github.com/mate-desktop/mate-panel/blob/master/.travis.yml#L73 with an extra commit in this PR for testing please?

Sure, will do.

> I can't commit to your branch :-)

Can't you?  That's a bummer, I check the option allowing maintainers though…

> I will enable `build pushed branches` than. Or do you have access to the switch from my previous screenshot? I am not sure if all people from core-team can do that.

I do have access.  I'm not sure if we really should re-enable this right away, because it'd be better to either
* tell @mate-desktop/core-team not to push PR branches to the main repo but work on a fork
* and/or wait to move the Debian runner to GitHub Actions if we're going that route.

What do you think?

> Maybe it make sense to include archlinux build to github-actions? Archlinux was part of the origin travis config and disabled when travis-ci wasn't free anymore a while ago. Build dependencies are listed in .build.yml.

Sure, IMO it makes sense to add any setup that adds value, and Arch is definitely both a fairly different setup, and a widely used one.

.github/workflows/build.yml
cwendling: Sure, will add this.  Also, maybe we want to add scan-build as well.  Unless we want to move it altogether form Travis to GHA?

cwendling: Done.

.github/workflows/build.yml
cwendling: I'm trying this, let's see if that works… if it does, it simplifies the build file :)

raveit65:
> I do have access. I'm not sure if we really should re-enable this right away, because it'd be better to either
> 
>     * tell @mate-desktop/core-team not to push PR branches to the main repo but work on a fork
> 
>     * and/or wait to move the Debian runner to GitHub Actions if we're going that route.
> 
> 
> What do you think?

At the moment we are only 5 active members who commit, so using a forked branch might be a interim solution.
Or team member disable `build pushed branches` for their PRs, which might be simpler.
In general build on debian docker container should move to github-actions.

.github/workflows/build.yml
raveit65: Honestly, i don't know what scan-build exactly means :)

raveit65: Leave it there for the moment. It is used for fedora container build

.github/workflows/build.yml
cwendling: Works and simplifies things, double win :)

.github/workflows/build.yml
cwendling: It's a [tool for running static code analysis](https://clang-analyzer.llvm.org/scan-build.html) easily.

.github/workflows/build.yml
raveit65: Cool, only one config :)

cwendling:
> At the moment we are only 5 active members who commit, so using a forked branch might be a interim solution.

From my POV it doesn't have to be interim, and that's how I'm used to do things everywhere else.  And normally GitHub even allows maintainers to push to PR branches in forks if the author allowed this in the PR (which is the default).  So I'm not sure if there's any downside, minus perhaps a little more bandwidth usage as you'd push some more commits to your fork (as other's commits wouldn't be there already, so pushing your PR branch would sync this) -- but that should not be much, it's just the few commits, and it's entirely transparent otherwise.

> Or team member disable `build pushed branches` for their PRs, which might be simpler.

Is it easy to do it per-branch?  I'm afraid it's easy to forget as well, and probably would still run a first unwanted duplicate, wouldn't it?

.github/workflows/build.yml
raveit65: I saw `scan-build checkers` use the defined checkers under vars. got it.
Might be a good idea to use it here.

raveit65:
I enabled `build pushed branches` for mate-panel. I like to neither both travis and actions can run together.

raveit65:
Travis-ci run nicely on this PR. I will disabled it again until this PR is ready go.

cwendling:
> Maybe it make sense to include archlinux build to github-actions? Archlinux was part of the origin travis config and disabled when travis-ci wasn't free anymore a while ago. Build dependencies are listed in .build.yml.

Added, although I had a bit of a hard time with pacman, it works now.

.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      # For ArchLinux
+      - name: Install dependencies
+        if: ${{ startsWith(matrix.container, 'archlinux:') }}
+        # don't upgrade, although told otherwise (see link below), because
+        # apparently in the container it doesn't quit work...
+        # https://wiki.archlinux.org/title/System_maintenance#Partial_upgrades_are_unsupported
+        run: |
+          pacman --noconfirm -Sy
+          pacman --noconfirm -S ${ARCH_BUILD_DEPS}
```
cwendling, line 130: If anybody knows better, we can try again `-Syu` as recommended, but I never used Arch and it seems it has issues in the docker container, and I had no idea what that meant.

raveit65: @zhuyaliang @yetist ^^^^

.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
...
+        run: |
+          cd mate-desktop
+          NOCONFIGURE=1 ./autogen.sh
+          { ./configure --prefix="$BASE_INSTALL_PATH" || { cat config.log; exit 1; } ; }
+
+      - name: Build mate-desktop
+        run: make -C mate-desktop -j ${{ env.JOBS }}
+
+      - name: Install mate-desktop
+        run: make -C mate-desktop -j ${{ env.JOBS }} install
```
cwendling, line 159: Ideally we'd cache all this, because it usually won't matter (well, until the distro updates in an incompatible manner maybe…)

.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: Build
+        run: make -C mate-panel -j ${{ env.JOBS }}
+
+      - name: Run Tests
+        run: make -C mate-panel -j ${{ env.JOBS }} check
+
+      - name: Run distcheck
+        if: ${{ startsWith(matrix.container, 'debian:') }}
+        run: make -C mate-panel -j ${{ env.JOBS }} distcheck
```
cwendling, line 175: As mentioned in the commit comment, I chose to only do distcheck on on system, as I see a lot less value in doing this everywhere.  But we can enable it back for every one if need be.

.github/workflows/build.yml
```diff
@@ -0,0 +1,195 @@
...
+    gobject-introspection
+
+jobs:
+  build:
+    name: Build on ${{matrix.container}} with in-process=${{matrix.in-process}}
+    runs-on: ubuntu-latest
+    container: ${{matrix.container}}
+
+    strategy:
+      fail-fast: false
```
cwendling, line 94: This line means that we'll keep other builds in the matrix running even if one fails.  Is it needed, though?  Upside is that we can see if it's distro-specific, downside is that if it's a plain code mistake it'll run other ones for nothing…

.github/workflows/build.yml
raveit65: Building a dep installer package might be a solution.
In gitlab-ci there was a caching option iwhre you could cache a result until it was overwritten.
githup and gitlab CI is very similar . i will search for this
 .

.github/workflows/build.yml
raveit65: I think it's sufficent to run run distcheck only one time.

.github/workflows/build.yml
raveit65: We should catch distro-specific issues, ihmo.

raveit65:
> Is it easy to do it per-branch? I'm afraid it's easy to forget as well, and probably would still run a first unwanted duplicate, wouldn't it?

You can only do it for the whole repository in settings, than it is disable for all PRs at this time, which isn't optimal.
I would simply comment out builds in .builds.yml for a pull request and remove that before merging ;)
But i am really thinking that 2 builds on fedora for every push in a PR is ok for our travis credits.

.github/workflows/build.yml
raveit65: I found this cache for actions https://github.com/actions/cache
reading.......

raveit65: I think this can be used. Actions  `artifacts` or `cache`. see https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#comparing-artifacts-and-dependency-caching

cwendling: Yeah I know it's doable, I just didn't spend the time yet to understand how to use it here, and it looks OK but not entirely trivial

.github/workflows/build.yml
cwendling: well, twice (one with in-process and out-of-process), as it might affect this, but yeah.

.github/workflows/build.yml
cwendling: In either case we will, the only question is whether one build failure cancels the other ones, or whether they all finish.

raveit65: I meant when eg. the job debian-testing in-process failed all other jobs should finished if possible.
Otherwise i have to look in all job logs to see which job failed. Or do i see something like canceled because a other job failed?

cwendling: I guess you'd see other jobs were cancelled (and possibly why?), but I'm not entirely sure.

.github/workflows/build.yml
raveit65: I think we have to use the cache option, which cache a result from master (main) branch to all child branches in a directory. This result can be restored with a key.
But they explained it better than me with examples.

cwendling: Thanks, I finally managed to get some caching working with the documentation you linked.
I hope the cache key is precise enough not to cause issues later (esp. will it be OK with rolling release distros?), but we can always change the cache key or drop it altogether it it causes issues.

But caching did shave off a bit of build time :)

raveit65: > ............ I hope the cache key is precise enough not to cause issues later (esp. will it be OK with rolling release distros?), but we can always change the cache key or drop it altogether it it causes issues.
> 
> But caching did shave off a bit of build time :)

Amazing :)
Using a developer release as build dependency is really a special case.
When ubuntu:rolling use a new mate major release eg. 1.28 we need to update github-ci config, and split out the updated distros in a own config part.
Or we need a condition in configuration. Something like....
`If package in repos >= 1.28 than do not use cached package`
Not sure if this is possible......only an idea.
With travis-ci we had the same issue in past. CI is a static config which needs interaction from time to time.
Usually Archlinux is the first distro (we don't use fedora-rawhide) which ship new Mate releases or new packages releases which can cause surprises.

raveit65:
Can you add more comments to parts of configuration?
This will make the config more understandable and it's easier to use for other team members as template for our other repos.
Personally i do not understand this part.
```
 - name: ccache
        uses: hendrikmuhs/ccache-action@v1.2
```

.github/workflows/build.yml
```diff
@@ -0,0 +1,200 @@
...
+    make
+    mate-common
+    meson
+    yelp-tools
+    iso-codes
+    gobject-introspection
+  # TODO
+  DEB_SCAN_BUILD_DEPS: |
+    clang
+    clang-tools
```
raveit65, line 62: Is scan-build really used?
I don't see it in build logs

.github/workflows/build.yml
raveit65: I would prefer to let other job finished. If other will be stopped because of an error of one job we don't know that all other jobs are free from errors.

.github/workflows/build.yml
raveit65: Ok, is see there is a ToDO :)
Can you mark the dependencies which are needed to build mate-desktop package?
Same what i did in travis config. https://github.com/mate-desktop/mate-panel/blob/master/.build.yml#L25

.github/workflows/build.yml
zhuyaliang: What is the error content

.github/workflows/build.yml
cwendling: Yeah, that was my thinking, so we can leave it as is

cwendling:
> Can you add more comments to parts of configuration? This will make the config more understandable and it's easier to use for other team members as template for our other repos. Personally i do not understand this part […]

Should be improved.  FWIW, it is to cache the [ccache](https://ccache.dev/)… cache, so later ccache runs can benefit from it.

.github/workflows/build.yml
cwendling: It was around there: https://github.com/mate-desktop/mate-panel/actions/runs/6867238053/job/18675135924?pr=1415#step:6:113.  Apparently it works, so either it was a transitive error, or I was doing something else wrong and though this was fatal (possible).
Anyway, it seems to work now, so we can probably resolve this.

cwendling:
OK so apart from whether or not we should run scan-build, I think everything that was raised has been addressed.

Cppcheck report has been improved a lot, but is quite slower now (5+ minutes, going from fastest to slowest); but I think the improved accuracy is worth it.

.github/workflows/build.yml
raveit65: Personally i read only build logs with their warnings in CI, so i can live pretty well without scan-build.
Of course it's enough to use it for one docker container if we want to use it.

raveit65:
@cwendling 
I this ready to go?

As i said before i can pretty well live without scanbuild.

raveit65 (approved):
LGTM

lukefromdc:
Any news on this one?

cwendling:
¾ or a year later and minimal updates & rebase, I think we can merge this.  I don't see any obvious things to change that haven't been mentioned (that is, potentially scan-build, see above), and it'd be useful in daily use I'd think.

lukefromdc:
Fine by me

PR #1418, opened 2023-11-15
Various cleanups

cwendling:
All this was mostly found by cppcheck while I tried to improve its output in #1415.  Not everything it reports is taken care of here, but a lot is.

### DIFF
--- a/mate-panel/panel-widget.c
+++ b/mate-panel/panel-widget.c
@@ -1504,7 +1504,6 @@ panel_widget_size_allocate(GtkWidget *widget, GtkAllocation *allocation)
 			AppletData *ad = list->data;
 			GtkRequisition chreq;
 			const char *id;
-			AppletInfo *info;
 			int position = ad->pos;
 			PanelObjectEdgeRelativity edge_relativity = PANEL_EDGE_START;
 			gboolean right_stuck = FALSE;
@@ -1526,7 +1525,7 @@ panel_widget_size_allocate(GtkWidget *widget, GtkAllocation *allocation)
 			id = mate_panel_applet_get_id_by_widget (ad->applet);
 			if (id)
 			{
-				info = mate_panel_applet_get_by_id (id);
+				AppletInfo *info = mate_panel_applet_get_by_id (id);
 				position = g_settings_get_int (info->settings,
 							       PANEL_OBJECT_POSITION_KEY);
 				edge_relativity = g_settings_get_enum (info->settings,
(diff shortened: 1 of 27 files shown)


lukefromdc (approved):
No build issues, and I tested panel lockdown, disabling force-quit, changing panel backgrounds (both image and color), the panel run dialog, adding/removing clock/calendar locations, and tested wayland as well. The multimonitor behavior was same as before: panel on primary monitor in x11, and on wayland on whichever monitor a terminal it is called from is on or the leftmost monitor if called by the session.

This is but a quick test, but it looks good on my end and addresses lots of "papercut" bugs in the code itself. Lots to test, but would be more if the same work was split into a dozen PR's that have to be pulled separately.

mate-panel/panel-widget.c
```diff
@@ -1693,7 +1690,7 @@ panel_widget_size_allocate(GtkWidget *widget, GtkAllocation *allocation)
 				prior_space = ad->constrained;
 				if (list->prev)
 				{
-					pad = list->prev->data;
+					AppletData *pad = list->prev->data;
```
raveit65, line 1693: This gives me a build warning in travis-ci (cppcheck) https://app.travis-ci.com/github/mate-desktop/mate-panel/jobs/613338178#L5099
```
mate-panel/panel-widget.c:1693:18: style: Variable 'pad' can be declared as pointer to const [constVariablePointer]

     AppletData *pad = list->prev->data;

                 ^
```

raveit65:
Beside from the new warning it seems that the panel runs fine in quick test.

lukefromdc:
What is the status of this? due to the number of file changes we may have to rewrite either this or other PR's to deal with merge conflicts, so I would prefer to have this finished and out of the way. The cleanups improve the quality of the code, and the cppcheck warning is but advice on how code could be further improved, not a build warning. Holding on merging this as dealing with that is an unanswered question at this time

mate-panel/panel-widget.c
cwendling: Should be fixed (although it's kind of unrelated, it's good to have less mutables)

cwendling:
@lukefromdc IMO it's ready.  I addressed @raveit65's comment although it's not really new from this PR, but my additional commit just marks a bunch of pointers const, so either the compiler will emit new warnings, or it's almost certainly free of behavioral changes.

lukefromdc:
This works fine, save for not containing the pager hack which for obvious reasons in in a different PR

lukefromdc:
Accidental closure on comment

PR #1419, opened 2023-11-20
wnck-pager: update workspace switcher aspect ratio 1.26

raveit65:
fixes https://github.com/mate-desktop/mate-panel/issues/1230 This is needed because of an update in libwnck
https://gitlab.gnome.org/GNOME/libwnck/-/commit/3456b747b6381f17d48629dd8fdd4d511e739b10

With that commit applets should be build out-of-process to avoid issues with windows-list and tray applets.
See https://github.com/mate-desktop/mate-panel/pull/1417 for more infos.

@cwendling @lukefromdc 
Please do a review because commit is a bit different to 1.27 version.
Not sure if we really should merge it into stable branch now until the issues with windows-list and tray applets  aren't fix, when applets are build in-process.
But i need this for fedora and i can use it without merging.

### DIFF
--- a/libmate-panel-applet/mate-panel-applet.c
+++ b/libmate-panel-applet/mate-panel-applet.c
@@ -1146,15 +1146,21 @@ mate_panel_applet_get_preferred_height (GtkWidget *widget,
 static GtkSizeRequestMode
 mate_panel_applet_get_request_mode (GtkWidget *widget)
 {
-	/* Do not use GTK_SIZE_REQUEST_HEIGHT_FOR_WIDTH
-	 * or GTK_SIZE_REQUEST_WIDTH_FOR_HEIGHT
-	 * to avoid problems with in-process applets
-	 * when the panel is not expanded
-	 * See https://github.com/mate-desktop/mate-panel/issues/797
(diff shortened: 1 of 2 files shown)


lukefromdc (approved):
Just tested this, behavior of the switcher and of the window list seems to be identical to that of the 1.27 PR. I see no reason not to merge this and build 1.26 with the window list out of process for distros, anyone running/testing with wayland is going to want the latest 1.27 versions anyway.

lukefromdc:
Also note that in switching between 1.26 and 1.27 versions you will get applet order changed, this will revert when you change back

raveit65:
I use this commit for fedora now and switched all panel applets to out-of-process.

lukefromdc:
Not a problem for 1.26, wayland support there is very minimal and not likely to be used

lukefromdc:
You could alternately use the code from https://github.com/mate-desktop/mate-panel/pull/1421 for this

lukefromdc:
Any updates on this?

raveit65:
Issue is already fixed by https://github.com/mate-desktop/mate-panel/commit/e284501d5707556f4fa2a885a3bfe5f6b59d599d

PR #1420, opened 2023-11-21
wnck-pager: Fix workspace switcher aspect ratio

### LINKED ISSUE #1230: Update to libwnck3 40.0-1 breaks mate-panel workspace switcher aspect ratio

cwendling:
This is a dirty hack, but it works both in-process and out-of-process, and doesn't break window list on non-expanded panels.

A better solution would be to actually fix the panel's layout code, but nobody seems to have enough time and experience with this to actually do this at the moment.

Based off an idea and code from Alberts Muktupāvels[^1], thanks.

Fixes #1230.

[^1]: https://github.com/mate-desktop/mate-panel/issues/1230#issuecomment-1046235088

### DIFF
--- a/applets/wncklet/Makefile.am
+++ b/applets/wncklet/Makefile.am
@@ -23,6 +23,8 @@ WNCKLET_SOURCES = \
 	workspace-switcher.h \
 	showdesktop.c \
 	showdesktop.h \
+	pager-container.c \
+	pager-container.h \
 	$(BUILT_SOURCES)
 
 WNCKLET_LDADD =						\
--- a/applets/wncklet/pager-container.c
+++ b/applets/wncklet/pager-container.c
@@ -1,0 +1,156 @@
+/* -*- Mode: C; tab-width: 8; indent-tabs-mode: nil; c-basic-offset: 8 -*- */
+/*
+ * Copyright (C) 2022 Alberts Muktupāvels
+ * Copyright (C) 2023 Colomban Wendling
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* See:
+ * https://github.com/mate-desktop/mate-panel/issues/1230#issuecomment-1046235088
+ */
+
+#ifdef HAVE_CONFIG_H
+#include <config.h>
+#endif
+
+#include "pager-container.h"
+#include <gtk/gtk.h>
+
+
+G_DEFINE_TYPE (PagerContainer, pager_container, GTK_TYPE_BIN)
+
+static gboolean
+queue_resize_idle_cb (gpointer user_data)
+{
+	gtk_widget_queue_resize (GTK_WIDGET (user_data));
+	return G_SOURCE_REMOVE;
+}
+
+static void
+pager_container_get_preferred_width (GtkWidget *widget,
+                                     int       *minimum_width,
+                                     int       *natural_width)
+{
+	PagerContainer *self;
+
+	self = PAGER_CONTAINER (widget);
+
+	if (self->orientation == GTK_ORIENTATION_VERTICAL)
+	{
+		/* self->size is panel width */
+		*minimum_width = *natural_width = self->size;
+	}
+	else
+	{
+		/* self->size is panel size/height, that will get allocated to pager, request width for this size */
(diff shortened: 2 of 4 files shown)


raveit65:
Sorry, i don't like the name hack in function plus mentioned name of co-author who makes always problems and and acted like a damned smartass here. Plus he involved people here.
Please change that.
I will never test or merge that like it is.

cwendling:
> Sorry, i don't like the name hack in function

I can change that, although it's a fair name.  I'll try to come up with something meaningful, but that might be tricky.  Any suggestion?

> plus mentioned name of co-author who makes always problems and and acted like a damned smartass here. Plus he involved people here.
> Please change that.
> I will never test or merge that like it is.

Isn't that a bit childish?  He did write the whole thing, so it's only fair to credit him, isn't it?  I get that you don't get along, but he also spent time suggesting alternatives solution for this issue, including this one.

I guess in practice he doesn't care, so I probably can just copy the copyright info from another file even though it's wrong (at first I had the other file's content directly in *workspace-switcher.c*, so no additional header), but I'm not entirely at ease with this.

lukefromdc (approved):
I can confirm that this works as intended, fixing the switcher without causing the last tray applet to be cut or causing problems for a nonexpanded panel.

The name "pager-hack" could be easily changed to "pager-wrapper" or something similar quite easily using Pluma's search and replace function.

As for giving credit, we have to give credit to the authors who write code if we use that code, and I don't have the skills to clean room reimplement this. Will leave to others the decision about what to do about this. I am OK with it as it stands, and open to other solutions

lukefromdc:
I pushed a separate commit renaming "pager-hack" to "pager-container" which is more descriptive. Did not squash or force-push so as to preserve the option of reverting the name change if someone doesn't like ti.

cwendling:
@lukefromdc Funnily enough I was in the process of doing the exact same renaming :smile: I updated your version stripping the "MATE" prefix because it's not common in those files and it makes the same needlessly long, plus updated 2 places that slipped through your replacing.  Otherwise, looks good :+1:

cwendling:
Regarding the crediting, or its removal thereof:

* you can do whatever you want, I won't fight anything
* what I'm more or less comfortable doing myself is one of:
  * keeping it as-is
  * de914273eea4caed66177551b3fb54d81edc7491: it's pretty ugly, but that's how I had it at first (minus the rename).  Everything is folded in *workspace-switcher.c*, so there's no additional header crediting anybody.
  * replacing the header with verbatim copies of the *workspace-switcher*'s ones.  It's incorrect, but at least coherent with the rest of the files.

I'll let you do what you like best here.

cwendling:
> I can confirm that this works as intended, fixing the switcher without causing the last tray applet to be cut or causing problems for a nonexpanded panel.

:+1: thanks for testing!

The best part is that it fixes both in-process and out-of-process IMO, as well as being a lot less intrusive -- at least before we have a thorough and proper fix for the underlying problem.

lukefromdc (approved):
This works as intended, I am fine with this or with the single-file change  at
https://github.com/mate-desktop/mate-panel/commit/1bcec01910c5005c6409103ec7e8b08b17dcb9f9
though the latter has the advantage of keeping the file list shorter

lukefromdc:
If we go with https://github.com/mate-desktop/mate-panel/commit/1bcec01910c5005c6409103ec7e8b08b17dcb9f9 we are not doing anything unusual concerning authorship or credit. If we introduce new files without credit to someone who wrote the bulk of the code, that creates problems and a troublesome precedent. 

I doubt that we have many source files that don't contain code suggested by someone other than the listed authors.

lukefromdc:
Note that neither this solution putting the container into new files nor
https://github.com/mate-desktop/mate-panel/commit/1bcec01910c5005c6409103ec7e8b08b17dcb9f9 
causes any merge conflicts with the newly merged cleanup work, so we get a break there

raveit65:
> Isn't that a bit childish? He did write the whole thing, so it's only fair to credit him, isn't it? I get that you don't get along, but he also spent time suggesting alternatives solution for this issue, including this one.

He brought pain to whole the Mate community for more than 2 years. And said in a lot of comments that he is part of gnome and he won't commit in Mate.
Also he insulted people in several discussions here in the last 6 years.
Sorry , i don't agree with with copyright in files because i don't want to reward his bad attitude.

I chooses the other solution for fedora with 1.26 and my role is downstream maintainer now.

lukefromdc:
See
https://github.com/mate-desktop/mate-panel/pull/1421
which does all of this in one file. Same as
https://github.com/mate-desktop/mate-panel/commit/1bcec01910c5005c6409103ec7e8b08b17dcb9f9
except rebased after https://github.com/mate-desktop/mate-panel/commit/b91376262abb56d95f9ec86993edaa8e9fb2cc1a

lukefromdc:
Superseded by
https://github.com/mate-desktop/mate-panel/pull/1421
which makes the same changes in one file and has been merged

PR #1421, opened 2023-11-21
wnck-pager: Fix workspace switcher aspect ratio in a single file

lukefromdc:
Put the workspace switcher into a container so it can be properly sized without changing anything else in the panel.
All changes are in workspace-switcher.c

### DIFF
--- a/applets/wncklet/workspace-switcher.c
+++ b/applets/wncklet/workspace-switcher.c
@@ -53,6 +53,153 @@
 
 #define WORKSPACE_SWITCHER_ICON "mate-panel-workspace-switcher"
 
+/* Container for the WnckPager to work around the sizing issues we have in the
+ * panel.  See
+ * https://github.com/mate-desktop/mate-panel/issues/1230#issuecomment-1046235088 */
+
+typedef struct _PagerContainer PagerContainer;
+typedef GtkBinClass PagerContainerClass;
+
+static GType pager_container_get_type (void);
+
+#define PAGER_CONTAINER_TYPE (pager_container_get_type ())
+#define PAGER_CONTAINER(obj) (G_TYPE_CHECK_INSTANCE_CAST ((obj), PAGER_CONTAINER_TYPE, PagerContainer))
+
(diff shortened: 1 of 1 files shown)


lukefromdc:
Alternative to https://github.com/mate-desktop/mate-panel/pull/1420 and works exactly the same way save this is kept in a single file for simplicity

applets/wncklet/workspace-switcher.c
```diff
@@ -522,7 +672,9 @@ static void display_workspace_names_changed(GSettings* settings, gchar* key, Pag
 
 static void all_workspaces_changed(GSettings* settings, gchar* key, PagerData* pager)
 {
-	gboolean value = g_settings_get_boolean (settings, key);
+	gboolean value = TRUE; /* Default value */
+
+	value = g_settings_get_boolean (settings, key);
```
cwendling, line 677: Seems like a rebase mishap

```suggestion
	gboolean value = g_settings_get_boolean (settings, key);
```

cwendling:
Otherwise LGTM, it's the exact same code I have in https://github.com/mate-desktop/mate-panel/commit/1bcec01910c5005c6409103ec7e8b08b17dcb9f9, so nothing new.  Only issue is that it accidentally reverts one change from #1418

applets/wncklet/workspace-switcher.c
lukefromdc: I see that, not sure how that happened but will fix it. Since` value` it will be unconditionally assigned whatever comes back from `gboolean value = g_settings_get_boolean (settings, key);` even if that somehow turns out to be garabge, the initialization to TRUE should never be used, so we save one assignment here

lukefromdc: Fixed and force-pushed, thanks for catching it!

cwendling (approved):
LGTM, works fine here.

cwendling:
Cherry-picket to 1.26: e284501d5707556f4fa2a885a3bfe5f6b59d599d

PR #1425, opened 2023-12-28
Update opacity when changed gsettings color value

### LINKED ISSUE #1408: Panel Properties ->  Background [tab]: gsettings does not update the "Style" (transparent-opaque) knob
#### Expected behaviour
The "Style" knob should be placed in the center.

#### Actual behaviour
The "Color" is updated, but not the "Style" knob:

Screenshot at 2023-10-05 20-34-12

Set solid color

Screenshot at 2023-10-05 20-33-35

* Run this from the terminal: `gsettings set org.mate.panel.toplevel.background:/org/mate/panel/toplevels/top/background/ color 'rgba(0,0,0,0.500)'`

#### MATE general version
1.26.0

#### Package version
1.26.2-1

#### Linux Distribution
Ubuntu MATE 22.04

zhuyaliang:
Fix #1408

### DIFF
--- a/mate-panel/panel-properties-dialog.c
+++ b/mate-panel/panel-properties-dialog.c
@@ -737,6 +737,21 @@ panel_properties_dialog_update_background_image (PanelPropertiesDialog *dialog,
 }
 
 static void
+panel_properties_dialog_update_opacity (PanelPropertiesDialog *dialog,
+                                        gdouble                percentage)
+{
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
This is simple and it works. The slider is now updated in realtime while showing if a color panel background is changed using gsettings from command line etc.

PR #1426, opened 2024-01-06
wayland: keep window list menu off space not used for buttons

lukefromdc:
*Fix the window list menu replacing the panel context menu between the end of the window list and the next applet when running under wayland


lukefromdc:
This has been a significant "paper cut" bug in running the wayland session, this fixes it

PR #1429, opened 2024-01-25
run-dialog: Don't needlessly override DISPLAY environment variable

### LINKED ISSUE #1387: wayland: panel run dialog cannot start xwayland apps
#### Expected behaviour
when running on wayland, starting an app such as GIMP (still on GTK 2) should bring it up in xwayland

#### Actual behaviour
non-wayland supporting apps won't open from the panel run dialog, if the panel is run from a terminal we get the same "cannot open display" messages we had in terminal until https://github.com/mate-desktop/mate-terminal/commit/79c6912d76612479a90f754983c00d3724722e28

Start a wlroots-supporting wayland session such as wayfire
start mate-panel within that session or set up the session to autostart mate-panel (same results)
add the run dialog applet to the panel
click on the run dialog applet
enter the name of any gtk2 or other non-wayland supporting app, it does not open 
and if the panel is running from a terminal look for the "cannot open display" message

#### MATE general version
1.27

#### Package version
mate-panel from git master as of 7-19-2023

#### Linux Distribution
Debian Unstable with locally compiled gtk 3.24.24 and wayfire 0.8 from git master

None as this is a local build

cwendling:
This code got introduced with 3c21f66bf30810fcb25dfe8c3c0960df270dc416 trying to mimic `gdk_spawn_on_screen()`, but does not: what `gdk_spawn_on_screen()` did (on X11) was to append the screen number to the display, so that it worked on a multi-screen setup where the app is not on the default X screen.

However, the code that got introduced just sets the display, which is gonna be already there anyway (yet possibly loosing the default screen in the process?), and AFAIK GDK dropped support for multi-screen setups anyway (not to be confused with multi-monitor with e.g. Xinerama).

We should actually be better of without any of this, as it most likely doesn't do what it is supposed to, nobody needs what it failed to do, and it breaks XWayland on Wayland.

Fixes #1387.

:warning: I didn't actually test that much, and not at all on Wayland.  But it should help.

### DIFF
--- a/mate-panel/panel-run-dialog.c
+++ b/mate-panel/panel-run-dialog.c
@@ -407,15 +407,6 @@ command_is_executable (const char   *command,
 	return TRUE;
 }
 
-/*
- * Set the DISPLAY variable, to be use by g_spawn_async.
- */
-static void
-set_environment (gpointer display)
-{
-  g_setenv ("DISPLAY", display, TRUE);
-}
-
 static void
 dummy_child_watch (GPid         pid,
 		   		   gint         status,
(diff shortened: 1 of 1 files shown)


cwendling:
CC @infirit as you were the one authoring that commit, you might have more insight (or not, I don't know)

lukefromdc (approved):
This works, and lets me use xwayland apps from the panel run dialog.This simplifies the code too. I have the run dialog button added to the panel since keybinding support on wayland is spotty

lukefromdc:
Just checked on x11, no problems there either. Will merge

PR #1439, opened 2024-04-22
wncklet: Remove icon loader function from window list applet

vkareh:
This function only loads the icon during the startup-notification, at
which point it gets replaced by the actual icon read in libwnck. Moving
forward the loading function should become part of libwnck.

Related: https://gitlab.gnome.org/GNOME/libwnck/-/merge_requests/55

### DIFF
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -781,55 +781,6 @@ static void applet_size_allocate(GtkWidget *widget, GtkAllocation *allocation, T
 		mate_panel_applet_set_size_hints(MATE_PANEL_APPLET(tasklist->applet), size_hints, len, 0);
 }
 
-#ifdef HAVE_X11
-/* Currently only used on X11, but should work on Wayland as well when needed */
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
I tested this in both the x11 and wayland sessions, didn't notice any visible change. In both cases, and both with and without this applied, I opened libreoffice to get a delay and watched the icon in the taskbar button. This gave first the generic "icon-missing" icon, then the general libreoffice icon, and finally the icon libreoffice uses with the filetype in question, a quick three-stage process. Not sure if all three or only two icons (one icon changover) in x11 as I only noticed the third one in the wayland session, but in no case did applying this PR change anything I could see. 

It is possible of course that I may not have startup notification working properly and would not notice this given its not used in wayland.

At any rate, this looks good to go on my end, and it's a code simplification as well

zhuyaliang:
@vkareh Can we merge this PR now

raveit65:
@vkareh 
libwnck-43.2 ist released. Can you update you're MR at gitlab please? https://gitlab.gnome.org/GNOME/libwnck/-/merge_requests/10
Or is this obsolete?

PR #1446, opened 2024-06-18
status-notifier: Plug large leak with icons from paths

### LINKED ISSUE #1444: notification_area applet memory leaks if the tray icon is updated constently
#### Expected behaviour
No memory leaks.

#### Actual behaviour
mate-panel will memory leak really fast until no ram is left.
Video showcasing the problem: https://www.youtube.com/watch?v=zjMsdzgMdvE

1. Clone and compile [runcat-tray](https://github.com/chux0519/runcat-tray)
2. Open runcat-tray while using mate-panel with the notification area applet.

runcat-tray will call `app_indicator_set_icon` every tick to update the tray icon, this will result in mate-panel memory leaking.
I haven't tested with others apps and/or checked the notification area applet code yet.

#### MATE general version
MATE 1.28.2

#### Package version
mate.mate-panel 1.28.1

#### Linux Distribution
NixOS unstable 24.11.20240612.57d6973 (Vicuña) x86_64

This issue doesn't seem to be distribution related, I will check on linux mint tomorrow and return results if necessary.

cwendling:
When loading an icon from a path rather than the icon theme, we scale it down to match the scale we need[^1], but we leaked the original pixbuf in this case.

Fixes #1444.

[^1]: See commit e5c8646d8f6167fe2d3634a4604a1bf9ec8ee5af

### DIFF
--- a/applets/notification_area/status-notifier/sn-item-v0.c
+++ b/applets/notification_area/status-notifier/sn-item-v0.c
@@ -287,7 +287,12 @@ update (SnItemV0 *v0)
           if (pixbuf && icon_size > 1)
             {
               /*An icon specified by path and filename may be the wrong size for the tray */
-              pixbuf = gdk_pixbuf_scale_simple (pixbuf, icon_size-2, icon_size-2, GDK_INTERP_BILINEAR);
+              GdkPixbuf *scaled_pixbuf;
+
+              scaled_pixbuf = gdk_pixbuf_scale_simple (pixbuf, icon_size-2, icon_size-2, GDK_INTERP_BILINEAR);
+              g_object_unref (pixbuf);
+              pixbuf = scaled_pixbuf;
+
               surface = gdk_cairo_surface_create_from_pixbuf (pixbuf, scale, NULL);
             }
           if (pixbuf)


lukefromdc (approved):
This builds and runs fine. Running an old test case than changes an indicator icon every  second for an unrelated bug, I did not see any rapid RAM use runup, though with other memory leaks anywhere on the system it's hard to be sure.

The code change however should definately work.

lukefromdc:
Just rebased against master for the travis fix, will merge if that passes as we now have a second successful test and one which verifies this fixed the memory leak at
**https://github.com/mate-desktop/mate-panel/issues/1444#issuecomment-2176307508**

lukefromdc:
Cannot update translations, the new tzm translation files still generate errors. Thus will release 1.28.2 with the existing translations. This has broken updating translations in every package

lukefromdc:
cherrypicked to 1.28 and release 1.28.2 with existing translations due to build failure if translations are updated.

lukefromdc:
Anyone have a problem with cherrypicking to 1.26 as well? Lots of distro versions shipping that won't be updating to 1.28. Please advise

cwendling:
@lukefromdc fine with me, it's a big enough leak that it's worth backporting, and it should apply cleanly there as well.

lukefromdc:
Will do, thanks

lukefromdc:
On trying to update travis.yml  so a release would be possible I got this error:
`The command "./docker-build --name ${DISTRO} --config .build.yml --install" failed and exited with 1 during`
Will try reverting the travis change and cherrypicking to the 1.26 branch and see what happens.

lukefromdc:
./autogen.sh  --with-in-process-applets=wncklet,clock,fish,notification-area --prefix=/usr libdir=/usr/lib/x86_64-linux-gnu --sysconfdir=/etc --enable-shared=yes --enable-static=no --disable-maintainer-mode --disable-schemas-compile

lukefromdc:
Same travis error with or without the deploy fix. I do not know how to fix that, so until that is fixed I cannot do a release for 1.26 carrying this cherrypick

raveit65:
@lukefromdc 
Try `dist: jammy`, see master/1.28 branch

lukefromdc:
That solved the problem, will release 1.26.5 again using the existing translations due to breakage in seemingly all repos with the new ones

lukefromdc:
Fixed, 1.26.1 (after the other travis fix, which I had to fiddle with) is out so all recent distros should pick this up

Issue #1451, opened 2024-10-04
Clock applet: 12-hour time format broken with musl libc


alterae:
#### Expected behaviour
Setting the time format option for the clock applet to twelve-hour display should display the time in a twelve-hour format.

#### Actual behaviour
Setting the time format option for the clock applet to twelve-hour display causes the time string to display as `???`.

#### Steps to reproduce the behaviour
1. Right click the clock applet.
2. Open the preferences menu.
3. Select twelve-hour time.

#### MATE general version
`mate 1.26.0_1`

#### Package version
`mate-panel 1.28.2_1`

#### Linux Distribution
[Void](https://voidlinux.org) **(with musl libc `musl 1.1.24_21`)**

#### Link to bugreport of your Distribution (requirement)
<https://github.com/void-linux/void-packages/issues/52494>

alterae:
Additional information:

I am able to work around the issue by setting a custom time format string through dconf. Strings containing the `%p` specifier (uppercase AM/PM) work as expected, while strings containing `%P` (lowercase am/pm, and a [GNU extension][1]) display as `???`.

[1]: https://sourceware.org/glibc/manual/latest/html_mono/libc.html#Formatting-Calendar-Time-1

sgn:
This report provides false analysis. The problematic format specifier is `%l`. 

https://github.com/mate-desktop/mate-panel/blob/1de955b0b3a80cbc1bd840bad312f8ad3f0ae99b/applets/clock/clock.c#L461

A possible patch is:
```diff
Index: mate-panel-1.28.2/applets/clock/clock-location-tile.c
===================================================================
--- mate-panel-1.28.2.orig/applets/clock/clock-location-tile.c
+++ mate-panel-1.28.2/applets/clock/clock-location-tile.c
@@ -428,6 +428,7 @@ format_time (struct tm   *now,
 
         if (local_now.tm_wday != now->tm_wday) {
                 if (clock_format == CLOCK_FORMAT_12) {
+#ifdef __GLIBC__
                         /* Translators: This is a strftime format string.
                          * It is used to display the time in 12-hours format
                          * (eg, like in the US: 8:10 am), when the local
@@ -435,6 +436,15 @@ format_time (struct tm   *now,
                          * (the %A expands to the weekday). The %p expands to
                          * am/pm. */
                         format = _("%l:%M <small>%p (%A)</small>");
+#else
+                        /* Translators: This is a strftime format string.
[...]

PR #1452, opened 2024-10-05
applets/clock: Fix compatibility with musl libc

sgn:
%l is an glibc extension for strftime, this extensions isn't supported by musl libc.

Replace %l with %_I, which is functionaly equivalence. %_I itself is also an extensions, however, %_I is supported by both glibc, musl, FreeBSD and deriviates, AIX, and Solaris.

Fix: #1451

### DIFF
--- a/applets/clock/clock-location-tile.c
+++ b/applets/clock/clock-location-tile.c
@@ -434,7 +434,7 @@ format_time (struct tm   *now,
                          * weekday differs from the weekday at the location
                          * (the %A expands to the weekday). The %p expands to
                          * am/pm. */
-                        format = _("%l:%M <small>%p (%A)</small>");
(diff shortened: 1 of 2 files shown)


lukefromdc:
I cannot test this except for breaking changes with glibc, we don't have musl libc on Debian Unstable

sgn:
Does it break for glibc? We have musl over here, and it seems to work correctly.

lukefromdc:
I can test under libglibc when I get home

lukefromdc (approved):
Didn't get any problems with this using glibc in Debian Unstable, so we seem to be good on that end

lukefromdc:
Merging as musl libc users need it and nothing broke in my tests. If this breaks anywhere it should break everywhere and it did not.

PR #1453, opened 2024-10-15
Clock Applet: Add scrollbar to Locations lists when the lists get too long

thesquash:
Before this pull request, the Clock Applet's popup window would expand off the screen if the user added more than a few locations to the Locations list. In addition, the list of locations in the Clock Applet Preferences could get extremely tall as well, and could even push the bottom part of the Preferences dialog off the screen.

This PR adds (vertical) scrollbars to both widgets.  Here's how the Clock Applet popup looks now:
mate-panel-clock-applet-locations-with-scrollbar

And here is how the Preferences dialog looks now:
mate-panel-clock-applet-preferences-locations-with-scrollbar

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1248,6 +1248,7 @@ location_tile_need_clock_format_cb(ClockLocationTile *tile, gpointer data)
 static void
 create_cities_section (ClockData *cd)
 {
+        GtkWidget *cities_box;
         GSList *node;
         GSList *cities;
         GSList *l;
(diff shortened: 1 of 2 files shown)


lukefromdc:
Issue confirmed here

lukefromdc (commented):
The scrollbar works, the total size of the calendar window still pushes about half of the top bar (where the "edit" button is found) off the screen
Top_Right_of_Screen_10-15-2024

lukefromdc:
Any updates on this? It would be good to fix the issue of cutting off half the top bar when the list is too long, but at least this ensures nobody loses access to the edit button and finds themselves unable to restore access to it. We do in fact need this one;

vkareh:
@lukefromdc I cannot reproduce your issue. I can keep adding a bunch of locations and all it does is grow the scroll box. The 'Edit' button remains in the same place as always:

Screenshot at 2025-04-23 08-47-35

lukefromdc:
Might be something specific to my system then. Note that I have the clock on a bottom panel
It's still an important fix even if others get what I  got because at least now the edit button is always available,
so I am OK with merging this as-is and worrying about the cutoff top of the button later

vkareh:
I tested with the applet on the bottom and it still works:

Screenshot at 2025-04-23 17-46-23

The only thing I can think of is that you have multiple screens. Not sure if you remember that issue where having a mate-panel on the top of the bottom screen (or on the bottom of the top screen) causes issues in calculating the area of maximised windows? I wonder if it's related to that - panel struts, I think it was called.

vkareh (approved):
LGTM

lukefromdc:
Pretty sure test was with one screen only BUT I might have found a wayland-only issue with that or even a sway/wayfire-only issue
with computing window dimensions. Considering this good to go as it prevents a one-way addition to the list that is not easily
reversed when the button is fully hidden with the current release/master code

lukefromdc (approved):
Just confirmed the problem I had with half the edit button getting cut off is wayland-only, this works fine in x11. Will merge, we can worry about a half button that still functions ion wayland later

PR #1454, opened 2024-10-15
ci: Bump dpl branches for Ruby 3 support

correctmost:
The CI error on master:
```
/home/travis/.rvm/gems/ruby-3.3.5/gems/cl-1.2.4/lib/cl/opts.rb:112:in `block in taint': undefined method `taint' for an instance of String (NoMethodError)

          [key, self[key] && self[key].secret? ? value.taint : value]
                                                      ^^^^^^
```

That error with Ruby 3 was fixed here: https://github.com/travis-ci/travis-cl/commit/e17a1b2bd2fe51eb9cc4fd90d19c8a12997717a0#diff-ee038a05d1feba327913a56ad3533ada6845400f2aa2a1b2b2d2584d4e2852ba

I'm hoping that bumping from https://github.com/travis-ci/dpl/releases/tag/v2.0.3-beta.4 to https://github.com/travis-ci/dpl/releases/tag/v2.0.5 pulls in the fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ before_deploy:
 deploy:
   - provider: pages
     edge:
-      branch: v2.0.3-beta.4
+      branch: v2.0.5
     token: $GITHUB_TOKEN
     keep_history: false
     committer_from_gh: true
@@ -59,7 +59,7 @@ deploy:
       condition: ${DISTRO} =~ ^fedora.*$
   - provider: script
     edge:
-      branch: v2.0.3-beta.4
+      branch: v2.0.5
     script: ./docker-build --verbose --config .build.yml --release github
     on:
       tags: true


lukefromdc (approved):
I tested this by copying the changes to a "test-travis" branch so Travis would run a job and it worked. I will know if it generates successful releases when I do a release with it, which will be the 1.28 release carrying the newest bugfix.

Will merge, thanks. Will have to copy this to every repo if it generates a successful release

lukefromdc:
Got an error that shows up only on releases:

`You have already activated logger 1.6.0, but your Gemfile requires logger 1.6.1. Since logger is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports logger as a default gem. (Gem::LoadError)`

lukefromdc:
Just did that, let's hope 1.28.4 gets out

lukefromdc:
Thanks, that worked and 1.28.4 is out

PR #1473, opened 2025-06-02
Update layout of clock window

luigifab:
Updates the layout of the list to fill the window.
Tested by rebuilding MATE 1.27 Debian source package.

before

### DIFF
--- a/applets/clock/clock.ui
+++ b/applets/clock/clock.ui
@@ -600,6 +600,7 @@
                 <child>
                   <object class="GtkBox" id="hbox54">
                     <property name="visible">True</property>
+                    <property name="vexpand">True</property>
                     <property name="can-focus">False</property>
                     <property name="spacing">6</property>
                     <child>


lukefromdc (commented):
No runtime issues with this, but with my font and theme I did not notice any visible change with this due presumably to how the widgets were already packing space-wise. At least in wayfire this is a non-resizable window.

If this fixes a problem in any theme or font combination I am fine with merging it

luigifab:
Maybe because you have a lot of places, and there is no scrollbar?

image

oz123:
I believe the change is modifying this:

image

See the gray area below the latest location to the bottom of the box above the buttons. The change suggested looks OK to me.

oz123 (approved):

lukefromdc:
Just retested, found out this is theme dependent. The bottom row buttons do not evenly expand in green-submarine, GreenLaguna, the Traditional themes, BlackMATE, or my own UbuntuStudio_Legacy theme which shares much code with BlackMATE. Works in all of the Menta themes, Adwaita, and the Yaru themes.

luigifab:
I added a before screenshot in in my first message. But I don't understand why it's theme dependent.

lukefromdc:
My guess would be something in a theme setting a padding between two of the buttons. Note that whatever
we set in the code is a minimum but is expandable by themes. Anyway this doesn't break anything

lukefromdc:
Any updates on this?

PR #1479, opened 2025-06-26
Make mate-panel build with X11 disabled (including notification area)

mark-herbert42:
Some proof of concept update to make mate-panel buildable (and usable) without X11 at all. 
Built and tested with X11 and wayland, wayland only and X11 only.

No functionality improvements  - just fixed applets that did work on wayland before but required X11 to compile by placing X11 specific code under #ifdef HAVE_X11 in *.c *.h  files and if ENABLED_X11 in makefiles.  

Also reimplemented GDK_IS_X11_DISPLAY using GDK_IS_WAYLAND_DISPLAY in the case if gdkx.h not there.

### DIFF
--- a/applets/Makefile.am
+++ b/applets/Makefile.am
@@ -1,7 +1,9 @@
 SUBDIRS =			\
 	clock			\
 	fish			\
-	wncklet
+	wncklet			\
+	notification_area
+
 
 if ENABLE_X11
 SUBDIRS += \
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -60,6 +60,11 @@
 #include <gdk/gdkx.h>
 #endif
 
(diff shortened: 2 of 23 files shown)


mark-herbert42:
I guess there will be more components that are not yet buildable with wayland only - so there might be more work to do. 
In fact the runtime display system is determined using  GDK_IS_X11_DISPLAY which is a part of gdkx.h, so all the components using it need to be reviewed.  But that's still work in progress - can not do all at one time.

I have mate-panel build with wayland only support, and managed to build mate-power-manager (useless - does not work with wayland but it contains applet which builds) , and mate-media - also containing applet (volume-control-applet)  that compiles and works.  So mate-panel with this merge did not affect the other components which are still requiring X11 libs and includes to build.

Issue #1477, opened 2025-06-25
Wayland - drag and drop for moving panel works terribly.


mark-herbert42:
On Xorg session the panel simply follow the cusor and can be dragged and dropped to any position.  With wayland panel is not shown - we get cursor hand and panel "appears" as a preview when cursor reaches suitable position (top, bottom, feft or right) - but panel has a tendency to land on top. Even if preview was at bottom before drop, and after several attempts to drag it from top it is much easier to go to menu and select orientation there isteand of multiple failed drag-drop attemps.

proposal: disable drag-drop on wayland totally, use only menu.

lukefromdc:
I just did some experimenting with the panel as currently installed, and what I found was that to "let go of" the panel
and drop it in one of the 4 positions made possible by using gtk-layer-shell to fix it to a border, the mouse has to be
in the exact center of the panel's width (for a horizontal panel). Otherwise it jumps around.

I also found that nonexpanded panels can only be dragged to one of the 4 centered positions, but that is the expected
behavior with how gtk-layer-shell is used-and with the math I had to use to position the menus and such over applets.

mark-herbert42:
The issue I experience that the panel drag is  - I grag it, and it appears at the bottom. I releae the mouse button - and it  appears on top. Same when I move it to the right edge. Actuallly if I drag it an apply little "pressure"  on my mouse it starts blinking lef-right-left-right. Or top-bottom-top-bottom. So to make this usable you need to position the cursor exactly on the edge. Not "pushing" it over the border and not leaving any space.  So you need to be a real gaming marksman , a god of computer mice.   But it is Ok when you want panel on top or left.

It also has something to do with virtual desktops maybe - but I do not have moving cursor to the edge switching my workspaces.  Without the panel drag cursor just stuck there on the edge so no idea why panel drag do not do the same.

lukefromdc:
Duplicate of 
https://github.com/mate-desktop/mate-panel/issues/1156

lukefromdc:
try
https://github.com/mate-desktop/mate-panel/pull/1480

PR #1480, opened 2025-07-04
wayland: stop panel "dodging" mouse and moving to top or left

lukefromdc:
Disable grabbing and moving expanded panels with the mouse in wayland due to tendency of panel other than on top to jump to top or left

### DIFF
--- a/mate-panel/panel-toplevel.c
+++ b/mate-panel/panel-toplevel.c
@@ -475,6 +475,16 @@ static void panel_toplevel_begin_grab_op(PanelToplevel* toplevel, PanelGrabOpTyp
 	GdkSeat *seat;
 	GdkSeatCapabilities capabilities;
 
+#ifdef HAVE_WAYLAND
+	/*FIXME: this disables dragging the panel in wayland entirely
+	 *But it hasn't worked properly in wayland ever (panel "jumps to top or left)
+	 *and a nonexpanded panel is forcibly centered by gtk-layer-shell so cannot
+	 *be dragged along the screen edge
+	 */
+	if (GDK_IS_WAYLAND_DISPLAY(gdk_display_get_default()))
+		return;
+#endif
+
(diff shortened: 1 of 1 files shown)


lukefromdc:
Fixes
https://github.com/mate-desktop/mate-panel/issues/1156
and
https://github.com/mate-desktop/mate-panel/issues/1477

Replacement for https://github.com/mate-desktop/mate-panel/pull/1391 which cannot be reopened by restoring the PR branch which had deleted here but which I had kept locally. Rebased against current master of course

lukefromdc:
Latest force-push added a FIXME to this code as suggested in 2023 by @raveit65 , does not build this code unless wayland is enabled at build time, and looks for running under wayland rather than for not running under x11 for compatability with
https://github.com/mate-desktop/mate-panel/pull/1479
in case we want to do that

lukefromdc:
If anyone objects to this I can make it optional by a gsettings value or (with more work) a menuitem in the panel properties dialog

mark-herbert42:
Tested on wayland - works fine for me. Panel does not ocasionally got moved to top or left and stays exactly where it is specified by properties dialog.

lukefromdc:
Just rebased against master, stopped the cursor from changing to a hand when the panel cannot be grabbed, and updated the inline comments to reflect that we cannot drag a nonexpanded wayland panel along a screen edge as we have its position hardcoded to the center using gtk-layer-shell.

FIXME's included to mark the spots we have grabbing the panel disabled in case anyone ever figures out how to make dragging the panel work in wayland without it jumping back to the top or left.

lukefromdc:
@raveit65 I believe your original issues with https://github.com/mate-desktop/mate-panel/pull/1391 have been fixed here

lukefromdc:
Merging, this affects only wayland and I've been using it a month with good results

PR #1481, opened 2025-07-05
Make mate-panel build with X11 disabled (including notification area)

mark-herbert42:
Same pull as before - just cleaning up branches in my fork.

### DIFF
--- a/applets/wncklet/showdesktop.c
+++ b/applets/wncklet/showdesktop.c
@@ -35,6 +35,11 @@
 #include <libwnck/libwnck.h>
 #endif
 
+#ifndef HAVE_X11
+#include <gdk/gdkwayland.h>
+#define GDK_IS_X11_DISPLAY(object)        !(G_TYPE_CHECK_INSTANCE_TYPE ((object), GDK_TYPE_WAYLAND_DISPLAY))
+#endif
+
 #include "wncklet.h"
 #include "showdesktop.h"
 
--- a/applets/wncklet/window-list.c
+++ b/applets/wncklet/window-list.c
@@ -25,6 +25,11 @@
 #include <libwnck/libwnck.h>
 #endif /* HAVE_X11 */
 
+#ifndef HAVE_X11
+#include <gdk/gdkwayland.h>
+#define GDK_IS_X11_DISPLAY(object)        !(G_TYPE_CHECK_INSTANCE_TYPE ((object), GDK_TYPE_WAYLAND_DISPLAY))
+#endif
+
 #ifdef HAVE_WAYLAND
 #include <gdk/gdkwayland.h>
 #include "wayland-backend.h"
--- a/mate-panel/panel-background.c
+++ b/mate-panel/panel-background.c
@@ -763,6 +763,7 @@ panel_background_free (PanelBackground *background)
 	background->default_pattern = NULL;
 }
 
+#ifdef HAVE_X11
 char *
 panel_background_make_string (PanelBackground *background,
 			      int              x,
@@ -799,6 +800,7 @@ panel_background_make_string (PanelBackground *background,
 
 	return retval;
 }
+#endif
 
 PanelBackgroundType
 panel_background_get_type (PanelBackground *background)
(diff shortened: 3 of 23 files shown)


lukefromdc (approved):
I got clean builds both with and without x11 support here. Also note that x11 dev libraries are successfully autodetected and if present x11 support defaults to yes as it should.

applets/wncklet/showdesktop.c
```diff
@@ -35,6 +35,11 @@
 #include <libwnck/libwnck.h>
 #endif
 
+#ifndef HAVE_X11
+#include <gdk/gdkwayland.h>
+#define GDK_IS_X11_DISPLAY(object)        !(G_TYPE_CHECK_INSTANCE_TYPE ((object), GDK_TYPE_WAYLAND_DISPLAY))
+#endif
+
```
cwendling, line 42: and again.  Still, usages of `GDK_IS_X11_DISPLAY()` are already guarded.  Best would be to update the guards to include `GDK_WINDOWING_X11`, but that's probably not usually a problem -- if it is, just do so -- and this patch isn't gonna fix it.

applets/wncklet/window-list.c
```diff
@@ -25,6 +25,11 @@
 #include <libwnck/libwnck.h>
 #endif /* HAVE_X11 */
 
+#ifndef HAVE_X11
+#include <gdk/gdkwayland.h>
+#define GDK_IS_X11_DISPLAY(object)        !(G_TYPE_CHECK_INSTANCE_TYPE ((object), GDK_TYPE_WAYLAND_DISPLAY))
+#endif
+
```
cwendling, line 32: And again. I'm gonna stop complaining about this in the rest of the diff, it's the same everywhere.

mate-panel/panel-background.c
```diff
@@ -763,6 +763,7 @@ panel_background_free (PanelBackground *background)
 	background->default_pattern = NULL;
 }
 
+#ifdef HAVE_X11
```
cwendling, line 766: What's that about (ant the other disabling of the background stuff)? Couldn't this just disable the problematic XLib piece, and still support at least color?

mate-panel/wayland-backend.c
```diff
@@ -71,6 +71,7 @@ wayland_panel_toplevel_update_placement (PanelToplevel* toplevel)
 		break;
 	case PANEL_ORIENTATION_BOTTOM:
 		anchor[GTK_LAYER_SHELL_EDGE_BOTTOM] = TRUE;
+		anchor[GTK_LAYER_SHELL_EDGE_RIGHT] = TRUE;
```
cwendling, line 74: This sound wrong, shouldn't this just follow the expand policy as done above? And if not for some reason, it feels like other orientations likely need similar treatment.

cwendling: Generally for legitimate non-obvious changes you probably should make it a bit clearer somewhere why that's done (commit message at least), and separate them from other changes (this ought not to be needed for a *build*, even if it was for a neat experience on Wayland)

cwendling: Re git/github: what you should do is create separate branches, and work on each thing separately.  For you WIP stuff, if you need the PR, just base your WIP branch on the PR branch, but don't push it.

To remove this change from this PR:
* modify the code to remove the change
* `git add mate-panel/wayland-backend.c` to record the modification
* `git commit --amend` to modify the topmost commit
* `git push -f origin wayland-only-buld` to publish the update to this PR (change "origin" if needed)

Then, create a new branch and add back your changes (just do `git checkout -b my-new-branch` and start hacking again).  If you need to modify this PR again, go back on this branch (`git switch wayland-only-buld`), hack on it, commit, push the new stuff for this PR.

lukefromdc:
Since all of the changes here are in a single commit, I never even spotted the functional change for a nonexpanded panel in wayland. An obvious reason that should NOT be here is that if it was ever necessary to revert that change for any reason (even to rewrite it)  it could not be done without also reverting the build system changes.

PR #1484, opened 2025-07-14
Fix dragging panel color

### LINKED ISSUE #870: Dragging color from caja backgrounds and emblems to mate-panel sets white background instead of the dragged color
#### Expected behaviour
As mentioned in mate-user-guide's panel help, dragging a color from a color selector or from Caja's backgrounds and emblems>colors dialog should turn the panel to that color. Dragging a pattern from there should set the panel background to that pattern,

#### Actual behaviour
Dragging a color from GIMP's color selector works, but dragging a color from Caja's backgrounds and emblems>colors dialog will set the panel background to white no matter what color is selected. Dragging a pattern works, and dragging the "return arrow" image restores the system background. Most likely this is from an incompatable color format.

Open caja's backgrounds and emblems dialog, select "colors." drag a color other than white to the panel. The panel background turns white

#### MATE general version
1.21

#### Package version
mate-panel 1.21.1 from git master current as of 9-13-2018

#### Linux Distribution
Debian Unstable with locally built GTK 3.24.0 and glib 2.58.0

None as I have all of MATE locally compiled.

luigifab:
Fix #870. Tested by rebuilding MATE 1.27 Debian source package.
Maybe the alpha value is no longer lost.

### DIFF
--- a/mate-panel/panel.c
+++ b/mate-panel/panel.c
@@ -454,7 +454,8 @@ set_background_image_from_uri (PanelToplevel *toplevel,
 
 static gboolean
 set_background_color (PanelToplevel *toplevel,
-		      guint16       *dropped)
+		      guint16       *dropped,
+		      gint          length)
 {
 	GdkRGBA color;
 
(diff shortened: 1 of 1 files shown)


lukefromdc (approved):
Builds clean and this works even in wayland where I thought all the restrictions would block this. I was able to drag both colors and patterns from the caja backgrounds and emblems dialog to the panel and they instantly became the panel background. Was not necessary to even open the panel context menu to do this. Mate-color-select works too

PR #1485, opened 2025-07-18
Wayland-only-build attempt 3 - changes from other branch removed

mark-herbert42:
Failed to clean up git mess - so recreated fork from scratcj and locally stored patches. Removed wayland backend anchors that got accidentally merged from other branch, and removed additions from workspace-switcher that is not compatible (and probably never be ) with wayland.

### DIFF
--- a/applets/Makefile.am
+++ b/applets/Makefile.am
@@ -1,7 +1,9 @@
 SUBDIRS =			\
 	clock			\
 	fish			\
-	wncklet
+	wncklet			\
+	notification_area
+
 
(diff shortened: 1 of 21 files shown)


lukefromdc (approved):
This built without issues both with and without x11 support enabled. Could not test with x11 development libraries removed as that's too complex a change on a live system

On a wayland session, running `GDK_BACKEND=x11 mate-panel --replace` failed as expected on the wayland-only build while it worked with the x11 and wayland supporting build, proving that this works

PR #1488, opened 2025-07-22
Fix GTK deprecations

vkareh:
- remove deprecated GtkStockItem: Also replace gtk_dialog_add_button with panel_dialog_add_button to ensure mnemonics and icons are preserved.
- Remove deprecated gdk_screen_get_* functions: Screen functions are replaced with their Display equivalents to remove deprecation warnings. Display functions take into account the entire display server connection, rather than a single physical monitor.
- panel-stock-icons: Remove deprecated GtkIconFactory and GtkIconSource: All applet menus already use theme icon names directly, making this legacy stock icon code obsolete. Removing it eliminates all the icon factory deprecation warnings.
- Bump GTK version to 3.24 (I suspect all distros that ship MATE are already at 3.24)


lukefromdc (approved):
This builds and runs without issue and is a nice cleanup. Thanks!

lukefromdc:
Let's see if we can get a second review on this if possible

luigifab (approved):
It looks good with 1.27.1.

PR #1507, opened 2025-09-03
Fix uninitialized and unused screen vars

vkareh:
A previous overzealous cleanup left some hanging screen variables and one instance of an uninitialized screen being used.


lukefromdc (approved):
Builds fine, and no runtime problems with either the wayland or x11 backend. Calendar popup and fish both work fine.

PR #1516, opened 2025-09-21
clock: Store calendar client in ClockData

vkareh:
This simplifies the evolution calendar client architecture and makes the popup show up faster, since we don't need to create a new one every time we open it.


vkareh:
I just pushed a new commit that adds a refresh function after a timeout. This fixes a thing where changes in Evolution data were not showing up whenever you opened the popup.

Apart from this PR, to test Evolution, you don't need to have an email account - you can just create a local calendar that has no online components: 
Screenshot at 2025-09-22 14-26-49
Then you can add whatever events and Evolution stores them locally.

For mail you can just enter a fake email (e.g. `nobody@nowhere`), select "Standard Unix mbox spool file", and point it to an empty file - basically the mail functionality becomes a no-op if you do that.

PR #1519, opened 2025-09-30
Register DBus method for interacting with applets

vkareh:
This PR registers an 'Activate' DBus method that emits an 'activate' signal that any applet can connect to and perform whatever action it means to activate in that applet.

This allows users to control certain applet functionalities through DBus, like registering new keybindings, or integrating them with other software.

As an example, I've added the signal to the clock applet in this PR, so that it can be opened/closed through a dbus command:
```
$ dbus-send --session --print-reply --dest=org.mate.panel.applet.ClockAppletFactory /org/mate/panel/applet/ClockApplet/0 org.mate.panel.applet.Applet.Activate string:"toggle-calendar" uint32:0
```
You can even add that as a custom keybinding to have a quick keyboard-centric way of viewing your calendar.

My plan is to add this to a bunch of other applets eventually.


lukefromdc:
This could be different in or out of process

lukefromdc:
This is still broken, but since the clock change can be applied in-process without breaking the problem seems to be in the libmate-panel-applet changes

lukefromdc:
Just tested applets from other packages with this.

Notifcation status (mate-notification-daemon) and the volume control applet work without a rebuild, presumably not calling whatever become ABI incompatable. The mate dictionary applet requires a rebuild, as does mate-power-manager's new mate-power-profiles applet and potentially either or both both of the other applets (not tested as running the wayland session at the moment).

I didn't find anything that was still broken after a rebuild

lukefromdc:
I got this error on running the dbus-send test command, probably a difference in how my panel layout named the applet:

`Error org.freedesktop.DBus.Error.UnknownMethod: Object does not exist at path “/org/mate/panel/applet/ClockApplet/0”`

lukefromdc:
I must have closed this trying to comment! I had absolutely NO intention of doing that. Assuming the
problem running the test command is local to my machine, this is probably ready to go

lukefromdc (commented):
running `dconf dump /org/mate/panel/applet/` on my system shows an empty directory. I found the applet instead under` /org/mate/panel/objects/object-63/` but trying to run the test command on this gave

`Object path was not valid: '/org/mate/panel/objects/object-63'`

Note that my panel configuration file has not been reset/remade from scratch in a great many years, doing so would take over an hour to reset all my customizations.

I don't know much about dbus, so I cannot really do a functional test here but I can confirm nothing breaks if all applets are recompiled with no code changes required.

vkareh:
@lukefromdc I pushed a new commit that bumps the ABI version. That way applets know to rebuild against the new change.

lukefromdc (approved):
I just got a good build of this with all my existing applets working after rebuilding mate-applets. Managing the transition of all this is out of my skillset, but the code itself seems to work so far as keeping existing functionality working.

Again, I really don't know much about how to use Dbus.

PR #1520, opened 2025-10-01
ci: Upgrade libsoup dependency from 2.4 to 3.0

vkareh:
mate-panel uses libsoup indirectly through libmateweather, which already requires libsoup-3.0. Evolution Data Server (used by the clock applet for calendar integration) also requires libsoup-3.0.

### DIFF
--- a/.build.yml
+++ b/.build.yml
@@ -51,7 +51,7 @@ requires:
     - libmateweather-dev
     - libpango1.0-dev
     - libsm-dev
-    - libsoup2.4-dev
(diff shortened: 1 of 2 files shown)


lukefromdc:
Any particular way of testing this or is this basically ready to go?

vkareh:
@lukefromdc we can manually run the github workflow scripts, but the CI already did it for us. The checks all passed (and they show libsoup-3 in the logs).

PR #1521, opened 2025-10-02
panel-toplevel: Add dynamic HiDPI scale change support

vkareh:
Panels now respond to HiDPI scale factor changes without requiring a restart. When the monitors-changed signal is received, the panel updates its scale factor and forces applets to redraw at the new scale.

The temporary size change (to 0 and back) is necessary because applets don't individually listen to scale change signals.

This eliminates the need for mate-settings-daemon to kill and restart mate-panel when the scale factor changes.

### DIFF
--- a/mate-panel/button-widget.c
+++ b/mate-panel/button-widget.c
@@ -34,10 +34,13 @@ struct _ButtonWidgetPrivate {
     guint             arrow         : 1;
     guint             dnd_highlight : 1;
     gboolean          needs_move;
+
+    gulong            scale_changed_handler;
 };
 
 static void button_widget_icon_theme_changed (ButtonWidget *button);
 static void button_widget_reload_surface (ButtonWidget *button);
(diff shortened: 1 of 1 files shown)


lukefromdc:
Builds fine. First observation from the wayland session (this mostly needs x11 testing) is no apparent change,
and the issue with launchers and only launchers not scaling with a wayland panel until a restart remains.
This of course is no surprise as in wayland the compositor does all the scaling.

lukefromdc (commented):
This works with the other two PR's applied and the panel responds to the scale change with a slight delay. On a few occasions the panel launchers did not rescale properly, going out of position and being cut off same as in a wayland scale change but most of the time they rescaled properly.

A fix for one of those issues is probably a fix for the other or at least guidance on how to fix it

lukefromdc:
Thanks, works really well now in x11. Didn't fix the wayland issue with launchers because we don't have the signal from GTK, I'll have to look into that one myself. I'll have to find a way to detect a compositor scale change and reload the buttons the same way. Merging

PR #1522, opened 2025-10-06
clock: Follow EDS calendar colors and highlight current event

vkareh:
Evolution tracks a color for each calendar. We can fetch this to differentiate events from multiple calendars in the popup.

This also marks the current event in bold and scrolls the window to make it visible to the user on load.

Screenshot at 2025-10-06 16-43-15


lukefromdc (commented):
An initial non-evolution test build showed no change of behavior, as it should be

lukefromdc (approved):
An evolution build  seems to work fine, though as I do not really know how to use evolution it took me a while to figure out how to get everything to show up. Got these results with some test entries:


Evolution_In_Calendar_10-7-2025

lukefromdc:
Found one minor nuisiance though this may be for another PR:  clicking the expander to hide the Appointments or the Tasks frame also changes the gsettings value, and to get the expander back you have to reopen the clock preferences dialog.

vkareh:
@lukefromdc 
> Found one minor nuisiance though this may be for another PR: clicking the expander to hide the Appointments or the Tasks frame also changes the gsettings value, and to get the expander back you have to reopen the clock preferences dialog.

Fixed by https://github.com/mate-desktop/mate-panel/pull/1524

PR #1524, opened 2025-10-07
clock: Separate expander state from section visibility preferences

vkareh:
The Appointments and Tasks expanders in the calendar popup were incorrectly linked to the show-calendar-events and show-tasks gsettings. This caused collapsing the section to toggle the preference, making the entire section disappear on next popup.

Fixed by adding a new expand-calendar-events and expand-tasks gsettings keys to track the UI state separately from whether the section is visible or not.


lukefromdc (approved):
This is simple and it works as intended. Merging

Issue #1531, opened 2025-11-19
Wayland window list: buttons overflow unless show desktop applet is also in the panel


lukefromdc:
A fix is being investigated for this, but for the moment if the window list is overflowing in wayland with mate-panel from recent git master, the WORKAROUND is to put the show desktop applet anywhere in the panel. I will have a fix for this as soon as I can folks!

Expected behavior:
Window list buttons in a wayland panel never overflow and push other applets off the end of the panel regardless of what other applets are in the panel

Actual behavior:
If the Show Desktop applet is in the panel, the window list works normally. If the Show Desktop applet is NOT in the panel however, the window list wayland backed (used for both applets) somehow cannot get the panel available space and the buttons are never shrunk, thus quickly overflowing the window list

MATE general version:
1.28 with all available 1.29 updates

To Duplicate:
1: Set up a mate-wayland session, using any 1.29 version of mate-panel supporting the show desktop applet. This commit was dated Sep 20
The problem probably came from here, as both applets re-use the same backend code but should spawn two instances.

2: Put the window list in the panel if it is not already there
3: Do NOT put the show desktop applet in the panel-the window list will overflow
4: Put the show desktop applet in the panel, restart the panel if necessary. The window list now will work properly

PR #1532, opened 2025-11-19
wl-window-list: ensure same behavior with and without show desktop...

lukefromdc:
applet in panel

*Note that in the wayland panel the show desktop applet works by using an invocation of the window list backend to send a signal to every window to mimimize or unminimize 
*This however causes every added button to be counted twice when the show desktop applet is also used 
*Detect the show desktop applet, and only divide the number of window list buttons by 2 when it is present

### DIFF
--- a/applets/wncklet/wayland-backend.c
+++ b/applets/wncklet/wayland-backend.c
@@ -68,6 +68,8 @@ typedef struct
 	gboolean fullscreen;
 } ToplevelTask;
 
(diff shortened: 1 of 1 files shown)


lukefromdc:
Closes https://github.com/mate-desktop/mate-panel/issues/1531

lukefromdc:
Merging, I've been using this with no problems and it has no effect on the x11 session

PR #1545, opened 2026-03-14
use libwnck-4.0 to be able to disable tooltips

mbkma:
Screenshot at 2026-03-14 11-20-39


mbkma:
ok, we should probably only do this when we switch to gtk4. I try to do a workaround maybe.

balazs-endresz:
Instead of a dropdown a separate checkbox for the tooltip might make more sense btw. Because it is useful to have both.

mbkma:
I do not think so. I think it looks bad, because the tooltip is on top of the preview which makes it hard to even see the preview. That is bad ui design.

Screenshot at 2026-03-14 14-05-48

vkareh:
Oh funny, I have a local branch from last year to add this but never pushed it. I like your idea of making thumbnail and toolitip mutually exclusive.

I can look at this on Monday when I'm in front of a computer

lukefromdc:
Moving to GTK4 would require re-implementing most of the deprecated widgets within
the mate-desktop package. GTK 4 is really not designed for the traditional UI we use,
as least they split out some of the GNOME-specific stuff into libadwaita which is used
only for GNOME style apps and in GNOME itself.

mbkma:
Does using libwnck-4.0 mean we can only use it when all our components are moved to gtk4? I mean locally it works, but I do not know if we run into problems when we depend on it in the future.

lukefromdc:
As of gnome-panel 3.58.1, not even gnome-panel from the gnome-flashback session (or whichever one that is)
as transitioned to libwnck-4.0 and is itself on GTK3.

Any future transition of libwnck-4.0 to GTK4 would probably bind us to either attempt an emergency transition,
revert, or fork libwnck (again). Also note that libwnck-4.0 is not showing up in the Debian Sid repo at this time

mbkma:
but in pluma for example we already use `gtksourceview4` which is also built on gtk4. why can we do it in pluma and not here?

mbkma:
actually without libwnck-4 https://github.com/GNOME/libwnck/commit/5ddd059ac86c59fb15774529c71e8ff63c810fb6 we do not even get @vkareh fix https://github.com/GNOME/libwnck/commit/86f2f7005573a335f803071cd1c98cfc92a3983a if I understand it correctly

lukefromdc:
Pretty sure gtksourceview isn't providing any actual GTK widgets in Pluma. The window list's "tasklist" by comparison
is a custom GTK widget provided directly by libwnck. EDIT: right now that's not a problem, but it would be if libwnck ever switches to GTK 4.0 and is not kept buildable againt GTK 3.0 with distros offering GTK3 builds.

At any rate, Debian doesn't even offer libwnck-4.0, as a packages, only  libwnck-3.0 so depending
on it would force Debian to either update to it or boot mate-panel (and probably all of MATE) out of the distro. Until they do, all users installing a panel depending on it have to build it themselves if they are on a distro that does not ship it.

EDIT: Did some digging, GNOME simply bumped API version to 4.0, so presumably 3.0 is going away at some point anyway unless 3.0 is branched off. As of now, 4.0 still uses GTK3, no idea their future plans

PR #1549, opened 2026-03-29
fix memory leaks

mbkma:


lukefromdc (approved):
Builds and runs fine here. The menubar and the clock can be freely addd and removed from the panel with no crash or error messages with this applied.

Pretty sure these memory leaks in master should be only when removing these applets from the panel, but every bit helps.

PR #1551, opened 2026-03-31
Weblate migration

evelyneharle:
update pot files
remove transifex configuration
remove "es_ES" as it is the same as "es" and prevent weblate usage prevent po updates when generating pot files (weblate does that for us)


evelyneharle:
This refers to https://github.com/mate-desktop/mate-terminal/issues/487

mbkma (approved):

PR #1552, opened 2026-04-01
fish: Fix alternating animation during April Fools

vkareh:
During April Fools in the morning, Wanda the fish is expected to show up upside down, with dirty-looking water, as though dead. However, because of the hour comparison, it ends up alternating between the standard animation and the dead fish.


lukefromdc:
On my setup I just tested the fish as coded in current master, got a static dead fish image not any animation. Same behavior
with this applied

vkareh:
@lukefromdc I was only able to reproduce the issue when the hour was still 12pm (but before 1pm), so the comparison between `hour >= 12` and `hour <= 12` kept toggling the `april_fools` flag between TRUE and FALSE on every cycle.

mbkma (approved):
LGTM

PR #1553, opened 2026-04-01
clock: fix memory leaks in EDS event list handling

vkareh:
The calendar popup was not freeing the list of events returned by the calendar client after populating the appointment and task views. This caused unbounded memory growth every time the popup was opened.


mbkma (approved):
looks good

PR #1554, opened 2026-04-05
Gh actions

mbkma:


.github/workflows/builds.sh
```diff
@@ -0,0 +1,38 @@
...
+	echo "::group::${1}"
+}
+infoend() {
+	echo "::endgroup::"
+}
+
+if [ -f autogen.sh ]; then
+	infobegin "Configure (autotools)"
+	NOCONFIGURE=1 ./autogen.sh
+	./configure --prefix=/usr --enable-compile-warnings=maximum || {
```
vkareh, line 19: We should also test with `--with-in-process-applets=all` (I think in addition to `--with-in-process-applets=none`), since this is important for Wayland builds. Maybe we do this only for a single distro (e.g. `${OS} == "debian"`, since that's what @lukefromdc uses for Wayland development I think?)

.github/workflows/builds.sh
```diff
@@ -0,0 +1,38 @@
...
+	}
+	infoend
+
+	infobegin "Build (autotools)"
+	make -j ${CPUS}
+	infoend
+
+	infobegin "Check (autotools)"
+	make -j ${CPUS} check || {
+		true
```
vkareh, line 31: Does this swallow errors - I think we do this in all our repos anyway, but wondering if we're missing stuff because of the `|| true`...

.github/workflows/mate-desktop.sh
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/bash
+
+set -e
+set -o pipefail
+
+NAME="mate-desktop"
+TEMP_DIR=$(mktemp -d)
+OS=$(cat /etc/os-release | grep ^ID | head -n 1 | awk -F= '{ print $2}')
```
vkareh, line 8: the `/etc/os-release` file sets the variables directly, you should be able to change this to `OS=$(source /etc/os-release && echo $ID)`, which may be a bit more robust

vkareh (changes requested):

lukefromdc:
Correct-we need to know what happens in-procrss. I am on Debian Unstable here

.github/workflows/builds.yml
```diff
@@ -0,0 +1,94 @@
...
+      volumes:
+        - /tmp/.cache
+        - /var/cache/apt
+
+    strategy:
+      fail-fast: false # don't cancel other jobs in the matrix if one fails
+      matrix:
+        container:
+          [
+            "debian:testing",
```
vkareh, line 36: Should we change this to `"debian:unstable"` since that's what @lukefromdc is using? This would likely (hopefully) solve the bad cert issue in GH Actions

lukefromdc: Fine by me, though we must remember that Debian Unstable can always get a bad or out of synch update. Matching what I use will hopefully reduce problems though

PR #1558, opened 2026-05-09
applet clock: Fix proportional font digits by tnum

### LINKED ISSUE #1338: Clock applet changes width as the time changes.
#### Expected behaviour

The clock applet should be exactly the same width all the time.

#### Actual behaviour

The clock's width is affected by the width of the time in the font used to render the time. As a result, the clock's width changes. 

This image shows the clock at two different times. The second time is wider, so the volume control, network dropdown, and weather applets all get pushed to the left.

image

Turn on the clock's second hand. Now watch as the entire upper-right panel jumps around every second. 

#### Package version

mate-desktop "1.20.5-1"

#### Linux Distribution

Debian

I didn't make a Debian bug report because the "report-bug" tool doesn't work.

2387skju:
Fixes #1338: The clock-time was jumping (and also the rest of the panel), due to different width of the digits.

This fix uses "tabular numbers" as fix.
The font must support this feature, otherwise this fix will not work.

Example Fonts, which will be fixed here: [Cantarell](https://en.wikipedia.org/wiki/Cantarell_(typeface)) and [Inter (variant of Adwaita Sans)](https://en.wikipedia.org/wiki/Inter_(typeface)#Variants)

---------
Perspective:
- I will send soon another separate fix for these non-supported fonts (e.g. font "Quicksand"): set the width of each single Glyph, if needed.
- I also planing to rework the current calculate_minimum_width(). Otherwise its maybe jumping between this views (but only at wrong defined set width): toggle_orient_clock_screen_2_16f

### DIFF
--- a/applets/clock/clock.c
+++ b/applets/clock/clock.c
@@ -1546,6 +1546,16 @@ create_main_clock_label (ClockData *cd)
         GtkWidget *label;
 
         label = gtk_label_new (NULL);
+
+        // Fix proportional font by font feature tabular numbers (tnum) (if supported by the font)
+        PangoAttribute *attr;
+        PangoAttrList *alist;
+        attr = pango_attr_font_features_new ("tnum=1");
+        alist = pango_attr_list_new ();
+        pango_attr_list_insert (alist, attr);
+        gtk_label_set_attributes (GTK_LABEL (label), alist);
+        pango_attr_list_unref (alist);
+
 /*Fixme-this is invalid for labels with any recent GTK3 version, maybe all of them*/
 /*
         g_signal_connect (label, "size-request",


lukefromdc (approved):
While I do not currently get this issue with the "sans bold" font I use (I don't have cantarell installed), I didn't get any adverse effects from this change.

lukefromdc:
Hoping for a review by someone using supported fonts

2387skju:
Another effected font is [Inter (= a variant of Adwaita Sans)](https://en.wikipedia.org/wiki/Inter_(typeface)#Variants)

lukefromdc:
Again not one I have installed. I limit unused fonts to limit filesystem sizw

vkareh (approved):

PR #1560, opened 2026-05-21
window-list: Add preview window decorations

vkareh:
This adds window decorations to the on-hover preview. It also fixes the thumbnail so that the content is captured fresh on each hover.


mbkma (approved):
Nice improvement. Pulling the frame geometry and extents into the preview path makes sense here, and capturing a fresh pixbuf on hover should keep the thumbnail aligned with the current window contents instead of showing stale decorations or border offsets.

PR #1565, opened 2026-06-10
Clock applet multiple data sources

oz123:
Building on top of the work for getting events from EDS, I added the option to read events from [vdir](https://vdirsyncer.readthedocs.io/en/stable/vdir.html).

This is useful for people who don't want to install evolution just for getting calendar events shown in the applet. Also, it requires less RAM (no background process, as vdirsyncher can be launched using cron or a systemd user unit files).

This is my 3rd attempt to implement this feature (previous attempts not published), I hope this time I got it right,


lukefromdc (commented):
I do not know how to use vdirsyncer, I normally use text files on the desktop to manage events (primitive but hard to forget) so without a gui option to invoke it from the clock, I was unable to test that functionally. Didn't interfere with Evolution, though I have an empty calendar so obscure problems might not show up.

The line
`password = "yourpassword"`
in the example config file is a security hazard if users are supposed to store their login password (which gives ROOT access if they have sudo allowed!) in it!

Without it, on running 
`vdirsyncer discover my_calendar`

I get 

```
Discovering collections for pair my_calendar
my_local:
warning: Failed to discover collections for my_remote, use `-vdebug` to see the full traceback.
error: Unknown error occurred: Cannot connect to host your-caldav-server:443 ssl:default [Name or service not known]
error: Use `-vdebug` to see the full traceback.
luke@ubuntu:~$ vdirsyncer discover my_calendar -vdebug
Usage: vdirsyncer discover [OPTIONS] [PAIRS]...
Try 'vdirsyncer discover -h' for help.

```
`vdirsyncer sync` asks me to run the previous command, and no change seen anywhere in the calendar.

oz123:
Thanks for testing!
Ok, vdirsyncher is confusing.
I will update the documentation of getting started with it.
As for storing the password - 
First, the files should be saved with `chmod 600` ( or by setting umask`. Both are actually steps that are overlooked or easily forgotten. We can add it too.
Additionally, vdirsyncher supports `password.fetch` so users running gnome-keyring can do:
```
$ keyring set mycalendar.com myuser
>TypePasswordinPrompt
```
Then in the configuration:

```
[storage foo]
type = "caldav"
url = "https://mycalendar.com/dav/"
username = "myuser"
password.fetch = ["command", "keyring", "get", "mycalendar.com/", "myuser"]
```

It replicates vdirsyncher documentation in here, but it's going to spare a few minutes for users of mate desktop and make it more secure. I can add it too.

lukefromdc:
For test purposes, is there any way to set this up to run without calling the user's
password at all? If not, being able to prompt for it and not retain it should be
enough BUT it remains an unnecessary risk. No guarantee I will be able to 
test this for function given the complexity of doing so.

After all, if the calendar is sensitive the whole /home directory 
needs to be encrypted.

lukefromdc:
Any updates on this?

lukefromdc:
Assuming this is a simple rebase?

oz123:
@vkareh would love to have your review here, as this work is based on the EDS backend you backported from gnome.
@lukefromdc I added docs that explicitly addresses your concerns about secrets leaking. I think I also found a quirk you bumped into. Documented the work-around a possible bug in vdirsync.

I link to test script in this gist. It will run a calendar using radicale, inject events, setup vdirsync, sync events, sync events, and finally show the calendar pop up with the events.

It should be something like this:

image

The gist: https://gist.github.com/oz123/da9bdb04ce47825be8206b4238e80e0c

applets/clock/calendar-vdir-provider.c
```diff
@@ -0,0 +1,921 @@
...
+  priv->display_name = read_metadata_file (directory, "displayname");
+  priv->color_string = read_metadata_file (directory, "color");
+
+  /* Resolve the local timezone via the system TZ name */
+  {
+    time_t    now   = time (NULL);
+    struct tm local = { 0, };
+    localtime_r (&now, &local);
+    char tz_name[64];
+    if (strftime (tz_name, sizeof (tz_name), "%Z", &local) > 0)
```
vkareh: This `%Z` can yield ambiguous tz names and end up falling back to UTC. You should use `system_timezone_get` like in the EDS file.

configure.ac
```diff
@@ -123,6 +123,35 @@ AM_CONDITIONAL(HAVE_EDS, test "x$have_eds" = "xyes")
...
+  have_libical=no
+fi
+
+# EDS implies libical (libecal depends on libical-glib)
+if test "x$have_eds" = "xyes" && test "x$have_libical" = "xno"; then
+  have_libical=yes
+  AC_DEFINE(HAVE_LIBICAL, 1, [Define if libical-glib is available])
+  LIBICAL_CFLAGS="$EDS_CFLAGS"
+  LIBICAL_LIBS=""
+fi
```
vkareh, line 153: This seems fragile. Can you just do a `PKG_CHECK_MODULES` for libical-glib if `have_eds = yes`, that way it guarantees it as an install dependency.

applets/clock/calendar-vdir-provider.c
```diff
@@ -0,0 +1,921 @@
...
+  } while (0)
+
+  /* ------------------------------------------------------------------
+   * RRULE expansion
+   * ------------------------------------------------------------------ */
+  prop = i_cal_component_get_first_property (comp, I_CAL_RRULE_PROPERTY);
+  if (prop != NULL)
+    {
+      ICalRecurrence   *rrule = i_cal_property_get_rrule (prop);
+      ICalRecurIterator *iter  = i_cal_recur_iterator_new (rrule, dtstart);
```
vkareh, line 233: Need a NULL check on this

applets/clock/calendar-vdir-provider.c
```diff
@@ -0,0 +1,921 @@
...
+
+      if (looks_like_vdir)
+        {
+          CalendarProvider *p = calendar_vdir_provider_new (subdir);
+          if (p != NULL)
+            providers = g_slist_prepend (providers, p);
+        }
+      else
+        {
+          /* Not a collection itself — recurse one level (vdirsyncer stores
```
vkareh, line 911: I think `recurse one level` here is not accurate - this seems like it recurs all the way down (which is probably fine, mostly the comment doesn't seem right).

applets/clock/calendar-client.c
```diff
@@ -131,1960 +85,640 @@ enum
...
+        /* Check whether this settings instance has the vdir key before
+         * calling get_strv (avoids a GLib critical if key is absent). */
+        gchar **all_keys  = g_settings_list_keys (settings);
+        gboolean has_key  = FALSE;
+        for (int ki = 0; all_keys[ki] != NULL; ki++)
+          if (g_strcmp0 (all_keys[ki], KEY_VDIR_CALENDAR_PATHS) == 0)
+            { has_key = TRUE; break; }
+        g_strfreev (all_keys);
+
+        if (has_key)
```
vkareh, line 458: You should be able to replace all this with a single `g_settings_schema_has_key` call...

applets/clock/calendar-client.c
```diff
@@ -131,1960 +85,640 @@ enum
...
+          {
+            gchar **paths = g_settings_get_strv (settings, KEY_VDIR_CALENDAR_PATHS);
```
vkareh, line 460: This needs a watch of some sort during init, otherwise you'd have to restart the applet to apply new changes...

vkareh (changes requested):
The `applets/clock/meson.build` file still references `calendar-sources.c`, which breaks meson builds. Also there seem to be some inconsistencies in the `if have_eds` gating and dependencies between both build systems...

I did a first pass, mostly jumping around between things. I'll review more parts as I go over them.

oz123:
@vkareh thanks for the review! I'll address the issues and push fixes.

applets/clock/calendar-vdir-provider.c
oz123: fixed.

lukefromdc (commented):
The vsyncdir documentation is still rather out of my understanding, but like I said earlier I have never used it. I have never seen a password-protected calendar before, note that I have only ever managed anything that sensitive locally never remotely so that may be why,

oz123:
> The vsyncdir documentation is still rather out of my understanding, but like I said earlier I have never used it. I have never seen a password-protected calendar before, note that I have only ever managed anything that sensitive locally never remotely so that may be why,

Have you tried the script in the gist?

lukefromdc:
First I saw that gist. You have the script set up for an x11 build, meaning I'd have to close all applications, log out, and switch to the x11 session to test that build. Building isn't the problem using it is.

lukefromdc:
At any rate we have another reviewer and if it works for them I am satisfied.

configure.ac
oz123: done.

lukefromdc:
@vkareh is this ready to go? This one is really out of my understanding

oz123:
No. It's not really as I haven't addressed all the issues point out by @vkareh. I'm occupied with other things right now, doing my best to address those asap.

lukefromdc:
No problem, work on in when you get time

PR #1567, opened 2026-07-10
Backport gnome-panel pack model for zone positioning

### LINKED ISSUE #62: align applets with relative position
Fix the somewhat "brokenness" of mate-panel (which led to the feature-trimming redesign of gnome-panel) in a way that the freedom for the user to place the applets and buttons wherever he likes is retained.

Problem specification:
When one resizes the screen resolution (and possible when other race conditions occur), the items on the panel tend to be shuffled around and be misaligned.

Possible solution: 
Position of items should be remembered in a smart and _relative_ way, regardless of the size of the panel.  That means if one puts an item on the very right side, it should _always_ be there, at the very end of the panel. All items placed directly side by side without spacing between them, should _always_ retain their order and null-spacing relative to the others. If the panel gets smaller, the items order should _never_ be touched, the space between the items should be adjusted instead. It should be tried to retain the relative position as good as possible under all circumstances.
If the size gets to small to place every item on the panel, a drop-down-mechanism or hide-and-show feature of groups of items depending on where one puts the mouse-cursor over, could be introduced. Maybe even some sort or user-setable prioritization of what items should _always_ be shown would be nice. For example the mate-menu is one that should be shown under all circumstances, it should default to that setting, but a user should be allowed to change that nevertheless. The systray (notification area) on the other hand, should default to collapse to one general icon if space on the panel gets scarce. User-added items should default to a higher-than-normal but lower-than-highest priority.
Bottom line, smart, sane defaults, user-configurable.

What do you think?

regards
Michael
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/1311866-align-applets-with-relative-position?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #183: Feature to place applets left-centre-right
It would be nice if users could place applets (using gsettings?) on the left-centre-right of the panel and the applet stayed in that position.
mate-panel
This is how I have my panel set up but if I change screens (to a larger screen)  the weather/clock/calender applet moves from the centre of MATE panel to the right like the screenshot below
mate-panel-2.
If there was a setting that users could just say "centre" the applet should, regardless of the screen size/resolution, stay in the centre. Currently we just drag the applet to where we want it and if we change screens (or resolutions, please see screenshot below for resolution change) it moves to the coordinates that the location corresponds to.
mate-panel-3
I hope I have explained this well enough
## 

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2121110-feature-to-place-applets-left-centre-right?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F314880&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #560: icons position on mate panel
#### Expected behaviour
icons should be in order as i put them

#### Actual behaviour
some icons reorder on next boot

put icon for showing desktop as most right icon beyond clock and calendar icon ..and they switch on next boot

#### MATE general version
mate for debian 1.26

#### Linux Distribution
debian

MATE-pult

### LINKED ISSUE #1108: Position of panel-applets neither saved nor locked
#### Expected behaviour

Panel applets should stay in place even after a reboot

#### Actual behaviour

They change position after reboot even when locked

#### MATE general version

1.24.0

#### Package version

1.24.0

#### Linux Distribution

Manjaro

### LINKED ISSUE #1456: Panel Icons change positions at boot even if locked
#### Expected behaviour
When you add apps to the panel and move them around and then lock them I expect the next time you boot they would be where you put them.

#### Actual behaviour
They should stay where I put them

Brand new installations of Linux Mint MATE v22. Just add apps to the panel and move them around an lock them. Then reboot.

#### MATE general version
1.26.2

#### Package version
mate-panel                      1.27.1-2build3     amd64

#### Linux Distribution
Linux Mint MATE 22

https://github.com/linuxmint/linuxmint/issues/736

New clean installs of Linux Mint MATE 22. Several machines not just one.
When 22 installs it puts the menu icon all the way to the left in the default bottom panel. Then it puts the show desktop app icon and then a separator. Then the other default app icons.
On most of the machines I move this around. First the Mint menu, then a separator, then the rest of the icons. I add other apps and arrange them the least used icons farthest left to the most used icons farthest right. The show desktop icon is usually the farthest right just left of the WindowsListApplet. When I get all the apps and the positions set I lock every one even the separators.
With all the machines that I do this to when I reboot the separator that was next to the mint menu moves right over an app or two. In addition some of the machines move the WindowsListApplet further left between other icons. So I diligently put everything back the way it was and lock them but the next time I reboot things move again. This goes on for about 6 or 8 reboots and then they stay where I put them. Like it heals itself.
To track this down I found out that I can output the mate-panel settings by doing dconf dump /org/mate/panel/ > ~/dconf-panel-settings in a terminal. You can look at that file and see each object settings including it's position. The order of the objects in that file are jumbled so it makes it difficult to see what is placed where.
So I enlisted Perplexity AI to write a python script to sort the objects first by panel, then by the relative-to-end setting, then by position left to right. When I looked at the resulting output of machines that were still dong this I can easily see what the problem is. There are identical duplicate objects and different objects with the same position setting.
So there is some bug in the mate-panel program having to do with either adding apps or moving them around that creates the problem objects. I can easily edit the settings in a text editor and re-import it back into mate-panel with dconf load /org/mate/panel/ < ~/dconf-panel-new. Once I do that everything says put.

vkareh:
This is a backport of gnome-panel's pack model, where applets are index-sorted in either START, CENTER, or END zones within the panel.

This is the list of commits backported:
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/446ff991d
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/f42b54636
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/8804c8b37
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/9480f4458
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/6db6be277
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/51757603d
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/cb29d47f5
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/1652218bb
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/ed1054799
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/2256381b6
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/533581210
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/f63d91d16
- https://gitlab.gnome.org/GNOME/gnome-panel/-/commit/6278658de

The main difference is that mate-panel still supports the right-stick and position as a fallback, but as soon as the applets load they'll get the new keys. That way we don't break existing layouts. We can make the old keys a no-op on the next major release.

Fixes #62
Fixes #183
Fixes #1108
Fixes #1456
Fixes https://github.com/mate-desktop/mate-desktop/issues/560

### DIFF
--- a/data/default.layout
+++ b/data/default.layout
@@ -11,43 +11,43 @@ size=24
 [Object menu-bar]
 object-type=menu-bar
 toplevel-id=top
-position=0
+pack-index=0
 locked=true
 
 [Object notification-area]
 object-type=applet
 applet-iid=NotificationAreaAppletFactory::NotificationArea
 toplevel-id=top
-position=10
-panel-right-stick=true
+pack-type=end
+pack-index=1
 locked=true
 
 [Object clock]
 object-type=applet
 applet-iid=ClockAppletFactory::ClockApplet
 toplevel-id=top
-position=0
-panel-right-stick=true
+pack-type=end
+pack-index=0
 locked=true
 
 [Object show-desktop]
 object-type=applet
 applet-iid=WnckletFactory::ShowDesktopApplet
 toplevel-id=bottom
-position=0
+pack-index=0
 locked=true
 
 [Object window-list]
 object-type=applet
 applet-iid=WnckletFactory::WindowListApplet
 toplevel-id=bottom
-position=20
+pack-index=1
 locked=true
 
 [Object workspace-switcher]
 object-type=applet
 applet-iid=WnckletFactory::WorkspaceSwitcherApplet
 toplevel-id=bottom
-position=0
-panel-right-stick=true
+pack-type=end
+pack-index=0
 locked=true
--- a/data/fedora.layout
+++ b/data/fedora.layout
@@ -11,72 +11,72 @@ size=24
 [Object menu-bar]
 object-type=menu-bar
 toplevel-id=top
-position=0
+pack-index=0
 locked=true
 
 [Object file-browser]
 object-type=launcher
 launcher-location=/usr/share/applications/caja-browser.desktop
 toplevel-id=top
-position=10
+pack-index=1
 locked=true
 
 [Object terminal]
 object-type=launcher
 launcher-location=/usr/share/applications/mate-terminal.desktop
 toplevel-id=top
-position=20
+pack-index=2
 locked=true
 
 [Object web-browser]
 object-type=launcher
 launcher-location=/usr/share/applications/firefox.desktop
 toplevel-id=top
-position=30
+pack-index=3
 locked=true
 
 [Object volume-control]
 object-type=applet
 applet-iid=GvcAppletFactory::GvcApplet
 toplevel-id=top
-position=20
-panel-right-stick=true
+pack-type=end
+pack-index=2
 locked=true
 
 [Object notification-area]
 object-type=applet
 applet-iid=NotificationAreaAppletFactory::NotificationArea
 toplevel-id=top
-position=10
-panel-right-stick=true
+pack-type=end
+pack-index=1
 locked=true
 
 [Object clock]
 object-type=applet
 applet-iid=ClockAppletFactory::ClockApplet
 toplevel-id=top
-position=0
-panel-right-stick=true
+pack-type=end
+pack-index=0
 locked=true
 
 [Object show-desktop]
 object-type=applet
 applet-iid=WnckletFactory::ShowDesktopApplet
 toplevel-id=bottom
-position=0
+pack-index=0
 locked=true
 
 [Object window-list]
 object-type=applet
 applet-iid=WnckletFactory::WindowListApplet
 toplevel-id=bottom
-position=20
+pack-index=1
 locked=true
 
 [Object workspace-switcher]
 object-type=applet
 applet-iid=WnckletFactory::WorkspaceSwitcherApplet
 toplevel-id=bottom
-position=0
-panel-right-stick=true
+pack-type=end
+pack-index=0
 locked=true
--- a/data/linuxmint.layout
+++ b/data/linuxmint.layout
@@ -7,35 +7,35 @@ size=24
 object-type=applet
 applet-iid=MintMenuAppletFactory::MintMenuApplet
 toplevel-id=bottom
-position=0
+pack-index=0
 locked=true
 
 [Object show-desktop]
 object-type=applet
 applet-iid=WnckletFactory::ShowDesktopApplet
 toplevel-id=bottom
-position=1
+pack-index=1
 locked=true
 
 [Object window-list]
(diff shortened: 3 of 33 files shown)


lukefromdc (approved):
The volume of code is more than I can give more than a cursory skim at the moment but I didn't see any blatant problems.

Unlike the previous attempts to fix issues with applet positioning in 1.27, this did not create any moved applets with my existing layout. No nasty surprises on reverting to git master either.

vkareh:
@lukefromdc thanks for testing it. My main fear was losing my panel configuration, which is in part why I kept the old code working instead of just replacing it entirely like GNOME did. For review it's a lot easier going commit by commit, although some are still on the longer side...

lukefromdc:
While I have not had runtime issues with applets from mate-applets with this, it looks like we change the API used by all other applets here. I don't know for sure if we actually do, but I do see changes in every applet shipped by mate-panel including the number of variables when we call functions like `panel_menu_bar_load_from_gsettings()` , `panel_profile_prepare_object()`  and others. Seems to me this should require updating all applets provided by other packages.

Haven't tried to rebuilt mate-applets, I think my test build of mate-utils (for the screenshot code) was with this installed and it worked. This doesn't change anything about this PR (which worked in my test), rather it indicates a potential need for fallback PR's. Even if we are catching this with the fallback to pixel positions, that could still cause some applets to move around for some I suspect.

Anyway, people have been screaming for years for a fix to this moving applets on restart issue. I don't have it, but I do remember it from the GNOME 2 days.

vkareh:
@lukefromdc those applets you see changing there are not so much "real" applets, but more internal applet "helpers": separator, drawer (which has a mini mate-panel inside), generic launcher, etc. These don't use the libmate-panel API, instead they are compiled directly into the mate-panel binary.

All other applets rely on the libmate-panel API, which doesn't actually change.

lukefromdc:
In that case this should be ready to go

PR #1568, opened 2026-07-18
[Wayland] Add support for ext-workspace-v1 in pager

Sunderland93:
This PR adds support for `ext-workspace-v1` protocol in workspace switcher (pager) using [libxfce4windowing](https://gitlab.xfce.org/xfce/libxfce4windowing/-/tree/main?ref_type=heads) library. Although there are currently a number of limitations of the protocol itself (e.g. there is no window preview in pager and no window-to-workspace mapping), basic implementation works pretty good. Tested on Labwc.

cc @lukefromdc


lukefromdc:
I tested this, it actually does work with labwc though having no experience setting up that compositor I was unable to get more than one workspace showing on it. It did NOT work with wayfire however as versions of wayfire new enough to contain the necessary support seem to be impossible to get a build to finish. Most recent successful build was 0.11.0 from git master 2025-09-19.

I do not know how long it will be before wayfire releases a 0.12 version (which will presumably build without issue on most systems)  but current git master is unbuildable on Debian Unstable.

If you get this ready for merge now, that's fine with me, we just need to
1: mark it experimental and not supported by all wlroots versions
2: set a minimum width so users can remove it if they add it and zero workspaces are shown. That currently yields a 1px vertical line on a horizontal panel

Sunderland93:
I think we might use libxfce4windowing as abstraction layer for ext-workspace and some other panel-related protocols. It's pretty trivial to implement it. What do you think?

lukefromdc:
I've never worked with  libxfce4windowing but if you can get this to work with wayfire i'm all for it.

lukefromdc:
I reported the wayfire build failures at
https://github.com/WayfireWM/wayfire/issues/3074

Sunderland93:
Done. Also I fixed a couple of UI issues with applet itself

lukefromdc:
I found the wayfire build issues: needed to disable using system wf-config or it tried to use an old version, and to load the submodules into the build directory with` git submodule update` not ` git submodule update --remote` or we get the dev version of wlroots (now at 0.21), which current git master wayfire is not compatable with.

lukefromdc:
Just built and installed this for testing. On wayfire we now show one workspace (always #1) but do not read which workspace we are actually on or show the other workspaces. The switcher does however look as we would expect it to look with one workspace and not being able to show windows within it (same as compiz with one workspace).

lukefromdc:
CI failures are in mate-submodules again. CI will also need a change to reflect the new dependency on libxfce4windowing-0 for wayland builds

lukefromdc:
Found out why it didn't work with wayfire. Turns out we get `Your compositor does not support the ext_workspace_manager_v1 protocol` when running the panel with the switcher in it under wayfire. In otther words wayfire isn't ready yet even with yesterday's build

lukefromdc:
Commenting at https://github.com/WayfireWM/wayfire/issues/2838#issuecomment-5030181953
and https://github.com/WayfireWM/wayfire/issues/2838#issuecomment-5030246110
on the usefulness of getting workspace support working in wayfire. There is a somewhat old 3ed party branch including it, but it's from 7 months ago and did not build. Master from his repo built fine, so presumably needs rebasing and updating

Sunderland93:
> CI will also need a change to reflect the new dependency on libxfce4windowing-0 for wayland builds

Done

Sunderland93:
@lukefromdc `ext-workspace` support is merged in Wayfire

lukefromdc:
And I have it working here. Wayfire just had their 0.11 release without this, but 1.0 is reported to be in the works and should be out prior to mate 1.30. Nothing breaks if the switcher applet is loaded with an older version of wayfire, it just only shows one workspace and cannot switch to anything. 

Will merge, and I should be able down the road to add theming support to make the switcher does as it does with compiz in x11

lukefromdc:
Theming required only theme file changes, not changes to this code. I have the switcher looking on my box exactly as it did with compiz and x11 (and without the recent wnck switcher button width bug) right now

PR #1571, opened 2026-07-31
Add Meson support

Sunderland93:


lukefromdc:
Got this error trying to configure:
`
mate-panel/mate-submodules/meson.build:1:0: ERROR: Second call to project().`

Sunderland93:
Should be fixed now

lukefromdc (approved):
A build with all applets in-process and a build with all applets out of process both finished fine. Help files built fine, an issue we had in meson support for another package.

lukefromdc:
Just tested building mate-applets with this installed, since we lose the .la files (libmate-panel-applet-4.la in this case) in all meson builds of anything. The build finished fine. Mate-media (which builds an applet) built fine with autotools, but we also have meson support there anyway.

lukefromdc:
We got this warning at ninja -C build:
```
[584/869] Compiling gschemas data/gsettings-compile-data
No schema files found: doing nothing.
```
An install into a test directory did in fact install the schema files, but glib-compile-schemas did not appear to have been run. Not sure if it would run when actually installing to the system or not, but testing that here would really need rebasing against master as I have the switcher in the panel plus removing my previous panel install.

In Debian packages and probably other distribution systems, this is run in postinstall scripts, but local installs with meson and ninja do need to compile schemas to run if the panel was not previously installed or any schema has changed

lukefromdc:
That seems to work, we get this code on installing to a test directory indicating schema compilation triggering, being bypassed as intended when DESTDIR is set to install to (in this case) a test directory

lukefromdc:
Confirmed on master with this squashed, rebased, and merged schemas compile fine

PR #1573, opened 2026-08-08
build: Fix Wayland auto-detection breaking X11-only builds

### LINKED ISSUE #1572: Can not build after commit 827ddfd - Use libxfce4windowing for handle panel-related procotols, applet fixes
### Steps to Reproduce

1. Try to build in RHEL 9 (any OS with glib < 2.72.0)
2. Build fails for libxfce4windowing < 4.19.4
3. Can not build libxfce4windowing >= 4.19.9 due to glib >= 2.72.0 requirement
4. Build with libxfce4windowing = 4.19.2 fails with error: 'XFW_CLIENT_TYPE_PAGER' undeclared

### Expected Behavior

Build for quite standard / popular OS like RHEL 9 succeed.

### Actual Behavior

Build for RHEL 9 fails.

### Error Output

```shell

```

### Version

1.29.0 git20260801.1b52783

### Distribution

Rocky Linux 9.8

### Additional Context

_No response_

### Reproducibility

Every time

### Checklist

- [x] I have searched existing issues to avoid duplicates
- [x] I have provided all the requested information above

vkareh:
libxfce4windowing check was a hard error even when Wayland support was only auto-detected. Any X11 without libxfce4windowing would fail to configure entirely.

This changes it to only error out if --enable-wayland=yes was explicit and otherwise it just falls back to X11.

Also bumped the version requirement to 4.19.4, since that's what XFW_CLIENT_TYPE_PAGER actually needs.

Fixes #1572

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -214,15 +214,21 @@ if test "x$enable_wayland" != "xno"; then
     ])
 fi
 
-AM_CONDITIONAL(ENABLE_WAYLAND, [test "x$have_wayland" = "xyes"])
-
 have_xfw=no
 if test "x$have_wayland" = "xyes"; then
-  AC_DEFINE(HAVE_WAYLAND, 1, [Have the Wayland development library])
-
-  PKG_CHECK_MODULES(XFW, libxfce4windowing-0 >= 4.19.0, have_xfw=yes, [
-    AC_MSG_ERROR([Wayland support requires libxfce4windowing-0 >= 4.19.0])
+  PKG_CHECK_MODULES(XFW, libxfce4windowing-0 >= 4.19.4, have_xfw=yes, [
+    if test "x$enable_wayland" = "xyes"; then
+      AC_MSG_ERROR([Wayland support requires libxfce4windowing-0 >= 4.19.4])
+    else
+      have_wayland=no
+    fi
   ])
+fi
+
+AM_CONDITIONAL(ENABLE_WAYLAND, [test "x$have_wayland" = "xyes"])
+
+if test "x$have_wayland" = "xyes"; then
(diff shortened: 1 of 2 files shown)


lukefromdc (approved):
Both meson and autotools builds with this on my setup with all necessary libraries for wayland support worked fine.

We did get an error with glib-compile-schemas bring invoked at the end of the meson build (not the install) but this is surely present in master too as nothing that would affect that changes in this PR. Appears to be a harmless waste invocation, as installing with `sudo ninja -c build install` successfully recompiled schemas
```

[589/869] Generating data/gschemas.compiled with a custom command
/home/luke/Desktop/Development/MATE_Development_Work/mate-panel_1.29.0+git20260801v2/build/data/org.mate.panel.object.gschema.xml:4:1  Error on line 4 char 1: <enum id='org.mate.panel.PanelObjectType'> not (yet) defined..  This entire file has been ignored.
```

lukefromdc:
Looks like this or something else fixed the CI issues too

lukefromdc:
We need a test on a system with the wayland/XCFE libs NOT installed if we can get one, so hoping for a second review

vkareh:
> We need a test on a system with the wayland/XCFE libs NOT installed if we can get one, so hoping for a second review

If it helps, my system has none of those libraries :P

lukefromdc:
A test on that system with default build options should be perfect. The corner case is an older wayland enabled system
with a compositor gtk-layer-shell etc but without the XFCE library. In that case, attempting to select wayland should
error out saying that library needs to be installed

lukefromdc:
Assuming this worked on your system and misinterpreted your comment. Merging

PR #1578, opened 2026-09-07
Add configurable panel scroll volume with single OSD

guideahon:
## Summary

Add optional native volume control to MATE panels.

## Details

- Scrolling up/down over a panel changes the default PulseAudio sink volume.
- Works with PipeWire setups exposing the PulseAudio compatibility server.
- Handles discrete and smooth scroll events.
- Uses one reusable OSD window, replacing the previous notification instead of stacking OSDs.
- Places the OSD on the monitor containing the pointer.
- Provides optional desktop-bell feedback.
- Adds GSettings for enabling the feature, step size, OSD and feedback.
- Keeps the feature disabled by default for compatibility.
- Uses optional `libpulse`; builds without it and leaves the feature unavailable rather than adding a hard dependency.

The existing MATE panel model already stores monitor and autohide state per `PanelToplevel`, with independent timers. No duplicate panel implementation was added because that behavior is already native in `mate-panel`.

## Validation

- `gcc -fsyntax-only` for `panel-volume.c` with libpulse enabled.
- `gcc -fsyntax-only` for `panel-volume.c` with libpulse unavailable.
- `git diff --check`.

A complete current-master build is blocked on this Ubuntu installation because the installed `mate-desktop` is 1.26.2 while this branch's current master requires >= 1.28.2.


L-U-T-i:
I've built mate-panel with this PR, but see no difference comparing to the previous build.

What should be different, and how exactly to test this PR?

lukefromdc:
This behavior would conflict with scrolling over other applets notable the window list as well as not being very discoverable. In x11 scrolling over the existing volume applet from mate-media already works. This would best be implemented as an alternative volume applet not whole panel scrolling

lukefromdc:
Why do I see changes to the applet pack system as part of this PR? Are they necessary to accomodate this?

lukefromdc:
At least one of the failed builds errored out with this message:
```
2026-09-13T02:53:37.9798301Z panel-volume.c: In function 'panel_volume_handle_scroll':
2026-09-13T02:53:37.9798915Z panel-volume.c:225:9: error: 'direction' undeclared (first use in this function); did you mean 'sigaction'?
2026-09-13T02:53:37.9799373Z   225 |         direction = 1;
2026-09-13T02:53:37.9799622Z       |         ^~~~~~~~~
2026-09-13T02:53:37.9800255Z panel-volume.c:225:9: note: each undeclared identifier is reported only once for each function it appears in
2026-09-13T02:53:37.9800802Z panel-volume.c:232:5: error: 'step' undeclared (first use in this function)
```

lukefromdc:
I see the the word "codex" in the repo name. I suspect this is generated by AI I do not have the time to review this volume of code line by line and trace the entire flow of it in my head. I will never merge AI generated code where nobody including the author has done this as AI is known to hallucinate. There are also potential copyright issues such the AI training on a mix of BSD licensed and GPL code, meaning the output cannot be released under the GPL.

Also note that this is cosmetically different from but does the same job as the volume applet we already have. Little benefit and could introduce new bugs-and is another marker of AI generated content. Closing until this is certified as human-written or as reviewed and understood line by line by the author

PR #1581, opened 2026-09-08
panel-applet-frame: Fix crash when reloading a crashed applet

vkareh:
When an applet crashes and the user clicks "Reload", the panel reads pack-style data from the applet's widget after mate_panel_applet_clean() has already destroyed it and freed the AppletInfo. Move the reads before the clean call so everything is captured while the data is still valid, otherwise this can crash the panel.


lukefromdc:
This can only be tested on out of process built applets, as a crashed applet in the same process as mate-panel 
takes down the entire panel.

L-U-T-i:
I have "out-of-process" builded panel and applets (everything the latest git snapshots, master branch) installed, I can try to test it.

How can it be tested (how to crash an applet and what exactly to do after)?

lukefromdc:
If you can identify the applet processes in matre-sytem-monitor's process tab you should be able to kill one from there

L-U-T-i:
Tested. Killed process, been asked what to do, reloaded OK. Seems to work as expected.

lukefromdc:
We have a successful functional test of this with out of process applets by another, and I can see exactly how this code works to read the necessary data before freeing a crashed applet. The code looks good. Merging

PR #1582, opened 2026-09-11
applets/window-selector: port to Wayland

Sunderland93:
The only restriction is window stacking ordering - Wayland doesn't support it (yet?), so windows return unordered. Tested in Wayfire, should works with all compositors that implements `wlr-foreign-toplevel-management` protocol.

### DIFF
--- a/applets/wncklet/Makefile.am
+++ b/applets/wncklet/Makefile.am
@@ -42,7 +42,9 @@ WNCKLET_SOURCES += \
 	wayland-backend.c \
 	wayland-backend.h \
 	wayland-workspace.c \
-	wayland-workspace.h
+	wayland-workspace.h \
+	wayland-window-menu.c \
+	wayland-window-menu.h
 
 WNCKLET_LDADD += \
 	$(XFW_LIBS)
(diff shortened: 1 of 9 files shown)


lukefromdc:
This built fine, actually adding the applet to the panel was fiddly, often having to restart the panel to get it to show up. Once it did. the icon would follow the last window opened, but the menu would come up as "no windows open." Image below was with several windows open

No_Windows_Open


If mate-panel was running from terminal with mate-panel --replace, I got both of these messages but no others:
```
** Message: 22:46:17.308: Wayland does not support discovering window stacking; windows returned are unordered
** Message: 22:46:17.308: Window<->Workspace association is not available on Wayland

```

Sunderland93:
Please check the latest commit

lukefromdc:
Mostly works with a few nuisance issues. I ran the panel from terminal and got this warning:

```
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug
```
The menu when opened the first time was empty and had to be loaded from the end with the scrolbar ends. It then never expanded fully, giving this result:

Window_Menu_applet_V2

Sunderland93:
Should be fixed now

lukefromdc (approved):
This now works as intended, the menu renders perfectly

lukefromdc:
I reran the CI, all the builds and cppcheck finished clean

PR #1583, opened 2026-09-12
wayland/tasklist: support task buttons grouping

Sunderland93:
Tested in "auto" and "always" mode. Not tested with vertical panels.

### DIFF
--- a/applets/wncklet/wayland-backend.c
+++ b/applets/wncklet/wayland-backend.c
@@ -1776,6 +1776,26 @@ tasklist_rebuild (TasklistManager *tasklist)
 
 	tasklist->rebuilding = FALSE;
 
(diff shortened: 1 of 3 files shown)


lukefromdc (commented):
This worked mostly as intended when I tested it, and changes were applied in realtime.

I did run into one issue when turning window grouping off with a lot of windows on the desktop. If all the buttons don't fit with both text and icons showing, instead of running adjust buttons and hiding the icons and in really extreme cases the labels, the window list overflows.

This however should be a simple fix, we probably need to run` adjust_buttons` from `rebuild_tasklist`

Sunderland93:
Should be fixed now

lukefromdc:
This now works flawlessly

PR #1584, opened 2026-09-18
[Wayland] Add support for autohide

Sunderland93:
Tested with 1, 2 and 4 panels. Not tested on multi-monitor setup


lukefromdc (commented):
This works mostly as expected. Note that in both x11 and wayland some themes do not show the buttons as visually distinct from the rest of the panel when not clicked on, but all show them with clicked on, some themes visually separate the buttons always, and in all cases the hide arrows are shown. 

One issue came up: testing with two panels showing, I found that when a top or left panel was hidden, wayfire's default background showed rather than caja's background.

Sunderland93:
Should be fixed now

lukefromdc:
Still getting the same behavior with this installed: wayfire's native background showing behind top and left
panels when hidden

